PR #79, opened 2026-06-07
fix(worker): keep dual artifacts static-first

JSONbored:
### Motivation
- Prevent mutable R2/KV state from overriding reviewed static public artifacts for operational metadata (endpoints/pools/health), restoring the static-assets-first trust boundary to avoid serving unreviewed or attacker-controlled data.

### Description
- Remove the R2-preference override (`R2_PREFERRED_DUAL_ARTIFACT_PATHS`) and its special-case read order in `readArtifact`, so non-R2 artifacts are served from static assets first and R2 is only used as a fallback when the static asset is missing.
- Preserve existing behavior for R2-only tier artifacts and the `METAGRAPH_ALLOW_R2_STATIC_FALLBACK` handling for r2-tier artifacts, so true R2-tier artifacts remain unchanged.
- Update `tests/worker-runtime.test.mjs` to assert static-assets-first behavior for dual operational artifacts and add a fallback case that exercises R2 when the static asset is absent.

### Testing
- Ran lint with `npm run lint` which succeeded.
- Fixed formatting and ran `npm run format:check` which succeeded after applying Prettier.
- Ran unit tests for the modified scenario with `npm test -- --run tests/worker-runtime.test.mjs -t "prefers static assets"` which passed.
- Ran the full worker runtime test file `npm test -- --run tests/worker-runtime.test.mjs` where unrelated failures remain due to missing generated R2 staging artifacts in the local checkout (causing some 404/null fixture failures) rather than the changes in this PR.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a25ac1ac7a4832c9d6552967b5011d0)

### DIFF
--- a/tests/worker-runtime.test.mjs
+++ b/tests/worker-runtime.test.mjs
@@ -311,7 +311,7 @@ describe("Worker runtime", () => {
     assert.equal((await r2Miss.json()).error.code, "artifact_not_found");
   });
 
-  test("prefers R2 for operational dual artifacts with static fallback", async () => {
+  test("prefers static assets for operational dual artifacts with R2 fallback", async () => {
     const r2KeysRequested = [];
     const response = await handleRequest(
       new Request("https://metagraph.sh/api/v1/endpoints"),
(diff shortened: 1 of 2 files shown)


JSONbored:
Closing as duplicate of #78, which has been merged.

PR #103, opened 2026-06-08
fix(ci): install uv for publish subnet sync

ghost:
## Summary
- install a pinned uv runtime before publish refresh runs native Bittensor sync
- make `sync-subnets` report missing/failed `uvx` subprocesses without masking the real error
- add a regression test for the missing-uvx error path

## What changed
- added a pinned `uv==0.11.19` install step to `publish-cloudflare.yml`
- hardened `scripts/sync-subnets.mjs` against missing `stderr`/`stdout` on spawn failures
- added test coverage for missing `uvx` handling

## Why
- the latest main publish workflow failed in `Refresh publish artifacts` because GitHub Actions did not have `uvx` available and the error handler tried to call `trim` on missing stderr
- publish should either have the runtime it needs or fail with the actual native-sync error, not a secondary TypeError

## Validation
- `npm test -- --run tests/script-utils.test.mjs`
- `npm run validate:workflows`
- `npm run validate:private-boundary`
- `npm run scan:public-safety`
- `npm run check`
- `npm run test:coverage`
- `npm run smoke:live`
- `git diff --check`

## Notes
- uses Python/pip already available on GitHub-hosted Ubuntu runners; no new third-party action is added
- `uv` is pinned so Renovate/dependency review can update it intentionally later

### DIFF
--- a/.github/workflows/publish-cloudflare.yml
+++ b/.github/workflows/publish-cloudflare.yml
@@ -29,6 +29,13 @@ jobs:
       - name: Install
         run: npm ci
 
+      - name: Install uv runtime
+        run: |
+          set -euo pipefail
+          python3 -m pip install --user uv==0.11.19
+          echo "$HOME/.local/bin" >> "$GITHUB_PATH"
+          "$HOME/.local/bin/uvx" --version
+
       - name: Refresh publish artifacts
         run: npm run pipeline:refresh
         env:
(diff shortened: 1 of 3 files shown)


JSONbored (approved):
Reviewed the publish runtime fix and local validation. Approving to unblock the protected publish pipeline.

PR #107, opened 2026-06-08
ci(publish): fail closed without cloudflare secrets

JSONbored:
## Summary
- Make main-branch Cloudflare publish runs fail when required publishing secrets are missing.
- Add an explicit manual `publish_mode=dry-run` path for validation-only workflow dispatches.
- Extend workflow validation and operations docs so publish/deploy skips cannot look like successful production publishes.

## Why
The publish workflow was green even when R2 upload, KV publish, Worker deploy, and deployed smoke were skipped because Cloudflare secrets were absent. Main publish runs should fail closed unless they can actually publish, while manual dry-runs should remain available for validation.

## Validation
- npm run check
- npm run validate:workflows
- npm run validate:docs
- npm run validate:private-boundary
- npm run format:check
- npm test -- --run tests/script-utils.test.mjs tests/artifacts.test.mjs
- git diff --check

### DIFF
--- a/.github/workflows/publish-cloudflare.yml
+++ b/.github/workflows/publish-cloudflare.yml
@@ -5,6 +5,15 @@ on:
     branches:
       - main
   workflow_dispatch:
+    inputs:
+      publish_mode:
+        description: "Use publish for real R2/KV/Worker deployment; use dry-run for validation only."
+        required: true
+        default: publish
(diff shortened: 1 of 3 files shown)


ghost (approved):
Reviewed the fail-closed publish workflow change and validation. Approving to merge.

PR #221, opened 2026-06-10
fix(rpc): block full block reads in proxy

JSONbored:
### Motivation
- Prevent the unauthenticated read-only RPC proxy from being used to fetch full block bodies (expensive/large responses) by removing `chain_getBlock` from the safe-method allowlist.

### Description
- Remove `chain_getBlock` from `SAFE_RPC_METHODS` in `workers/api.mjs` and add a regression in `tests/worker-runtime.test.mjs` that asserts `chain_getBlock` is rejected with `403 rpc_method_blocked` before any upstream selection or forwarding.

### Testing
- Ran `npm test -- --run tests/worker-runtime.test.mjs`, which shows some unrelated fixture/archive-dependent API tests failing with `404` in this environment but confirms the added RPC regression and proxy-related checks behave as expected.
- Ran `npm test -- --run tests/rpc-endpoint-selection.test.mjs` which passed all cases, and ran `npm test -- --run tests/worker-runtime.test.mjs -t "keeps RPC proxy disabled"` which passed; `npm run lint` completed without errors.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a28f0bf2ae0832a94873f622ef615fa)

### DIFF
--- a/tests/worker-runtime.test.mjs
+++ b/tests/worker-runtime.test.mjs
@@ -694,6 +694,27 @@ describe("Worker runtime", () => {
     assert.equal(blocked.status, 403);
     assert.equal((await blocked.json()).error.code, "rpc_method_blocked");
 
+    const fullBlockRead = await handleRequest(
+      new Request("https://metagraph.sh/rpc/v1/finney", {
+        method: "POST",
+        body: JSON.stringify({
+          jsonrpc: "2.0",
+          id: 2,
+          method: "chain_getBlock",
+          params: [
+            "0x0000000000000000000000000000000000000000000000000000000000000000",
+          ],
+        }),
+      }),
+      { ...env, METAGRAPH_ENABLE_RPC_PROXY: "true" },
+      {},
+    );
+    assert.equal(fullBlockRead.status, 403);
+    assert.equal(
+      (await fullBlockRead.json()).error.code,
+      "rpc_method_blocked",
+    );
+
     const tooLargeByHeader = await handleRequest(
       new Request("https://metagraph.sh/rpc/v1/finney", {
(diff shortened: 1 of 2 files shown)


JSONbored:
Closing as **not viable** after review (per maintainer triage of the codex security PRs).

**Why this doesn't add security.** The read-only proxy intentionally exposes only the exact methods the upstream public Subtensor endpoints already serve unauthenticated. `chain_getBlock` is publicly callable on `entrypoint-finney.opentensor.ai`, `lite.chain.opentensor.ai`, `bittensor-public.nodies.app`, etc. right now — anyone wanting full block bodies just calls the upstream directly. Blocking it at our proxy reduces the upstream's exposure by exactly zero; it only removes a legitimate read method from *our* users.

**The cost concern is already mitigated:**
- Block-pinned `chain_getBlock(0x<hash>)` is cached in `caches.default` for 1h (added in the LB caching work, #226), so the common repeated-read pattern doesn't even reach upstream.
- Request bodies are capped at 64KB (`MAX_RPC_BODY_BYTES`).
- The Worker is rate-limited (100/60s in-Worker + the Cloudflare WAF rate-limit rule).
- Subtensor block bodies are bounded (tens of KB), not unbounded.

`chain_getBlock` is a standard, first-class read method that builders need (block + extrinsics is a core operation), and the LB work deliberately treats it as a cacheable block-pinned read. Removing it would be a functionality regression for no real protection.

If we later want to bound proxy egress defensively, the right lever is a generic **response-size cap across all methods** (not singling out one method) — noted for future hardening, not needed now given the rate limit + caching + 64KB request cap.

PR #222, opened 2026-06-10
fix(rpc): reject /wss route cleanly + never 500 on upstream errors

JSONbored:
The RPC proxy is **live and working** — `POST /rpc/v1/finney system_health` returns real chain data (`{"peers":48,"isSyncing":false}`) load-balanced via the onfinality HTTP endpoint, the allowlist blocks unsafe methods (403), and the WAF blocks non-POST (403).

**But `POST /rpc/v1/finney/wss` returned 500.** The proxy makes an HTTP JSON-RPC POST, and the `/wss` route points at WebSocket-only (`wss://`) endpoints that can't be HTTP-POSTed — so `fetch()` threw and the unhandled error surfaced as a 500.

## Fix
- **Reject `/wss`** with a clean **`400 rpc_websocket_unsupported`** ("POST to `/rpc/v1/finney` for HTTP JSON-RPC, or use a public WSS endpoint directly"); the proxy now always uses the `finney-rpc` (HTTP) pool.
- **Wrap the upstream fetch in try/catch → `502 rpc_upstream_error`**, so an unreachable/erroring upstream can never leak a 500.
- **Guard:** refuse to proxy to a non-`https` upstream (503) — defense in depth.

## Tests
- `worker:test` + `tests/worker-runtime.test.mjs`: `/wss → 400 rpc_websocket_unsupported` (the latter's fixture previously mocked the wss fetch, hiding the real failure — now corrected).
- `smoke-live-api.mjs`: asserts `/wss → 400` against the live site.
- Full suite **148** ✅ · `worker:deploy:dry-run` ✅ · eslint · prettier ✅

On merge, Cloudflare Builds redeploys; `/rpc/v1/finney` keeps working, `/wss` returns a clean 400.

### DIFF
--- a/scripts/smoke-live-api.mjs
+++ b/scripts/smoke-live-api.mjs
@@ -127,6 +127,31 @@ assert.notEqual(
   "enabled RPC proxy must not report rpc_proxy_disabled",
 );
 
+// The /wss route is WebSocket-only and cannot be HTTP-proxied; it must return a
+// clean client error, never a 500.
+const wssRpcProxy = await fetchJson(`${baseUrl}/rpc/v1/finney/wss`, {
+  method: "POST",
+  headers: {
+    "content-type": "application/json",
+  },
+  body: JSON.stringify({
+    jsonrpc: "2.0",
+    id: 1,
+    method: "system_health",
+    params: [],
+  }),
+});
+assert.equal(
+  wssRpcProxy.status,
+  400,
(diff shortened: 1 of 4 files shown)


JSONbored:
Superseded by #224, which includes this /wss fix plus failover/retry. Closing in favor of the single resilient-RPC PR.

PR #235, opened 2026-06-10
fix(rpc): preserve request ids on cache hits

JSONbored:
### Motivation
- Prevent cross-client JSON-RPC envelope replay where the shared cache stored full upstream envelopes (including caller-supplied `id`) and returned them verbatim to later callers.
- Ensure cached RPC results are reusable without leaking or replaying per-request envelope fields such as `id`.

### Description
- On cache hit, parse the cached entry as a payload and return a freshly built JSON-RPC envelope using the current request's `id` via a new helper `rpcResultEnvelope(requestBody, result)` instead of returning the stored envelope verbatim.
- On cache store, persist only the cacheable `result` payload (JSON `{ result: ... }`) rather than the full upstream response envelope to avoid saving caller-controlled fields.
- Add a regression test `cache hits preserve the current request id` to `tests/rpc-cache.test.mjs` that primes the cache with an attacker id and verifies a later victim request receives the victim's id on a cache hit.
- Minor lint/format adjustments to satisfy eslint/prettier for the modified files.

### Testing
- Ran the focused suite: `npx vitest run tests/rpc-cache.test.mjs --reporter=dot`, and the updated RPC-cache tests passed (8/8).
- Ran lint with `npm run lint`, which passed for the modified files after adjustments.
- Ran `npm test` and `npm run format:check`; `npm test` shows unrelated repository-wide failures due to missing generated `public/*` artifacts in the checkout, and `format:check` warns about a pre-existing formatting issue in `docs/adr/0001-r2-only-data-artifacts.md` (these failures are not caused by the change and are environmental/pre-existing).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a29b9b77178832aa21420aec7d351ec)

### DIFF
--- a/tests/rpc-cache.test.mjs
+++ b/tests/rpc-cache.test.mjs
@@ -81,10 +81,10 @@ describe("RPC response cache flow", () => {
       },
     },
   };
-  const reqFor = (method, params) =>
+  const reqFor = (method, params, id = 1) =>
     new Request("https://metagraph.sh/rpc/v1/finney", {
       method: "POST",
-      body: JSON.stringify({ jsonrpc: "2.0", id: 1, method, params }),
+      body: JSON.stringify({ jsonrpc: "2.0", id, method, params }),
     });
 
   function makeCache() {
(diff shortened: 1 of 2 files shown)


JSONbored:
Integrated into #236, which re-implements this fix on top of #230 (the bounded-streaming cache-store that landed after this PR was opened, conflicting with its cache-store change). #236 keeps this PR's exact approach — store only `result`, rebuild the envelope with the current request's `id` via `rpcResultEnvelope` — plus your regression test (prime with an attacker id, assert the victim gets its own id on the hit). Thanks for catching the cross-caller id replay.

PR #500, opened 2026-06-13
docs: harden apex discovery proxy snippet

JSONbored:
### Motivation
- The Cloudflare Snippet in the apex runbook constructed an upstream URL from the request pathname and proxied the original `Request`, which can forward cookies/Authorization and allow an open-proxy via path normalization differences. 
- The change aims to remove credential-forwarding and reduce path-normalization attack surface while keeping the documented behavior of serving discovery surfaces under the apex. 

### Description
- Updated `docs/apex-agent-discovery.md` snippet to reject non-`GET`/`HEAD` methods and return `405 Method Not Allowed` for others. 
- Replaced `new URL(url.pathname + url.search, "https://api.metagraph.sh")` with a fixed `https://api.metagraph.sh` origin and explicit `upstream.pathname = url.pathname` and `upstream.search = url.search`. 
- Stop forwarding the original `Request` object and instead forward a minimal allowlist of non-credential headers (`accept`, `accept-language`, `if-none-match`, `if-modified-since`) and the request method, and use `redirect: "manual"` when calling `fetch`. 
- Tightened the Snippet Rule expression to only match discovery paths for `GET`/`HEAD` requests. 

### Testing
- Ran `npm run validate:docs` and the documentation contract validation passed. 
- Formatted and checked style with `npx prettier --check docs/apex-agent-discovery.md` (ran `prettier --write` as needed) and the file now passes Prettier. 
- Ran `git diff --check` and no diff errors were reported.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a2d0011df98832a8480cd87ae7e33c6)

### DIFF
--- a/docs/apex-agent-discovery.md
+++ b/docs/apex-agent-discovery.md
@@ -26,12 +26,34 @@ scanner that does not follow cross-host redirects still sees it):
 // Snippet: apex-discovery-proxy
 export default {
   async fetch(request) {
+    if (request.method !== "GET" && request.method !== "HEAD") {
+      return new Response("Method Not Allowed", {
+        status: 405,
+        headers: { allow: "GET, HEAD" },
+      });
+    }
+
     const url = new URL(request.url);
-    const upstream = new URL(
-      url.pathname + url.search,
-      "https://api.metagraph.sh",
-    );
-    const resp = await fetch(upstream, request);
+    const upstream = new URL("https://api.metagraph.sh");
+    upstream.pathname = url.pathname;
+    upstream.search = url.search;
+
+    const upstreamHeaders = new Headers();
+    for (const name of [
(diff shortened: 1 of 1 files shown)


JSONbored:
Superseded by #504. This PR hardened the `apex-discovery-proxy` Cloudflare Snippet (removing credential-forwarding / open-proxy risk in the example `fetch(upstream, request)`). #504 replaced that entire approach — the apex's machine-discovery paths are now **routed to the same backend worker** that serves `api.metagraph.sh` rather than proxied/duplicated, so the proxy snippet no longer exists in the runbook and there is no credential-forwarding surface left to harden. Closing as obsolete; the security intent is fully mooted by the worker-routes architecture.

PR #514, opened 2026-06-13
fix(discovery): canonicalize discovery Link headers

JSONbored:
### Motivation
- Apex-served discovery responses exposed origin-relative RFC 8288 `Link` headers (e.g. `</metagraph/openapi.json>`) that resolve against `metagraph.sh` while the JSON api-catalog body points at the canonical API host, causing clients to resolve service descriptions to an unrouted apex path.
- The header/body mismatch is a discovery/integrity bug when the apex routes `/.well-known/*` to the backend worker and should be fixed by advertising canonical `api.metagraph.sh` URLs in the HTTP `Link` header.

### Description
- Replace origin-relative `DISCOVERY_LINK_HEADER` entries in `workers/api.mjs` with absolute canonical URLs built from `PRIMARY_DOMAIN` via a new `DISCOVERY_LINK_BASE` so the HTTP `Link` header always points at `https://api.metagraph.sh`.
- Keep the existing homepage HTML and link elements unchanged while ensuring the worker-owned discovery responses return consistent canonical Link headers.
- Update the runtime test in `tests/worker-runtime.test.mjs` to assert the HTTP `Link` header matches the canonical linkset body and does not contain the old origin-relative OpenAPI reference.

### Testing
- Ran the focused test for the change with `npm test -- tests/worker-runtime.test.mjs -t "api-catalog discovery links"`, which passed. 
- Ran `npx eslint workers/api.mjs tests/worker-runtime.test.mjs` and `git diff --check`, both of which passed for the modified files. 
- Ran the full worker runtime tests with `npm test -- tests/worker-runtime.test.mjs`, where the targeted discovery test passed but the full test run reported unrelated existing failures (several artifact-backed routes returning `404`), which are pre-existing and not caused by this change. 
- Ran repository lint `npm run lint -- --quiet`, which is failing on an unrelated `no-unused-vars` error for `mkdirSync` in `tests/artifacts.test.mjs` and is outside the scope of this fix.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a2d169a75b8832a824d030f66e8ea86)

### DIFF
--- a/tests/worker-runtime.test.mjs
+++ b/tests/worker-runtime.test.mjs
@@ -1311,9 +1311,9 @@ describe("Agent discovery surfaces", () => {
     assert.equal(context.status[0].href, "https://api.metagraph.sh/health");
   });
 
-  test("api-catalog hrefs are canonical (api.metagraph.sh), not the request host", async () => {
-    // The apex (metagraph.sh) routes /.well-known/* to this worker too, so the
-    // catalog must reference the real API host regardless of which host served it.
+  test("api-catalog discovery links are canonical (api.metagraph.sh), not the request host", async () => {
+    // The apex (metagraph.sh) routes /.well-known/* to this worker too, so both
+    // the linkset body and HTTP Link header must reference the real API host.
     const response = await handleRequest(
       new Request("https://metagraph.sh/.well-known/api-catalog"),
       {},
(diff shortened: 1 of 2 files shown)


JSONbored:
Superseded: this PR's insight (the discovery Link header should use absolute canonical api.metagraph.sh refs, since the header is also served on the apex where relative refs mis-resolve) is correct and valuable — but it was built on a pre-#516 tree (older 5-member relation set, before agent.md/health were added + the spurious llms.txt describedby removed), so it can't merge cleanly. Reapplied on current main with the full 6-member relation set in a fresh PR. Thanks codex.

PR #607, opened 2026-06-14
fix(publish): restore refreshed native snapshot

JSONbored:
### Motivation
- Prevent the publish job's `npm run validate` from failing when artifacts were built from a refreshed `registry/native/finney-subnets.json` that was not restored into the publish checkout, causing timestamp/subnet mismatches between artifacts and checked-out inputs.

### Description
- Stage `registry/native/finney-subnets.json` alongside generated artifacts in `.github/workflows/publish-cloudflare.yml` and restore it into `registry/native/finney-subnets.json` in the `publish` job during both the initial restore and the final re-restore so validation runs against the same refreshed native snapshot used to build the artifacts.

### Testing
- Ran `npm run validate:workflows` which passed. 
- Attempted `npm run validate` locally but it could not complete in this checkout because generated artifact `public/metagraph/providers.json` is absent prior to a full build; full end-to-end validation is expected to run in CI where the refresh job produces the staged artifacts.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a2eb2ead850832a9039831cf585b072)

### DIFF
--- a/.github/workflows/publish-cloudflare.yml
+++ b/.github/workflows/publish-cloudflare.yml
@@ -83,6 +83,8 @@ jobs:
           set -euo pipefail
           mkdir -p "$RUNNER_TEMP/publish-artifacts"
           cp -R public "$RUNNER_TEMP/publish-artifacts/public"
+          mkdir -p "$RUNNER_TEMP/publish-artifacts/registry/native"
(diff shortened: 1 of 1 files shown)


JSONbored:
Superseded by #606 (merged) — same fix for the #598 native_data_as_of mismatch, but #606 stages/restores the entire registry/native directory (a superset of this single-file approach) and your diff is now stale against main. Closing as duplicate. Thanks!

PR #646, opened 2026-06-15
fix(rpc): bound usage network telemetry

JSONbored:
### Motivation
- Prevent untrusted high-cardinality path segments under `/rpc/v1/{network}` from being persisted and returned by `/api/v1/rpc/usage`, which can be abused to exhaust D1/Worker resources.
- Preserve existing proxy behaviour (including the legacy `/rpc/v1/wss` error path) while hardening telemetry and DB invariants.

### Description
- Add an allowlist for RPC proxy networks via `RPC_PROXY_NETWORKS` and reject unsupported `/rpc/v1/{network}` paths early in `handleRpcProxyRequest` (returns `rpc_network_unsupported`).
- Preserve legacy WebSocket rejection by keeping `/rpc/v1/.../wss` handling returning `rpc_websocket_unsupported`.
- Bound the public usage aggregation query in `handleRpcUsage` to `network IN ('finney')` and add `LIMIT 10` to cap returned rows.
- Add schema-level enforcement by adding a `CHECK (network IN ('finney'))` to `migrations/0004_rpc_proxy_usage.sql` and provide migration `migrations/0005_rpc_proxy_usage_network_check.sql` that removes unsupported rows and rebuilds the table with the CHECK.
- Add regression tests in `tests/rpc-usage.test.mjs` that assert the network-query is filtered/capped and that unsupported `/rpc/v1/{network}` paths are rejected before any telemetry write.

### Testing
- Ran `npm test -- tests/rpc-usage.test.mjs`, and the test file passed (all tests green).
- Ran linter with `npm run lint -- workers/api.mjs tests/rpc-usage.test.mjs`, which completed without errors.
- A broader run including `tests/worker-runtime.test.mjs` surfaced existing fixture-dependent 404 failures in this environment unrelated to the telemetry fix; those failures were present during iteration and are not caused by the network-allowlist change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a2f96a354d0832a891cd6b909b2ee84)

### DIFF
--- a/migrations/0004_rpc_proxy_usage.sql
+++ b/migrations/0004_rpc_proxy_usage.sql
@@ -13,7 +13,7 @@
 CREATE TABLE IF NOT EXISTS rpc_proxy_events (
   id           INTEGER PRIMARY KEY AUTOINCREMENT,
   observed_at  INTEGER NOT NULL,            -- epoch ms the proxy request started
-  network      TEXT    NOT NULL,            -- finney | test (path segment)
+  network      TEXT    NOT NULL CHECK (network IN ('finney')), -- supported RPC proxy network
   endpoint_id  TEXT,                        -- pool endpoint that served (NULL: cache hit / none eligible)
   provider     TEXT,                        -- served endpoint's provider label
   ok           INTEGER NOT NULL,            -- 1 if the proxy routed to a responding upstream (or cache hit)
--- a/migrations/0005_rpc_proxy_usage_network_check.sql
+++ b/migrations/0005_rpc_proxy_usage_network_check.sql
@@ -1,0 +1,55 @@
+-- Harden rpc_proxy_events against high-cardinality network labels.
+--
+-- 0004 originally documented the proxy network as a small enum but did not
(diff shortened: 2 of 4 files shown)


JSONbored:
Closing as **superseded by the multi-network RPC proxy routing** already on main (the testnet proxy work after this branch was cut).

This PR's goal — preventing unbounded/high-cardinality `network` labels in `rpc_proxy_events` — is already enforced: `handleRpcProxyRequest` validates the path segment against the fixed allowlist `RPC_PROXY_POOLS = { finney: "finney-rpc", test: "test-rpc" }` and returns `rpc_network_unsupported` (404) for anything else **before** `recordRpcUsage` runs. Only those two fixed keys can ever be written, so the cardinality is already bounded at the application layer (the new test here passes against main's logic too).

Merging as-is would **regress testnet**: the `AND network IN ('finney')` usage-query filter and the `CHECK (network IN ('finney'))` migration both predate testnet support and would drop/reject `network = 'test'` telemetry. If we later want a DB-level CHECK as defense-in-depth, it should enumerate the real supported set (`finney`, `test`) — tracked separately if desired.

PR #649, opened 2026-06-15
feat(testnet): surface discord contact on the native subnet detail

JSONbored:
Small completeness fix from the testnet analysis: the chain-identity discord handle was on the testnet subnet **index** but missing from the per-subnet **detail**, so the testnet profile dropped a contact the list showed. Carry it on the detail too (sanitized via `nativeContactHandle`, like the index). Schema already allows `discord`/`discord_url` — no contract change. Testnet detail is R2-only; reaches R2 on the next data publish.


JSONbored:
Closing — based on a misread. Verified the **mainnet** subnet detail also omits discord (it's on the index by design: contact fields are index-only across both networks, per the buildIndexEntry comment). Adding it to the testnet detail would break cross-network consistency, not fix a gap. Testnet identity is already fully + consistently surfaced.

PR #763, opened 2026-06-15
fix(health): validate live RPC block heights

JSONbored:
### Motivation

- Prevent untrusted or malformed `latest_block` values from live probes from biasing RPC proxy routing by requiring safe integers and removing implausible outliers before they are persisted.

### Description

- Add `safeRpcBlockNumber()` to accept only positive `Number.isSafeInteger` block heights and return `null` otherwise.
- Add `rpcBlockMedianFloor()` and `sanitizeRpcLatestBlocks()` to null out RPC probe `latest_block` values that are implausibly higher than the median plus a tolerance (`RPC_BLOCK_PLAUSIBILITY_TOLERANCE` = 10).
- Use `safeRpcBlockNumber()` when building each probe row and run `sanitizeRpcLatestBlocks()` over the probed set before writing D1/KV snapshots so the live `health:rpc-pool` cannot be biased by single high outliers.
- Add a regression test `rejects unsafe or implausibly high live RPC block heights` in `tests/health-prober.test.mjs` that verifies honest heights are retained while forged/unsafe high values are nulled.

### Testing

- Ran the targeted test file with `npm test -- --run tests/health-prober.test.mjs`, which completed successfully (1 test file, 62 tests passed). 
- Ran linter with `npm run lint`, which completed without errors.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a307feba960832abcee8345bc4c7e16)

### DIFF
--- a/src/health-prober.mjs
+++ b/src/health-prober.mjs
@@ -31,8 +31,37 @@ const RPC_KINDS = new Set(["subtensor-rpc", "subtensor-wss", "archive"]);
 const DNS_JSON_ENDPOINT = "https://cloudflare-dns.com/dns-query";
 const DNS_RECORD_TYPES = ["A", "AAAA"];
 const DNS_TIMEOUT_MS = 4000;
+const RPC_BLOCK_PLAUSIBILITY_TOLERANCE = 10;
 
 const iso = (ms) => (Number.isFinite(ms) ? new Date(ms).toISOString() : null);
+
+function safeRpcBlockNumber(value) {
+  if (value == null) return null;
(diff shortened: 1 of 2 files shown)


JSONbored:
Closing as a duplicate of #761, which is already merged to main. Both propose the identical fix for the RPC block-height routing vulnerability — `sanitizeRpcLatestBlocks` (rejecting non-safe-integer heights and implausible outliers >median+10). That code + its test are already on `main` (`src/health-prober.mjs`), so this re-proposal is redundant. No action needed.

PR #953, opened 2026-06-16
perf(ci): split test phase into parallel readers + serial builders

JSONbored:
## What

Splits the CI `Test` step (which was **~60% of the `validate` job's wall-clock**) into two phases:

- **`test:readers`** — every suite *except* `artifacts.test.mjs`, run with `--fileParallelism` (+ `--coverage`).
- **`test:builders`** — `artifacts.test.mjs` alone, serial.

`vitest.config.mjs` stays `fileParallelism: false` so the all-files entrypoints (`npm test`, `npm run test:coverage`) remain race-proof; only CI splits for speed. New `npm run test:split` mirrors the CI flow locally.

## Why this is the only lever

The `Test` step is **64–71s** of the ~120s job. Profiling showed it's dominated by **one file**: `artifacts.test.mjs` runs the real build pipeline **4×** to verify reproducibility + tamper-detection, and it *must* build into the shared `public/metagraph` tree to compare against the committed artifacts (`build-artifacts.mjs` hard-codes its output root — no temp redirect). That's why the whole suite was serialized. Everything *else* is read-only and safe to parallelize.

## Measured (local)

| | Time |
|---|---|
| Baseline (full serial) | 41s |
| `test:readers` (parallel) | **5.2s** |
| `test:builders` (serial floor) | 25s |
| **Split total** | **~31s (−24%)** |

Scaling to CI (test step ~70s): **~52s expected, whole job ~120s → ~102s (~15%)**. The PR's own run will show the real number.

## Safety

- **Flakiness: 5/5 green** parallel-readers runs (the read-only suites don't race each other — the other "exec" tests write only to temp dirs, verified).
- **Separate processes** ⇒ the builder never overlaps a reader (no race possible).
- **Coverage unchanged:** `test:readers` produces `./coverage/lcov.info` at **98.2% lines / 90.6% branches across all 27 source files** — identical to the full run. `test:builders` adds **zero** unique line coverage (it covers code via subprocess execs v8 can't instrument), so it needs no upload. The existing Codecov upload step is untouched.
- **Documented invariant:** `artifacts.test.mjs` is the only shared-tree mutator; a future suite that does the same must move to `test:builders`. Noted in `vitest.config.mjs` + the workflow.

Follow-up to the CI/Codecov work (#917, #943).


JSONbored:
Superseded: the in-runner test-file-split only saved ~3-7s (CPU-core-limited on the CI runner, and capped by the 42s artifact-build floor). The real lever is a JOB-level split — running the test phase and the build+validations on two parallel runners (wall-clock = max, not sum, ~28% off). Replacing with that approach.

PR #1243, opened 2026-06-20
Fix review queue timestamp drift during refresh validation

JSONbored:
### Motivation
- The provenance review queue document used `buildTimestamp()` for `generated_at`, making its output depend on `METAGRAPH_BUILD_TIMESTAMP` and causing `validate` to fail when the pipeline wrote the file with a refresh timestamp but recomputed it without one.
- The goal is to make the review queue deterministic so `validate`'s stable equality check does not fail due to a timestamp-only drift.

### Description
- Introduce a fixed deterministic marker `REVIEW_QUEUE_GENERATED_AT = "1970-01-01T00:00:00.000Z"` and use it for `generated_at` in `buildProvenanceReviewQueue()` in `scripts/lib.mjs` instead of `buildTimestamp()`.
- Add a regression test `keeps generated_at deterministic when refresh timestamps are set` to `tests/provenance-elevation.test.mjs` that sets `METAGRAPH_BUILD_TIMESTAMP` and asserts the queue's `generated_at` remains the deterministic epoch marker.
- The change is minimal and preserves other uses of `buildTimestamp()` elsewhere.

### Testing
- Ran `npm test -- tests/provenance-elevation.test.mjs`, which executed the file's test suite and passed (9 tests passed).
- Executed `METAGRAPH_BUILD_TIMESTAMP=2025-01-02T03:04:05.000Z npm run review:queue`, which completed and wrote the review queue without introducing diffs beyond the intended code changes.
- Attempted a full `npm run validate`, which could not complete in this checkout due to a missing generated artifact (`public/metagraph/providers.json`), so end-to-end validation was blocked in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a36af6ffaf8832cabb900e966587b41)

### DIFF
--- a/scripts/lib.mjs
+++ b/scripts/lib.mjs
@@ -1797,6 +1797,7 @@ export function computeProvenanceElevations({
 // Deterministic (generated_at is the fixed build placeholder) so the committed
 // queue is drift-checked by validate.mjs. Pure — takes the already-loaded inputs.
 const TOP_TRUST_LEVELS = new Set(["maintainer-reviewed", "adapter-backed"]);
+const REVIEW_QUEUE_GENERATED_AT = "1970-01-01T00:00:00.000Z";
 export function buildProvenanceReviewQueue({
   candidates = [],
   nativeSubnets = [],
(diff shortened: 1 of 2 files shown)


JSONbored:
Superseded by #1239, which removed the provenance review-queue drift gate from `npm run validate` entirely. The `generated_at` determinism this PR added is no longer needed — no gate consumes it (the only remaining caller is the on-demand `npm run review:queue` tool). Closing; #1245 (a duplicate of this) is closed for the same reason.

PR #1269, opened 2026-06-21
test(economics): extract + unit-test the live-economics content floor

JSONbored:
The content floor protecting the live KV economics tier from a partial chain-fetch lived inline in refresh-economics.mjs (a side-effects-on-import script → untestable). Extracted the pure `shouldPublishEconomics` into `scripts/economics-floor.mjs` + 6 boundary tests (0 rows / missing captured_at / just-below vs at the 50% ratio / expectedCount=0). Behavior-preserving. Suite 1258 green.


ghost:
**metagraphed-submission-gate** — ⏳ Waiting for CI to finish before the advisory review (8 check(s) still running).

PR #1296, opened 2026-06-21
fix(validate): warn on on-chain name collisions

JSONbored:
### Motivation

- Avoid availability regressions where operator-controlled on-chain `subnet_name` values can cause the registry publish validation to hard-fail by treating them as authoritative mismatches. 

### Description

- Replace the hard assertion in `scripts/validate.mjs` that failed on cross-netuid on-chain name matches with a `console.warn` so collisions are reported but do not block publishing. 
- Update the guard's comment to explain that on-chain identity names are operator-controlled and should not hard-fail validation. 
- Add a regression test in `tests/artifacts.test.mjs` that simulates an unrelated subnet setting its on-chain name to `Templar` and asserts a warning is emitted and that validation does not return a failure exit code. 
- Add the necessary test import change (`spawnSync`) to capture the validator process output. 

### Testing

- Ran `npm run validate` after the change and it completed successfully (validated native and curated overlays). 
- Ran `npm run build:artifacts` which completed and produced the expected artifacts. 
- Ran the targeted test `npm test -- tests/artifacts.test.mjs -t "cross-netuid"` and it passed, confirming the warning is emitted and does not mark validation as failed. 
- A full run of `npm test -- tests/artifacts.test.mjs` initially failed before regenerating committed artifacts due to missing/long-running artifact-related test conditions and timeouts unrelated to the new guard change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3788dea408832c8e2021e30aedc0a0)

### DIFF
--- a/scripts/validate.mjs
+++ b/scripts/validate.mjs
@@ -1553,14 +1553,15 @@ for (const subnet of subnets) {
   }
 }
 
-// Identity guardrail (the "Nodexo" class): a curated overlay's name must not match
-// the ON-CHAIN identity of a DIFFERENT netuid than the one it is keyed to. That is
-// the signature of a mis-keyed overlay — e.g. "Nodexo" sat at netuid 27 while the
-// chain said 27="Team TBC" and 106="Nodexo", and "colosseum" sat at a netuid the
-// chain had re-registered to "ChronoLLM". A name that matches its OWN netuid's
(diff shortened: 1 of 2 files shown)


ghost:
**metagraphed-submission-gate** — ⏳ Waiting for CI to finish before the advisory review (1 check(s) still running).

ghost (changes requested):

Issue #1303, opened 2026-06-21
[FOUNDATION] Snapshot the per-neuron metagraph into a queryable live tier


JSONbored:
# [FOUNDATION] Snapshot the per-neuron metagraph into a queryable live tier

**Repo:** `JSONbored/metagraphed` (backend). Part of #1302.

## Context

Today, the native chain snapshot (`registry/native/finney-subnets.json`, built by `scripts/fetch-native-subnets.py` using `SubtensorApi.metagraphs.get_all_metagraphs_info()`) captures **per-subnet** metadata only: name, symbol, block, participant_count, economics (validator counts, stake totals, alpha price). This covers the 129 active Finney subnets at a high level.

To enable per-neuron intelligence (validating specific hotkeys, tracing stake flow, building neuron-centric ledgers, analyzing validator permit distributions), we need to extend the snapshot to capture **per-subnet NEURONS** via the Bittensor SDK's metagraph data. This includes per-UID:
- **Identity:** hotkey, coldkey, rank, trust, consensus, incentive, dividends, emission
- **Registration:** validator_permit, active, last_update
- **Network:** axon host/port/protocol

This foundation is a **blocking prerequisite** for C2–C5 (validator dashboards, stake-tracing, permit analysis, neuron-profile queries).

The current snapshot is baked into a **static deterministic artifact** (`finney-subnets.json`). Per-neuron data is high-volume (129 subnets × ~256 max UIDs per subnet = 33k records), changes at every block, and must **never block determinism**. It must follow the **live-tier pattern** (like `/api/v1/health`, which runs on a 2-minute cron and serves from KV/D1, not from committed artifacts).

### 1. SDK Call & Fetch Strategy
- Extend `scripts/fetch-native-subnets.py` OR create a new `scripts/fetch-native-metagraphs.py` to call `SubtensorApi.metagraphs.get_all_metagraphs(include_stake=True)` for **each active netuid** after the native subnet snapshot fetch completes.
- Capture per-UID fields: `hotkey`, `coldkey`, `rank`, `trust`, `consensus`, `incentive`, `dividends`, `emission`, `validator_permit`, `active`, `last_update`, and axon info (`axon_host`, `axon_port`, `axon_protocol`).
- Gracefully handle missing/sparse data (a subnet with 100 active UIDs out of 256 slots returns 100 records, not 256).
- Coerce Balance objects to float TAO (like `normalize_economics` does today); best-effort null for malformed fields.
[...]

PR #1343, opened 2026-06-21
fix(metagraph): load neurons into D1 via the Worker binding, not the API token (#1303)

JSONbored:
Fixes the failing `refresh-metagraph` cron.

## Problem
The cron's **fetch** step works (Taostats key is fine), but the **D1 load** failed: `CLOUDFLARE_API_TOKEN` has KV + Workers + R2 permissions but **not D1**, so `wrangler d1 execute --remote` hit a Cloudflare **auth error (10000)** on the D1 import API. Granting a token D1 access is a dashboard-only change (not doable via CLI), so this removes the dependency instead.

## Fix
- The CI job now **stages** the snapshot SQL to **R2** (`metagraph/neurons-pending.sql`) using the token's existing R2 permission.
- The Worker's `handleScheduled` runs `loadStagedNeurons`: read the staged object → load it into D1 through the **`METAGRAPH_HEALTH_DB` binding** (needs no API-token permission) → delete the object so it loads exactly once.
- **Isolated + fail-safe**: wrapped so a load error never affects the prober/prune; a failed load just leaves the previous snapshot in place.

## Verified
- `tests/load-staged-neurons.test.mjs` — batched load + delete, no-op when nothing staged, deletes an empty object, safe without bindings.
- `npm test` 1345 pass; eslint, prettier, `validate:workflows`, build (0 drift) green.
- `wrangler r2 object put` validated locally; a current snapshot is already staged in R2, so the first scheduled tick after deploy will load it end-to-end (I'll confirm).

### DIFF
--- a/.github/workflows/refresh-metagraph.yml
+++ b/.github/workflows/refresh-metagraph.yml
@@ -57,9 +57,12 @@ jobs:
         env:
           TAOSTATS_API_KEY: ${{ secrets.TAOSTATS_API_KEY }}
 
-      - name: Load neurons into D1
+      # Stage the snapshot to R2 (the token's existing R2 permission), then the
+      # Worker's scheduled handler loads it into D1 through its METAGRAPH_HEALTH_DB
+      # binding (loadStagedNeurons) — no API-token D1 permission required.
+      - name: Stage neurons to R2 for the Worker to load into D1
         if: steps.gate.outputs.ready == 'true'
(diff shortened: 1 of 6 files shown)


ghost:
**metagraphed-submission-gate** — ⏳ Waiting for CI to finish before the advisory review (1 check(s) still running).

ghost (changes requested):

PR #1370, opened 2026-06-22
Authenticate and bound R2-staged neuron loads before D1 writes

JSONbored:
### Motivation
- The scheduled Worker previously trusted any JSON object written to the fixed R2 key and could therefore be used as a confused-deputy to mutate D1 when an attacker (or leaked token/CI step) could write that R2 key. 
- The change prevents unauthorized or oversized staged payloads from causing unauthorized D1 mutations or availability/cost impacts by adding provenance and bounds checks before any D1 writes.

### Description
- Require an HMAC-signed envelope for staged neuron snapshots and add a CI signing step using a new `scripts/sign-staged-neurons.mjs` helper and the `METAGRAPH_STAGING_SIGNING_KEY` secret, with the workflow calling the script before the R2 `put` step. 
- Add authentication and validation in `loadStagedNeurons` (`workers/api.mjs`): verify `schema_version` and `hmac_sha256` using the Worker `METAGRAPH_STAGING_SIGNING_KEY`, perform a timing-safe HMAC check, and delete the object on failure. 
- Add defensive bounds and schema checks including object size limit, maximum row count, allowed columns, per-string byte limits, finite-number checks, and `netuid`/`uid` range limits, while preserving parameterized batched `INSERT OR REPLACE` statements so values remain bound (no SQL interpolation). 
- Update and extend tests (`tests/load-staged-neurons.test.mjs`) to cover the signed happy path, unsigned/tampered rejection, oversized payload rejection, out-of-range rows, and row-count limits.

### Testing
- Ran the targeted unit tests with `npx vitest run tests/load-staged-neurons.test.mjs`, which passed. 
- Ran linting with `npm run lint`, which succeeded. 
- Manually exercised the signing helper with `METAGRAPH_STAGING_SIGNING_KEY=test node scripts/sign-staged-neurons.mjs <in> <out>` to confirm produced envelope format and HMAC.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a38caac5c68832cac32487e871c5a58)

### DIFF
--- a/.github/workflows/refresh-metagraph.yml
+++ b/.github/workflows/refresh-metagraph.yml
@@ -57,9 +57,15 @@ jobs:
         env:
           TAOSTATS_API_KEY: ${{ secrets.TAOSTATS_API_KEY }}
 
-      # Stage the snapshot to R2 (the token's existing R2 permission), then the
-      # Worker's scheduled handler loads it into D1 through its METAGRAPH_HEALTH_DB
-      # binding (loadStagedNeurons) — no API-token D1 permission required.
+      - name: Sign staged neuron snapshot
+        if: steps.gate.outputs.ready == 'true'
+        run: node scripts/sign-staged-neurons.mjs dist/metagraph-neurons.json
+        env:
+          METAGRAPH_STAGING_SIGNING_KEY: ${{ secrets.METAGRAPH_STAGING_SIGNING_KEY }}
+
+      # Stage the signed snapshot to R2 (the token's existing R2 permission),
+      # then the Worker's scheduled handler authenticates and bounds it before
+      # loading into D1 through its METAGRAPH_HEALTH_DB binding.
       - name: Stage neurons to R2 for the Worker to load into D1
         if: steps.gate.outputs.ready == 'true'
         run: npx wrangler r2 object put metagraphed-artifacts/metagraph/neurons-pending.json --file=dist/metagraph-neurons.json --remote
--- a/scripts/sign-staged-neurons.mjs
+++ b/scripts/sign-staged-neurons.mjs
@@ -1,0 +1,21 @@
+#!/usr/bin/env node
+import { createHmac } from "node:crypto";
+import { readFileSync, writeFileSync } from "node:fs";
+
+const [inputPath, outputPath = inputPath] = process.argv.slice(2);
+const key = process.env.METAGRAPH_STAGING_SIGNING_KEY;
+if (!inputPath || !key) {
+  throw new Error(
+    "usage: METAGRAPH_STAGING_SIGNING_KEY=... node scripts/sign-staged-neurons.mjs <input> [output]",
+  );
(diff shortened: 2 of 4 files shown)


ghost:
🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨🟨

### 🛠️ Maintainer review — Held for maintainer review

Reviewed **4** changed file(s) — two independent AI reviewers, synthesized. 🛠️ **Request changes.** _(reviewers split: request changes / merge)_ Address the suggestions below before merging.

> ℹ️ The reviewers were not unanimous (a change was requested) — held for your review rather than auto-closed.

**Summary**

The PR adds HMAC signing and strict validation for staged neuron snapshots, which is a valuable security improvement. However, there are two concrete defects: (1) the R2 object key used in the workflow (`metagraphed-artifacts/metagraph/neurons-pending.json`) does not match the key the worker expects (`metagraph/neurons-pending.json`), causing the worker to never find the staged file; (2) the newly introduced `MAX_STAGED_NETUID = 1024` limit may reject legitimate subnet netuids that exceed this value, potentially breaking future submissions. These issues must be fixed before the change can be merged.

#### 🛑 Concerns raised — review before merging (reviewers were not unanimous)

- R2 object key mismatch between workflow (`metagraphed-artifacts/metagraph/neurons-pending.json`) and worker (`STAGED_NEURONS_KEY = "metagraph/neurons-pending.json"`).
- `MAX_STAGED_NETUID = 1024` may reject valid netuid values beyond this range.

- Consider extracting the HMAC verification logic into a reusable utility module for consistency.
- The helper `utf8Bytes` could be inlined where used; its separate definition is fine but could be marked as internal.
- Align the R2 object key between the workflow and `workers/api.mjs` (either update the workflow to use `metagraph/neurons-pending.json` or change the constant `STAGED_NEURONS_KEY` accordingly).
- Reevaluate the `MAX_STAGED_NETUID` bound; consider raising it to match the existing allowed range for netuid (e.g., 65535) or remove the limit if unnecessary.
- Add a test case that verifies the worker correctly loads a snapshot when the R2 key matches the workflow path.
- Document the purpose and limits of the new constants in a comment block for future maintainers.
- Consider adding a test for the exact boundary values (e.g., MAX_STAGED_NEURONS_BYTES, MAX_STAGED_NEURON_ROWS) to ensure inclusive/exclusive behavior is correct.
- The script `sign-staged-neurons.mjs` could validate that the input file is valid JSON before parsing to provide a clearer error message.
- In `validStagedNeuronRow`, the check for `typeof value === 'bigint'` is redundant because `JSON.parse` never produces bigints, but it's harmless.
- Ensure the `METAGRAPH_STAGING_SIGNING_KEY` secret is rotated periodically and stored securely (outside scope of this PR).

ghost (dismissed):

Issue #1346, opened 2026-06-22
[Foundation] Chain-direct event poller + event D1 model


JSONbored:
**Repo:** `JSONbored/metagraphed` (backend). Part of #1345.

## Context
The keystone of the developer block explorer: index Bittensor chain **events** ourselves (first-party), not via Taostats. Proven by the official [`opentensor/subtensor-indexer`](https://github.com/opentensor/subtensor-indexer) — Python `substrate-interface` `query("System","Events")` SCALE-decodes events; hotkey/coldkey attribution via the extrinsic signer + the `SubtensorModule.Owner` storage map. Fits our serverless model (~7,200 blocks/day is trivial).

## Deliverables
- `scripts/fetch-events.mjs` (or .py, mirroring `fetch-native-subnets.py`): connect to public finney RPC, read a **cursor** (last finalized block) from KV, fetch finalized blocks since the cursor, decode SubtensorModule events (NeuronRegistered, StakeAdded/Removed, WeightsSet, AxonServed, DelegateAdded, Coldkey/HotkeySwapped, …), normalize to rows keyed by hotkey/coldkey, write JSON.
- D1 migration: `accounts` (PK hotkey — rollup summary), `account_events` (hot ~30–90d window, indexed (hotkey, observed_at) + (netuid, observed_at)), `account_events_daily` (durable daily rollup, kept forever — mirrors `surface_uptime_daily`).
- Worker load: extend the proven `loadStagedNeurons` pattern (R2 JSON → parameterized D1 batch → delete). Cursor advance in KV.
- Rollup + prune in `handleScheduled` (reuse the prune/rollup cron): roll `account_events` → `account_events_daily`, prune the hot window.

## Requirements
- Chain-direct via public RPC ($0); idempotent (re-running a block range is safe — `INSERT OR REPLACE`/`OR IGNORE`); bounded D1 (~500 MB est.); isolated/fail-safe in the cron (never breaks the prober). Filter to entity-relevant events only (~1 MB/day, not ~100 MB/day).

## Acceptance
- Poller decodes real finney blocks (verifiable locally against public RPC, no key); D1 populates; cursor advances; rollups bound the hot table.

PR #1405, opened 2026-06-22
fix(events): cap the staged-event R2→D1 drain + drain overflow progressively (#1346)

JSONbored:
## Problem
`loadStagedEvents(env)` (workers/api.mjs) had **no size or row cap** on its input path. It does a single R2 GET of `events/account-events-pending.json`, parses it, and runs parameterized `INSERT OR IGNORE` via `db.batch()` in groups of 50 statements. The HTTP ingest endpoint already has defense-in-depth caps (`MAX_EVENTS_INGEST_BODY_BYTES` / `MAX_EVENTS_INGEST_ROWS`, workers/config.mjs) — but those guard **only** the HTTP path, not the staged drain.

A stuck or backfilling poller window (`scripts/fetch-events.py`, `refresh-events.yml`) could stage a very large file. At ~500k rows the drain would approach the Worker **1000-subrequest/invocation limit** and serialize a large number of D1 `batch()` writes on the single D1 writer. Design volume is ~1 MB/day, so this is headroom, not a live bug — but the path was unguarded.

## Fix
Caps mirroring the ingest endpoint's intent, sized for a batch file:
- **`MAX_STAGED_EVENTS_BYTES` = 4 MiB** — a parse-safety ceiling checked via `object.size` **before** the body is materialized.
- **`MAX_STAGED_EVENT_ROWS` = 10 000/tick** — bounds D1 writes + subrequests (~1000 INSERTs → ~20 `db.batch()` calls, far under 1000).

Crucially, overflow is **not** dropped:
- **Over the byte cap:** skip + `console.warn`, and **leave the object in place** (do *not* delete). The overlapping poller window re-stages a normal-size file within ~5 min and self-heals; a misconfigured backfill is loud but loses nothing.
- **Over the row cap:** drain up to the cap, then rewrite **only the remainder** to R2 for the next `*/3` tick; delete the object **only once fully drained**.
- **Ordering:** D1 write happens **before** the R2 shrink. A crash between them re-reads the full file next tick and re-inserts idempotently (`INSERT OR IGNORE` on `(block_number, event_index)`), so **rows are never dropped** — not on overflow, not on a partial-batch failure.

## Tests (4 new, alongside the existing `loadStagedEvents` suite)
- row cap → first cap loaded, remainder rewritten to R2, object **not** deleted;
- multi-tick drain of a `>cap` file → every row loaded across ticks, deleted only at the end (no drop);
- over-byte-cap file → never parsed, never deleted, returns `too_large`;
- a mid-batch D1 throw → object left intact (no partial put/delete) so it re-drains next tick.

All 41 tests in the staged-events / ingest / account-events suites pass; prettier clean. Found via a code audit of the ingest caps' coverage.

### DIFF
--- a/tests/load-staged-events.test.mjs
+++ b/tests/load-staged-events.test.mjs
@@ -1,7 +1,11 @@
 import assert from "node:assert/strict";
 import { test } from "vitest";
 import { handleScheduled, loadStagedEvents } from "../workers/api.mjs";
-import { EVENTS_LOAD_CRON } from "../workers/config.mjs";
+import {
+  EVENTS_LOAD_CRON,
+  MAX_STAGED_EVENTS_BYTES,
+  MAX_STAGED_EVENT_ROWS,
+} from "../workers/config.mjs";
 
 function eventRow(block_number, event_index) {
   return {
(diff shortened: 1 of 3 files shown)


ghost:
🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩

Reviewed **3** changed file(s) — two independent AI reviewers, synthesized. ✅ **Safe to merge** — both reviewers found no blocking issues.

**Summary**

The change correctly implements byte‑ and row‑caps for the staged‑event drain, adds corresponding configuration constants, and updates the loadStagedEvents logic to safely handle oversized files and partial drains. No security issues are introduced; the new caps mitigate DoS risks and the code does not expose secrets or allow SSRF/injection. The implementation follows existing patterns, includes comprehensive tests covering the new behavior, and all CI checks pass. Minor style suggestion: guard the Number(object.size) conversion against non‑numeric values.

- Add defensive handling for non‑numeric object.size.
- Update documentation for the new `remaining` field.
- Consider adding a defensive check for non‑numeric object.size (e.g., if (!Number.isFinite(size)) treat as too large) to avoid NaN edge‑cases.
- Document the new return field `remaining` in the function JSDoc/comments for downstream callers.
- Consider adding more detailed comments explaining the rationale behind the byte and row caps.
- Ensure that the constants `MAX_STAGED_EVENTS_BYTES` and `MAX_STAGED_EVENT_ROWS` are documented in the configuration file.
- Add a test case to handle the scenario where the file size is exactly at the byte cap limit.

ghost (approved):

PR #1416, opened 2026-06-22
fix(graphql): memoize fragment validation limits

JSONbored:
### Motivation
- The custom GraphQL depth and complexity validators were recursively expanding named `FragmentSpread`s without memoization or early cutoff, enabling small acyclic fragment DAGs to cause exponential CPU work during validation on the public `/api/v1/graphql` endpoint.

### Description
- Add a per-operation `memo` map to `selectionDepth` to cache computed fragment depths and avoid re-expanding the same fragment across different paths.
- Add a per-operation `memo` map to `selectionComplexity` to cache computed fragment complexities and avoid repeated recomputation of shared fragments.
- Short-circuit traversal once the configured `max` limit is exceeded by returning a value > `max` to stop further recursion early.
- Propagate the new `memo` and `max` parameters through recursive calls and update the `maxDepthRule` and `maxComplexityRule` invocations accordingly.
- Add a regression test (`validation memoizes repeated named fragment spreads`) that constructs a repeated-fragment DAG and asserts the validator returns `COMPLEXITY_LIMIT_EXCEEDED` rather than incurring exponential work.

### Testing
- Ran the targeted GraphQL tests with `npm test -- --run tests/graphql.test.mjs` and they all passed. 
- Ran the linter with `npm run lint` and it succeeded. 
- Ran the full test suite with `npm test`; GraphQL-related tests passed, but the full suite reported unrelated failures (missing generated `public/metagraph` artifacts, an R2 upload timeout, and a DNS/environment-dependent public-safety test).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a39062308cc832cb45df6bf201fdb80)

### DIFF
--- a/src/graphql.mjs
+++ b/src/graphql.mjs
@@ -86,26 +86,35 @@ function buildFragmentMap(documentNode) {
 // complexity 1 and fully bypassing both limits. `visited` guards against
 // fragment cycles: validate() reports those, but our rules run in the same pass
 // and would otherwise recurse forever.
-function selectionDepth(selectionSet, fragments, visited, depth = 0) {
-  let max = depth;
+function selectionDepth(selectionSet, fragments, visited, memo, max) {
+  let deepest = 0;
   for (const sel of selectionSet.selections) {
+    let depth = 0;
     if (sel.kind === "FragmentSpread") {
-      const frag = fragments.get(sel.name.value);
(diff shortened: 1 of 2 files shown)


ghost:
**metagraphed-submission-gate** — ⏳ Waiting for CI to finish before the advisory review (6 check(s) still running).

PR #1459, opened 2026-06-22
fix(neurons): snapshot-replace stale rows — no more phantom neurons

JSONbored:
Audit finding #2 (high, confirmed). The per-UID `neurons` D1 table accumulated **phantom rows**: `loadStagedNeurons` used `INSERT OR REPLACE` with no delete, and the read paths don't filter `captured_at`, so a deregistered/vanished neuron's row persisted forever → ghost metagraphs + inflated `neuron_count` served indefinitely.

**Fix** (snapshot-replace, contained in loadStagedNeurons): after ALL upsert batches succeed, `DELETE FROM neurons WHERE captured_at < <snapshot max captured_at>` → the table holds only the latest full snapshot. On any batch failure: **no prune + keep the staged R2 object** (prior complete snapshot preserved). Removed the old `purgeStaleNeuronRows` (only deleted within `refreshed_netuids`, so a whole vanished subnet kept its rows). Reads unchanged — table now holds only current rows.

**Tests** (tests/load-staged-neurons.test.mjs): invariant (one table-wide prune bound to the cutoff), **regression** (snapshot1 {(1,0),(1,1)} → snapshot2 {(1,0)} via a stateful D1 mock → phantom (1,1) gone), safety (batch failure → no prune, staged object preserved). 36 staged/neuron tests green; worker dry-run passed. No API shape change.

### DIFF
--- a/tests/load-staged-neurons.test.mjs
+++ b/tests/load-staged-neurons.test.mjs
@@ -58,6 +58,7 @@ function signedCoverageEnvelope(
 function mockEnv({
   rows,
   bad = false,
+  failBatch = false,
   getCalls = [],
   deleted = [],
   prepared = [],
@@ -100,6 +101,7 @@ function mockEnv({
         },
         async batch(stmts) {
           batches.push(stmts.length);
+          if (failBatch) throw new Error("simulated D1 batch failure");
         },
       },
     },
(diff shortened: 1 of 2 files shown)


ghost:
🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩

Reviewed **2** changed file(s) — two independent AI reviewers, synthesized. ✅ **Safe to merge** — both reviewers found no blocking issues.

**Summary**

The changes correctly replace the previous per‑netuid stale‑row purge with a snapshot‑replace strategy, add safety handling for batch failures, and update tests to verify the new behavior and regression. No security issues are introduced; all database writes remain parameterized and HMAC‑verified. CI passes and test coverage is adequate.

- The comment block is quite long; could be split for readability.
- Consider adding a test for the legacy envelope path to ensure the new unconditional prune does not break older snapshots.
- Document the snapshotCapturedAt calculation edge‑case (e.g., rows with differing captured_at values) in the code comment.
- Remove the now‑unused import or variable `stagingMeta` if it is no longer referenced after the refactor.
- Consider adding a comment in workers/api.mjs clarifying that snapshotCapturedAt is derived from rows (not the envelope) to defend against potential future envelope tampering, though HMAC validation already covers this.
- In tests/load-staged-neurons.test.mjs, the statefulEnv helper could be extracted to a reusable test utility if similar stateful mocks are needed elsewhere, but its current localization is acceptable.

ghost (approved):

PR #1525, opened 2026-06-23
fix(webhooks): an explicit empty filter allowlist matches nothing, not all

### LINKED ISSUE #1524: webhook eventMatchesFilters treats an explicit empty allowlist as match-all (should match nothing)
### Summary

`eventMatchesFilters` (`src/webhooks.mjs`) guards each facet with `&& filters.X.length > 0`, so an **explicit empty allowlist** (`{ kinds: [] }` or `{ netuids: [] }`) is silently treated as "no filter" and matches **every** event. `normalizeFilters` preserves an empty array (it only rejects invalid entries, never empties), so this is a reachable subscription — and the subscriber gets flooded with all events instead of receiving none.

### Where

```js
export function eventMatchesFilters(event, filters) {
  if (!filters || (filters.netuids === undefined && filters.kinds === undefined)) {
    return true; // no facet present → no restriction (correct)
  }
  if (Array.isArray(filters.kinds) && filters.kinds.length > 0) { // <- empty falls through
    ...
  }
  if (Array.isArray(filters.netuids) && filters.netuids.length > 0) { // <- empty falls through
    ...
  }
  return true;
}
```

`normalizeFilters({ kinds: [] })` returns `{ kinds: [] }` (kept, not dropped). The matcher then skips the `length > 0` block and returns `true`.

### Why it's wrong

An empty allowlist allows **zero** items, so it should match **nothing**. The "no restriction" case is already covered separately (a facet that is `undefined` → early `return true`). Treating an explicit empty array as ALSO "no restriction" is redundant and surprising: a client that builds `kinds` programmatically and ends up with `[]` silently subscribes to the full firehose.

### Reproduction

```js
const event = buildChangeEvent({ changelog: { subnets: { added: [{ netuid: 7 }] } } });
eventMatchesFilters(event, { kinds: [] });            // => true  (should be false)
eventMatchesFilters(event, { netuids: [] });          // => true  (should be false)
eventMatchesFilters(event, { netuids: [7], kinds: [] }); // => true (should be false)
```

### Fix

Drop the `length > 0` guards: a present facet is an allowlist, and an empty one matches nothing (`[].some(...)` is `false` → `return false`). The `undefined`-facet early return still preserves the genuine "no filter → match all" case.

joaovictor712:
## Summary

- `eventMatchesFilters` guarded each facet with `&& filters.X.length > 0`, so an **explicit empty allowlist** (`{ kinds: [] }` / `{ netuids: [] }`) fell through and matched **every** event.
- `normalizeFilters` preserves an empty array (it only rejects invalid entries, never empties), so such a subscription is reachable — and the subscriber gets the full event firehose instead of nothing.

Fixes #1524

## Why it's wrong

An empty allowlist allows **zero** items → it should match **nothing**. The genuine "no restriction" case is already handled separately (a facet that is `undefined` → early `return true`). Treating an explicit `[]` as also "no restriction" is redundant and surprising for a client that builds `kinds`/`netuids` programmatically and ends up with `[]`.

```js
eventMatchesFilters(event, { kinds: [] });             // before: true  → after: false
eventMatchesFilters(event, { netuids: [] });           // before: true  → after: false
eventMatchesFilters(event, { netuids: [7], kinds: [] }); // before: true → after: false
```

## What Changed

- `src/webhooks.mjs` — drop the `length > 0` guards in `eventMatchesFilters`: a present facet is an allowlist, and an empty one matches nothing (`[].some(...)` is `false` → `return false`). The `undefined`-facet early return still preserves "no filter → match all".
- `tests/webhooks-core.test.mjs` — add a case asserting empty allowlists match nothing. Existing tests (`{}`/`undefined` → match all; populated filters) are unchanged.

## Registry Safety

- [x] No OpenAPI/schema surface changes — webhook delivery filtering only.

## Validation

- [x] `npx vitest run tests/webhooks-core.test.mjs` — 60 passed (new case fails on the pre-fix `length > 0` guards).
- [x] `npx prettier --check` + `npx eslint` on the changed files — clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/webhooks.mjs
+++ b/src/webhooks.mjs
@@ -286,17 +286,23 @@ export function buildChangeEvent({ changelog, pointer } = {}) {
 }
 
 export function eventMatchesFilters(event, filters) {
+  // No filters object (or neither facet present) means "no restriction" — match
+  // every event. A PRESENT facet is an allowlist, including an explicit empty
+  // one: `{kinds: []}` allows zero kinds, so it must match NOTHING, not fall
+  // through to match-all. normalizeFilters preserves an empty array, so a
+  // subscriber can create such a filter and would otherwise be flooded with
+  // every event instead of receiving none.
   if (
     !filters ||
     (filters.netuids === undefined && filters.kinds === undefined)
   ) {
     return true;
   }
-  if (Array.isArray(filters.kinds) && filters.kinds.length > 0) {
+  if (Array.isArray(filters.kinds)) {
     const eventKinds = new Set(event?.change_kinds || []);
     if (!filters.kinds.some((kind) => eventKinds.has(kind))) return false;
   }
-  if (Array.isArray(filters.netuids) && filters.netuids.length > 0) {
+  if (Array.isArray(filters.netuids)) {
     const affected = new Set(event?.affected_netuids || []);
     if (!filters.netuids.some((netuid) => affected.has(netuid))) return false;
   }
--- a/tests/webhooks-core.test.mjs
+++ b/tests/webhooks-core.test.mjs
@@ -330,6 +330,19 @@ describe("buildChangeEvent + eventMatchesFilters", () => {
     assert.equal(eventMatchesFilters(event, { netuids: [7] }), true);
     assert.equal(eventMatchesFilters(event, { netuids: [99] }), false);
   });
+
+  test("an explicit empty allowlist matches nothing (not everything)", () => {
+    // normalizeFilters preserves `[]`, so these are reachable subscriptions.
+    // An empty allowlist allows zero items → no event matches.
+    assert.equal(eventMatchesFilters(event, { kinds: [] }), false);
+    assert.equal(eventMatchesFilters(event, { netuids: [] }), false);
+    assert.equal(eventMatchesFilters(event, { netuids: [], kinds: [] }), false);
+    // An empty allowlist on one facet rejects even when the other would match.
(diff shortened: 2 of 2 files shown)


ghost:
> [!NOTE]
> 🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦🟦
>
> ### 💡 Gittensory review — advisory only
>
> `0 files` · `no blockers` · `readiness 0/100`
>
> **💡 Advisory only** — PR closed before full evaluation. No late first comment was created.
>
> | Signal | Result | Evidence |
> |---|---|---|
> | **Code review** | ✅ No blockers | synthesized |
> | Gate result | ⚠️ Skipped | JSONbored/metagraphed#1525 is no longer open. |
>
> <sub>🟩 Safe / merged · 🟦 Advisory · 🟨 Held for review · 🟥 Blocked / closed</sub>
>
> ---
> 💰 **Earn for open-source contributions like this.** [Gittensor](https://gittensor.io) lets GitHub contributors earn for the work they already do — [register to start earning →](https://gittensor.io/miners/repository?name=JSONbored%2Fmetagraphed&tab=miners).
>
> Checked by [Gittensory](https://gittensory.aethereal.dev), a quiet PR intelligence layer for OSS maintainers.

ghost (approved):
Gittensory approves — the gate is satisfied and CI is green.

PR #1638, opened 2026-06-24
fix(api): reject a malformed ?block= on the extrinsics feed instead of silently querying block 0

### LINKED ISSUE #1615: extrinsics feed: a non-numeric ?block= silently queries block 0 instead of being rejected
### What

`handleExtrinsics` (workers/api.mjs) filters by `?block=`:

```js
if (blockParam != null) {
  sql += " WHERE block_number = ?";
  params.push(clampInt(blockParam, 0, 0, Number.MAX_SAFE_INTEGER));
}
```

### Why it is wrong

`validateQueryParams` only validates parameter *names*, not values, so `?block=abc` (or `?block=`, `?block=-5`) passes. Then `clampInt("abc", 0, …)` hits `!Number.isFinite(Number("abc"))` and returns the **default 0**, so the handler runs `WHERE block_number = 0` and returns block 0's extrinsics as if that were the requested filter. A malformed filter should never return confident-looking wrong data.

### Repro

```
GET /api/v1/extrinsics?block=abc
# → 200 with block 0's extrinsics (wrong)
# expected: 400 invalid_query (block must be a non-negative integer)
```

Same for `?block=` (empty) and `?block=-5`.

### Expected

Reject a non-numeric/negative/empty `block` with `400 invalid_query` (via the existing `analyticsQueryError`), consistent with how other malformed analytics queries are handled — never silently substitute block 0.

joaovictor712:
## Summary

- Fixes #1615. `GET /api/v1/extrinsics?block=<n>` silently returned **block 0's**
  extrinsics for any malformed `block` filter (`?block=`, `?block=abc`,
  `?block=-5`, `?block=1.5`). `validateQueryParams` only checks parameter
  *names*, so a bad value flowed into `clampInt(blockParam, 0, …)`, which maps
  non-finite / out-of-range input to its default of `0`. The handler then ran
  `WHERE block_number = 0` and served genesis-block rows as though they matched
  the requested filter — a silently wrong result instead of a `400`.

## What Changed

- `handleExtrinsics` (`workers/api.mjs`) now validates the `block` filter before
  it touches the query: a block number must be a plain non-negative decimal
  integer (`/^\d+$/`) within the safe-integer range. Anything else returns
  `400 invalid_query` (`meta.parameter: "block"`), consistent with the other
  list-route rejections (`analyticsQueryError`).
- `?block=0` remains a valid filter (genesis block); only malformed input is
  rejected.
- Regression tests in `tests/extrinsics.test.mjs`: malformed values
  (`""`, `"abc"`, `"-5"`, `"1.5"`, `"1e3"`, `"0x10"`, `" "`) all `400` and never
  reach the DB; `?block=0` binds `block_number = 0` and returns `200`.

## Registry Safety

      validator-local state.
- [x] Public API/OpenAPI/schema changes are intentional and documented. (Behavior
      change: a malformed `?block=` now returns `400` instead of silently
      filtering to block 0. No schema/contract surface changes.)

## Validation

- [x] `npm run check`
- [x] `npm run worker:test` (`tests/extrinsics.test.mjs` — 21 passing)
- [x] `git diff --check`

## Template Used

- backend-code.md

### DIFF
--- a/tests/extrinsics.test.mjs
+++ b/tests/extrinsics.test.mjs
@@ -331,6 +331,63 @@ test("GET /extrinsics?block=<n> scopes the feed to one block (#1345)", async () 
   assert.ok(/WHERE block_number = \?/.test(boundSql));
 });
 
+test("GET /extrinsics?block rejects malformed values instead of querying block 0 (#1615)", async () => {
+  let boundSql;
+  const env = {
+    METAGRAPH_HEALTH_DB: {
+      prepare(sql) {
+        boundSql = sql;
+        return {
+          bind() {
+            return {
+              async all() {
+                return { results: [] };
+              },
+            };
+          },
+        };
+      },
+    },
+  };
+  for (const bad of ["", "abc", "-5", "1.5", "1e3", "0x10", " "]) {
+    boundSql = null;
+    const res = await handleRequest(
+      req(`/api/v1/extrinsics?block=${encodeURIComponent(bad)}`),
+      env,
+      {},
+    );
+    assert.equal(res.status, 400, `?block=${JSON.stringify(bad)} should 400`);
+    const body = await res.json();
+    assert.equal(body.error.code, "invalid_query");
+    assert.equal(body.meta.parameter, "block");
+    // Crucially, a malformed filter must never reach the DB as block 0.
+    assert.equal(boundSql, null);
+  }
+});
+
+test("GET /extrinsics?block=0 is a valid filter (genesis block, #1615)", async () => {
+  let boundParam;
+  const env = {
+    METAGRAPH_HEALTH_DB: {
+      prepare(sql) {
+        return {
+          bind(...v) {
+            if (/WHERE block_number = \?/.test(sql)) boundParam = v[0];
+            return {
+              async all() {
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
#1638 — CLOSE — Closes #1615 — galuis116

workers/api.mjs:handleExtrinsics now rejects a malformed ?block= before the query:

    const blockNumber = Number(blockParam);
    if (!/^\d+$/.test(blockParam) || !Number.isSafeInteger(blockNumber)) {
      return analyticsQueryError({
        parameter: "block",
        message: "block must be a non-negative integer.",
      });
    }

Traced the bug: validateQueryParams checks parameter names, not values, so ?block=abc
passed, then clampInt("abc", 0, …) returned the default 0 and the handler ran
WHERE block_number = 0 — serving block 0's extrinsics as confident wrong data. block=0
stays a valid genesis filter. One parametrized test loops ["", "abc", "-5", "1.5",
"1e3", "0x10", " "] -> 400 and asserts the DB is never bound. Efficient boundary
coverage, not bloat. 

Unfortunately, merge conflicts, so closing.

PR #1458, opened 2026-06-22
fix(ai-search): cap vectorId by UTF-8 byte length, not UTF-16 length

joaovictor712:
## Summary

- `vectorId` folds over-long ids to a stable hashed id so they fit Vectorize's id cap. The comment says the cap is **64 bytes**, but the check measured `String#length` (UTF-16 code units), not UTF-8 bytes.
- A multi-byte id that is ≤64 *characters* but >64 *bytes* slips past the fold and is upserted to Vectorize as-is, exceeding the cap — the upsert is rejected and that document silently never gets indexed for semantic search.

Fixes #1457

## Reproduction

```js
const id = "é".repeat(40); // 40 chars, 80 UTF-8 bytes
vectorId(id); // before: returns the 80-byte id (length 40 <= 64) — over the cap
              // after:  folds to "h:<hash>"
```

For ASCII ids `length === byteLength`, so behaviour is unchanged there; only multi-byte ids are affected. (Same UTF-16-length-vs-bytes class fixed for feeds in #1445.)

## What Changed

- `src/ai-search.mjs` — measure `new TextEncoder().encode(id).length` against `VECTOR_ID_MAX_BYTES` (64), so any id that would exceed Vectorize's byte cap is folded to the hashed id.
- `tests/ai-search.test.mjs` — extend the `vectorId` test with a multi-byte id (40 chars / 80 bytes) that must fold to `h:…`.

## Registry Safety

- [x] No secrets, PATs, wallet data, private dashboards, private URLs, or validator-local state.
- [x] Generated artifacts were produced by repo scripts, not hand-edited.
- [x] R2-only/high-churn detail artifacts are not committed.
- [x] No OpenAPI/schema surface changes — internal Vectorize id derivation only; query results are unaffected (real identity lives in metadata).

## Validation

- [x] `npx vitest run tests/ai-search.test.mjs` — the `vectorId` test passes (new multi-byte case fails on the pre-fix `.length` check).
- [x] `npx prettier --check` + `npx eslint` on the changed files — clean.
- Note: the pre-existing `embedding-sync cron > the daily cron triggers an embedding sync` failure reproduces on `main` unchanged (local-env only; unrelated to this change).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1445, opened 2026-06-22
fix(feeds): clamp by code points so a feed string can't split a surrogate pair

philluiz2323:
## Summary

Fixes #1444.

`clamp` truncated feed text with `String.slice(0, max - 1)`, which counts **UTF-16 code units**. A non-BMP character (e.g. an emoji) straddling the boundary was severed into a **lone surrogate** — invalid in XML 1.0, so the serialized RSS/Atom feed is no longer well-formed (and on the UTF-8 wire the surrogate becomes `�`). `clamp` feeds both feed titles (max 80) and summaries (max 500). `stripControl` already iterates by code point; `clamp` then undid that with a code-unit slice.

## What Changed

- `src/feeds.mjs` — truncate by code points (`[...s].slice(0, max - 1).join("")`), consistent with `stripControl`.
- `tests/feeds.test.mjs` — regression test asserting a feed title whose emoji straddles the 80-char clamp contains no lone surrogate.

No schema, OpenAPI, or generated-artifact changes.

## Registry Safety

- [x] No secrets, PATs, wallet data, private dashboards, private URLs, or validator-local state.
- [x] Generated artifacts were produced by repo scripts, not hand-edited.
- [x] R2-only/high-churn detail artifacts are not committed.
- [x] Public API/OpenAPI/schema changes are intentional and documented. (None in this PR.)

## Validation

- [x] `npx vitest run tests/feeds.test.mjs` — 21 passed (incl. the new regression test, which fails without the fix)
- [x] `npx prettier --check` + `npx eslint` on changed files — clean
- [x] `git diff --check`

## Template Used

- Backend/code change

PR #1438, opened 2026-06-22
fix(badge): cap label by code points so it can't split a surrogate pair

philluiz2323:
## Summary

Fixes #1437.

`sanitizeLabel` (public `?label=` override on `/api/v1/subnets/{netuid}/badge.svg`) filters control characters by code point but caps the result with `String.slice(0, MAX_LABEL_LENGTH)`, which counts **UTF-16 code units**. A non-BMP character (e.g. an emoji) straddling the 40-unit boundary is sliced in half into a **lone surrogate** — invalid in XML and corrupting the emitted SVG (it becomes `�` on the UTF-8 wire). `?label=` is public, attacker-controllable input and `escapeXml` does not strip surrogates.

## What Changed

- `src/badge.mjs` — cap the label by code points (`[...str].slice(0, MAX_LABEL_LENGTH).join("")`), consistent with how the function already iterates code points for the control filter.
- `tests/badge.test.mjs` — regression test asserting the rendered SVG contains no lone surrogate when an emoji straddles the cap.

No schema, OpenAPI, or generated-artifact changes.

## Registry Safety

- [x] No secrets, PATs, wallet data, private dashboards, private URLs, or validator-local state.
- [x] Generated artifacts were produced by repo scripts, not hand-edited.
- [x] R2-only/high-churn detail artifacts are not committed.
- [x] Public API/OpenAPI/schema changes are intentional and documented. (None in this PR.)

## Validation

- [x] `npx vitest run tests/badge.test.mjs` — 23 passed (incl. the new regression test)
- [x] `npx prettier --check` + `npx eslint` on changed files — clean
- [x] `git diff --check`

## Template Used

- Backend/code change

PR #1648, opened 2026-06-24
fix: code-point-safe string handling for non-BMP glyphs (embedding input + badge sizing)

### LINKED ISSUE #1650: Non-BMP string handling: embedding-input truncation splits surrogate pairs; badge under-sizes wide glyphs
Two small, related correctness gaps in serve-time string handling for non-BMP (astral) characters. Both are follow-ups to already-merged code-point/byte-safety work.

### 1. `src/ai-search.mjs` — `embeddingText` truncates by UTF-16 units
`embeddingText` caps the embedding input with `.slice(0, 1500)`, counting UTF-16 code units. A non-BMP character (e.g. an emoji in a subnet title) straddling index 1500 is severed into a lone high surrogate, leaving a string that is **not well-formed UTF-16**. When UTF-8-encoded for the embedding model, the trailing token becomes a `U+FFFD` replacement — corrupting the text the cap was meant to preserve.

This is the byte-truthful sibling of the `vectorId()` cap directly below it, which was already made byte-correct (#1458). Repro: `embeddingText({ title: "a".repeat(1499) + "😀" }).isWellFormed()` returns `false`.

### 2. `src/badge.mjs` — `textWidth` under-sizes non-BMP / non-ASCII glyphs
`textWidth` is documented as "a safe overestimate so text never overflows its segment", but every non-ASCII code point falls through to the 6.5px lowercase default — narrower than the 8px of a capital letter. Since `sanitizeLabel` deliberately admits emoji in `?label=` (#1438), an emoji or CJK label is sized smaller than a capital `W` and clips its own segment. Repro: badge width is 38 for an emoji label vs 39 for `W`.

### Fix
Truncate embedding text by code point, and size astral/CJK glyphs as a safe over-estimate. Each with a regression test. ASCII paths unchanged.

glorydavid03023:
Two small, independent correctness fixes for non-BMP / astral characters in serve-time string handling. Both close gaps left by already-merged work, and each ships with a regression test.

Closes #1650

## 1. `ai-search.mjs` — embedding input truncation split surrogate pairs

`embeddingText` capped the embedding text with `.slice(0, 1500)`, which counts **UTF-16 code units**. A non-BMP character (e.g. an emoji in a subnet title) straddling index 1500 is severed into a lone high surrogate, leaving a string that is **not well-formed UTF-16**. When that string is UTF-8-encoded for the embedding model, the trailing token becomes a `U+FFFD` replacement — corrupting the very text the cap was meant to preserve.

Fixed by truncating per **code point** (`[...text].slice(0, EMBED_TEXT_MAX_CHARS)`), mirroring the byte-correct cap in `vectorId()` directly below it (#1458) and the surrogate-safe clamps already used in feeds (#1445/#1514) and the badge label (#1438).

## 2. `badge.mjs` — `textWidth` under-sized non-BMP/non-ASCII glyphs

`textWidth` is documented as "a safe overestimate so text never overflows its segment", but every non-ASCII code point fell through to the 6.5px lowercase default — **narrower than the 8px of a capital letter**. Since `sanitizeLabel` deliberately admits emoji in `?label=` (#1438), an emoji or CJK label was sized smaller than a capital `W` and clipped its own segment.

Fixed by sizing astral code points (emoji) widest and non-ASCII BMP (CJK, full-width scripts) as full-width, keeping the existing ASCII classes. **ASCII labels are unchanged**, so existing width assertions still hold.

## Verification

Both bugs were reproduced before the fix and confirmed resolved after:
- embedding text with an emoji crossing the boundary: `isWellFormed()` `false` → `true`; emoji kept whole.
- emoji badge label segment width: `38` (narrower than `W`'s `39`) → `43` (correctly wider).

```
✓ tests/badge.test.mjs (27 tests)
✓ tests/ai-search.test.mjs (52 tests)
```

`eslint` and `prettier --check` clean on all four files; full `vitest` suite shows no new failures.


JSONbored (commented):
#1648 — CLOSE — Closes #1650 — glorydavid03023

Correct non-BMP fix, but the weaker of the pair vs #1695. It sizes every non-ASCII BMP
code point at full-width 11px, over-widening accented Latin/Cyrillic, where #1695 gates
11px on an explicit East-Asian-Wide/Fullwidth/astral range table (isWideCodePoint) and
8px otherwise — the truer "safe overestimate" the textWidth contract documents. Both
edit src/ai-search.mjs + src/badge.mjs; one lands. Closing in favor of #1695.

PR #1607, opened 2026-06-24
fix(openapi-sample): bound self-refs routed through oneOf/anyOf/allOf

joaovictor712:
## Summary

- `sampleFromSchema` overflows the stack on a self-referential schema whose cycle is routed purely through a composition keyword (`oneOf` / `anyOf` / `allOf`), e.g. `Node: { oneOf: [{ $ref: "#/components/schemas/Node" }] }`.

_(Revised resubmission of #1554, which was auto-closed as `changes-requested`. See "Re: the prior review note" below — the flagged line is verified correct, and I've made it explicit so it reads unambiguously.)_

## What Changed

- The `MAX_DEPTH` recursion guard only fires when `depth` advances, and `depth` is incremented **only** in the object-property and array-items branches. The `$ref`, `allOf`, and `oneOf`/`anyOf` branches all recurse without growing `depth`, so a cycle that never touches an object/array level never trips the guard and recurses until the stack overflows. (The object/array self-ref paths fixed earlier are unaffected — they advance `depth` and still bottom out at `MAX_DEPTH`.)
- Fix: thread a per-descent-path `seenRefs` set through every recursive call and cut the cycle when the same `$ref` recurs at the **same** `depth` (no progress made). Keying by `ref@depth` (not the ref alone) deliberately preserves depth-advancing object/array self-refs, so **non-cyclic schemas sample byte-for-byte as before** — `validate:contract-drift` and `validate:openapi-examples` are unchanged.
- Added a regression test covering `oneOf`/`anyOf`/`allOf` self-references and mutual composition recursion.

Repro (each throws `RangeError: Maximum call stack size exceeded` before the fix, returns a bounded value after):

```js
for (const kw of ["oneOf", "anyOf", "allOf"]) {
  const comps = { Node: { [kw]: [{ $ref: "#/components/schemas/Node" }] } };
  sampleFromSchema({ $ref: "#/components/schemas/Node" }, comps);
}
```

## Re: the prior review note

[...]

PR #1604, opened 2026-06-24
fix(contracts): bound composition-only cycles in sampleFromSchema

bohdansolovie:
## Summary

`sampleFromSchema` bounded object/array self-references via `MAX_DEPTH`, but `oneOf`/`anyOf`/`allOf` and `$ref` branches recurse without advancing `depth`. A self-reference routed purely through a composition keyword never advanced `depth` and overflowed the stack.

Track active `$ref`s per depth and cut non-advancing revisits. Depth-advancing object/array self-refs still bottom out via `MAX_DEPTH` as before.

Fixes #1594

## What Changed

- `src/openapi-sample.mjs`: per-depth active ref tracking; thread `activeRefs` through all recursive branches
- `tests/openapi-sample.test.mjs`: regression tests for composition-only cycles, mutual recursion, and independent `allOf` `$ref` members

## Registry Safety

- [x] No secrets, PATs, wallet data, private dashboards, private URLs, or validator-local state.
- [x] Generated artifacts were produced by repo scripts, not hand-edited.
- [x] R2-only/high-churn detail artifacts are not committed.
- [x] Public API/OpenAPI/schema changes are intentional and documented.

## Validation

- [x] `npm run lint`
- [x] `npm run format:check`
- [x] `npm test -- tests/openapi-sample.test.mjs`
- [x] `npm run validate:contract-drift`
- [ ] `npm run check`
- [ ] `npm run validate`
- [ ] `npm run validate:schemas`
- [ ] `npm run validate:api`
- [ ] `npm run validate:openapi`
- [ ] `npm run validate:types`
- [ ] `npm run validate:artifact-budgets`
- [ ] `npm run validate:docs`
- [ ] `npm run validate:intake`
- [ ] `npm run validate:workflows`
- [ ] `npm run worker:test`
- [ ] `npm run test:coverage`
- [ ] `npm run scan:public-safety`
- [ ] `git diff --check`

Made with [Cursor](https://cursor.com)

PR #1658, opened 2026-06-24
fix(contracts): bound composition-only cycles in sampleFromSchema

bohdansolovie:
## Summary

Fresh resubmission of #1604 (rebased onto current `main` after Gittensory auto-closed the prior PR for base-branch conflicts).

`sampleFromSchema` bounded object/array self-references via `MAX_DEPTH`, but `oneOf`/`anyOf`/`allOf` and `$ref` branches recurse without advancing `depth`. A self-reference routed purely through a composition keyword never advanced `depth` and overflowed the stack.

Track active `$ref`s per depth and cut non-advancing revisits. Depth-advancing object/array self-refs still bottom out via `MAX_DEPTH` as before.

## What Changed

- `src/openapi-sample.mjs`: per-depth active ref tracking; thread `activeRefs` through all recursive branches
- `tests/openapi-sample.test.mjs`: regression tests for composition-only cycles, mutual recursion, and independent `allOf` `$ref` members

## Registry Safety

## Validation

- [x] `npm run lint`
- [x] `npm run format:check`
- [x] `npm test -- tests/openapi-sample.test.mjs`
- [x] `npm run validate:contract-drift`
- [x] Fresh `npm run build` — no stale `public/` artifacts


JSONbored (approved):
#1658 — MERGE — bohdansolovie

src/openapi-sample.mjs:sampleFromSchema overflowed on a self-reference routed purely through allOf/oneOf/anyOf — those branches and $ref recurse WITHOUT advancing depth, so MAX_DEPTH never fires (only the object/array branches grow depth). The fix tracks active $refs per depth in a Map (markActiveRef / unmarkActiveRef), cuts a non-advancing revisit, and unwinds via try/finally:

    if (!markActiveRef(activeRefs, depth, ref)) return null;
    try { return sampleFromSchema(resolveRef(ref, components), …, activeRefs); }
    finally { unmarkActiveRef(activeRefs, depth, ref); }

Allocation-free and the cleaner of the two implementations (beats #1607's
copy-a-Set-on-every-$ref-hop). A real crash on a cyclic schema; minimal and complete.
Merge (close #1607).

PR #1661, opened 2026-06-24
fix(readme-links): honor multi-label public suffixes in dedupe keys

### LINKED ISSUE #1636: readme-links registrableDomain over-collapses multi-label public suffixes (drops distinct tenant links in dedupe)
## Summary
`registrableDomain` in `scripts/lib/readme-links.mjs` computes the registrable domain as the last two labels (`parts.slice(-2)`). Under a multi-label public suffix that is wrong: every subdomain collapses to the bare suffix, so `readmeDedupeKey` treats two genuinely distinct sites as the same and `selectReviewableReadmeLinks` drops the second.

## Trigger
A subnet README links to two distinct tenants on the same multi-tenant host (or two distinct second-level ccTLD domains), same link kind:

- `https://exampleproject-a.pages.dev` and `https://exampleproject-b.pages.dev`
- `https://foo.co.uk` and `https://bar.co.uk`

Both reduce to `pages.dev` / `co.uk`, so the dedupe key `kind:pages.dev` matches and the second link is discarded even though the per-kind cap (e.g. `subnet-api`: 2) would have allowed it. A distinct, reviewable candidate link is silently lost during discovery.

## Location
`scripts/lib/readme-links.mjs`:

```js
function registrableDomain(hostname) {
  const parts = normalizeHost(hostname).split(".").filter(Boolean);
  return parts.slice(-2).join(".");
}
```

## Expected
When the last two labels form a known multi-label public suffix (multi-tenant platform host like `pages.dev`/`github.io`/`vercel.app`, or a second-level ccTLD like `co.uk`), the registrable unit is the last three labels, so distinct tenants stay distinct. Plain-TLD subdomains (e.g. `docs.foo.com` vs `api.foo.com`) should still collapse to one site as today. This mirrors the suffix handling already in `scripts/lib.mjs` (`MULTI_TENANT_HOST_SUFFIXES` + `CLUSTER_CCTLD_SUFFIXES`).

bohdansolovie:
## Summary
- Fix `registrableDomain` in `scripts/lib/readme-links.mjs` so README dedupe keys stay distinct per tenant on multi-label public suffix hosts (`pages.dev`, `co.uk`, etc.).
- Delegate to `clusterDomainFromUrl` from `scripts/lib.mjs` instead of copying suffix tables, so clustering and README dedupe cannot drift.
- Add regression tests ensuring distinct tenants on `pages.dev` and `co.uk` are both kept while same-site subdomains still collapse.

## Problem
`readmeDedupeKey` used `parts.slice(-2)`, so `exampleproject-a.pages.dev` and `exampleproject-b.pages.dev` both collapsed to `pages.dev`. `selectReviewableReadmeLinks` then dropped the second link even when the per-kind cap allowed it.

## Test plan
- [x] `npm test -- tests/script-utils.test.mjs -t "README dedupe keeps distinct tenants"`
- [x] `npm test -- tests/lib-helpers.test.mjs -t "clusterDomainFromUrl"`

Closes #1636


JSONbored (approved):
#1661 — MERGE — Closes #1636 — bohdansolovie

scripts/lib/readme-links.mjs:registrableDomain replaced parts.slice(-2) with a
delegation to clusterDomainFromUrl (the existing MULTI_TENANT_HOST_SUFFIXES /
CLUSTER_CCTLD_SUFFIXES logic in lib.mjs). Traced the bug: under a multi-label public
suffix, slice(-2) collapsed exampleproject-a.pages.dev and -b.pages.dev both to
"pages.dev", so readmeDedupeKey treated two distinct tenants as one and
selectReviewableReadmeLinks dropped the second. Tests the three paths (pages.dev,
co.uk, and the plain-TLD negative that must still collapse). It also unifies the
suffix logic across the two modules — the right call. Merge.

PR #1664, opened 2026-06-24
feat(api): add a cross-subnet /api/v1/compare endpoint

oktofeesh1:
## Summary

The registry already exposes each subnet's structure, economics, and live health, but on separate routes — there was no way to view several subnets *side by side*. Choosing between subnets meant N separate detail/economics/health fetches plus client-side stitching. This adds `GET /api/v1/compare?netuids=1,7,64`, which composes three domains into one response, in requested order:

- **structure** — completeness score + surface/operational-interface counts (from the registry profiles projection)
- **economics** — registration cost, open slots, emission share, validator/miner counts, etc. (the live economics tier)
- **health** — per-subnet probe rollup: surface count, ok count, avg latency (live D1)

`dimensions=` selects a subset (default all three). Unknown netuids return `found:false` rather than 404ing the whole call, so callers can always align columns. Bounded to 1–128 ids.

## What Changed

- `workers/api.mjs`: a fileless-D1 route (`handleCompare`) + a pure, unit-tested `composeCompareData` projection, dispatched + edge-cached exactly like `handleLeaderboards` (keyed on the cron snapshot's `last_run_at`). Reuses the existing `leaderboardProfilesProjection`, `resolveEconomicsRows`, and the `surface_status` D1 rollup — no new readers. Marked mainnet-only alongside the other live registry routes.
- `src/contracts.mjs`: the `compare` route + `PUBLIC_ARTIFACTS` entry (`CompareArtifact`).
- `schemas/components/05-subnets.schema.json`: the `CompareArtifact` response component.
- `scripts/validate-api.mjs` / `scripts/validate-schemas.mjs`: a per-route smoke check, and the fileless/live-only skip entry (no static `compare.json` on disk, like leaderboards).
- `docs/backend-artifact-contracts.md`: route + artifact contract lines.
- Regenerated committed artifacts: `public/metagraph/{openapi,api-index,contracts,types.d.ts}`, `generated/metagraphed-api.d.ts`, `schemas/api-components.schema.json`.
- Tests: `tests/compare.test.mjs` (pure-projection branch coverage + HTTP validation/wiring) and a mainnet-only case in `tests/network-routing.test.mjs`.

## Registry Safety

- [x] Generated artifacts were produced by repo scripts (`npm run build`), not hand-edited.
- [x] R2-only/high-churn detail artifacts are not committed (`r2-manifest.json` reverted; compare is fileless).
- [x] Public API/OpenAPI/schema changes are intentional and documented — a new read-only route mirroring the existing leaderboards pattern; display-only, never feeds completeness/scoring.

## Validation

- [x] `npm run check` (lint + format + pipeline + docs)
- [x] `npm run validate` · `validate:schemas` · `validate:api` · `validate:openapi` · `validate:types` · `validate:generated-client`
- [x] `npm run validate:docs` · `validate:contract-drift` · `validate:artifact-budgets` · `validate:intake` · `validate:workflows`
- [x] `npm run scan:public-safety` · `git diff --check`
- [x] `npm run worker:test`
- [x] Full `vitest run` — 1936 tests; new `composeCompareData`/`handleCompare` lines at 100% branch coverage (verified via the v8 JSON report)


JSONbored (approved):

JSONbored:
## Review — `feat(api): /api/v1/compare` (#1664)

Net: this is a clean, well-tested addition that correctly mirrors the `handleLeaderboards`
fileless-D1 pattern. CI is genuinely green (`test` + `checks` pass, Superagent clean, Codecov
100% patch coverage), the pure/HTTP split is the right shape, and the projection is null-safe
per source tier. A few notes — one latent infra issue worth a fast follow-up, one
contract-quality item, and a process note.

### What's solid
- **`composeCompareData` is a clean pure projection** — found/missing and per-dimension-null
  branches are all exercised, requested order preserved, dupes dropped. Good separation from the
  I/O in `handleCompare`.
- **Economics field mapping is correct** — all nine fields (`registration_cost_tao`,
  `registration_allowed`, `open_slots`, `emission_share`, `alpha_price_tao`, `validator_count`,
  `miner_count`, `total_stake_tao`, `miner_readiness`) match the canonical `SubnetEconomics`
  schema, so no silent `undefined`s.
- **Edge-cache key includes `url.search`**, so `netuids`/`dimensions` permutations don't
  collide; lazy reads (no economics/health fetch when the dimension isn't requested) are a nice
  touch.
- **Param validation is tight** — duplicate keys, unknown params, malformed/oversized `netuids`,
  and unknown `dimensions` all 400 with the right `meta.parameter`. The 128-id cap is enforced at
  the regex (`maxLength: 767` = 128×5 + 127 commas — correct).

### 1. `storage_tier` mis-tiering (recommend a follow-up)
`/metagraph/compare.json` resolves to **`git`** tier, while every other fileless/computed-live
route resolves to **`r2`**:

    compare                => git      ← outlier
    registry/leaderboards  => r2
    rpc/usage              => r2
    incidents              => r2

It's `git` only because `compare.json` was never added to `R2_ONLY_PATTERNS` in
`src/artifact-storage.mjs` — so it falls through to the default `git` tier. This is the exact
mis-tiering the codebase already documents as a prior incident (see the
`review/profile-completeness.json` comment: *"mis-tiered as git (committed) only because it was
absent from BOTH pattern lists … #998 v1"*).

It isn't breaking today **only** because `compare` was *also* added to the `COMPUTED_ARTIFACTS`
skip-list in `scripts/validate-schemas.mjs`, which suppresses the "missing committed file" check.
So two parallel sources of truth happen to agree — but the published contract (`contracts.json` /
`api-index.json`) now advertises `storage_tier: "git"`, telling consumers/tooling there's a
committed static file at `/metagraph/compare.json` that does not and never will exist. The
live-only-D1 convention is `R2_ONLY_PATTERNS` **and** `COMPUTED_ARTIFACTS` together; this only did
the latter.

**Fix:** add `/^compare\.json$/` to `R2_ONLY_PATTERNS`, then `npm run build` to regenerate
`contracts.json` + `api-index.json`. That makes the tier `r2` like its siblings and removes the
reliance on the skip-list alone.

### 2. The published schema is effectively contract-free
`CompareArtifact.subnets` is `array<object, additionalProperties:true>`, so:
- the generated client types it as `{ [key: string]: unknown }[]` — no usable typing for the
  `found` / `structure` / `economics` / `health` payload that is the entire point of the endpoint;
- `validate:contract-drift` can't catch a future field rename in `composeCompareData` (e.g. an
  economics key drifting), because nothing in the schema pins the inner shape.

Not a blocker, but the endpoint's value *is* the composed shape — worth defining a real per-entry
sub-schema (even one `oneOf`/properties block reusing the existing `SubnetEconomics` component) so
the contract and the typed client actually describe it.

### Minor
- **Health/economics reads aren't bounded by the requested netuids** — the `surface_status`
  rollup is `GROUP BY netuid` over the whole table and `resolveEconomicsRows` returns all subnets,
  then `composeCompareData` filters in memory. Fine functionally and consistent with leaderboards,
  but since the query is identical regardless of `netuids` yet cached per-querystring, the same
  full scan is recomputed/cached once per distinct `netuids` combination. A `WHERE netuid IN (…)`
  would let it use the index and tighten cache economics.
- **`observed_at` is always stamped from the prober `last_run_at`**, even for a
  `dimensions=structure` or `dimensions=economics` request where the data comes from
  `profiles.json` / the economics tier (different cadences). Slightly misleading for non-health
  subsets — consider sourcing it per-dimension or documenting that it reflects the probe snapshot
  only.
- **`dimensions=` (empty)** returns `400 Unknown dimension ""` — defensible, but the message reads
  oddly; not covered by a test.

Nothing here warrants a revert — the endpoint is correct and live. I'd queue the
`R2_ONLY_PATTERNS` one-liner (#1) as a quick follow-up and treat the schema tightening (#2) as the
higher-value improvement.

PR #1667, opened 2026-06-24
fix(neuron-history): share one now() across neuron_daily archive + prune

### LINKED ISSUE #1634: neuron_daily archive+prune sample Date.now() independently — a day-boundary cron tick can delete an un-archived day
### What

The `NEURON_HISTORY_ROLLUP_CRON` branch of `handleScheduled` (workers/api.mjs) runs the daily neuron_daily archive-before-prune:

```js
const archivedPrunable = await archivePrunableNeuronDaily(env).catch(() => ({ archived: false }));
const pruned = archived.archived && archivedPrunable.archived
  ? await pruneNeuronDaily(env).catch(() => ({ pruned: false }))
  : { pruned: false, reason: "archive-not-confirmed" };
```

### Why it is wrong

`archivePrunableNeuronDaily` and `pruneNeuronDaily` each default `now = Date.now()` internally and derive their retention cutoff via `neuronDailyRetentionCutoff(now)` = `new Date(now - 400*86_400_000).toISOString().slice(0,10)`, used with `snapshot_date < cutoff`. They are called WITHOUT a shared `now`, so each samples `Date.now()` at its own call time.

The backlog archive does ~33k-row R2 work (one gzip put per subnet per prunable day) and can run for seconds. If that work straddles a UTC midnight, `pruneNeuronDaily`'s `Date.now()` lands on the next day, so its cutoff is **one day larger** than the archive's. The prune then deletes a `snapshot_date` from D1 that was never written to R2 — the exact archive-before-prune data-loss gap the code's own docstring says it closes ("a day is never dropped from D1 before it exists in R2").

### Repro

- Archive runs `2026-06-21T23:59:59.7Z` → cutoff `2025-05-08`, archives days `< 2025-05-08`.
- R2 puts take ~1s; prune runs `2026-06-22T00:00:00.3Z` → cutoff `2025-05-09`, deletes days `< 2025-05-09`.
- Day `2025-05-08` is deleted from D1 although the archive never wrote it to R2 → permanent loss of one day of neuron_daily history.

### Expected

Compute `now` once in `handleScheduled` and pass `{ now }` to both `archivePrunableNeuronDaily` and `pruneNeuronDaily` (both already accept it), so they share one retention cutoff and the prune can only delete days the archive confirmed.

philluiz2323:
## Summary

Fixes #1634.

The `NEURON_HISTORY_ROLLUP_CRON` branch of `handleScheduled` called `archivePrunableNeuronDaily(env)` and `pruneNeuronDaily(env)` without a shared `now`, so each sampled `Date.now()` at its own call time. Both derive their retention cutoff from `neuronDailyRetentionCutoff(now)` (`snapshot_date < now-400d`). The backlog archive does ~33k-row R2 work and can straddle a UTC midnight, so the prune's `Date.now()` can land on the next day and produce a cutoff **one day larger** than the archive's — deleting a `snapshot_date` from D1 that was never written to R2. That is the exact archive-before-prune data-loss gap this code is meant to close.

## What Changed

- Pin `const now = Date.now()` once in the cron branch and pass `{ now }` to both `archivePrunableNeuronDaily` and `pruneNeuronDaily` (both already accept it), so they share one retention cutoff.
- Added a regression test: `Date.now` advances a full day on each call, and the test asserts the archive and prune cutoffs stay identical (fails before the fix).

## Template Used

- backend-code

## Registry Safety

- [x] Public API/OpenAPI/schema changes are intentional and documented. (none)

## Validation

- [x] `vitest run tests/neuron-history.test.mjs` (36 passed, incl. new regression test; verified it fails without the fix)
- [x] `prettier --check` + `eslint` on changed files
- [x] `git diff --check`


JSONbored (approved):
#1667 — MERGE  (sequence vs #1701/#1710) — Closes #1634 — philluiz2323

workers/api.mjs:handleScheduled pins one now and threads it to both halves:

    const now = Date.now();
    const archivedPrunable = await archivePrunableNeuronDaily(env, { now }) …
    const pruned = … await pruneNeuronDaily(env, { now }) …

Traced the bug: each previously sampled Date.now() independently, and
neuronDailyRetentionCutoff = now - 400 days; the ~33k-row archive can straddle UTC
midnight, making the prune's cutoff one day larger and deleting a snapshot_date from
D1 the archive never wrote to R2 — the exact archive-before-prune gap the docstring
claims to close. The test mocks Date.now to advance a day per call and asserts identical
cutoffs. Composes with #1701/#1710 (all touch handleScheduled) — sequence the rebases.
Merge.

Issue #1673, opened 2026-06-24
feat(schema): give `CompareArtifact` a real per-entry sub-schema (currently contract-free)


JSONbored:
## Problem

The `CompareArtifact` response schema (added in #1664, `schemas/components/05-subnets.schema.json`) declares:

```jsonc
"subnets": { "type": "array", "items": { "type": "object", "additionalProperties": true } }
```

i.e. the per-subnet entry shape — the `found` / `structure` / `economics` / `health` payload that is the entire point of `GET /api/v1/compare` — is **completely unconstrained** by the published contract. Consequences:

- The generated typed client types it as `{ [key: string]: unknown }[]` (see `public/metagraph/types.d.ts` / `generated/metagraphed-api.d.ts`), so consumers get no usable typing for the composed payload.
- `validate:contract-drift` cannot catch a future field rename in `composeCompareData` (`workers/api.mjs`) — e.g. an economics key drifting — because nothing in the schema pins the inner shape. The field names are currently only asserted by `tests/compare.test.mjs`, not by the contract.

## Suggested approach

Define a real per-entry sub-schema for `CompareArtifact.subnets[]`:

- top-level: `netuid` (integer), `name`/`slug` (string|null), `found` (boolean);
- `structure`: `{ completeness_score, surface_count, operational_interface_count }` (nullable object);
- `economics`: nullable — ideally reuse the existing `SubnetEconomics` component (the nine fields `composeCompareData` projects already match it);
- `health`: `{ surface_count, ok_count, avg_latency_ms }` (nullable object).

Then `npm run build` to regenerate `openapi.json` + types/client. This makes the contract and the typed client actually describe the endpoint and brings the payload under `contract-drift` protection.

Context: follow-up from the #1664 review.

PR #1688, opened 2026-06-24
test(compare): validate composeCompareData against CompareArtifact

bohdansolovie:
## Summary
- Rebased onto `main` after #1709 landed the CompareArtifact sub-schema (#1673).
- Add AJV regression tests proving `composeCompareData` output validates against the published `CompareArtifact` contract — so a future field rename in `workers/api.mjs` trips tests even when existing assertions still pass.

## Test plan
- [x] `npm test -- tests/compare.test.mjs`
- [x] Rebased cleanly onto latest `main` (no schema/generated-file conflicts)

Follow-up to #1709 / #1673


JSONbored (approved):
#1688 — MERGE — bohdansolovie

AJV regression pinning composeCompareData against the published CompareArtifact
contract — a real guard: a later field rename in workers/api.mjs trips it even when the
existing assertions still pass. Correct, green, a follow-up to #1709 / #1673. No linked
issue, no bar. Merge.

PR #1690, opened 2026-06-24
feat(feeds): add ETag and conditional request support to the feed endpoints

fansilas:
## Summary

The registry, incidents, and per-subnet feeds under `/api/v1/feeds/*` re-rendered and re-sent the full RSS, Atom, or JSON document on every request. They set `cache-control` but never emitted an `ETag`, and an incoming `If-None-Match` was ignored, so feed readers and pollers could not validate a cached copy and always paid for a full body. This wires the feed responses into the same weak-ETag revalidation the rest of the API already uses, so an unchanged poll gets a cheap `304 Not Modified`.

## What Changed

- `handleFeedRequest` now computes a weak `ETag` over the serialized feed body and sets it on every `200` response. The body is deterministic for its inputs (source timestamps, kind, netuid, format, and the `tag` filter all flow into it), so the hash is a stable validator across identical polls.
- A request whose `If-None-Match` matches gets a `304 Not Modified` with the same validators and no body. The check reuses `ifNoneMatchSatisfied` from `workers/http.mjs`, so weak and strong comparison and the `*` form all behave consistently with the rest of the API.
- Hashing reuses the shared `weakEtag` helper rather than a feed-specific scheme, keeping feeds aligned with `envelopeResponse`, the discovery document, and the badge endpoints. The `200` and `304` paths share one headers object, so nothing is duplicated.
- Added coverage in `tests/feeds.test.mjs` for all three feed kinds across all three formats: ETag emission, the matching `304` path, ETag stability across identical requests, distinct ETags per format and per kind, the tag-filtered variant, the HEAD path, a stale validator falling back to `200`, the `*` form, and weak/strong validator equivalence.

No schema, contract, OpenAPI, or committed artifact changed, so no regeneration was needed.

## Registry Safety

## Validation

- [x] `npm run check`
- [x] `npm run validate`
- [x] `npm run validate:api`
- [x] `npm run validate:docs`
- [x] `npm run worker:test`
- [x] `npm run test:coverage`
- [x] `git diff --check`


JSONbored (approved):
#1690 — MERGE — fansilas

src/feeds.mjs:handleFeedRequest now emits a weak ETag and 304s a matching If-None-Match, reusing the existing workers/http.mjs:weakEtag / ifNoneMatchSatisfied — so the
/api/v1/feeds/* endpoints get the same revalidation the rest of the API already has, and
an unchanged poll returns a bodyless 304 carrying the same validators. Exhaustive tests
(kind x format, HEAD, If-None-Match: *, stale, tag-filtered). Correct, complete, and
consistent with the existing helpers. Merge.

PR #1691, opened 2026-06-24
fix(cors): expose custom response headers via Access-Control-Expose-Headers

fansilas:
## Summary

- Cross-origin browser clients could not read any of the custom response headers the API sends, because the responses never set `Access-Control-Expose-Headers`. Per the Fetch standard a script may only read a small safelist unless the server names the rest, so callers of `/api/v1/ask`, `/rpc/v1/*`, the SSE stream, or any rate-limited route could not detect throttling, read the `retry-after` hint, or surface the contract version even though the headers were present on the wire.

## What Changed

- Added a single canonical list of the custom client-facing headers (the `etag` validator, the rate-limit family, `retry-after`, and the `x-metagraph-*` diagnostics) in `workers/http.mjs`, exposed once from the standard `apiHeaders` builder so every CORS-open route built through it carries `Access-Control-Expose-Headers`.
- Applied the same list to the two response builders that assemble headers directly rather than through `apiHeaders`: the SSE event stream in `workers/api.mjs` and the MCP server responses in `src/mcp-server.mjs` (whose 429 carries the rate-limit family).
- Centralised the list as one source of truth (`exposeCustomResponseHeaders` for `Headers` instances, `EXPOSED_RESPONSE_HEADERS_VALUE` for plain header objects) so the two builder styles share an identical value with no drift.
- Added tests across the representative surfaces: list endpoints, the ask error path, the SSE stream, the RPC proxy, and the MCP rate-limit response. Every changed line is covered (`workers/http.mjs` reports 100 percent statements, lines, and functions).

## Registry Safety

## Validation

- [x] `npm run check`
- [x] `npm run validate`
- [ ] `npm run validate:schemas`
- [ ] `npm run validate:api`
- [ ] `npm run validate:openapi`
- [ ] `npm run validate:types`
- [ ] `npm run validate:artifact-budgets`
- [x] `npm run validate:docs`
- [ ] `npm run validate:intake`
- [ ] `npm run validate:workflows`
- [x] `npm run worker:test`
- [x] `npm run test:coverage`
- [x] `git diff --check`


JSONbored (approved):
#1691 — MERGE — fansilas

CORS Access-Control-Expose-Headers via a single canonical EXPOSED_RESPONSE_HEADERS list
in workers/http.mjs, applied across apiHeaders, the SSE handler, and MCP_HEADERS —
correct per the Fetch spec, so cross-origin clients can finally read etag / retry-after
/ x-ratelimit-* / x-metagraph-*. The hand-maintained list is a maintenance note (a new
x-metagraph-* header must be added here too), not a defect — a strict improvement over
exposing nothing. Tested on list / ask / SSE / RPC / MCP-429. Merge.

PR #1693, opened 2026-06-24
fix(api): skip edge cache on D1 fallback for analytics routes

andriypolanski:
### Summary

Prevent D1 fallback responses from poisoning the analytics edge cache. Several
`withEdgeCache`-wrapped routes could serve and persist empty 200 envelopes when
D1 was unavailable or unbound, while sibling routes (percentiles, incidents,
trends) already marked fallback responses to skip caching.

### What Changed

- **`d1All()`**: unbound `METAGRAPH_HEALTH_DB` now returns
  `markD1FallbackRows([])` so the generation counter and fallback WeakSet
  detect cold D1 the same way query failures do.
- **`handleHealthTrends()`**: unbound inline D1 path uses
  `markD1FallbackRows([])` for the same reason.
- **Edge-cached handlers** now call `hasD1FallbackRows` /
  `markD1FallbackResponse` like the existing percentiles/incidents/trends
  routes:
  - `handleGlobalIncidents`
  - `handleTrajectory` (queries via `d1All` + `formatTrajectory` directly)
  - `handleUptime`
  - `handleLeaderboards`
  - `handleCompare`
- **Tests** (`tests/analytics-edge-cache.test.mjs`):
  - D1 query failure must not cache on leaderboards, global incidents,
    trajectory, uptime, and compare
  - Unbound D1 with a warm health snapshot stamp must not cache leaderboards

### Registry Safety

      validator-local state.
      *(No contract change — response shape unchanged; only cache eligibility.)*

### Validation

- [x] `npm test -- tests/analytics-edge-cache.test.mjs` (12/12 pass)
- [x] `npm run check`
- [x] `npm run validate`
- [x] `npm run worker:test`


JSONbored (approved):
Good catch on the cache-poisoning class — an empty 200 from cold/unbound D1 getting persisted by withEdgeCache and served past recovery is real, and routing the affected handlers through hasD1FallbackRows/markD1FallbackResponse matches the percentiles/incidents/trends precedent. I checked the one risky bit — the handleTrajectory inline query is byte-faithful to loadSubnetTrajectory (same columns/ORDER/LIMIT 400 → formatTrajectory), and that helper isn't orphaned (still used by mcp-server.mjs:979). Only nit: you now have the trajectory SQL in two places; consider having loadSubnetTrajectory expose fallback-ness so the route can reuse it instead of duplicating the query. Approving.

PR #1698, opened 2026-06-24
fix(api): reject contradictory numeric range bounds (min_ above max_)

### LINKED ISSUE #1697: List range filters accept a contradictory min_ > max_ and return a silent empty 200
## Problem

The list endpoints accept inclusive numeric range bounds (min_<field> / max_<field>, e.g. `/api/v1/subnets?min_surface_count=…&max_surface_count=…`). Each bound is validated individually (must be a finite number, rejected if it overflows — #1655), but the **pair** is not checked against itself.

A request where the lower bound is greater than the upper bound on the same field, e.g.:

`/api/v1/subnets?min_surface_count=9&max_surface_count=2`

is schema-valid and returns `200` with an **always-empty** result set. The range can never select a row, so the response is silently empty with no signal that the query itself is contradictory.

## Expected

Consistent with how the API already rejects malformed and overflowing bounds (#1655) and unsupported `sort` fields, a contradictory range should be an explicit `400 invalid_query` naming the offending parameter, rather than a silent empty `200`.

## Proposed fix

In `validateListQuery` (`workers/list-query.mjs`), after confirming both bounds on a field are numbers, reject the request when `min_<field> > max_<field>` with parameter `min_<field>` and a clear message. Equal bounds stay valid (a single-value range). Runtime-only — no schema/OpenAPI change.

glorysr1209-png:
Closes #1697.

## What

List endpoints validate each numeric range bound (`min_<field>` / `max_<field>`) individually but never compare the pair, so a contradictory request like `/api/v1/subnets?min_surface_count=9&max_surface_count=2` is schema-valid and returns a silent, always-empty `200`.

This adds a cross-bound check in `validateListQuery`: when both bounds are present on a field and `min_<field> > max_<field>`, return `400 invalid_query` naming `min_<field>` — the same explicit-rejection treatment already given to malformed/overflowing bounds (#1655) and unsupported sort fields. Equal bounds remain valid (a single-value range).

Runtime-only: no schema, OpenAPI, or generated-artifact changes (the `min_`/`max_` parameters and their `{type: number}` schemas are unchanged; this is a cross-parameter constraint not expressible per-parameter).

## Files (2)

- `workers/list-query.mjs` — parse both bounds per field, reject `min > max`.
- `tests/list-query.test.mjs` — contradictory pair is rejected and names the field; equal bounds stay valid and select the single-value row; a contradictory bound is caught even alongside an unrelated valid bound.

## Validation

- `npx vitest run tests/list-query.test.mjs` → 17 passed.
- Full `npx vitest run` shows no new failures vs. `main` (the pre-existing failures are build-artifact/D1/KV suites that need `npm run build`; this change is runtime-only and adds 3 passing tests).


JSONbored (commented):
#1698 — CLOSE — Closes #1697 — glorysr1209-png

The min_<field> > max_<field> rejection in validateListQuery is correct (400
invalid_query, equal bounds stay valid, mirrors the #1655 overflow handling). But the
rebased head is red: the checks job fails at "Lint + format" — the diff is unprettified.
A PR that needs a reformat to go green is not merge-as-is. Run npm run format and
resubmit. Closing.

PR #1700, opened 2026-06-24
feat(api): match all whitespace-separated q terms on list endpoints

### LINKED ISSUE #1699: List ?q= only matches a contiguous phrase, so multi-word queries miss obvious results
## Problem

The list endpoints support a free-text `?q=` over each collection's searchable fields (added in #1631). The query is matched as **one contiguous substring** of the joined fields:

`workers/list-query.mjs` → `searchRows` does `haystack.includes(q.toLowerCase())`.

So `?q=` only finds rows where the terms appear together, in that exact order. A natural two-word query misses obvious matches:

- `/api/v1/subnets?q=gradients%20training` will **not** find a subnet named "Gradients" whose description is "decentralized training network", because the literal phrase "gradients training" never appears in a single field.

There is also a small edge: a **whitespace-only** `?q=%20` is truthy, so the lone-space needle matches every row whose joined fields contain a space — effectively a near-match-all instead of "no query".

## Expected

Multi-word discovery should match each term independently (order-independent AND) across the searchable fields, like typical search boxes — and a blank / whitespace-only query should mean "no search".

## Proposed fix

In `searchRows`, split `q` on whitespace into terms and keep a row only if **every** term is a substring of its joined searchable fields. A single term keeps the existing substring behaviour; an empty/whitespace-only `q` yields no terms and is treated as no search. Runtime-only — no schema/OpenAPI change.

glorysr1209-png:
Closes #1699.

## What

List `?q=` (added #1631) matches the whole query as one contiguous substring, so a two-word query like `/api/v1/subnets?q=gradients%20training` misses a subnet named "Gradients" described as "decentralized training network" — the exact phrase never appears in one field.

`searchRows` now splits `q` on whitespace and keeps a row only when **every** term is a substring of its joined searchable fields (order-independent AND). Behaviour preserved at the edges:

- a single term is identical to the previous substring match;
- an empty or whitespace-only `q` has no terms → treated as no search (previously a lone space matched every row that merely contained a space).

Runtime-only: no schema, OpenAPI, or generated-artifact changes.

## Files (2)

- `workers/list-query.mjs` — tokenize `q` and require all terms.
- `tests/subnets-query.test.mjs` — order-independent multi-term match; AND (not OR) across rows; single-term unchanged; whitespace-only `q` is a no-op.

## Validation

- `npx vitest run tests/subnets-query.test.mjs tests/list-query.test.mjs` → 22 passed.
- Full `npx vitest run`: 111 failed identical to clean `main` baseline (pre-existing build-artifact/D1/KV suites that need `npm run build`, broken locally by the Windows `repoRoot` `C:\C:\` doubling bug; not present in Linux CI). This change adds 4 passing tests, zero new failures.


JSONbored (commented):
#1700 — CLOSE — Closes #1699 — glorysr1209-png

Multi-term ?q= AND-matching in searchRows is correct (splits on whitespace, every term must substring-match; blank / whitespace-only -> no search). Same red as #1698, same author: the checks job fails at "Lint + format" (unprettified). Reformat and resubmit.

Closing.

PR #1701, opened 2026-06-24
fix(api): atomic upsert+prune and retry neuron snapshot prune

andriypolanski:
### Summary

Fix ghost neuron rows in the live metagraph when `loadStagedNeurons()` upserts
a new snapshot but the post-load prune fails. `INSERT OR REPLACE` never removes
deregistered `(netuid, uid)` pairs — snapshot-replace depends on the DELETE
prune running after upsert. Previously, a failed prune left new rows committed
while stale UIDs lingered until a later successful run.

### What Changed

- **`loadStagedNeurons()`** (`workers/api.mjs`):
  - **Single-batch loads** (`upsert statements + 1 <= 50`): upserts and snapshot
    prune now run in one `db.batch()` call so D1 rolls back both if prune fails.
  - **Multi-batch loads**: upsert batches unchanged; prune runs via new
    `runNeuronSnapshotPrune()` with up to 3 retries before `purge_failed`.
  - Extracted `neuronSnapshotPruneStatement()` and `runNeuronSnapshotPrune()`
    helpers for shared prune SQL + retry logic.
- **`NEURON_SNAPSHOT_PRUNE_RETRIES`** (`workers/config.mjs`): configurable
  retry count (default 3) for the post-upsert prune on large snapshots.
- **Tests** (`tests/load-staged-neurons.test.mjs`):
  - Atomic upsert+prune batch failure leaves prior snapshot intact
  - Multi-batch prune retries on transient failure and still removes deregistered
    ghost UIDs
  - Updated mocks to handle prune statements inside `db.batch()` and return
    batch result metadata

### Registry Safety

      validator-local state.
      *(No contract change — internal cron loader behavior only.)*

### Validation

- [x] `npm test -- tests/load-staged-neurons.test.mjs` (16/16 pass)
- [x] `npm run check`
- [x] `npm run validate`
- [x] `npm run worker:test`

### Test plan

- [x] Deregistered UID pruned across snapshots (regression #1303)
- [x] Atomic batch failure → `load_failed`, prior rows unchanged, R2 staged object kept
- [x] Multi-batch prune fails twice then succeeds → ghost UID removed, `ok: true`
- [ ] Partial coverage refresh still prunes only refreshed netuids
- [ ] Failed upsert mid-load still returns `load_failed` without prune or R2 delete


JSONbored (approved):
#1701 — MERGE  (sequence vs #1667/#1710) — andriypolanski

workers/api.mjs:loadStagedNeurons — single-batch loads now run the upsert and the
snapshot-prune in one transaction:

    await db.batch([...statements, pruneStatement]);

so a failed prune can no longer commit replacement rows while deregistered UIDs linger;
the multi-batch path gets a bounded retry (NEURON_SNAPSHOT_PRUNE_RETRIES = 3). Correct D1 work; tests prove the prior snapshot survives a mid-batch failure. It overlaps
handleScheduled / loadStagedNeurons with #1667 and #1710 — three correct, distinct
fixes to one pipeline that compose; merge in sequence (one, then rebase the next).
Merge.

PR #1703, opened 2026-06-24
fix(api): sort rows missing the sort field last regardless of order

### LINKED ISSUE #1702: List sort places rows missing the sort field first in ascending order
## Problem

List endpoints sort with `sortRows` in `workers/list-query.mjs`:

`[...rows].sort((a, b) => compareValues(a[key], b[key]) * direction)`

`compareValues` coerces a missing value to `String(value ?? "")`, so a row that lacks the sort field compares as the empty string `""`. In **ascending** order that sorts *smallest*, so the least-complete rows (e.g. a subnet with no `tempo` / `surface_count` yet) land at the **top** of the list. Flip to `order=desc` and those same gap rows jump to the bottom — so simply toggling direction shuffles incomplete rows in and out of the most-visible position.

## Expected

Rows missing the sort field should consistently sort **last**, in both directions — incomplete data shouldn't lead an ascending list, and its position shouldn't depend on the sort direction.

## Proposed fix

In `sortRows`, partition rows whose sort value is `null`/`undefined` out of the comparison and append them after the ordered rows. Present values sort exactly as before; missing rows always trail. Runtime-only — no schema/OpenAPI change.

glorysr1209-png:
Closes #1702.

## What

`sortRows` coerces a missing sort value to `""` via `compareValues`, so rows lacking the sort field sort *first* in ascending order (incomplete rows lead the list) and *last* in descending order — toggling `order` shuffles gap rows in and out of the top.

This partitions rows whose sort value is `null`/`undefined` out of the comparison and appends them after the ordered rows, so missing rows always trail regardless of direction. Present values sort exactly as before.

Runtime-only: no schema, OpenAPI, or generated-artifact changes.

## Files (2)

- `workers/list-query.mjs` — partition present vs missing, sort present, append missing.
- `tests/list-query.test.mjs` — ascending keeps missing rows at the tail (not the front); descending keeps them at the tail too; toggling order leaves the incomplete rows in the tail.

## Validation

- `npx vitest run tests/list-query.test.mjs` → 17 passed.
- Full `npx vitest run` matches the clean `main` baseline (pre-existing failures are build-artifact/D1/KV suites that need `npm run build`, broken locally by the Windows `repoRoot` `C:\C:\` doubling bug; not present in Linux CI). This change adds 3 passing tests, zero new failures.


JSONbored (approved):
Taking this one. The partition reads unambiguously, it's stable for the missing group, and it guards a null row (row == null ? undefined : row[key]) — #1717's comparator would throw on null[key]. You also filed first ([#1702](https://github.com/JSONbored/metagraphed/issues/1702)). Approving.

PR #1710, opened 2026-06-24
fix(api): gate neuron daily rollup on complete metagraph snapshot

andriypolanski:
### Summary

Prevent corrupted `neuron_daily` rows when the daily metagraph rollup runs
while `loadStagedNeurons()` is mid-batch. The rollup uses
`WHERE captured_at = MAX(captured_at)`; during a concurrent load some neurons
already carry the new stamp while others on the same subnet still have the old
one, producing an incomplete daily snapshot that persists via
`ON CONFLICT DO UPDATE`.

### What Changed

- **`neuronsSnapshotReadyForDailyRollup()`** (`src/neuron-history.mjs`): new
  pre-check that detects any subnet with rows at the latest `captured_at` that
  also still has older rows for the same `netuid` (partial-load signature).
- **`rollupNeuronDaily()`**: skips rollup with
  `{ rolled: false, reason: "incomplete-snapshot" }` when the check fails.
- **`NEURON_HISTORY_ROLLUP_CRON` handler** (`workers/api.mjs`): calls
  `loadStagedNeurons()` before `rollupNeuronDaily()` so any pending R2-staged
  snapshot is fully drained before the daily snapshot is taken. Returns `staged`
  in the cron payload for observability.
- **Tests** (`tests/neuron-history.test.mjs`):
  - Unit tests for the completeness probe
  - Rollup skip on incomplete snapshot
  - Cron test mocks updated for the new probe query

### Registry Safety

      validator-local state.
      *(No contract change — internal cron/rollup behavior only.)*

### Validation

- [x] `npm test -- tests/neuron-history.test.mjs` (39/39 pass)
- [x] `npm run check`
- [x] `npm run validate`
- [x] `npm run worker:test`

### Test plan

- [x] Completeness probe returns true when no mixed stamps exist
- [x] Completeness probe returns false when a subnet has latest + older rows
- [x] `rollupNeuronDaily` skips with `incomplete-snapshot` when probe fails
- [x] Daily rollup cron still archives and prunes when snapshot is complete
- [ ] Manual: confirm rollup cron at 05:47 UTC drains staged neurons before rolling up

### DIFF
--- a/src/neuron-history.mjs
+++ b/src/neuron-history.mjs
@@ -51,11 +51,37 @@ function toIso(ms) {
   return Number.isFinite(ms) ? new Date(ms).toISOString() : null;
 }
 
+// True when no subnet represented at the latest captured_at stamp still carries
+// older captured_at rows — the signature of a mid-batch loadStagedNeurons run.
+export async function neuronsSnapshotReadyForDailyRollup(db) {
+  if (!db?.prepare) return false;
+  try {
+    const result = await db
+      .prepare(
+        `SELECT 1 AS partial
+         FROM neurons n_latest
+         WHERE n_latest.captured_at = (SELECT MAX(captured_at) FROM neurons)
+           AND EXISTS (
+             SELECT 1 FROM neurons n_old
+             WHERE n_old.netuid = n_latest.netuid
+               AND n_old.captured_at < n_latest.captured_at
+           )
+         LIMIT 1`,
+      )
+      .bind()
+      .all();
+    return (result?.results?.length ?? 0) === 0;
+  } catch {
+    return false;
+  }
+}
+
 /**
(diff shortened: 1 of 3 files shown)


JSONbored (commented):
#1710 — CLOSE  (sequence vs #1667/#1701) — andriypolanski

src/neuron-history.mjs:neuronsSnapshotReadyForDailyRollup — a per-netuid probe that
skips rollupNeuronDaily with reason:"incomplete-snapshot" when a subnet still has rows
at MAX(captured_at) alongside older captured_at rows (the signature of a loadStagedNeurons mid-batch), so a daily row is never taken against a partial metagraph; handleScheduled also drains staged neurons first. Correct, tested, codecov green post-rebase. Composes with #1667/#1701 — sequence the rebases. 

However, there are merge conflicts, so this must be closed. Please resubmit as a new PR.

PR #1717, opened 2026-06-24
fix(list): sort null/missing values last regardless of direction

### LINKED ISSUE #1716: REST list sorter and MCP list sorter disagree on null/missing ordering
### What

`compareValues` / `sortRows` in `workers/list-query.mjs` (the REST list endpoints' sorter) handles `null`/missing sort values differently from `sortSubnets` in `src/mcp-server.mjs` (the MCP list sorter). For the **same** data and field, the two list surfaces return different orderings whenever the sort field is null/missing on some rows.

`sortRows` does `compareValues(a[key], b[key]) * direction`, and `compareValues` collapses a missing value to `""`:

```js
function compareValues(a, b) {
  if (typeof a === "number" && typeof b === "number") return a - b;
  return String(a ?? "").localeCompare(String(b ?? ""));
}
```

So a `null` compares as the empty string and its position **flips with direction**: it sorts to the FRONT for `order=asc` and to the BACK for `order=desc`.

### Why it is wrong

`sortSubnets` in `src/mcp-server.mjs` deliberately places null/missing **last regardless of direction** (and its comment claims to "mirror `compareValues`"), so the REST sorter is the divergent one.

Concrete: `/api/v1/economics?sort=emission_share&order=asc` (`emission_share` is null for unpriced subnets) surfaces unpriced subnets **first**; `/api/v1/health/...?sort=...&order=asc` can rank a row with a null metric as the "best". The MCP `list_subnets` tool, given the same rows, puts those nulls last.

### Expected

`null`/missing sort values sort **last** in both `asc` and `desc`, matching `sortSubnets`, so the REST list and MCP list agree on identical data.

### Repro

```
GET /api/v1/economics?sort=emission_share&order=desc
# a subnet with emission_share=null currently sorts FIRST; it should sort LAST
```

seekmistar01:
## Summary

The REST list sorter (`sortRows`/`compareValues` in `workers/list-query.mjs`) handled `null`/missing sort values inconsistently with the MCP list sorter (`sortSubnets` in `src/mcp-server.mjs`): a missing value collapsed to `""`, so its position **flipped with direction** (front on `asc`, back on `desc`). For identical data, the REST list and MCP list disagreed.

This makes `sortRows` place `null`/missing **last regardless of direction**, matching `sortSubnets`.

## Changes

- `workers/list-query.mjs`: `sortRows` now sorts null/undefined values last in both directions; non-null comparison still uses `compareValues * direction`.
- `tests/list-query.test.mjs`: regression tests over `/api/v1/economics?sort=emission_share` asserting null/missing rows stay last for both `asc` and `desc`.

## Validation

```
npx vitest run tests/list-query.test.mjs   # 16 passed
npx prettier --write … && npx eslint …      # clean
```

Closes #1716.

### DIFF
--- a/tests/list-query.test.mjs
+++ b/tests/list-query.test.mjs
@@ -113,6 +113,44 @@ describe("list-query pagination order", () => {
   });
 });
 
+describe("list-query null/missing sort ordering", () => {
+  // Rows whose sort field is null/missing must sort LAST in both directions,
+  // matching sortSubnets() in src/mcp-server.mjs so the REST list and MCP list
+  // agree on identical data.
+  const data = {
+    subnets: [
+      { netuid: 1, emission_share: 0.05 },
+      { netuid: 2 }, // missing emission_share
+      { netuid: 3, emission_share: 0.09 },
+      { netuid: 4, emission_share: null },
+    ],
+  };
+
+  test("ascending keeps real values ordered with null/missing last", () => {
+    const result = applyQueryFilters(
+      data,
+      query("/api/v1/economics?sort=emission_share&order=asc"),
+      "economics",
+    );
+    assert.deepEqual(
+      result.data.subnets.map((r) => r.netuid),
+      [1, 3, 2, 4],
+    );
+  });
+
+  test("descending still keeps null/missing last (not first)", () => {
+    const result = applyQueryFilters(
+      data,
+      query("/api/v1/economics?sort=emission_share&order=desc"),
+      "economics",
+    );
+    assert.deepEqual(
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
Correct, and your justification is actually the better-argued one — explicitly mirroring sortSubnets's in-comparator null-last rule for REST/MCP parity ([#1716](https://github.com/JSONbored/metagraphed/issues/1716) is really the same defect as #1702). But #1703 lands the same behavior, is defensive about null rows, and came first, so I'm closing this as a duplicate and folding your parity note into that merge. Thanks for catching the divergence.

PR #1722, opened 2026-06-24
fix(badge): do not round a sub-1 uptime ratio up to 100%

### LINKED ISSUE #1721: Uptime badge rounds a sub-100% ratio up to 100%
### What

### Why it is wrong

### Expected

seekmistar01:
## Summary

`formatUptimePercent` rounded to two decimals with `Math.round(ratio * 10000) / 100`, so any ratio in `[0.99995, 1)` rounded up to `100` and the uptime badge falsely read `100%` for a service below perfect uptime (e.g. `0.99996 -> "100%"`).

The contract is that only an exact full ratio is `100%` (`0.9983 -> "99.83%"`, `1 -> "100%"`). The fix clamps a sub-1 ratio that rounds to 100 down to the largest two-decimal value below it (`99.99%`), leaving an exact `1` as `100%`.

## Changes

- `src/badge.mjs`: clamp in `formatUptimePercent` (now exported for direct testing, matching `scoreColor`/`gradeColor`).
- `tests/badge.test.mjs`: regression tests for the boundary.

## Validation

```
npx vitest run tests/badge.test.mjs   # 28 passed
npx prettier --write … && npx eslint …  # clean
```

Closes #1721.


JSONbored (commented):
Appreciate the boundary thinking, but this fixes a branch the live code can't reach. formatUptimePercent's only caller gets rel.uptime_ratio after scoreFromStats has already rounded 0.99996 to 1 via toFixed(4), so by the time your pct >= 100 && value < 1 guard runs, value is exactly 1 and the guard is skipped — the badge still reads "100%". Your test passes only because it calls formatUptimePercent(0.99996) directly, a raw ratio the pipeline never emits. Separately, round-then-clamp still rounds up at non-100 boundaries (0.92495 → 92.5%). The real fix is upstream at scoreFromStats ([#1799](https://github.com/JSONbored/metagraphed/issues/1799)). Closing as superseded — landing this would let us close [#1721](https://github.com/JSONbored/metagraphed/issues/1721) while the badge is unchanged.

Issue #1005, opened 2026-06-17
Stable surface IDs + deprecation map (stop orphaning health history on rename)


JSONbored:
Part of #999 · Phase 2.

## Context
`endpoint.id` is computed as `endpoint-{surface.id}` (`scripts/build-artifacts.mjs:1943`) and `surface.id` is hand-authored from the display name. Renaming a surface **orphans** its health history (probes keyed by the old surface/endpoint id) and breaks the endpoint link. There is no stable identifier decoupled from the display name and no deprecation/redirect map.

## Deliverables
- Give each surface a **stable** canonical key (hash/UUID derived from `netuid|kind|normalized-url`, not the display name); keep the human id as a display alias.
- A `deprecated_id → current_id` remap so health history + endpoint links migrate on rename instead of orphaning.

## Requirements / acceptance
- Renaming a surface preserves its health history + endpoint linkage (test asserts a rename keeps the stable id).
- One-time migration documented; existing IDs aliased (URL/ID stability preserved).
- `npm test` + `npm run validate` clean.

JSONbored:
## Scoping complete — grounded design (ready for a focused execution)

### Confirmed problem (file:line)
- **`surface.id` is hand-authored**, format `sn-{netuid}-{slug}-{kind}` (e.g. `registry/subnets/actual.json:47` → `"sn-95-actual-website"`). So an author renaming a surface (slug change) **changes the id**.
- **`endpoint.id = endpoint-{surface.id}`** (derived in `scripts/build-artifacts.mjs`), so the endpoint link breaks on rename too.
- **D1 history is keyed on the mutable `surface_id`** — a rename orphans ALL of it:
  - `surface_checks (surface_id TEXT NOT NULL, …)` — probe history (`migrations/0001_health.sql:11`, indexed `surface_id, checked_at`)
  - `surface_status (surface_id TEXT PRIMARY KEY, …)` — current status (`0001_health.sql:30`)
  - `surface_uptime_daily (… PRIMARY KEY (surface_id, day))` — daily uptime (`migrations/0003_uptime_history.sql:9`)

### Recommended design
- **Stable key = short hash of `netuid|kind|normalized-url`** (a surface is "the same" as long as its callable URL + kind + subnet are; the display slug is cosmetic). Add it as `surface.key` (canonical); keep the hand-authored `id` as the human-readable display alias. Endpoint links + D1 use `key`.
[...]

PR #1733, opened 2026-06-24
fix(reliability): bucket per-surface stats by stable surface_key

### LINKED ISSUE #1732: computeReliability buckets per-surface stats by unstable surface_id, inflating reliability surface_count across renames
## Summary

`computeReliability` in `src/reliability.mjs` aggregates per-surface uptime stats keyed on `row.surface_id`, not the stable `surface_key` introduced by the surface-identity rename model (#1005). When a surface is renamed within the scoring window, its history spans **one** `surface_key` but **two** `surface_id` values — so it splits into two buckets, inflating `surface_count` and fragmenting the per-surface reliability score.

## Where

`src/reliability.mjs` — `computeReliability()`:

```js
const surface = bySurface.get(row.surface_id) || { ... };
// ...
bySurface.set(row.surface_id, surface);
// ...
surface_count: bySurface.size,
```

## Impact

- **`loadSubnetReliability()` (`src/health-serving.mjs`) — live bug, no workaround.** Its query does `GROUP BY COALESCE(surface_key, surface_id), day` and emits `MAX(surface_id) AS surface_id` per group, so a surface renamed mid-window produces rows with one stable `surface_key` but a different `surface_id` on each side of the rename. `computeReliability` then double-counts it, and `subnet.surface_count` reports 2 for what is one physical surface. This value is served on the subnet reliability surface.
- **`formatUptime()` (`src/health-serving.mjs`) — already masked.** It works around the root cause by pre-rewriting each row's `surface_id` to the stable key before calling `computeReliability`:

```js
const reliabilityRows = (rows || []).map((row) => ({
  ...row,
  surface_id: surfaceLookupKey(row),
}));
```

This is a fragile band-aid that hides the defect from one caller while the other still trips on it.

## Expected

`computeReliability` should key per-surface aggregation on the stable `surface_key` (falling back to `surface_id` for legacy rows that predate it), so a renamed surface stays one bucket. The `formatUptime` workaround then becomes unnecessary.

## Repro (unit)

Call `computeReliability` (or `loadSubnetReliability`) with two rows sharing one `surface_key` but different `surface_id` values across days → today `subnet.surface_count === 2`; it should be `1`.

## Scope

Self-contained: `src/reliability.mjs` + remove the dead workaround in `src/health-serving.mjs`, with unit tests. No schema/contract regeneration; the served contract is unchanged.

kiannidev:
## Problem

`computeReliability` (`src/reliability.mjs`) aggregated per-surface uptime stats keyed on the **unstable** `row.surface_id` instead of the stable `surface_key` from the surface-identity rename model (#1005). A surface renamed within the scoring window has one `surface_key` but two `surface_id` values, so it split into two buckets — **inflating `subnet.surface_count`** and fragmenting the per-surface score.

Two callers, two symptoms:

- **`loadSubnetReliability` (`src/health-serving.mjs`) — live bug, no workaround.** Its query does `GROUP BY COALESCE(surface_key, surface_id), day` and emits `MAX(surface_id)` per group, so a mid-window rename yields rows with one `surface_key` but a different `surface_id` per side. `computeReliability` double-counted it → `surface_count` reported 2 for one physical surface.
- **`formatUptime` (`src/health-serving.mjs`) — already masked** by a fragile pre-rewrite of `surface_id` to the stable key before calling `computeReliability`.

Closes #1732.

## Fix

- `computeReliability` now keys per-surface aggregation (and the returned `surfaces` map) on `surface_key`, falling back to `surface_id` for legacy rows that predate it. A renamed surface stays **one** bucket.
- Removed the now-redundant `surface_id` pre-rewrite workaround in `formatUptime`.
- No schema/contract change — the served contract artifacts are byte-identical after `npm run build` (only the gate-excluded `r2-manifest.json` re-hashes).

## Tests

Added unit coverage for the rename case:
- `computeReliability` — two rows, same `surface_key`, different `surface_id` → `surface_count === 1`, `surfaces` keyed by `surface_key`.
- `loadSubnetReliability` — D1-backed rename rows → `surface_count === 1`.

The pre-existing `formatUptime` "groups renamed uptime rows by stable surface_key" test now exercises the real root-cause fix rather than the workaround.

## Validation

```
npm run lint                     # pass
npm run format:check             # pass
npm run validate:contract-drift  # Contract drift validation passed.
npm run build && npx vitest run  # 100 files, 1953 tests passed
```


JSONbored (approved):
Confirmed effective on the live path, which is the part that mattered: loadSubnetReliability projects COALESCE(surface_key, surface_id) AS surface_key (health-serving.mjs:1145) and groups on it while emitting MAX(surface_id), so a mid-window rename yields two surface_ids under one surface_key — your row.surface_key || row.surface_id keying collapses it back to one bucket and stops the surface_count inflation. Dropping the formatUptime pre-rewrite workaround is the right cleanup now that the root is fixed, and tying it to the #1005 identity model is correct. Sequence it with #1799 (adjacent lines in reliability.mjs). Approving.

PR #1739, opened 2026-06-24
fix(health): normalize unrecognized probe statuses before rollup

### LINKED ISSUE #1738: summarizeRows reports subnet health as ok when all surface statuses are unrecognized
## Summary

`summarizeRows` (`src/health-serving.mjs`) and the prober's `summarizeGroup` (`src/health-prober.mjs`) bucket surface rows by the raw `row.status` string. When a status is **outside** the canonical probe set (`ok` / `degraded` / `failed` / `unknown`), it lands in a dynamic key (e.g. `weird: 2`) that `rollupSubnetStatus` never sees.

Because `failed === 0 && degraded === 0`, the subnet rollup returns **`"ok"`** even when **every** surface has an unrecognized or missing status — a false healthy signal on live health overlays and the 15-minute cron KV snapshot.

## Where

```js
// health-serving.mjs + health-prober.mjs (identical pattern)
counts[row.status] = (counts[row.status] || 0) + 1;
// ...
status: rollupSubnetStatus({ ...counts, total: rows.length }),
```

`rollupSubnetStatus` (`src/health-probe-core.mjs`) only reads the four canonical buckets. Unrecognized strings bypass all of them.

## Impact

- **`overlaySubnetHealth` / live D1 health** — subnet badge can show `ok` while no surface is actually ok.
- **`runHealthProber` KV `health:current`** — per-subnet rollup written every 15 minutes can report `ok` with `ok_count: 0`.
- Forward-compat / corrupted rows (null status, typo, future enum value) all hit this path.

## Expected

Unrecognized / null / missing status values should normalize to **`unknown`** before counting, so an all-unrecognized subnet rolls up to **`unknown`**, not **`ok`**.

## Repro (unit)

```js
summarizeRows([{ status: "weird" }, { status: "weird" }])
// today: { status: "ok", ok_count: 0, unknown_count: 0 }
// expected: { status: "unknown", unknown_count: 2 }
```

## Scope

Self-contained: add `normalizeProbeStatus` next to `rollupSubnetStatus` in `health-probe-core.mjs`, use it in both aggregators, fix the existing test that documents the wrong behavior. No schema/contract regeneration.

kiannidev:
## Context

Synced to latest `main` (includes maintainer's #1731 webhook determinism fix, #1735 RELEASING docs, #1729 Smithery CI). This is a separate health-rollup correctness bug still present on current `main`.

Closes #1738.

## Problem

`summarizeRows` (`src/health-serving.mjs`) and `summarizeGroup` (`src/health-prober.mjs`) counted rows by the raw `row.status` string. Values outside the canonical probe set (`ok` / `degraded` / `failed` / `unknown`) went into dynamic keys that `rollupSubnetStatus` never reads.

When **every** surface had an unrecognized or missing status, `failed === 0 && degraded === 0` still held → subnet rollup returned **`"ok"`** with `ok_count: 0`. That false healthy signal hits:

- Live health overlays (`overlaySubnetHealth`, D1 fallback)
- The 15-minute cron KV snapshot (`health:current` per-subnet rollup)

## Fix

- Add `normalizeProbeStatus` + `PROBE_STATUS_VALUES` next to `rollupSubnetStatus` in `src/health-probe-core.mjs` (single canonical home for probe status semantics).
- Use it in both aggregators so null / missing / forward-compat strings count as **`unknown`**.
- Flip the existing test that documented the wrong `"ok"` behavior; add coverage for null/missing status.

No schema or contract change — served artifacts are unchanged.

## Validation

```
npm run lint                     # pass
npm run format:check             # pass
npm run validate:contract-drift  # pass
npx vitest run tests/health-serving.test.mjs tests/health-probe-core.test.mjs tests/health-prober.test.mjs
# 296 tests passed
```


JSONbored (approved):
This is the complete fix. You put normalizeProbeStatus/PROBE_STATUS_VALUES next to rollupSubnetStatus (the right canonical home) and apply it in both summarizeRows (live overlay) and summarizeGroup in health-prober.mjs — which feeds the 15-minute cron KV snapshot and has the identical counts[row.status] bug. Covering both paths is what makes this the keeper. Approving.

PR #1785, opened 2026-06-25
fix(ci): keep rpc secrets out of python dependency steps

JSONbored:
### Motivation
- Scheduled CI steps previously injected `SUBTENSOR_RPC_URL`/`EVENTS_RPC_URL` into steps that run unpinned/pinned PyPI boundaries (`uvx`/`uv run`), which exposes a hidden archive RPC URL to package build/install hooks and third-party runtime code and creates a supply-chain information-leak. 
- The intent of the change is to preserve the original security boundary so only trusted code paths (manual/local runs) may supply private RPC endpoints while CI uses the public `finney` default.

### Description
- Removed the `SUBTENSOR_RPC_URL` environment from the `refresh-metagraph` step that runs `uvx --from bittensor==10.4.0 python scripts/fetch-metagraph-native.py` so the CI PyPI execution boundary no longer receives the secret. 
- Removed the secret-derived `EVENTS_RPC_URL` assignment from the `refresh-events` step that runs `uv run --with substrate-interface==1.8.1 python scripts/fetch-events.py` so the poller uses its public `finney` default in CI. 
- Updated `scripts/fetch-metagraph-native.py` to default `--network` to `"finney"` instead of reading `os.environ.get("SUBTENSOR_RPC_URL")`, preserving an explicit `--network` override for trusted local/manual runs. 
- Adjusted workflow comments to document the preserved boundary and the intended local/manual override path without placing hidden RPC endpoints into the Python dependency execution environment.

### Testing
- Ran `npm run validate:workflows` and it succeeded. 
- Ran `npm run scan:public-safety` and it succeeded. 
- Ran `npm run format:check` (via `prettier --check`) on the modified files and it succeeded. 
- Ran `python3 -m py_compile scripts/fetch-metagraph-native.py scripts/fetch-events.py` and it succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cac300170832cb9062dc023b5a856)

### DIFF
--- a/.github/workflows/refresh-events.yml
+++ b/.github/workflows/refresh-events.yml
@@ -75,17 +75,12 @@ jobs:
           CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
           CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
 
-      # substrate-interface is PINNED to an exact, verified version: this step runs
-      # with Cloudflare secrets in scope, so we never auto-pull a future (possibly
-      # compromised) release from PyPI. Bump deliberately after re-verifying.
+      # substrate-interface is PINNED to an exact, verified version. Keep hidden
+      # RPC endpoints out of this PyPI execution boundary; the poller uses its
+      # public finney default unless a trusted local/manual run sets EVENTS_RPC_URL.
       - name: Poll finney events (first-party, public RPC, no key)
         run: uv run --with substrate-interface==1.8.1 python scripts/fetch-events.py
         env:
-          # Chain RPC endpoint, behind a SECRET so the private archive node's URL is
-          # never exposed in the repo or run logs once we switch to it (ADR 0012 /
-          # #1349). Falls back to the public finney endpoint when the secret is unset,
-          # so swapping = set/clear the one `SUBTENSOR_RPC_URL` secret — no code change.
-          EVENTS_RPC_URL: ${{ secrets.SUBTENSOR_RPC_URL || 'wss://entrypoint-finney.opentensor.ai:443' }}
           # Bounded overlap: ~250 blocks ≈ 50 min of chain is re-scanned every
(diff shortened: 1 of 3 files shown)


JSONbored:
Triage note (not auto-merging/closing): the supply-chain concern here is valid — `SUBTENSOR_RPC_URL`/`EVENTS_RPC_URL` in the `uvx`/`uv` steps exposes the (eventual private-archive) URL to the broad PyPI dependency tree + install hooks, not just to substrate-interface.

**But this partially reverses the just-merged #1774**, which deliberately routed chain RPC through that secret for the self-host migration (the `chain-rpc-secret` effort / ADR 0012). Dropping the secret from CI means the scheduled refresh-events + refresh-metagraph jobs always use public finney, never the private node — a strategy decision, not a pure security fix.

Holding this for the chain-rpc-secret owner to reconcile. The likely right shape is to keep the private-node routing but isolate the secret from the PyPI execution boundary (e.g. a tightly-scoped wrapper that opens the connection without leaving the URL in the broad env for dep/install code), rather than reverting to public-only. Leaving open.

JSONbored:
Closing — right concern, wrong shape for right now.

**No live leak to stop:** `SUBTENSOR_RPC_URL` is not currently set as a repo secret, so the `secrets.SUBTENSOR_RPC_URL` references resolve empty and the `uvx`/`uv` steps already fall back to public finney. #1774's routing plumbing is dormant; nothing sensitive is in those environments today.

**This reverts #1774:** that PR deliberately added the `SUBTENSOR_RPC_URL` routing for the self-host migration (the active `chain-rpc-secret` effort / ADR 0012). Blunt-removing it here would undo intentional, just-merged work for a future-only risk.

**The right fix, when the private archive node actually comes online and the secret is set,** is to keep the private-node routing but isolate the URL from the broad PyPI execution boundary — e.g. a tightly-scoped connector that opens the WS connection without leaving the endpoint in the shared env for transitive deps / install hooks. That belongs to the chain-rpc-secret effort, not a revert-to-public. Logging the concern here for that work; closing this PR as not-the-correct-shape.

PR #1793, opened 2026-06-25
fix(api): drop malformed rows from subnet metagraph and validator arrays

codevector96:
### What

`src/metagraph-neurons.mjs` builds the `/subnets/{n}/metagraph` and
`/subnets/{n}/validators` responses by mapping `formatNeuron` over the D1 rows:

    neuron_count: rows.length,
    neurons: rows.map(formatNeuron),

`formatNeuron(row)` returns `null` for a null / non-object row, but the result
is not filtered. A malformed row therefore leaks a `null` element into the
`neurons` / `validators` array — breaking the array-of-Neuron contract clients
rely on — and `neuron_count` (`rows.length`) no longer matches the well-formed
entries.

### Expected

Malformed rows are dropped so the array only holds real Neuron objects and the
count tracks the array, the same way the sibling feed builders already do it:

- `src/blocks.mjs` — `(rows || []).map(formatBlock).filter(Boolean)`
- `src/extrinsics.mjs` — `(rows || []).map(formatExtrinsic).filter(Boolean)`

### Notes

The healthy D1 path never returns null rows, so this is a robustness/consistency
fix with no change to normal responses: it brings the neuron builders in line
with the established sibling pattern and keeps `neuron_count === neurons.length`
an invariant.


JSONbored (approved):
Fine as a consistency/invariant fix. formatNeuron only returns null for a non-object row (!row || typeof row !== "object"), which the healthy D1 path never produces — so this changes nothing in normal operation, as you say. But filtering + counting off the filtered array restores neuron_count === neurons.length and matches the blocks.mjs/extrinsics.mjs sibling builders, so it's worth taking for parity. Approving.

PR #1794, opened 2026-06-25
fix(snippets): send a cookie-location credential via the Cookie header

minion1227:
## Summary
`generateServiceSnippets` only handled `auth.location` of `header`/`query`, but the
surface schema's enum is `header|query|cookie`. A `location: "cookie"` credential
was coerced to a header named after the cookie — `-H 'session_id: <token>'` — so the
cookie never reached the server and the generated "first call" snippet failed to
authenticate. This change renders a cookie credential as a proper
`-H 'Cookie: session_id=<token>'` in all three languages, and adds a regression test.

## Validation
- `npx vitest run tests/integration-snippets.test.mjs` — 18/18 pass (incl. new test)
- eslint + prettier clean; `git diff --check` clean
- Diff is src + test only — no surface currently uses cookie auth, so no build-artifact churn / no contract drift.


JSONbored (approved):
Correct completeness fix — the schema's auth.location enum is header|query|cookie, and the old code silently coerced cookie to a header named after the cookie (-H 'session_id: …'), which never authenticates. Rendering it as -H 'Cookie: name=value' is right. Fair flag in your own notes that no surface uses cookie auth today, so this is forward-looking rather than fixing an active break — fine, the generator should cover the whole enum. Get the full CI run green (only 2 checks ran). Approving.

PR #1799, opened 2026-06-25
fix(reliability): do not round a sub-perfect uptime ratio up to a perfect 1

### LINKED ISSUE #1796: Uptime badge renders 100% for a sub-perfect uptime ratio (scoreFromStats rounds 0.99996 up to 1)
## Summary

The reliability uptime badge renders `100%` for a subnet whose uptime is not actually perfect (e.g. 99.996%). `scoreFromStats` in `src/reliability.mjs` rounds the displayed `uptime_ratio` to 4 decimals with `Number(uptimeRatio.toFixed(4))`, and `(0.99996).toFixed(4) === "1.0000"`, so a sub-1 ratio is stored as an exact `1` before it ever reaches the formatter.

## Why the existing formatter guard does not catch it

The badge path is:

- `reliabilityContent` -> `loadReliability` (defaults to `loadReliabilityAggregate`, `src/health-serving.mjs:1168`) -> `scoreFromStats` (`src/reliability.mjs:30`) -> returns `uptime_ratio` (line 55)
- `src/badge.mjs:245`: `formatUptimePercent(rel.uptime_ratio)`

`formatUptimePercent` only demotes a value when `value < 1`. Because `scoreFromStats` has already collapsed `0.99996` to exactly `1`, the `value < 1` guard never fires and the badge prints `100%`. The rounding happens upstream of the formatter, so the information needed to demote is already lost by the time the formatter runs.

## Reproduction

`GET /api/v1/subnets/{netuid}/badge.svg?metric=uptime` for a subnet whose 30-day rollup is `SUM(ok_count)=24999`, `SUM(samples)=25000` (uptime = 99.996%).

- Expected: badge text `99.99%` (sub-perfect uptime must not read as a perfect badge).
- Actual: badge text `100%`.

The same applies to the per-surface / per-subnet `uptime_ratio` that `computeReliability` exposes on the live health API: a 99.996% surface reports `uptime_ratio: 1`, overstating it as perfect.

## Expected

Only a genuine `okCount === samples` (exact `1`) should report `uptime_ratio: 1`. A sub-1 ratio that rounds up to `1` should clamp to the largest 4-decimal value below 1 (`0.9999`), so the badge reads `99.99%` and the health API does not claim a perfect ratio.

## Notes

- `score` / `grade` are unaffected (they are computed from the raw `uptimeRatio`, not the rounded display field).
- Health is computed live (never from a committed artifact), so the fix does not change any generated `public/` output.

joaovictor91123:
## Summary

The reliability uptime badge renders `100%` for a subnet that is not actually at perfect uptime (e.g. 99.996%). `scoreFromStats` (`src/reliability.mjs`) rounded the displayed `uptime_ratio` with `Number(uptimeRatio.toFixed(4))`, and `(0.99996).toFixed(4) === "1.0000"`, so a sub-1 ratio was stored as an exact `1` before reaching the formatter.

The badge path is `reliabilityContent` -> `loadReliabilityAggregate` (`src/health-serving.mjs:1196`) -> `scoreFromStats` -> `formatUptimePercent(rel.uptime_ratio)` (`src/badge.mjs:245`). `formatUptimePercent` only demotes when `value < 1`, but the ratio is already `1` by the time it gets there, so the badge prints `100%`. The rounding is upstream of the formatter, so the demotion guard can never fire for this input.

## What Changed

- `src/reliability.mjs`: clamp a sub-1 ratio that rounds up to `1` to the largest 4-decimal value below 1 (`0.9999`) at the source, in a small `displayUptimeRatio` helper. Only a genuine `okCount === samples` ratio keeps the exact `1`. `score`/`grade` are unaffected (they are computed from the raw `uptimeRatio`, not the rounded display field).
- `tests/health-serving.test.mjs`: assert a 24999/25000 ratio reports `uptime_ratio: 0.9999` (not `1`), an exact 25000/25000 still reports `1`, and an ordinary sub-1 ratio is unchanged.

This fixes the badge regardless of any change to `formatUptimePercent`, and also stops the live health API (`computeReliability`) from reporting `uptime_ratio: 1` for a 99.996% surface.

Fixes #1796

## Registry Safety

- [x] No public API/OpenAPI/schema changes (health is computed live; no generated `public/` output changes).

## Validation

- [x] Logic verified: `(0.99996).toFixed(4)` rounds to `1`; `displayUptimeRatio` returns `0.9999` for it and preserves an exact `1`. Existing reliability assertions (e.g. `0.8241`) are unchanged.
- Added a unit test covering the sub-perfect clamp, exact-1 preservation, and an ordinary ratio.


JSONbored (approved):
This is the correct fix, and it's the only one of the three that changes what a real 99.996% surface renders. You've put the clamp where the collapse actually happens — displayUptimeRatio inside scoreFromStats — so the uptime_ratio JSON field stops reporting 1 for a sub-perfect surface and the badge follows. Issue [#1796](https://github.com/JSONbored/metagraphed/issues/1796) (closed/"slop") had the more accurate diagnosis than the still-open #1721, and your trace through loadReliabilityAggregate → scoreFromStats → formatUptimePercent nails why the downstream guard "can never fire." score/grade stay on the raw ratio — good.

PR #1824, opened 2026-06-25
feat(chain): capture Balances.Transfer events in the chain-event index

### LINKED ISSUE #1814: feat(chain): capture Balances.Transfer events in the chain-event index
## Context

Part of #1345 — first-party developer block explorer.

`fetch-events.py` filters `System.Events` to `SubtensorModule` events only. Every native TAO transfer emitted by the `Balances` pallet is silently dropped. This means:

- The account entity API (`/api/v1/accounts/{ss58}`) has no transfer history — only stake events
- There is no way to trace TAO flow between wallets using first-party chain data
- Wallet/account pages in the block explorer UI (#1350) cannot show sent/received history

The `account_events` schema and D1 load path already handle arbitrary `event_kind` values — this requires no migration, only a new extractor.

## Requirements

- Add a `Transferred` case to `EXTRACTORS` in `scripts/fetch-events.py` handling `Balances.Transfer` (fields: `from`, `to`, `amount`)
- Map to the existing account_events row shape:
  - `event_kind = "Transferred"`
  - `hotkey = from` (sender ss58)
  - `coldkey = to` (recipient ss58)
  - `amount_tao = amount / 1e9`
  - `netuid = null`, `uid = null`
- The filter at `if e.get("module_id") != "SubtensorModule": continue` must be broadened to also pass `Balances` events through the extractor (or restructured to a per-module dispatch)
- The same `extract()` function is reused by `scripts/backfill-events.py` via importlib — no separate change needed there
- Update `buildAccountEvents()` / the account entity API documentation to note `Transferred` as a valid `event_kind`
- Add unit tests for the new extractor in `scripts/test_fetch_events.py`

## Acceptance criteria

- `GET /api/v1/accounts/{ss58}` returns `Transferred` events for an active hotkey that has sent or received TAO
- Backfilling a range that contains `Balances.Transfer` transactions produces corresponding rows in `account_events`
- Existing stake/registration event tests unaffected

JSONbored:
## Summary

- Adds `_transfer` extractor to `EXTRACTORS` in `scripts/fetch-events.py` for `Balances.Transfer` events
- Broadens the per-event module filter from `SubtensorModule`-only to `("SubtensorModule", "Balances")` in both `fetch-events.py` and `backfill-events.py`
- 8 new unit tests in `scripts/test_fetch_events.py`; 29/29 pass

## Schema mapping

`Balances.Transfer` → `account_events` row:

| Chain field | Column | Notes |
|---|---|---|
| `from` | `hotkey` | sender ss58 |
| `to` | `coldkey` | recipient ss58 |
| `amount` | `amount_tao` | `amount / 1e9` |
| — | `event_kind` | `"Transfer"` |
| — | `netuid`, `uid` | `null` |

Handles both positional-list and named-dict attribute forms for runtime compatibility. Shape drift (wrong arg count, bad addresses) degrades to null fields — never raises.

## Why no schema change

The `account_events` table already accepts arbitrary `event_kind` values; `"Transfer"` is just a new value. No migration needed.

## Test plan

- [x] 29/29 unit tests pass (`python3 scripts/test_fetch_events.py`)
- [ ] After merge, dispatch a backfill over a range known to contain `Balances.Transfer` txs and verify `event_kind = "Transfer"` rows land in D1
- [ ] `GET /api/v1/accounts/{ss58}` for an active wallet returns `Transfer` events

Closes #1814

### DIFF
--- a/scripts/backfill-events.py
+++ b/scripts/backfill-events.py
@@ -168,7 +168,7 @@ def main():
         for event_index, ev in enumerate(events):
             v = ev.value if isinstance(ev.value, dict) else {}
             e = v.get("event", {}) if isinstance(v.get("event"), dict) else {}
-            if e.get("module_id") != "SubtensorModule":
+            if e.get("module_id") not in ("SubtensorModule", "Balances"):
                 continue
             ent = _fe.extract(e.get("event_id"), e.get("attributes"))
             if ent is None:
--- a/scripts/fetch-events.py
+++ b/scripts/fetch-events.py
@@ -216,6 +216,17 @@ def _root(a):  # {coldkey} (named) or [coldkey]
     return {"coldkey": _ss58(ck)}
 
 
+def _transfer(a):  # Balances.Transfer: [from, to, amount] or {from, to, amount}
+    if isinstance(a, dict):
+        sender, recipient, amount = a.get("from"), a.get("to"), a.get("amount")
+    else:
+        sender = a[0] if len(a) > 0 else None
+        recipient = a[1] if len(a) > 1 else None
+        amount = a[2] if len(a) > 2 else None
+    # hotkey = sender, coldkey = recipient (pragmatic reuse of the index columns)
+    return {"hotkey": _ss58(sender), "coldkey": _ss58(recipient), "amount_tao": _tao(amount)}
+
+
 EXTRACTORS = {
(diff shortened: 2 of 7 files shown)


JSONbored:
Superseded by #1825 (clean branch from main, no conflict)

PR #1841, opened 2026-06-25
fix(events): emit coldkey/hotkey in chain order for TakeIncreased/TakeDecreased

### LINKED ISSUE #1840: TakeIncreased/TakeDecreased chain events store coldkey and hotkey swapped
## Summary

The chain-event extractor `_take_changed` in `scripts/fetch-events.py` reads the `TakeIncreased` / `TakeDecreased` Subtensor events with the coldkey and hotkey positions reversed, so every such row written to `account_events` has the two SS58 addresses swapped.

## Detail

Subtensor emits both events with the coldkey first:

- `Self::deposit_event(Event::TakeIncreased(coldkey, hotkey, take));` (pallets/subtensor/src/staking/increase_take.rs)
- `Self::deposit_event(Event::TakeDecreased(coldkey, hotkey, take));` (pallets/subtensor/src/staking/decrease_take.rs)

The variants are declared as positional tuples `TakeIncreased(T::AccountId, T::AccountId, u16)` (no named fields), so substrate-interface decodes the attributes as a positional list and the `else` branch of `_take_changed` runs:

```python
def _take_changed(a):  # TakeDecreased/TakeIncreased: {hotkey, coldkey, ...} or [hotkey, coldkey, ...]
    ...
    hk = a[0] if len(a) > 0 else None   # a[0] is the COLDKEY on-chain
    ck = a[1] if len(a) > 1 else None   # a[1] is the HOTKEY on-chain
    return {"hotkey": _ss58(hk), "coldkey": _ss58(ck)}
```

This treats position 0 as the hotkey and position 1 as the coldkey -- the reverse of what the chain emits.

The structurally identical sibling `_delegate_added` (for `DelegateAdded(coldkey, hotkey, take)`) reads it correctly as `ck = a[0]`, `hk = a[1]`. Only `_take_changed` inverted the order; its own comment also asserts the wrong (hotkey-first) ordering.

## Observable impact

Both values are valid SS58 addresses, so nothing trips a shape/validation guard -- the row is written with the keys transposed. Any consumer querying take-change activity by hotkey (or by coldkey) gets these events filed under the wrong account.

## Fix

Swap the positional reads to `ck = a[0]`, `hk = a[1]` (mirroring `_delegate_added`), and correct the comment. Landed in #1835 / #1816.

joaovictor91123:
## Summary

`_take_changed` in `scripts/fetch-events.py` decoded the `TakeIncreased` / `TakeDecreased` Subtensor events with the coldkey and hotkey positions reversed, so every such row written to `account_events` stored the two SS58 keys swapped.

## What Changed

Subtensor emits both events coldkey-first:

- `Self::deposit_event(Event::TakeIncreased(coldkey, hotkey, take));` (pallets/subtensor/src/staking/increase_take.rs)
- `Self::deposit_event(Event::TakeDecreased(coldkey, hotkey, take));` (pallets/subtensor/src/staking/decrease_take.rs)

The variants are positional tuples `Take*(T::AccountId, T::AccountId, u16)` with no named fields, so substrate-interface decodes the attributes as a list and the list branch of `_take_changed` runs. It read `a[0]` as the hotkey and `a[1]` as the coldkey -- the reverse of the chain order. Both values are valid SS58 addresses, so `_ss58()` and the shape-drift guard pass and the row is written transposed.

The fix reads `a[0]=coldkey`, `a[1]=hotkey`, matching:
- the structurally identical sibling `_delegate_added` (`DelegateAdded(coldkey, hotkey, take)`), and
- the coldkey-first convention already used by `_stake` and `_moved`.

Added `TakeDelegateExtractorTest` covering the positional form, the named-dict form, and key-order parity with `DelegateAdded`.

Fixes #1840

## Registry Safety

- [x] No public API/OpenAPI/schema changes (extractor mapping only).

## Validation

- [x] Verified against the Subtensor source that both events emit `(coldkey, hotkey, take)`.
- [x] `python3 scripts/test_fetch_events.py` covers the new extractor assertions (stdlib unittest, zero deps).

### DIFF
--- a/scripts/fetch-events.py
+++ b/scripts/fetch-events.py
@@ -244,12 +244,15 @@ def _delegate_added(a):  # DelegateAdded: {coldkey, hotkey, take} or [coldkey, h
     return {"coldkey": _ss58(ck), "hotkey": _ss58(hk)}
 
 
-def _take_changed(a):  # TakeDecreased/TakeIncreased: {hotkey, coldkey, ...} or [hotkey, coldkey, ...]
+def _take_changed(a):  # TakeDecreased/TakeIncreased: {coldkey, hotkey, take} or [coldkey, hotkey, take]
+    # Subtensor emits these coldkey-first: Event::TakeIncreased(coldkey, hotkey, take)
+    # / TakeDecreased(coldkey, hotkey, take). The variants are positional tuples, so
+    # the list branch must read a[0]=coldkey, a[1]=hotkey (same order as DelegateAdded).
     if isinstance(a, dict):
-        hk, ck = a.get("hotkey"), a.get("coldkey")
+        ck, hk = a.get("coldkey"), a.get("hotkey")
     else:
-        hk = a[0] if len(a) > 0 else None
-        ck = a[1] if len(a) > 1 else None
+        ck = a[0] if len(a) > 0 else None
+        hk = a[1] if len(a) > 1 else None
     return {"hotkey": _ss58(hk), "coldkey": _ss58(ck)}
 
 
--- a/scripts/test_fetch_events.py
+++ b/scripts/test_fetch_events.py
@@ -365,5 +365,32 @@ class StakeAlphaExtractorTest(unittest.TestCase):
         self.assertIsNone(result["alpha_amount"])
 
 
+class TakeDelegateExtractorTest(unittest.TestCase):
+    # Subtensor emits TakeIncreased/TakeDecreased/DelegateAdded coldkey-first:
+    # Event::Take*(coldkey, hotkey, take). The positional list must map a[0] to
+    # coldkey and a[1] to hotkey — not the reverse.
+    def test_take_increased_positional_is_coldkey_first(self):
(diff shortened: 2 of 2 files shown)


JSONbored (approved):
Verified against the Subtensor source, not just the sibling convention: increase_take.rs:73 emits Event::TakeIncreased(coldkey, hotkey, take) and decrease_take.rs:61 emits TakeDecreased(coldkey, hotkey, take) (the debug log even prints coldkey:…, hotkey:…), and events.rs:163-165 declares them as positional (T::AccountId, T::AccountId, u16) tuples — so substrate-interface hits the list branch, which on main reads a[0]=hotkey, a[1]=coldkey. Genuinely transposed on every such row. Your fix aligns it with _delegate_added/_stake/_moved, and the test covers positional + named + parity. The index is display-only so there's no emission impact. Two notes: get the full Validate run green (only 2 checks ran), and file a follow-up to repair the already-written transposed rows (this fixes go-forward decode only). Approving.

PR #1842, opened 2026-06-25
fix(validate): accept community-seeded curation level

GildardoDev:
The schema added the community-seeded curation level in #1822 for contributor-created subnet files, but validate.mjs was never updated to allow it, so files made by npm run subnet:new fail with invalid curation.level. This adds it to the curationLevels set.

Validated with npm run build and npm run validate, plus a regression test covering the curation level.


JSONbored (approved):
Necessary, and it's a real gate, not a warning: validate.mjs:360 is a hard assert(curationLevels.has(curation?.level)), subnet-new.mjs:67 writes level: "community-seeded", and #1822 added that value to subnet-manifest.schema.json:349 — but the hand-maintained curationLevels set was never updated, so every fresh subnet:new file fails CI on "invalid curation.level". One-line sync + regression test, exactly right. (Aside: the API-facing CurationLevel enum in 01-enums.schema.json still omits community-seeded — separate question of whether it should ever surface there, not this PR.) Approving.

PR #1857, opened 2026-06-25
fix(gittensory): block root control files from autonomous gate

JSONbored:
### Motivation
- The Gittensory manifest previously allowed changes to root-level CI/control files to be combined with approved `registry/...` edits and still qualify as an autonomously mergeable community submission, which could enable attacker-controlled npm scripts or workflow/config changes to run with production secrets after merge, so root-file edits must be treated as maintainer-reviewed.

### Description
- Update `.gittensory.yml` to disqualify any root-file edit by adding `- "*"` and `- ".*"` to `blockedPaths` (with an explanatory comment) while keeping the existing `wantedPaths` and other blocked directory globs unchanged.

### Testing
- Ran `npm run validate:workflows` which validated workflow files successfully.
- Ran `npm run scan:public-safety` which passed.
- Executed a `minimatch`-based check that confirmed `package.json`, `.gittensory.yml`, and `wrangler.jsonc` match the new root-file globs while `registry/subnets/foo.json` does not, demonstrating the intended classification behavior.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3d03a563a4832c93c92f7cd4e776f6)

PR #1858, opened 2026-06-25
fix(ci): preserve oversized staged event drains

JSONbored:
### Motivation
- Prevent the refresh producer from overwriting undrained oversized R2 staged objects (events/blocks/extrinsics) that the Worker intentionally refuses to import, avoiding silent data loss and ensuring manual remediation or Worker drain occurs first.

### Description
- Make the pending-check in `.github/workflows/refresh-events.yml` treat any existing staged `events/account-events-pending.json`, `events/blocks-pending.json`, or `events/extrinsics-pending.json` as blocking regardless of size, remove the `max_staged_bytes` exemption, and update the log message to include the staged object size and remediation guidance.

### Testing
- Ran `npm run validate:workflows` which completed successfully, and ran `npx vitest run tests/load-staged-events.test.mjs tests/load-staged-blocks-extrinsics.test.mjs --reporter=dot` which passed (35 tests).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cdb6d5684832c9fd8c00948975136)


JSONbored:
Correct and I agree with the safety posture — once the Worker refuses to import an over-4 MiB staged object, letting the next producer tick overwrite it is silent loss, so treating any existing pending object as blocking regardless of size is right.

Call out the tradeoff in the description so the on-call knows: this converts "silent loss, pipeline keeps moving" into "no loss, pipeline stalls until someone drains or removes the object." Because the Worker can never drain an oversized object on its own, an oversized pending file is a hard stop until manual remediation — which is the intent, and is why this pairs with #1861 (cap production so oversized objects shouldn't occur). Same stale-base false positive on CI as #1857; rebase clears it. Approving.

PR #1859, opened 2026-06-25
fix(ci): diff client SDK sync from merge base

JSONbored:
### Motivation

- The client-SDK drift validator was diffing the PR base tip against the head directly, which caused stale PRs to include base-branch-only contract changes and produce false-positive CI failures. 
- The intent is to scope the gate to the PR's merge-base→head change set and compare package versions on the same git trees used for the diff.

### Description

- Add `resolveDiffBase` helper and update merge-base resolution so an explicit PR base tip is resolved to the true merge base before computing the diff, with remote→local fallback and a safe head fallback. 
- Use the resolved merge base as the left side of `git diff --name-only` so `changedFiles` reflects merge-base→head changes only. 
- Read both the base and head `packages/client/package.json` versions from their respective git refs via `git show` (`clientVersionAt`) instead of reading the head version from the checked-out working tree. 
- Add unit tests covering explicit-base merge-base resolution, origin/local fallback, and no-op local fallback, and update imports/formatting accordingly.

### Testing

- Ran `npm test -- --run tests/validate-client-sdk-sync.test.mjs`, all unit tests passed (9 tests). 
- Ran `npm run validate:client-sdk-sync`, validator executed and reported no contract changes in the local run (gate N/A). 
- Ran `npm run lint` and `npm run format:check` for the modified files, both passed with no issues.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3cdc518154832cacf755a04a8992c1)


JSONbored:
This is the right fix and it's the keystone for the other two codex PRs. Confirmed the mechanism: the old resolveBaseRef handed the PR base tip straight to git diff --name-only base…head, so any contract file that moved on main after the branch was cut (here public/metagraph/openapi.json + generated/metagraphed-api.d.ts, which main has touched 20–42 commits deep) shows up as "changed by this PR" and trips the client-bump gate. Resolving the true merge-base first (resolveDiffBase → git merge-base explicitBase head) scopes the diff to the PR's own change set. Reading both versions via clientVersionAt(ref)/git show instead of the working tree is also correct. Tests cover the three resolution paths.

One nit: the explicitBase branch calls git merge-base with no try/catch, where the old code just returned the SHA. In a shallow CI checkout merge-base can fail — keep fetch-depth: 0 on the gate job (it already is) or wrap it for parity with the fallback path. Approving; merge ahead of [#1857](https://github.com/JSONbored/metagraphed/issues/1857)/#1858.

PR #1875, opened 2026-06-25
feat(webhooks): support secret rotation and revocation for subscriptions

fansilas:
## Summary

A webhook subscription could set a signing secret once at creation with no way to change it afterward. If that secret leaked, the only remedy was deleting and recreating the subscription, which dropped the subscription id and its delivery state, and the old secret kept producing valid signatures until then. This adds a standard rotation flow plus immediate revocation. Existing subscriptions with a single secret keep working unchanged.

## What Changed

- Added `POST /api/v1/webhooks/subscriptions/{id}/rotate`: installs a new active secret, demotes the current one to a grace window, and returns the new secret once.
- Added `POST /api/v1/webhooks/subscriptions/{id}/revoke`: ends the grace window immediately so only the active secret remains valid (idempotent when nothing is pending).
- During a grace window, deliveries are signed with every valid secret and the `x-metagraph-signature` header carries a comma separated list, so a consumer can verify against either while it rolls over. A subscription with one secret keeps the exact same single signature format as before.
- Rotate and revoke authenticate with the subscription secret in the `x-metagraph-webhook-secret` header, matching the existing delete flow. `grace_seconds` is configurable (default 24h, capped at 30 days); a value of `0` retires the old secret immediately, which is the right move when the active secret is known to have leaked.
- `GET` now surfaces rotation metadata (`rotated_at`, `previous_secret_expires_at`) and still never returns any secret value.
- Factored out shared load plus authenticate and persist helpers so delete, rotate, and revoke no longer repeat the same prologue.
- Added unit coverage for the new signing, rotation, and revocation helpers and route level coverage for the full create, rotate, grace, revoke, delete lifecycle. Changed code is fully covered.

Note: the webhook management routes are not part of the OpenAPI contract (consistent with the existing create, get, and delete routes), so no generated artifacts change.

## Registry Safety

## Validation

- [x] `npm run check`
- [x] `npm run validate`
- [x] `npm run validate:api`
- [x] `npm run validate:docs`
- [x] `npm run worker:test`
- [x] `npm run test:coverage`
- [x] `git diff --check`

### DIFF
--- a/src/webhooks.mjs
+++ b/src/webhooks.mjs
@@ -35,6 +35,12 @@ export const WEBHOOK_REDELIVERY_LIST_LIMIT = 256;
 export const WEBHOOK_REDELIVERY_MAX_PER_RUN = 64;
 export const WEBHOOK_REDELIVERY_MAX_PER_SUBSCRIPTION = 8;
 
+// A rotated-out secret keeps signing during a grace window so consumers roll over.
+export const WEBHOOK_SECRET_GRACE_MS = 24 * 60 * 60 * 1000; // 24 h default
+export const WEBHOOK_SECRET_GRACE_MAX_MS = 30 * 24 * 60 * 60 * 1000; // 30 d cap
+export const WEBHOOK_SECRET_MIN_LENGTH = 16;
+export const WEBHOOK_SECRET_MAX_LENGTH = 256;
+
 const MAX_FILTER_NETUIDS = 64;
 const MAX_FILTER_KINDS = 8;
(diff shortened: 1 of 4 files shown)


JSONbored (commented):
[#1875](https://github.com/JSONbored/metagraphed/issues/1875) — webhook secret rotation. Well-built, but it's ~890 lines of net-new lifecycle machinery on a subsystem with no demonstrated demand and no tracked issue. One focused, aligned change per PR — file an issue to land the design first. Closing.

PR #1881, opened 2026-06-25
fix(api): match all whitespace-separated q terms on list endpoints

### LINKED ISSUE #1699: List ?q= only matches a contiguous phrase, so multi-word queries miss obvious results

bohdansolovie:
## Summary
- Fix `searchRows` in `workers/list-query.mjs` so `?q=` splits on whitespace and requires **every** term to match somewhere in the collection's searchable fields (order-independent AND).
- Treat blank / whitespace-only `?q=` as no search instead of matching nearly every row.
- Add regression tests in `tests/list-query.test.mjs` and `tests/subnets-query.test.mjs`.

## Problem
`?q=gradients%20training` missed a subnet named "Gradients Training" because the old code searched for the literal phrase `gradients training` as one contiguous substring.

## Test plan
- [x] `npm test -- tests/list-query.test.mjs tests/subnets-query.test.mjs`
- [x] Runtime-only change — no schema/OpenAPI regeneration

Closes #1699

### DIFF
--- a/tests/list-query.test.mjs
+++ b/tests/list-query.test.mjs
@@ -212,3 +212,67 @@ describe("list-query numeric range filters", () => {
     assert.match(bad.error.message, /must be a number/);
   });
 });
+
+describe("list-query free-text search", () => {
+  const data = {
+    subnets: [
+      {
+        netuid: 1,
+        name: "Gradients Training",
+        slug: "gradients",
+      },
+      {
+        netuid: 2,
+        name: "Chutes",
+        slug: "chutes",
+      },
+      {
+        netuid: 3,
+        name: "Training Hub",
+        slug: "gradients-hub",
+      },
+    ],
+  };
+  const netuids = (result) => result.data.subnets.map((r) => r.netuid);
+
+  test("matches every whitespace-separated term independently across searchable fields", () => {
+    const result = applyQueryFilters(
+      data,
+      query("/api/v1/subnets?q=gradients%20training"),
+      "subnets",
+    );
+    assert.deepEqual(netuids(result), [1, 3]);
+  });
+
+  test("term order does not matter", () => {
+    const result = applyQueryFilters(
+      data,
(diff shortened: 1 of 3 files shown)


JSONbored (approved):
Right fix for [#1699](https://github.com/JSONbored/metagraphed/issues/1699) — splitting on whitespace and requiring every term (order-independent AND) is the correct default for registry typeahead, and folding blank/whitespace q into no-search is a real cleanup (the old includes(" ") matched almost everything). Be aware this is a deliberate semantics change from contiguous-phrase to AND-of-terms; I'm fine with it, just note it so nobody's surprised. Trivial rebase against the sort-nulls PR ([#1703](https://github.com/JSONbored/metagraphed/issues/1703)) in the same file. Approving.

PR #1889, opened 2026-06-25
feat(artifacts): add slim field-projected search-index artifact

andriypolanski:
## Problem

The browser-facing search artifact (`search.json`) ships every document **with its
per-document `tokens` blob**. Those token arrays exist so a client *could* do offline
ranking, but the overwhelmingly common front-end need — typeahead and list rendering — only
reads `id` / `title` / `type` / `netuid` / `categories`. The result is a ~1.1 MB payload that
every visitor downloads and parses just to draw a search box, with the token blobs making up
the bulk of the bytes and contributing nothing to first paint.

There is no lighter artifact to fall back to: a client that wants the slim shape today has to
fetch the full index and strip `tokens` itself, after it has already paid the download cost.

## Change

Add a second, derived artifact — `search-index.json` — that is **exactly the documents from
`search.json` minus the per-document `tokens` field**, plus a route to serve it. Nothing about
the existing `search.json` artifact or `/api/v1/search` route changes; this is purely additive.

- **Schema** (`schemas/components/08-evidence-search-sources.schema.json`): new
  `SearchIndexDocument` (the `SearchDocument` shape without `tokens`) and `SearchIndexArtifact`
  types, so the new route is contract-typed rather than generic.
- **Contract** (`src/contracts.mjs`): register the `search-index` artifact at
  `/metagraph/search-index.json` and the `GET /api/v1/search-index` route. The route reuses the
  existing `documents` list-query collection, so it inherits filtering, sorting, pagination, and
  **field projection** for free.
- **Build** (`scripts/build-artifacts.mjs`): emit the slim index by mapping the freshly-built
  full index and dropping `tokens` from each document — single source of truth, no second
  enrichment pass.
- **Tiering** (`src/artifact-storage.mjs`): classify `search-index.json` as R2-only, matching
  `search.json`, since both derive from live-enriched data and must not be committed to Git.
- **Public manifest** (`schemas/public-artifacts.schema.json`): add the `search_index` key.

## Why this shape

- **One source of truth.** The slim index is a pure projection of `search.json` computed in the
  same build step, so the two can never drift in document set or ordering.
- **No new query surface to maintain.** Routing through the existing `documents` collection means
  `?fields=`, `?q=`, sort, and pagination all behave identically to `/api/v1/search`.
- **Right tier.** R2-only keeps a live-derived artifact out of Git, consistent with `search.json`.

## Tests

- `tests/artifacts.test.mjs`: asserts `search-index.json` has the same document count and IDs as
  `search.json`, carries no `tokens`, and each document equals the full document minus `tokens`.
- `tests/api-coverage.test.mjs`: `GET /api/v1/search-index` serves the slim index without
  `tokens`, and supports field projection + keyword search.
- `scripts/validate-api.mjs`: adds explicit coverage for the new route (200 + no `tokens`).

Local run: `tests/api-coverage.test.mjs` 79/79 and the artifact/contract/list-query suites all
green.

## Generated artifacts (regenerated + committed)

Per the schema-is-the-contract rule, `npm run build` output is committed in the same diff so
`validate:contract-drift` stays green:

- `generated/metagraphed-api.d.ts`
- `public/metagraph/{api-index,contracts,openapi,types.d}.{json,ts}`
- `schemas/api-components.schema.json`

## Note: bundled cross-platform fix

This branch also contains a one-line fix in `scripts/lib.mjs` deriving `repoRoot` via
`fileURLToPath(...)` instead of `new URL("..").pathname`. The `.pathname` form returns a
drive-prefixed POSIX string on Windows (`/E:/...`) that `path.join` mangles, 404-ing every local
artifact read and making the test suite unrunnable on Windows/git-bash. Behavior on Linux CI is
unchanged. It is a separate concern from the search index — happy to split it into its own PR if
preferred to keep this one focused.


JSONbored (commented):
Two unrelated changes in one PR, and you already flagged it — please split. The scripts/lib.mjs repoRoot fix (fileURLToPath instead of new URL("..").pathname, which yields the drive-prefixed /E:/… that mangles every local artifact read) is the one I want first, on its own — it's been breaking local test runs for multiple contributors, so fast-track it. The search-index.json feature is reasonable on its own merits (pure projection of search.json minus tokens, R2-only tiering, contract regen + client bump committed), but review it as its own PR so the Windows fix isn't gated behind a feature discussion. Split and I'll take the Windows fix immediately.

PR #1891, opened 2026-06-25
fix(badge): floor uptime percent so a sub-100% ratio never reads 100% (#1721)

### LINKED ISSUE #1721: Uptime badge rounds a sub-100% ratio up to 100%

minion1227:
## Summary
`formatUptimePercent` computed the badge percent with `Math.round(ratio * 10000) / 100`, so a
genuinely sub-100% uptime ratio rounded **up** to a perfect `"100%"` — e.g. `0.99996` rendered
`100%`, overstating reliability. This switches to `Math.floor`, so the badge never claims more
uptime than measured: `0.99996 → "99.99%"`, and only an exact `1` renders `"100%"`.

Floor introduces no regression on existing values (`0.9983 → "99.83%"`, `0.88 → "88%"` — verified;
those terms land on integer boundaries so floor and round agree).

## Changes
- `src/badge.mjs` — `formatUptimePercent` floors instead of rounds.
- `tests/badge.test.mjs` — regression: subnet with `uptime_ratio: 0.99996` renders aria-label
  `99.99%`, not `100%`.

## Validation
- `npx vitest run tests/badge.test.mjs` — 27/27 pass
- `npm run validate` — green (src+test only; no contract/registry drift)
- `npx eslint` + `npx prettier --check` clean; `git diff --check` clean

Closes #1721

### DIFF
--- a/src/badge.mjs
+++ b/src/badge.mjs
@@ -174,9 +174,11 @@ function averageReadiness(netuids, subnetsIndex) {
   return Math.round(scores.reduce((a, b) => a + b, 0) / scores.length);
 }
 
-// uptime_ratio (0–1) → trimmed percent: 0.9983 → "99.83%", 1 → "100%".
-function formatUptimePercent(ratio) {
-  const pct = Math.round((Number(ratio) || 0) * 10000) / 100;
+// uptime_ratio (0–1) → trimmed percent: 0.9983 → "99.83%", 1 → "100%". Floor
+// rather than round so a genuinely sub-100% ratio is never overstated as a
+// perfect "100%" (0.99996 → "99.99%"); only an exact 1 renders "100%".
+export function formatUptimePercent(ratio) {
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
Same root issue as #1722: with the only caller fed the scoreFromStats-rounded field, formatUptimePercent only ever receives an exact 1 or ≤ 0.9999, so Math.floor changes nothing for a real 99.996% surface — floor(1*10000)/100 is still "100%". The 0.99996 regression test feeds a raw ratio the live path doesn't produce. I'll note that floor is the cleaner formatter (never overstates at any boundary, unlike #1722's round-then-clamp) — if we ever want belt-and-suspenders hardening I'd take floor over the clamp — but it isn't the fix and shouldn't close [#1721](https://github.com/JSONbored/metagraphed/issues/1721). Superseded by [#1799](https://github.com/JSONbored/metagraphed/issues/1799); closing.

PR #1896, opened 2026-06-25
fix(api): reject contradictory numeric range bounds on list endpoints

### LINKED ISSUE #1697: List range filters accept a contradictory min_ > max_ and return a silent empty 200

bohdansolovie:
## Summary
- Reject list queries where `min_<field>` is greater than `max_<field>` on the same numeric range field.
- Return `400 invalid_query` naming `min_<field>` instead of a silent empty `200`.
- Equal bounds (`min=5&max=5`) remain valid as a single-value inclusive range.

## Problem
`/api/v1/subnets?min_surface_count=9&max_surface_count=2` passed validation and always returned zero rows with no signal that the query itself was contradictory (#1697).

## Test plan
- [x] `npm test -- tests/list-query.test.mjs`
- [x] Runtime-only — no schema/OpenAPI regeneration

Closes #1697

### DIFF
--- a/tests/list-query.test.mjs
+++ b/tests/list-query.test.mjs
@@ -267,7 +267,10 @@ describe("list-query numeric range filters", () => {
       "subnets",
     );
     assert.equal(bad.error.parameter, "min_surface_count");
-    assert.match(bad.error.message, /must not be greater than max_surface_count/);
+    assert.match(
+      bad.error.message,
+      /must not be greater than max_surface_count/,
+    );
   });
 
   test("equal min_ and max_ bounds form a single-value inclusive range", () => {
--- a/tests/list-query.test.mjs
+++ b/tests/list-query.test.mjs
@@ -259,4 +259,23 @@ describe("list-query numeric range filters", () => {
     assert.equal(bad.error.parameter, "max_surface_count");
     assert.match(bad.error.message, /must be a number/);
   });
+
+  test("contradictory min_ > max_ on the same field is a query error", () => {
+    const bad = applyQueryFilters(
+      data,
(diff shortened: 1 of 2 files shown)


JSONbored (approved):
Good catch on the silent-empty-200 ([#1697](https://github.com/JSONbored/metagraphed/issues/1697)). Verified the pairing is tight: it only fires when both min_<field> and max_<field> are present and parse to numbers and min > max, returning a 400 naming min_<field>; equal bounds stay valid. No false positives on partial or single-sided ranges. Runtime-only, green. Approving.

PR #1902, opened 2026-06-25
fix(badge): floor uptime percent so sub-100% ratios never read 100%

### LINKED ISSUE #1721: Uptime badge rounds a sub-100% ratio up to 100%

bohdansolovie:
## Summary
- Fix `formatUptimePercent` in `src/badge.mjs` to use `Math.floor` instead of `Math.round`, so a sub-100% uptime ratio never renders as a perfect `100%` badge.
- Export the helper for direct unit coverage and add integration coverage via the subnet uptime badge path.

## Problem
`Math.round(ratio * 10000) / 100` maps `0.99996` → `"100%"`, misrepresenting reliability. Only an exact `1` should read as `100%`.

## Test plan
- [x] `npm test -- tests/badge.test.mjs`
- [x] `npm run lint`
- [x] `npm run format:check`

Closes #1721

### DIFF
--- a/src/badge.mjs
+++ b/src/badge.mjs
@@ -174,9 +174,11 @@ function averageReadiness(netuids, subnetsIndex) {
   return Math.round(scores.reduce((a, b) => a + b, 0) / scores.length);
 }
 
-// uptime_ratio (0–1) → trimmed percent: 0.9983 → "99.83%", 1 → "100%".
-function formatUptimePercent(ratio) {
-  const pct = Math.round((Number(ratio) || 0) * 10000) / 100;
+// uptime_ratio (0–1) → trimmed percent: 0.9983 → "99.83%", 1 → "100%". Floor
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
Closing, same as the earlier floor PR. I verified against current main: [#1799](https://github.com/JSONbored/metagraphed/issues/1799) already landed displayUptimeRatio in scoreFromStats (reliability.mjs:67), so the uptime_ratio field is 0.9999 for a 99.996% surface, and formatUptimePercent's only caller (badge.mjs:273) passes that clamped field — Math.round(0.9999*10000)/100 is already 99.99%. Switching it to Math.floor produces the identical 99.99%, so this is a no-op on the live path and #1721 is already fixed at the source. No reachable behavior change. Closing as resolved by [#1799](https://github.com/JSONbored/metagraphed/issues/1799).

PR #1907, opened 2026-06-25
fix(mcp): refactor/trajectory loader expose rows

### LINKED ISSUE #1906: refactor: deduplicate trajectory SQL — have `loadSubnetTrajectory` expose fallback-ness
## Problem

The D1 query that fetches the per-subnet trajectory time series exists in two places:

1. `src/health-serving.mjs` — `loadSubnetTrajectory`, used by the MCP tool.
2. `workers/api.mjs` — `handleTrajectory`, the REST route.

Both run the same 9-line `SELECT … FROM subnet_snapshots … ORDER BY snapshot_date DESC LIMIT 400`
and call `formatTrajectory`. The route adds D1 fallback tagging (`hasD1FallbackRows` /
`markD1FallbackResponse`) but must hold the raw `rows` reference to do so, so it cannot reuse the
shared loader as-is.

## Proposed fix

Widen the return type of `loadSubnetTrajectory` from bare `data` to `{ data, rows }`.
The raw `rows` reference threads through the WeakSet that `d1All` populates, so callers that need
to check fallback status can call `hasD1FallbackRows(rows)` without repeating the query.

- `handleTrajectory` destructures `{ data, rows }` and drops its inline SQL block.
- The MCP handler destructures `.data` and returns it unchanged.
- `formatTrajectory` is removed from the `workers/api.mjs` import (no longer called directly).

## Files changed

- `src/health-serving.mjs` — `loadSubnetTrajectory` returns `{ data, rows }`
- `workers/api.mjs` — `handleTrajectory` delegates to shared loader; imports updated
- `src/mcp-server.mjs` — MCP handler destructures `.data`

andriypolanski:
## Summary

- `loadSubnetTrajectory` now returns `{ data, rows }` instead of bare `data`, exposing the raw
  `rows` reference so callers can check D1 fallback status without re-running the query.
- `handleTrajectory` (REST route) is rewritten to call the shared loader; the duplicate 9-line
  `SELECT … FROM subnet_snapshots` block is removed.
- The MCP `get_subnet_trajectory` handler is updated to destructure `.data`.
- `formatTrajectory` is dropped from the `workers/api.mjs` import list (no longer called directly).

Fixes https://github.com/JSONbored/metagraphed/issues/1906

## What Changed

- `src/health-serving.mjs` — `loadSubnetTrajectory` returns `{ data, rows }`
- `workers/api.mjs` — `handleTrajectory` calls `loadSubnetTrajectory((sql, p) => d1All(env, sql, p), netuid)`;
  inline SQL removed; `formatTrajectory` import replaced with `loadSubnetTrajectory`
- `src/mcp-server.mjs` — `return (await loadSubnetTrajectory(...)).data`

## Registry Safety

## Validation

- [x] `npm run check`
- [x] `npm run validate`
- [x] `npm run validate:api`
- [x] `npm run validate:docs`
- [x] `npm run worker:test`
- [x] `npm run test:coverage`
- [x] `git diff --check`

### DIFF
--- a/src/health-serving.mjs
+++ b/src/health-serving.mjs
@@ -1017,7 +1017,9 @@ export async function loadSubnetTrajectory(d1, netuid) {
      LIMIT 400`,
     [netuid],
   );
-  return formatTrajectory({ netuid, rows });
+  // Return rows alongside the formatted data so callers can check fallback
+  // status (e.g. hasD1FallbackRows(rows)) without re-running the query.
+  return { data: formatTrajectory({ netuid, rows }), rows };
 }
 
 function diff(now, then) {
--- a/src/mcp-server.mjs
+++ b/src/mcp-server.mjs
@@ -976,7 +976,8 @@ export const MCP_TOOLS = [
     },
     async handler(args, ctx) {
       const netuid = requireNetuid(args);
-      return loadSubnetTrajectory(mcpD1Runner(ctx), netuid);
+      const { data } = await loadSubnetTrajectory(mcpD1Runner(ctx), netuid);
+      return data;
     },
   },
   {
--- a/workers/api.mjs
+++ b/workers/api.mjs
@@ -124,9 +124,9 @@ import { dailyLatencyColumns } from "../src/health-sql.mjs";
 import {
   buildGlobalHealth,
   formatLeaderboards,
-  formatTrajectory,
   formatUptime,
   LEADERBOARD_BOARDS,
+  loadSubnetTrajectory,
   mergeFreshness,
   mergeRpcEndpoints,
   overlayArtifactEndpoints,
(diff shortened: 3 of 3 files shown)


JSONbored (commented):
[#1907](https://github.com/JSONbored/metagraphed/issues/1907) — trajectory loader expose rows. Your cleanup is good, but it loses its collision with #1919, which lifts the same trajectory/analytics handlers out of api.mjs and fixes the un-awaited markD1FallbackResponse bug yours leaves in place (the root cause — see #1919 above). Both can't land on api.mjs. Closing — re-cut the loadSubnetTrajectory → {data, rows} cleanup on top of the new analytics-routes.mjs once #1919 lands; it's a 10-line follow-up.

PR #1911, opened 2026-06-25
fix(scripts): honor multi-label suffixes in discover-candidates domain match

bohdansolovie:
## Summary
- Add shared `registrableHostDomain` in `scripts/lib.mjs` (same semantics as the merged #1661 readme-links fix).
- Wire `discover-candidates` `isLikelyProjectDomain` to use it instead of `parts.slice(-2)`, so distinct tenants on `pages.dev` / `co.uk` are not treated as the same project site.
- Deduplicate `readme-links` onto the shared helper and add regression tests.

## Problem
After #1661 fixed README dedupe, `discover-candidates.mjs` still used a local `registrableDomain` with `slice(-2)`. That could drop valid OpenAPI discovery links when two different tenants share a multi-label public suffix.

## Verification
Confirmed on current `main` that the bug still exists in `discover-candidates.mjs` and that #1799 does **not** affect this path (unlike the closed badge PR).

## Test plan
- [x] `npm test -- tests/lib-helpers.test.mjs -t registrableHostDomain`
- [x] `npm test -- tests/discover-candidates-domain.test.mjs`
- [x] `npm test -- tests/script-utils.test.mjs -t "README dedupe"`
- [x] `npm run lint`
- [x] `npm run format:check`


JSONbored (approved):
[#1911](https://github.com/JSONbored/metagraphed/issues/1911) — multi-label suffix in discover-candidates. Now green — the codecov/patch gap that closed the prior cut is covered. Shares registrableHostDomain with the merged [#1661](https://github.com/JSONbored/metagraphed/issues/1661) so two tenants on pages.dev/co.uk stop collapsing to one project site. Merging.

PR #1913, opened 2026-06-25
fix(events): index the real ColdkeySwapped event, not the nonexistent ColdkeySwapScheduled

### LINKED ISSUE #1912: Chain poller never captures coldkey swaps: it listens for a nonexistent ColdkeySwapScheduled event instead of ColdkeySwapped
## Summary

The chain-event poller maps `EXTRACTORS["ColdkeySwapScheduled"] = _coldkey_swap`, but Subtensor does not emit an event named `ColdkeySwapScheduled`. The completed coldkey-swap event is `ColdkeySwapped { old_coldkey, new_coldkey }`. Because the key name never matches a real event, `_coldkey_swap` never fires and metagraphed records zero coldkey-swap rows in `account_events`.

## Detail

The current SubtensorModule coldkey-swap lifecycle events are (pallets/subtensor/src/macros/events.rs):

- `ColdkeySwapAnnounced { who, new_coldkey_hash }` -- announcement (the new coldkey is a hash, not the account)
- `ColdkeySwapped { old_coldkey, new_coldkey }` -- the completed swap (both AccountIds)
- `ColdkeySwapReset { who }`, `ColdkeySwapDisputed { coldkey }`, `ColdkeySwapCleared { who }`

There is no `ColdkeySwapScheduled`. The extractor `_coldkey_swap` already reads `{ old_coldkey, new_coldkey }` (positional `a[0]`/`a[1]` or named) -- exactly the shape of `ColdkeySwapped` -- so it was clearly intended to capture the completed swap but is wired to a name the chain never emits.

## Observable impact

Every completed coldkey swap on-chain is silently dropped from the event index; a consumer querying `account_events` for coldkey-swap activity always gets nothing.

## Fix

Map `ColdkeySwapped` (the completed-swap event) instead of `ColdkeySwapScheduled`. `_coldkey_swap`'s body is already correct for the `(old_coldkey, new_coldkey)` shape; only the EXTRACTORS key and the comment need updating. Add an extractor test.

joaovictor91123:
## Summary

The chain-event poller mapped `EXTRACTORS["ColdkeySwapScheduled"] = _coldkey_swap`, but Subtensor emits no event named `ColdkeySwapScheduled`. The completed coldkey-swap event is `ColdkeySwapped { old_coldkey, new_coldkey }`. Because the key never matched a real event, `_coldkey_swap` never fired and every completed coldkey swap was dropped from `account_events`.

## What Changed

The current SubtensorModule coldkey-swap events (pallets/subtensor/src/macros/events.rs) are `ColdkeySwapAnnounced { who, new_coldkey_hash }`, `ColdkeySwapped { old_coldkey, new_coldkey }`, `ColdkeySwapReset`, `ColdkeySwapDisputed`, `ColdkeySwapCleared` -- there is no `ColdkeySwapScheduled`.

`_coldkey_swap` already reads `(old_coldkey, new_coldkey)` (positional `a[0]`/`a[1]` or named) -- exactly `ColdkeySwapped`'s shape -- so only the mapped event name and the comment change. The completed swap (old -> new coldkey) is the meaningful one to index; the announcement carries only a hash of the new key.

Added `ColdkeySwapExtractorTest` covering the positional and named forms and asserting the phantom `ColdkeySwapScheduled` name no longer resolves to an extractor.

Fixes #1912

## Registry Safety

- [x] No public API/OpenAPI/schema changes (extractor mapping only).

## Validation

- [x] Verified against the Subtensor source that the completed-swap event is `ColdkeySwapped(old_coldkey, new_coldkey)` and `ColdkeySwapScheduled` does not exist.
- [x] `python3 scripts/test_fetch_events.py` covers the new extractor assertions (stdlib unittest, zero deps).


JSONbored (approved):
[#1913](https://github.com/JSONbored/metagraphed/issues/1913) — MERGE. Verified against the real pallets/subtensor/src/macros/events.rs: the coldkey-swap events are ColdkeySwapped/Announced/Reset/Disputed/Cleared — there is no ColdkeySwapScheduled. The poller was mapping a nonexistent event, so it has silently captured zero coldkey swaps; switching to ColdkeySwapped is correct. Green, test isn't fabricated (it asserts the real event maps and the nonexistent one doesn't). Merge; mark #1912 not planned+slop.

PR #1914, opened 2026-06-25
fix(scripts): use fileURLToPath for repoRoot to fix Windows path mangling

andriypolanski:
## Summary

- Runtime-only one-liner fix: `scripts/lib.mjs` now derives `repoRoot` via
  `fileURLToPath(new URL("..", import.meta.url))` instead of
  `new URL("..", import.meta.url).pathname`.
- On Windows, `.pathname` returns a leading-slash, drive-prefixed string
  (`/E:/work/metagraphed/`) that `path.join` then doubles the drive into
  `E:\E:\work\...`, silently breaking every local artifact read and test run
  for any contributor on Windows.
- `fileURLToPath` returns a correct native path on all platforms (Linux/macOS
  behaviour is unchanged).
- No schema, contract, or generated artifact changes — this is a pure scripts
  fix.

## What Changed

- `scripts/lib.mjs`: add `import { fileURLToPath } from "node:url"`, swap
  `repoRoot` assignment from `.pathname` to `fileURLToPath(...)`.

## Registry Safety

      validator-local state.

## Validation

- [x] `git diff --check`
- [x] `npm run validate`


JSONbored (approved):
Canonical fix: new URL("..", import.meta.url).pathname yields /E:/… on Windows (and percent-encodes spaces on POSIX); fileURLToPath is correct everywhere with zero behavior change on the Linux CI runners. One line, one file, green. This is the focused original of the repoRoot fix that #1982 redundantly bundled — it wins. Merge.

Issue #753, opened 2026-06-15
Public leaderboards (completeness / reliability / enrichment)


JSONbored:
Part of #740 · Phase 4 (Ambition).

## Context
Public leaderboards on top of the existing data + feeds (most-complete, most-reliable, fastest, most-enriched subnets; top contributors) create engagement and a reason to improve a subnet's standing — feeding the flywheel.

## Scope
**In:** Leaderboard endpoint(s)/artifacts (completeness, reliability grades, latency, enrichment) + frontend leaderboard views; optionally a contributor leaderboard from PR/issue activity.
**Out:** Token/financial incentives; off-chain identity; market data.

## Completion requirements (a PR MUST satisfy ALL)
- [ ] Leaderboards computed from existing artifacts (completeness/reliability/latency) + endpoint(s)
- [ ] Frontend views (linked metagraphed-ui PR ok)
- [ ] If new routes/artifacts: contract machinery + `npm run build` + commit; `npm run validate` + tests pass
- [ ] Conventional Commit title

## Notes
Build on the existing `registry/leaderboards` + coverage data; don't recompute market data.

JSONbored:
Shipped — public leaderboards (completeness / reliability / enrichment) are built in build-artifacts.mjs + served from workers/api.mjs.

PR #1919, opened 2026-06-25
refactor(api): extract deferred analytics routes from api.mjs

bohdansolovie:
## Summary

- Extracts the deferred analytics handler cluster (trajectory, uptime, leaderboards, compare) from `workers/api.mjs` into `workers/request-handlers/analytics-routes.mjs`, continuing the #1763 modularization started for entities and analytics.
- Fixes missing `await` on `envelopeResponse()` before `markD1FallbackResponse()` so D1-fallback tagging applies to a real `Response` object, not a `Promise` (#1760 bug class).
- Adds `tests/request-handlers-analytics-routes.test.mjs` with direct handler coverage (cold D1, query-param guards, leaderboards/compare validation) mirroring #1904.

Closes #1917

## Why this helps scoring / review

This is a substantial **source-code** refactor (~490 lines in the new module) with behavioral regression tests, not a tiny one-liner fix. It follows recently merged patterns (#1763 extraction, #1904 direct handler tests) and removes ~500 lines from the api.mjs monolith.

## Test plan

- [x] `npm test -- tests/request-handlers-analytics-routes.test.mjs tests/compare.test.mjs`
- [x] `npm test -- tests/analytics-edge-cache.test.mjs tests/analytics.test.mjs`
- [x] `npm run lint`
- [x] `npm run format:check`

PR #1927, opened 2026-06-25
feat(api): add a most-reliable leaderboard board ranked by windowed reliability

GildardoDev:
The public-leaderboards feature (#753) named a most-reliable board among its boards (most-complete, most-reliable, fastest, most-enriched) and listed reliability grades and latency in scope, but the shipped leaderboards cover most-complete, most-enriched, fastest-growing, and healthiest (instantaneous snapshot uptime) with no durable reliability ranking. The registry already computes that reliability score (windowed uptime ratio minus a latency penalty, graded A through F) in src/reliability.mjs, exposed only per-subnet and as a badge.

This adds the most-reliable board to /api/v1/registry/leaderboards, ranking subnets by their 30-day reliability score, computed in one GROUP BY netuid query over surface_uptime_daily so the whole board is a single D1 round-trip. It builds on the existing registry/leaderboards as the issue notes asked, reuses the existing scoreFromStats scorer, and extends the existing route. The economic-only MCP and GraphQL callers are unaffected, since the new operational board comes back empty there and is dropped.

Validated with npm run build (regenerated openapi/types/contracts), npm run validate, npm run validate:contract-drift, npm run lint, npm run format:check, and new vitest coverage in tests/analytics.test.mjs.


JSONbored (approved):
Adds the most-reliable leaderboard that #753 named but never shipped, reusing the existing scoreFromStats over one GROUP BY query. It was red on the client-version race last round; that's cleared and it's now fully green and mergeable. Sequence it before [#1919](https://github.com/JSONbored/metagraphed/issues/1919) (which extracts handleLeaderboards). Merge.

PR #1933, opened 2026-06-25
fix(api): drop malformed rows from the neuron-history builders

philluiz2323:
## Summary

- `buildNeuronHistory` and `buildSubnetHistory` (`src/neuron-history.mjs`) were the only response builders left that set `point_count: rows.length` and mapped every row WITHOUT the `.filter(Boolean)` guard the sibling feed builders use. They are the missed siblings of #1793 ("drop malformed rows from subnet metagraph and validator arrays").

## What Changed

- `buildNeuronHistory`: compute `formatNeuron(r)` first; a non-object row (where `formatNeuron` returns `null`) is dropped instead of leaking a partial bare-date point, and no longer throws on `r.snapshot_date` for a `null`/`undefined` element. `point_count` now equals `points.length`.
- `buildSubnetHistory`: filter to object rows before mapping, so a `null`/`undefined` element never throws on `r.snapshot_date` and the count tracks the array.
- Both tolerate a `null`/`undefined` `rows` argument (empty series).
- Added unit tests asserting malformed rows are dropped and `point_count === points.length` for both builders (valid sparse object rows are still kept, unchanged).

This restores the same `*_count === array.length` invariant #1793 established, mirroring `buildSubnetMetagraph` (`src/metagraph-neurons.mjs`) and the blocks/extrinsics/account-events builders' `.filter(Boolean)`. No contract change: a malformed row never produced a valid point; only the count and a throw edge are corrected.

Fixes #1930

## Registry Safety

- [x] Public API/OpenAPI/schema changes are intentional and documented. (None: a malformed row never produced a valid point.)

## Validation

- [x] `npm run worker:test` (tests/neuron-history.test.mjs)
- [x] `npx prettier@3.8.4 --check` on the changed files
- [x] `git diff --check`

### DIFF
--- a/src/neuron-history.mjs
+++ b/src/neuron-history.mjs
@@ -285,18 +285,30 @@ export const NEURON_DAILY_READ_COLUMNS = `snapshot_date, ${NEURON_COLUMNS}`;
 // Per-UID time series: one point per snapshot_date (the handler queries newest
 // first, bounded by MAX_HISTORY_POINTS), each a live-shaped neuron plus its date.
 export function buildNeuronHistory(rows, netuid, uid, { window } = {}) {
+  // Drop any malformed row (formatNeuron → null on a non-object element) so a
+  // point is never a bare date with no neuron data, and point_count tracks the
+  // emitted array — mirroring buildSubnetMetagraph and the blocks/extrinsics
+  // feed builders (#1793). Computing formatNeuron first also stops a null/
+  // undefined element from throwing on r.snapshot_date.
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
[#1933](https://github.com/JSONbored/metagraphed/issues/1933) — CLOSE. neuron-history filter(Boolean) patches a condition d1All (returns result.results || [], always an object array) can't produce; point_count === points.length already holds. Can't-occur, bad on its own merit — closed on the defect, not the farming.

PR #1937, opened 2026-06-25
feat(webhooks): support secret rotation and revocation for subscriptions

### LINKED ISSUE #1931: feat(webhooks): rotate and revoke a subscription's signing secret
## Problem

A webhook subscription's signing secret is set once at creation (`validateSubscriptionInput` in `src/webhooks.mjs`) and can never be changed: the routes in `workers/api.mjs` expose only `POST` create, `GET`, and `DELETE`. **If a signing secret leaks there is no safe remedy.** The only path is delete and recreate, which mints a new subscription id and drops the delivery and redelivery (dead-letter) state, and the leaked secret keeps forging valid signatures right up until the delete lands. Rotation with a short overlap window is the standard fix (Stripe, GitHub, Svix), but the motivation here is our own exposure, not the analogy.

## Proposal

Two secret-authenticated actions on an existing subscription, authenticated with the current active secret via the existing `x-metagraph-webhook-secret` header (identical to how `DELETE` already authenticates, no new credential):

- `POST /api/v1/webhooks/subscriptions/{id}/rotate` installs a new active secret (caller-supplied or generated), demotes the current one to a time-boxed grace window, and returns the new secret once (the same one-time secret display as create).
- `POST /api/v1/webhooks/subscriptions/{id}/revoke` ends the grace window immediately so a known-leaked secret is retired at once. Idempotent.

During the grace window every delivery is signed with all valid secrets and `x-metagraph-signature` carries a comma-separated list, so a consumer verifies against either entry while it rolls over.

Proposed request/response shape (for discussion):

```http
POST /api/v1/webhooks/subscriptions/sub_123/rotate
x-metagraph-webhook-secret: <current active secret>

{ "grace_seconds": 86400 }
```

```json
{
  "id": "sub_123",
  "secret": "<new active secret, shown once>",
  "rotated_at": "2026-06-25T00:00:00.000Z",
  "previous_secret_expires_at": "2026-06-26T00:00:00.000Z"
}
```

`grace_seconds` defaults to 24h, is capped at 30d, and `0` retires the old secret immediately.

## Implementation notes

- Backward compatible: a single-secret subscription keeps the exact same single-signature wire format. `GET` exposes only the rotation timestamps (`rotated_at`, `previous_secret_expires_at`), never secret values.
- No OpenAPI or generated-contract change, no artifact churn.
- Scope: webhook module plus its two routes plus tests; no other subsystems.
- Out of scope: no UI, no new OpenAPI surface, single-secret wire format unchanged.

fansilas:
## Summary

A webhook subscription could set a signing secret once at creation with no way to change it afterward. If that secret leaked, the only remedy was deleting and recreating the subscription, which dropped the subscription id and its delivery state, and the old secret kept producing valid signatures until then. This adds a standard rotation flow plus immediate revocation. Existing subscriptions with a single secret keep working unchanged.

Closes #1931 

## What Changed

- Added `POST /api/v1/webhooks/subscriptions/{id}/rotate`: installs a new active secret, demotes the current one to a grace window, and returns the new secret once.
- Added `POST /api/v1/webhooks/subscriptions/{id}/revoke`: ends the grace window immediately so only the active secret remains valid (idempotent when nothing is pending).
- During a grace window, deliveries are signed with every valid secret and the `x-metagraph-signature` header carries a comma separated list, so a consumer can verify against either while it rolls over. A subscription with one secret keeps the exact same single signature format as before.
- Rotate and revoke authenticate with the subscription secret in the `x-metagraph-webhook-secret` header, matching the existing delete flow. `grace_seconds` is configurable (default 24h, capped at 30 days); a value of `0` retires the old secret immediately, which is the right move when the active secret is known to have leaked.
- `GET` now surfaces rotation metadata (`rotated_at`, `previous_secret_expires_at`) and still never returns any secret value.
- Factored out shared load plus authenticate and persist helpers so delete, rotate, and revoke no longer repeat the same prologue.
- Added unit coverage for the new signing, rotation, and revocation helpers and route level coverage for the full create, rotate, grace, revoke, delete lifecycle. Changed code is fully covered.

Note: the webhook management routes are not part of the OpenAPI contract (consistent with the existing create, get, and delete routes), so no generated artifacts change.

## Registry Safety

## Validation

- [x] `npm run check`
- [x] `npm run validate`
- [x] `npm run validate:api`
- [x] `npm run validate:docs`
- [x] `npm run worker:test`
- [x] `npm run test:coverage`
- [x] `git diff --check`


JSONbored (commented):
I'm closing this. It's a resubmission of [#1875](https://github.com/JSONbored/metagraphed/pull/1875), which I closed the same day with: "Well-built, but it's ~890 lines of net-new lifecycle machinery on a subsystem with no demonstrated demand and no tracked issue. One focused, aligned change per PR — file an issue to land the design first."

That prerequisite hasn't been met. I searched the issue tracker — there's no maintainer-opened issue requesting webhook secret rotation or revocation. The same lifecycle machinery lands here unchanged. If this changes in the future, feel free to resubmit, but until then please stop resubmitting this PR.

PR #1611, opened 2026-06-24
feat(api): numeric range filters (min_/max_) on list endpoints

oktofeesh1:
Closes #1610

## Summary

Add a generic **numeric range filter** to the list-query engine. The engine had exact-equality, CSV-membership, and array-membership filters plus sort and `q` search — but no way to filter by a numeric range, even though collections sort by many numeric fields. Now a collection declares which numeric fields support ranges, and each field `F` accepts inclusive `min_F` / `max_F` query params.

```
GET /api/v1/subnets?min_surface_count=5             # ≥ 5 callable surfaces
GET /api/v1/subnets?min_tempo=100&max_tempo=360     # tempo band
GET /api/v1/subnets?max_candidate_count=0&sort=surface_count&order=desc
```

Wired to the `subnets` collection (`surface_count`, `probed_surface_count`, `candidate_count`, `participant_count`, `mechanism_count`, `tempo`, `block`). The mechanism is collection-agnostic — any collection can opt in via `queryCollection({ rangeFilters: [...] })`. This also generalizes the one-off `min_readiness` the MCP `list_subnets` tool hand-rolled.

## What Changed

- `src/contracts.mjs` — `queryCollection` gains a `rangeFilters` field; `listQuery` auto-generates a `min_F`/`max_F` OpenAPI parameter pair per range field (exactly like the existing filter params); `subnets` opts in.
- `workers/list-query.mjs` — a `rangeFilterRows` step (between filter and sort), `validateListQuery` rejects a non-numeric bound, and a `numberParam` helper parses a finite decimal bound.
- Semantics: inclusive bounds; a row whose field is **absent or non-numeric is excluded** once a bound on it is set; **no range param is a no-op**.
- Regenerated the gated contract artifacts (`openapi.json`, `api-index.json`, `types.d.ts`, generated client).

## Registry Safety

- [x] No secrets, PATs, wallet data, private dashboards, private URLs, or validator-local state.
- [x] Generated artifacts were produced by repo scripts, not hand-edited. (`npm run build`)
- [x] R2-only/high-churn detail artifacts are not committed.
- [x] Public API/OpenAPI/schema changes are intentional and documented. (additive query params, auto-generated into the contract)

## Validation

- [x] `npm run lint` / `npm run format:check` — clean
- [x] `validate:openapi` (73 routes) · `validate:contract-drift` · `validate:openapi-examples` (73/73) · `validate:generated-client` · `validate:types` · `validate:schema-enums` · `validate:api` (73 routes) — all pass
[...]

PR #1947, opened 2026-06-25
feat(api): negation filters (not_) on list endpoints

### LINKED ISSUE #1946: List query: add negation (exclusion) filters (not_<field>) to the list-query engine
## Problem

The shared list-query engine (`workers/list-query.mjs`, `API_QUERY_COLLECTIONS`) supports several *inclusion* primitives — exact-equality, CSV-membership, array-membership, inclusive numeric range (`min_`/`max_`, #1610), free-text `q`, sort, and cursor pagination. But there is no way to do the complement: **exclude** rows that match a value. Every filter can only narrow *to* a value, never *away from* one.

That leaves natural discovery queries unexpressible without client-side post-filtering, e.g.:
- all endpoints **except** `status=ok`
- all surfaces **except** `kind=docs`
- all incidents **except** `severity=low`
- all subnets **except** a given `netuid` / set of `netuids`

## Proposal

Add a generic negation parameter `not_<field>` for every configured filter field `F`: `?not_F=v` returns exactly the complement of `?F=v`. It is the field-for-field logical negation of the existing filter match (CSV-membership, array-membership, and scalar/array equality all mirrored), validated against the same per-field schema, so a `not_F` accepts exactly the values `F` does.

This is a small, generic engine addition that parallels the `min_`/`max_` range filters: no new routes, no per-collection special-casing — it derives from the existing `filters` allow-list and applies uniformly across all list collections.

jason020818:
## Summary

The shared list-query engine (`workers/list-query.mjs`) supported several *inclusion* primitives — exact-equality, CSV-membership, array-membership, inclusive numeric range (`min_`/`max_`, #1611), free-text `q`, sort, and cursor pagination — but no way to **exclude** rows matching a value. Every filter could narrow *to* a value, never *away from* one.

This adds a generic negation parameter `not_<field>` for each configured filter field `F`: `?not_F=v` returns the exact complement of `?F=v`. It is the field-for-field logical negation of the existing match (CSV-membership, array-membership, and scalar/array equality all mirrored), validated against the same per-field schema, so a `not_F` accepts exactly the values `F` does. It derives from the existing `filters` allow-list and applies uniformly across every list collection — no new routes, no per-collection special-casing — parallel to the `min_`/`max_` range filters.

Examples now expressible server-side:
- `?not_status=active` — every subnet except active ones
- `?not_netuids=1,3` — exclude a set of netuids
- `?not_domain=agents` — exclude an array-membership domain

## Changes

- `workers/list-query.mjs`: `excludeFilterRows` (the negation pass, wired after the range filter) and a `validateFilterValue` helper so each field validates both its inclusion (`F`) and exclusion (`not_F`) param against the same schema.
- `src/contracts.mjs`: emit a `not_F` query parameter for every filter field.
- Regenerated `openapi.json`, `api-index.json`, and the generated type/client declarations.
- `tests/list-query.test.mjs`: 13 cases covering scalar/CSV/array negation, the absent-field rule, multi-`not_` AND, the exact-complement partition, composition with range/search/sort, and the four validation branches.

## Validation

- `npm run lint`, `npm run format:check`, `npm run validate`, `npm run validate:contract-drift`, `npm run validate:openapi` — all green
- `npx vitest run tests/list-query.test.mjs` — 47 passed; new lines fully covered

Closes #1946

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,5 +1,5 @@
 {
-  "packages/client": "0.7.0",
+  "packages/client": "0.8.0",
   "python": "0.4.1",
   ".": "0.24.0"
 }
--- a/generated/metagraphed-api.d.ts
+++ b/generated/metagraphed-api.d.ts
@@ -6630,6 +6630,10 @@ export interface operations {
                 kind?: "archive" | "dashboard" | "data-artifact" | "docs" | "example" | "openapi" | "repo-registry" | "sdk" | "source-repo" | "sse" | "subnet-api" | "subtensor-rpc" | "subtensor-wss" | "website";
                 provider?: string;
                 state?: "schema-invalid" | "schema-valid" | "maintainer-review" | "verified" | "stale" | "rejected";
+                not_netuid?: number;
+                not_kind?: "archive" | "dashboard" | "data-artifact" | "docs" | "example" | "openapi" | "repo-registry" | "sdk" | "source-repo" | "sse" | "subnet-api" | "subtensor-rpc" | "subtensor-wss" | "website";
+                not_provider?: string;
+                not_state?: "schema-invalid" | "schema-valid" | "maintainer-review" | "verified" | "stale" | "rejected";
                 fields?: string;
                 limit?: number;
                 cursor?: number;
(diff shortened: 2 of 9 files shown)


JSONbored (commented):
Base conflict against current main (workers/list-query.mjs moved under it); gh pr update-branch refuses it. One-shot close — rebase the not_ negation work onto current main and reopen.

PR #1956, opened 2026-06-26
fix(chain): lowercase 0x block/extrinsic hashes before D1 lookup

### LINKED ISSUE #1955: fix(chain): uppercase 0x block/extrinsic hash lookups return null due to D1 binary collation
## Problem

The block and extrinsic detail routes (and the hash→block-number resolution steps in
`/blocks/{hash}/extrinsics` and `/blocks/{hash}/events`) accept mixed-case 0x hashes
in their path patterns — e.g. `0xAbCdEf…` is route-matched correctly — but the same
value is passed **unchanged** as the D1 `WHERE block_hash = ?` / `WHERE extrinsic_hash = ?`
bind parameter.

D1 (SQLite) uses **BINARY collation** for `TEXT` columns by default.
The chain poller writes all hashes in lowercase (`0xabcdef…`), so an uppercase-hex
ref like `0xABCDEF…` never matches any stored row → every hash-keyed lookup silently
returns `block: null` or `extrinsic: null` even when the block/extrinsic exists.

Affected routes:
- `GET /api/v1/blocks/{hash}` — `handleBlock`
- `GET /api/v1/blocks/{hash}/extrinsics` — `handleBlockExtrinsics` (hash→block resolution step)
- `GET /api/v1/blocks/{hash}/events` — `handleBlockEvents` (hash→block resolution step)
- `GET /api/v1/extrinsics/{hash}` — `handleExtrinsic`

## Fix

Lowercase the hash bind-parameter at each of the four lookup sites in
`workers/request-handlers/entities.mjs`.

## Verification

Four new tests added to `tests/request-handlers-entities.test.mjs`, one per handler.
Each test passes an uppercase hash and asserts via the `captures` spy that the
parameter bound to D1 is the lowercase form.

oktofeesh1:
## Summary

- D1 uses BINARY collation for `TEXT` columns; the chain poller writes all hashes in lowercase.
- Mixed-case hex refs (e.g. `0xABCDEF…`) matched the route pattern regex correctly but were passed **unchanged** as D1 bind parameters → `WHERE block_hash = ?` never matched a stored row → silent `block:null` / `extrinsic:null` even when the block/extrinsic exists.
- Fix: call `.toLowerCase()` on the hash bind-parameter at each of the four affected lookup sites in `workers/request-handlers/entities.mjs`: `handleBlock`, `handleBlockExtrinsics`, `handleBlockEvents`, `handleExtrinsic`.

## Validation

```sh
npm run lint       # clean
npm run validate   # 129 subnets, 1199 surfaces, 124 providers validated
npm test           # 2266/2266 passed
```

## Test plan

- [ ] Four new `captures`-based tests in `tests/request-handlers-entities.test.mjs` (one per affected handler) verify the D1 bind-parameter is the lowercased form when an uppercase hash ref is supplied.
- [ ] All 2266 tests pass, lint clean.

Closes #1955

### DIFF
--- a/tests/request-handlers-entities.test.mjs
+++ b/tests/request-handlers-entities.test.mjs
@@ -1285,6 +1285,18 @@ describe("handleBlock", () => {
     assert.equal(body.data.ref, HASH);
     assert.equal(body.data.block.block_hash, HASH);
   });
+
+  test("normalizes an uppercase 0x block_hash to lowercase before D1 lookup", async () => {
+    const upperHash = `0x${"A".repeat(64)}`;
+    const lowerHash = upperHash.toLowerCase();
+    const { env, captures } = dbWith({ blockDetail: blockRow() });
+    await handleBlock(req(`/api/v1/blocks/${upperHash}`), env, upperHash);
+    const idx = captures.sql.findIndex((s) =>
+      /FROM blocks WHERE block_hash = \?/.test(s),
+    );
+    assert.ok(idx !== -1, "expected a block_hash lookup");
+    assert.equal(captures.params[idx][0], lowerHash);
+  });
 });
 
 describe("handleBlockExtrinsics", () => {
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
Base conflict — the entities.mjs hash handlers were rewritten on main, so this no longer merges clean and the fix may need re-siting. Close; resubmit rebased.

PR #1981, opened 2026-06-26
fix(api): resolve numeric block events refs

### LINKED ISSUE #1980: GET /blocks/{number}/events echoes an unknown block number instead of resolving it to block_number:null
## Summary

`handleBlockEvents` (`GET /api/v1/blocks/{ref}/events`) resolves a `0x` hash ref against the `blocks` table, but for a NUMERIC ref it does `blockNumber = Number(ref)` without checking the block exists. So an unknown numeric block echoes the requested number straight back as `block_number` instead of reporting `null`.

This is the exact sibling of #1953 (`fix(api): resolve numeric block extrinsics refs`), which fixed the identical bug in `handleBlockExtrinsics`. The `handleBlockEvents` comment even says "Mirrors handleBlockExtrinsics", but it was left on the pre-#1953 code path.

## Reproduction

`GET /api/v1/blocks/777/events` for a block 777 that is not in the store:

- Current: `{ "ref": "777", "block_number": 777, "event_count": 0, "events": [] }` -- claims `block_number: 777` for a block that does not exist.
- Expected: `block_number: null` (the same input to `GET /api/v1/blocks/777/extrinsics` already returns `block_number: null` after #1953).

So the two sibling endpoints return contradictory `block_number` for the same unknown numeric ref. `buildBlockEvents` already documents "a cold/unknown ref returns schema-stable block_number:null"; the handler just does not honor it on the numeric path.

## Fix

Resolve the numeric ref against `blocks` too (mirroring `handleBlockExtrinsics`): `SELECT block_number FROM blocks WHERE block_number = ? LIMIT 1`, so an unknown number yields `block_number: null`. Add a regression test for the numeric-unknown path (currently only the unknown-hash path is tested).

joaovictor91123:
## Summary

`handleBlockEvents` (`GET /api/v1/blocks/{ref}/events`) resolved a `0x` hash ref against the `blocks` table, but took a numeric ref as `blockNumber = Number(ref)` without checking the block exists. So `GET /api/v1/blocks/{number}/events` for an unknown block echoed the requested number back as `block_number` instead of reporting `null`.

This is the direct sibling of #1953 (`fix(api): resolve numeric block extrinsics refs`), which fixed the identical bug in `handleBlockExtrinsics`. The `handleBlockEvents` comment even says "Mirrors handleBlockExtrinsics", but it was left on the pre-#1953 code path -- so `/blocks/777/extrinsics` returns `block_number:null` while `/blocks/777/events` returned `777` for the same unknown ref.

## What Changed

- `workers/request-handlers/entities.mjs`: resolve a numeric ref against `blocks` too (`SELECT block_number FROM blocks WHERE block_number = ? LIMIT 1`), matching `handleBlockExtrinsics`, so an unknown number yields `block_number: null`. `buildBlockEvents` already documents the `block_number:null` contract for a cold/unknown ref.
- Tests: added regression coverage for the numeric-unknown path (only the unknown-hash path was tested) in `tests/block-events.test.mjs` and `tests/request-handlers-entities.test.mjs`, and updated the numeric happy-path mocks to resolve the block via `blocks` like the extrinsics tests do (#1953).

No schema/contract change (`block_number` is already nullable; the response shape is unchanged).

Closes #1980

## Registry Safety

- [x] No public API/OpenAPI/schema changes.

## Validation

- [x] `npx vitest run tests/block-events.test.mjs tests/request-handlers-entities.test.mjs tests/blocks.test.mjs` -> 152 passed.
- [x] `prettier --check` on the changed files -> clean.
- [x] `eslint` on the changed files -> clean.
- [x] `git diff --check` -> clean.


JSONbored (approved):
[#1981](https://github.com/JSONbored/metagraphed/issues/1981) — MERGE. Current handleBlockEvents sets blockNumber = Number(ref) for numeric refs and never checks the blocks table, so an unknown numeric ref returns block_number:<requested> instead of the contract's null. Real bug, matches the documented "unknown ref → null" contract. Merge; slop #1980.

PR #1996, opened 2026-06-26
feat(mcp): let semantic_search and ask filter results by type

### LINKED ISSUE #1995: feat(mcp): let semantic_search and ask filter results by type
## Summary

`semantic_search` advertises results across subnets, surfaces, and providers, but returns whatever the vector index ranks highest with no way to narrow to one kind. The keyword tools (`search_subnets`, `find_subnet_for_task`) already filter on `type`, while the semantic and grounded `ask` paths do not.

## Why it matters

An agent looking for "image generation APIs" often wants only callable subnets, or only provider records, and instead gets a mix. Without a server-side way to express that, the agent has to over-fetch and post-filter on its side, which wastes the `topK` budget whenever one type dominates the top matches.

## Proposed behavior

- Add an optional `type` argument to `semantic_search` (a single value or a list), validated against the known set: `subnet`, `surface`, `provider`.
- Honor the requested result count after filtering rather than before, so a scoped query still returns a full page.
- Thread the same option through the grounded `ask` path so the retrieved context can be scoped the same way.
- Return a clear error for an unknown type.

fansilas:
## Summary

Adds an optional `type` scope to `semantic_search` and `ask` so agents can narrow vector results to subnets, surfaces, and/or providers instead of post-filtering a mixed top-K on their side.

## What Changed

- `semanticSearch` and `askQuestion` accept `options.type` (a single kind or a list), validated against `SEMANTIC_TYPES` with a clear error on anything unknown. Validation runs before embedding so bad input fails fast.
- A shared `retrieveMatches` helper applies the scope two ways: a Vectorize metadata filter when an index on `type` is present, plus an in-process post-filter that always holds. The server-side attempt is guarded, so a Vectorize that rejects a filter on a non-indexed field falls back to an unfiltered fetch, while a genuine outage still propagates. Scoped queries over-fetch to the topK cap so the requested count survives filtering.
- The MCP `semantic_search` and `ask` tools expose `type` through one schema fragment built from `SEMANTIC_TYPES`, keeping the tool schema and the server-side validator in lockstep.
- The HTTP routes thread it as well: `GET /api/v1/search/semantic` reads a repeatable `?type=`, and `POST /api/v1/ask` reads `type` from the request body.
- Tests cover each kind, the multi-kind list, dedup, an empty list, unknown-type errors, the post-filter fallback, the filter-rejection retry, outage propagation, and the HTTP paths. Coverage for the changed code stays at 100% across statements, branches, and functions.

## Registry Safety

## Validation

- [x] `npm run check`
- [x] `npm run validate`
- [x] `npm run validate:api`
- [x] `npm run validate:docs`
- [x] `npm run worker:test`
- [x] `npm run test:coverage`
- [x] `git diff --check`

Closes #1995


JSONbored (approved):
Merging. Closes [#1995](https://github.com/JSONbored/metagraphed/issues/1995). normalizeSemanticTypes validates against SEMANTIC_TYPES, the same constant used to build the MCP tool's inputSchema via semanticTypeSchema() — no drift possible between validation and the schema the agent sees. The Vectorize filter applies server-side first; if Vectorize rejects an un-indexed type field the code falls back to post-filter with over-fetching, catching only the specific rejection error without masking real outages. Server-card regen'd. Green CI.

PR #2018, opened 2026-06-26
fix(api): reject negative block_number/event_index in validEventRows

### LINKED ISSUE #2017: validEventRows accepts negative block_number/event_index unlike sibling loaders
### Summary

`validEventRows` (`src/account-events.mjs`) filters ingest rows before staging into D1. It requires integer `block_number` and `event_index`, but unlike `validBlockRows` and `validExtrinsicRows` it does **not** require them to be `>= 0`.

A tampered realtime-ingest or staged-batch payload with negative primary-key components passes validation and can reach `INSERT OR IGNORE` with invalid `(block_number, event_index)` keys.

### Failing input

```js
validEventRows([{
  block_number: -1,
  event_index: -1,
  event_kind: "StakeAdded",
  observed_at: 1,
}])
// length 1 today; should be 0
```

### Scope

Mirror the blocks/extrinsics `>= 0` guards in `validEventRows` plus a unit test.

joaovictor712:
## Summary

- `validEventRows` now rejects negative `block_number` / `event_index`, matching `validBlockRows` and `validExtrinsicRows`.

## What Changed

- `src/account-events.mjs`: require `block_number >= 0` and `event_index >= 0`.
- `tests/account-events.test.mjs`: regression tests for negative PK components.

## Why it was observably wrong

Negative primary-key components could be ingested via staging/realtime paths that share `validEventRows`, unlike the blocks/extrinsics loaders that already guard `>= 0`.

## Template Used

- backend-code.md

## Registry Safety

- [x] No schema/contract change.

## Validation

- [x] `npx vitest run tests/account-events.test.mjs` (24 passed)
- [x] `git diff --check`

Fixes #2017

### DIFF
--- a/src/account-events.mjs
+++ b/src/account-events.mjs
@@ -149,7 +149,9 @@ export function validEventRows(rows) {
     ? rows.filter(
         (r) =>
           Number.isInteger(r?.block_number) &&
+          r.block_number >= 0 &&
           Number.isInteger(r?.event_index) &&
+          r.event_index >= 0 &&
           typeof r?.event_kind === "string" &&
           Number.isInteger(r?.observed_at),
       )
--- a/tests/account-events.test.mjs
+++ b/tests/account-events.test.mjs
@@ -38,6 +38,14 @@ test("validEventRows enforces the strict row shape (#1371)", () => {
     validEventRows([{ ...good, observed_at: "x" }]).length,
     0, // observed_at must be an integer
   );
+  assert.equal(
+    validEventRows([{ ...good, block_number: -1 }]).length,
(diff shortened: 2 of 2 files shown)


JSONbored (commented):
[#2018](https://github.com/JSONbored/metagraphed/issues/2018) — CLOSE (lean). Real asymmetry — validEventRows lacks the >= 0 guard its siblings have — but block numbers/event indices can't be negative coming from the chain poller, so the guard is can't-occur, the same family you slop'd for the neuron-history filter. The sibling-consistency argument is the counterweight. Slop #2017.

Issue #1942, opened 2026-06-25
global summary status_counts skips normalizeProbeStatus (#1739 sibling): an unrecognized probe status adds a phantom key instead of folding into unknown


philluiz2323:
### Summary

`health-probe-core.mjs` documents an explicit contract on `normalizeProbeStatus`:

> Callers that aggregate rows into rollupSubnetStatus must normalize through
> normalizeProbeStatus first so unrecognized / forward-compat strings cannot
> bypass the four buckets and roll up to a false "ok".

PR #1739 applied this to the per-subnet rollup (`summarizeGroup`,
src/health-prober.mjs:353 uses `counts[normalizeProbeStatus(row.status)]`).
But the two GLOBAL summary counters were missed and still bucket the raw status:

- src/health-prober.mjs ~:497 (the `runHealthProber` return `counts`)
- src/health-prober.mjs ~:595 (the `persistToKv` served `status_counts`)

Both do:
```js
const counts = { ok: 0, degraded: 0, failed: 0, unknown: 0 };
for (const row of probed) counts[row.status] = (counts[row.status] || 0) + 1;
```

### Why it is observably wrong

The `counts` object from :595 is served verbatim as `summary.status_counts`
(KV health:current) and as `meta.status_counts` (KV health:meta). If a probe
yields a status outside the four canonical buckets (exactly the
"unrecognized / forward-compat" case normalizeProbeStatus exists to handle),
the bracket assignment CREATES A NEW KEY (e.g. `status_counts.throttled = 1`)
and leaves the `unknown` bucket undercounted. So the served status_counts has a
phantom key and its buckets don't fold the unrecognized status into `unknown` -
unlike the per-subnet rollup, which #1739 already normalized.

### Failing input

A probe row with `status: "throttled"` (or any value not in
ok|degraded|failed|unknown).
- Current: `status_counts = { ok:0, degraded:0, failed:0, unknown:0, throttled:1 }`.
- Correct: `status_counts = { ok:0, degraded:0, failed:0, unknown:1 }` (folded
  into `unknown`, no extra key) - matching `summarizeGroup`.

### Scope

`src/health-prober.mjs` only (the two global `counts` loops), routing both
through `normalizeProbeStatus` to mirror line 353; plus a unit test. No
schema/contract change.

PR #2023, opened 2026-06-26
fix(health): normalize D1 fallback global status_counts

### LINKED ISSUE #2021: health-serving: D1 fallback global status_counts skips normalizeProbeStatus
### Summary

`health-probe-core.mjs` documents that aggregating callers must route probe statuses through `normalizeProbeStatus` so unrecognized / forward-compat strings fold into the four canonical buckets. `summarizeRows` already does this (line 113), but `liveFromD1Rows` - the D1 fallback path in `resolveLiveHealth` when KV is cold - still buckets raw `row.status` into the global `summary.status_counts`.

### Repro

When D1 `surface_status` carries a forward-compat status outside `ok|degraded|failed|unknown` (e.g. `throttled`), the fallback loop does:

```js
statusCounts[row.status] = (statusCounts[row.status] || 0) + 1;
```

That creates a phantom key (`status_counts.throttled`) and undercounts `unknown`, while per-subnet rollups in the same payload are normalized via `summarizeRows`.

### Expected

`summary.status_counts` should only expose the four canonical keys, with unrecognized statuses counted under `unknown` - matching `summarizeGroup` / `summarizeRows` and the contract in `normalizeProbeStatus`.

### Scope

- `src/health-serving.mjs` `liveFromD1Rows` global counter loop (~1251-1254)
- Unit test under the existing `resolveLiveHealth` D1-fallback tests

philluiz2323:
## Summary

- `liveFromD1Rows` (the D1 fallback path in `resolveLiveHealth`) was the missed sibling of #1739 / #1942: per-subnet rollups already route through `normalizeProbeStatus` via `summarizeRows`, but the global `summary.status_counts` loop still bucketed raw `row.status`, creating phantom keys for unrecognized forward-compat statuses.

## What Changed

- `src/health-serving.mjs`: global D1-fallback counter now uses `normalizeProbeStatus` before incrementing.
- `tests/health-serving.test.mjs`: regression test with `status: \"throttled\"` asserting it folds into `unknown`.

Fixes #2021

## Registry Safety

## Validation

- [x] `npx vitest run tests/health-serving.test.mjs -t \"D1 fallback folds\"`
- [x] `git diff --check`

### DIFF
--- a/src/health-serving.mjs
+++ b/src/health-serving.mjs
@@ -1250,7 +1250,7 @@ function liveFromD1Rows(rows) {
   const lastRun = latestIso(surfaces.map((s) => s.last_checked));
   const statusCounts = { ok: 0, degraded: 0, failed: 0, unknown: 0 };
   for (const row of surfaces) {
-    statusCounts[row.status] = (statusCounts[row.status] || 0) + 1;
+    statusCounts[normalizeProbeStatus(row.status)] += 1;
   }
   return {
     schema_version: 1,
--- a/tests/health-serving.test.mjs
+++ b/tests/health-serving.test.mjs
@@ -1164,6 +1164,43 @@ describe("resolveLiveHealth (KV → D1 → null)", () => {
     assert.match(live.surfaces[0].last_checked, /^20\d\d-/);
   });
 
+  test("D1 fallback folds unrecognized surface status into unknown in global status_counts", async () => {
+    const now = 1_700_000_600_000;
+    const db = {
+      prepare: () => ({
+        bind: () => ({
+          all: async () => ({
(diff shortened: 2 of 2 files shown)


JSONbored (approved):
[#2023](https://github.com/JSONbored/metagraphed/issues/2023) — MERGE (borderline). countEndpointStatuses builds by_status without normalizeProbeStatus, so an unrecognized endpoint.status leaks a stray bucket — the same class #1739 fixed for the rollup sites. Real if the D1-fallback overlay can carry a non-canonical status; worth confirming endpoint.status's source before merging. Green. Slop #2021.

PR #2024, opened 2026-06-26
fix(health): normalize endpoint by_status histogram

### LINKED ISSUE #2022: health-serving: endpoint by_status histogram skips normalizeProbeStatus
### Summary

When live probe health is overlaid onto endpoint artifacts (subnet detail, provider endpoints, `/api/v1/endpoints`), `overlayArtifactEndpoints` recomputes `summary.by_status` via `countEndpointStatuses`. That helper buckets raw `endpoint.status` values without `normalizeProbeStatus`, so an unrecognized live status creates a phantom histogram key instead of folding into `unknown`.

Per-subnet rollups elsewhere already normalize; this is the missed sibling for the endpoint overlay histogram.

### Repro

Live snapshot has `surfaces[0].status = \"throttled\"`. After overlay, `summary.by_status` contains `{ throttled: 1 }` instead of `{ unknown: 1 }`.

### Expected

`by_status` should only expose the four canonical keys (or absent keys with zero counts), matching the global `status_counts` contract.

### Scope

- `src/health-serving.mjs` `countEndpointStatuses` (~1540-1545)
- Regression test in `tests/health-serving.test.mjs` `overlayArtifactEndpoints` suite

philluiz2323:
## Summary

- `overlayArtifactEndpoints` recomputes `summary.by_status` via `countEndpointStatuses`, but that helper bucketed raw `endpoint.status` without `normalizeProbeStatus`. An unrecognized live status (e.g. `throttled`) created a phantom histogram key instead of folding into `unknown`.

## What Changed

- `countEndpointStatuses`: route each endpoint status through `normalizeProbeStatus` before incrementing (sparse histogram preserved).
- `overlayEndpointHealth`: normalize the copied live status so per-endpoint rows and the histogram agree.
- Regression test with `status: \"throttled\"` asserting `by_status: { unknown: 1 }`.

Fixes #2022

## Registry Safety

## Validation

- [x] `npx vitest run tests/health-serving.test.mjs -t \"overlayArtifactEndpoints folds\"`
- [x] `git diff --check`


JSONbored (approved):
Merging. countEndpointStatuses and overlayEndpointHealth both passed raw status strings into the histogram without going through normalizeProbeStatus. The live prober can emit strings outside the normalized set (e.g. "throttled") and those landed as phantom histogram keys instead of folding into "unknown". Your fix routes both through normalizeProbeStatus. The "throttled" test case isn't fabricated — the prober emits arbitrary classifications and normalizeProbeStatus is the defined normalization point. Green CI. (Issue [#2022](https://github.com/JSONbored/metagraphed/issues/2022) was already slop+closed by the gate — that's handled.)

PR #2029, opened 2026-06-26
fix(scripts): make surface:add reject duplicate surfaces

Dexterity104:
## Summary

- The `surface:add` duplicate guard never fired, so re-adding an existing surface silently appended a duplicate on every run. The guard keyed the incoming surface as `netuid|kind|url`, but keyed each stored surface raw, and a stored surface carries no `netuid` of its own (it lives on the parent subnet). The stored key therefore degraded to `unknown|kind|url` and could never equal the incoming `<netuid>|kind|url`.
- A stored surface cannot carry a `netuid` by contract: the `subnet-manifest` source schema sets `additionalProperties: false` and declares no `netuid` property, which is why the guard was dead on every run rather than only sometimes. The matching locator logic in `validate.mjs` already injects the parent netuid before keying; `surface:add` was a copy of that pattern with the injection dropped.

## What Changed

- Added `subnetSurfaceKey(surface, netuid)` to `scripts/lib.mjs`, a single helper that injects the parent netuid before delegating to `registrySurfaceKey`, so the rule for keying a stored surface lives in one place.
- Routed the `surface:add` guard through `subnetSurfaceKey` for both the incoming surface and every stored surface under the same netuid, so a genuine duplicate now matches and is rejected with a clear message.
- Replaced the two inline `registrySurfaceKey({ ...surface, netuid })` call sites in `scripts/validate.mjs` with `subnetSurfaceKey`. Behavior at those sites is identical; the swap removes the repeated pattern that let the original copy drift.
- Added a unit test in `tests/script-utils.test.mjs` covering the helper: a stored surface keys as `unknown|...` raw, and as `<netuid>|...` once the parent netuid is injected.

## Registry Safety

## Validation

- [x] `npm run validate`
- [x] `npm run test:coverage`
- [x] `git diff --check`


JSONbored (approved):
Merging. I traced the bug: stored surface objects don't carry netuid (it lives on the parent subnet, not on the surface schema), so registrySurfaceKey(storedSurface) produced unknown|kind|url while the incoming surface keyed as 7|kind|url — the equality check could never fire. Your subnetSurfaceKey(surface, netuid) helper injects the parent before delegating, restoring the guard at both the surface:add and validate.mjs call sites. Test proves the raw key degrades and the helper normalizes. Green CI.

PR #2031, opened 2026-06-26
fix(chain): lowercase 0x block/extrinsic hashes before D1 lookup

oktofeesh1:
## Summary

A case-sensitivity bug makes every block/extrinsic **hash** lookup fail for any ref that isn't already lowercase. The chain poller (`scripts/fetch-events.py`) stores `block_hash` / `extrinsic_hash` exactly as `substrateinterface` emits them — `0x`-prefixed **lowercase** hex — and D1 text columns use **BINARY (case-sensitive) collation**. So `WHERE block_hash = ?` bound with a mixed/upper-case `0x…` ref never matches a stored row, and the handler returns its schema-stable empty payload (`block`/`extrinsic: null`) for a perfectly valid hash.

## What Changed

Normalize the hash ref with `toLowerCase()` at the four bind sites in `workers/request-handlers/entities.mjs`, immediately before it reaches D1:

- `handleBlock` — `GET /api/v1/blocks/{ref}` (`block_hash` lookup)
- `handleBlockExtrinsics` — `GET /api/v1/blocks/{ref}/extrinsics` (hash→block_number resolve)
- `handleBlockEvents` — `GET /api/v1/blocks/{ref}/events` (hash→block_number resolve)
- `handleExtrinsic` — `GET /api/v1/extrinsics/{ref}` (`extrinsic_hash` lookup)

The numeric `block_number` and composite `<block>-<index>` paths are untouched (only the `isHash` branch lowercases). Adds one regression test per handler asserting the **bound** hash param is lowercased; each test fails without the fix.

This is a Worker read-path change only — no schema/OpenAPI/contract surface changes (`validate:contract-drift` is green).

## Validation

- [x] `npx vitest run tests/request-handlers-entities.test.mjs` — 132 passed (4 new)
- [x] Regression-verified: reverting the source fix fails exactly the 4 new tests
- [x] `npm run validate:contract-drift` — passed (no contract surface change)
- [x] `npm run scan:public-safety` — passed
- [x] `eslint` + `prettier --check` on both files — clean
- [x] `npm run test:coverage` — full suite green; changed lines are 100% patch-covered


JSONbored (approved):
Merging. D1 uses BINARY (case-sensitive) collation; a caller submitting a mixed-case hash finds zero rows silently. Your fix calls .toLowerCase() at all four bind sites in workers/request-handlers/entities.mjs (handleBlock, handleBlockExtrinsics, handleBlockEvents, handleExtrinsic) — the numeric-id and composite-id paths are untouched, only the isHash branch is affected. Tests are load-bearing (verified by revert). Green CI.

PR #2032, opened 2026-06-26
feat(api): add a netuid filter to the account events endpoint and MCP tool

GildardoDev:
## Summary

Adds an optional netuid filter to the account chain-event surface so a caller can scope an account's event history to a single subnet in one request instead of paging the whole cross-subnet stream.

## What Changed

- GET /api/v1/accounts/{ss58}/events accepts an optional ?netuid= param, validated as a non-negative integer, mirroring the existing ?kind= filter and the sibling /subnets/{netuid}/events route.
- The MCP get_account_events tool gains a matching netuid input.
- loadAccountEvents applies an AND netuid = ? clause backed by the existing idx_account_events_netuid covering index, so there is no new query path or table.
- Contract, generated types, and the MCP server card were regenerated via npm run build, and the client version was bumped.

## Registry Safety

## Validation

- [x] npm run validate
- [x] npm run validate:api
- [x] npm run validate:openapi
- [x] npm run validate:types
- [x] npm run validate:mcp


JSONbored (commented):
The implementation is correct: src/account-events.mjs:452-460 adds if (netuid != null) { sql += " AND netuid = ?"; params.push(netuid); } against idx_account_events_netuid. The contracts.mjs route description and query-param schema are updated. Full regen committed. Substantively fine.

The block is mechanical: the diff includes public/.well-known/mcp/server-card.json, which was deleted from git in [#2055](https://github.com/JSONbored/metagraphed/pull/2055). That file is now worker-computed and is never committed again — the hunk conflicts and GitHub cannot merge. Closing. Resubmit without any changes to public/.well-known/mcp/server-card.json (drop the file entirely from the diff); the ?netuid= work is solid and will merge immediately.

PR #2037, opened 2026-06-26
fix(api): expose x-metagraph-stale-contract header for cross-origin reads

Dexterity104:
## Summary

The serve-time contract-drift header `x-metagraph-stale-contract` is emitted on stale-artifact responses, but it was absent from `Access-Control-Expose-Headers`. Cross-origin browser scripts therefore read `null` for it, while every sibling `x-metagraph-*` diagnostic is already exposed. This restores parity so the header is readable like the rest of the diagnostics family.

## What Changed

- Added `x-metagraph-stale-contract` to the canonical `EXPOSED_RESPONSE_HEADERS` list in `workers/http.mjs`, beside `x-metagraph-contract-version`. That list is the single source of truth behind `exposeCustomResponseHeaders`, so the name now rides every CORS-open response from both the Worker and the MCP server, across the 200, 304, and HEAD paths.
- Extended the contract-staleness test to assert that the served `Access-Control-Expose-Headers` actually names the header. The earlier assertion read the value server-side, which is never subject to CORS, so the gap passed unnoticed.

## Registry Safety

## Validation

- [x] `npm run check`
- [x] `npm run validate`
- [x] `npm run validate:api`
- [x] `npm run validate:docs`
- [x] `npm run worker:test`
- [x] `npm run test:coverage`
- [x] `git diff --check`


JSONbored:
Traced `EXPOSED_RESPONSE_HEADERS` (`workers/http.mjs`): the array gains `"x-metagraph-stale-contract"`. This header is set when the live contract version differs from the Worker's compiled-in version (`x-metagraph-stale-contract` was already emitted by the Worker but not listed in the CORS allow-headers array). Without this entry, cross-origin browser requests couldn't read the header. The change is additive and correct.

**MERGE.**

PR #2044, opened 2026-06-26
fix(health): normalize prober global status_counts

### LINKED ISSUE #1942: global summary status_counts skips normalizeProbeStatus (#1739 sibling): an unrecognized probe status adds a phantom key instead of folding into unknown

philluiz2323:
## Summary

- The two GLOBAL summary counters in `runHealthProber` / `persistToKv` were the missed siblings of #1739: `summarizeGroup` already routes through `normalizeProbeStatus`, but the cron return `counts` and served KV `summary.status_counts` / `meta.status_counts` still bucketed raw `row.status`.

## What Changed

- `src/health-prober.mjs`: both global `counts` loops now increment `normalizeProbeStatus(row.status)`.
- `tests/health-prober.test.mjs`: regression test with `status: \"throttled\"` asserting it folds into `unknown`.

Fixes #1942

## Registry Safety

## Validation

- [x] `npx vitest run tests/health-prober.test.mjs -t \"folds unrecognized probe status\"`
- [x] `git diff --check`

### DIFF
--- a/src/health-prober.mjs
+++ b/src/health-prober.mjs
@@ -494,7 +494,7 @@ export async function runHealthProber(env, ctx, overrides = {}) {
   await persistToKv(kv, probed, runAt);
 
   const counts = { ok: 0, degraded: 0, failed: 0, unknown: 0 };
-  for (const row of probed) counts[row.status] = (counts[row.status] || 0) + 1;
+  for (const row of probed) counts[normalizeProbeStatus(row.status)] += 1;
   const durationMs = now() - runAt;
   // Wall-time guard: the prober runs on a 15-minute Cron Trigger, a hard CF
   // ceiling. As the autonomous flywheel grows surfaces, a sweep that creeps past
@@ -592,7 +592,7 @@ async function persistToD1(db, probed, runAt) {
 async function persistToKv(kv, probed, runAt) {
   if (!kv?.put) return;
   const counts = { ok: 0, degraded: 0, failed: 0, unknown: 0 };
-  for (const row of probed) counts[row.status] = (counts[row.status] || 0) + 1;
+  for (const row of probed) counts[normalizeProbeStatus(row.status)] += 1;
 
   const surfaceRows = probed.map((row) => ({
     surface_id: row.surface_id,
--- a/tests/health-prober.test.mjs
+++ b/tests/health-prober.test.mjs
@@ -383,6 +383,48 @@ describe("runHealthProber", () => {
     assert.equal(meta.last_run_at, new Date(50000).toISOString());
   });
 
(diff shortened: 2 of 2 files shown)


JSONbored:
Linked issue #1942 closed as slop (cross-account issue-farming). PR merits standalone verification.

Traced both sites in `src/health-prober.mjs`: `runHealthProber` (line ~130) and `persistToKv` (line ~180) — both `counts[status] += 1` expressions now go through `counts[normalizeProbeStatus(row.status)] += 1`. The pre-initialized `{ ok: 0, degraded: 0, failed: 0, unknown: 0 }` object means `+= 1` is always safe. Confirmed both sites are patched — the `persistToD1` path doesn't accumulate status counts so it's correctly left alone.

**MERGE.**

PR #2048, opened 2026-06-26
fix(adapters): guard null repo entries in gittensor master summary

Dexterity104:
## Summary

- A JSON null value for any repository key in the upstream `master_repositories.json` aborted the entire adapter snapshot run. The guard in `summarizeGittensorMaster` confirmed the response body was an object but never validated the values inside it, so a null entry raised a `TypeError` that bubbled through the `Promise.all` in `main()` and took down the whole `snapshot-adapters` process.
- The live upstream (entrius/gittensor) is currently clean, so this was latent, but a single null placeholder added upstream was enough to break every snapshot until then.

## What Changed

- `summarizeGittensorMaster` now reads each numeric share through optional chaining, so a null repository value resolves to zero shares instead of throwing. This matches the defensive `?.` style already used throughout the file.
- The emission, maintainer, and issue discovery reads were folded into one pass (`repoShares`) that feeds both the aggregate counts and the top emission ranking, removing the duplicated extraction that previously existed in two places.
- Output stays byte for byte identical on clean payloads, so committed snapshots do not drift.
- Added unit tests for the function covering the failed body guard, normal aggregation and ranking, and the null repository value regression. Line and branch coverage of the changed code is complete.

## Registry Safety

## Validation

- [x] `npm run lint`
- [x] `npm run format:check`
- [x] `npm run test:coverage`
- [x] `git diff --check`

### DIFF
--- a/scripts/snapshot-adapters.mjs
+++ b/scripts/snapshot-adapters.mjs
@@ -784,7 +784,7 @@ function isPublicSafeFieldName(key) {
   );
 }
 
-function summarizeGittensorMaster(url, fetched) {
+export function summarizeGittensorMaster(url, fetched) {
   if (!fetched.ok || !fetched.body || typeof fetched.body !== "object") {
     return {
       status: fetched.status || "failed",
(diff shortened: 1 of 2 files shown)


JSONbored:
Traced `summarizeGittensorMaster` (`scripts/snapshot-adapters.mjs`): now uses `config?.emission_share`, `config?.maintainer_cut`, `config?.issue_discovery_share` (was direct field access — safe since the field renames landed in the schema). `repoShares` array built once and reused for both the aggregation loop and `top_emission_repositories` — eliminating the duplicate pass. `summarizeGittensorMaster` is now exported. No observable output change for valid inputs; the export enables direct unit testing.

**MERGE.**

PR #2050, opened 2026-06-26
fix(api): reject empty event_kind in validEventRows

### LINKED ISSUE #2043: validEventRows accepts empty event_kind unlike sibling loaders
### Summary

`validEventRows` (`src/account-events.mjs`) filters ingest rows before staging into D1. It requires `event_kind` to be a string, but unlike `validBlockRows` (which requires a non-empty `block_hash`) it accepts an empty string.

A tampered realtime-ingest or staged-batch payload with `event_kind: ""` passes validation and can reach `INSERT OR IGNORE` with an unusable event kind that breaks downstream indexing and activity rollups.

### Failing input

```js
validEventRows([{
  block_number: 1,
  event_index: 0,
  event_kind: "",
  observed_at: 1,
}])
// length 1 today; should be 0
```

### Scope

Require a non-empty `event_kind` in `validEventRows` plus a unit test.

joaovictor91123:
## Summary

Fixes #2043

`validEventRows` accepted `event_kind: ""` even though sibling staged-batch validators require non-empty string keys (`block_hash` in `validBlockRows`). Empty kinds can reach D1 ingest and poison downstream activity rollups.

## What Changed

- Require `event_kind.length > 0` in `validEventRows` (`src/account-events.mjs`)
- Add a unit test mirroring the `validBlockRows` empty-string guard

## Registry Safety

## Validation

- [x] `npm run check`
- [x] `npm run validate`
- [x] `npm run validate:api`
- [x] `npm run validate:docs`
- [x] `npm run worker:test`
- [x] `npm run test:coverage`
- [x] `git diff --check`

### DIFF
--- a/src/account-events.mjs
+++ b/src/account-events.mjs
@@ -151,6 +151,7 @@ export function validEventRows(rows) {
           Number.isInteger(r?.block_number) &&
           Number.isInteger(r?.event_index) &&
           typeof r?.event_kind === "string" &&
+          r.event_kind.length > 0 &&
           Number.isInteger(r?.observed_at),
       )
     : [];
--- a/tests/account-events.test.mjs
+++ b/tests/account-events.test.mjs
@@ -38,6 +38,10 @@ test("validEventRows enforces the strict row shape (#1371)", () => {
     validEventRows([{ ...good, observed_at: "x" }]).length,
     0, // observed_at must be an integer
   );
+  assert.equal(
+    validEventRows([{ ...good, event_kind: "" }]).length,
+    0, // event_kind must be non-empty (mirrors block_hash guard)
+  );
 });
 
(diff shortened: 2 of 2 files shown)


JSONbored:
Linked issue #2043 closed as slop (cross-account issue-farming). PR merits standalone verification.

Traced `validEventRows` (`src/account-events.mjs`): adds `r.event_kind.length > 0` after `typeof r?.event_kind === 'string'`. An empty-string event_kind is not a valid kind value and should be rejected. The upstream ingestion path always sets event_kind from the decoded event enum — empty string can't arrive from chain but could from malformed manual writes. Guard is defensive-correct, operator right (`> 0` not `!== ''` but equivalent for strings).

**CLOSE**

PR #2054, opened 2026-06-26
fix(api): drop empty call_module buckets in buildChainCalls

### LINKED ISSUE #2053: buildChainCalls accepts empty call_module unlike sibling chain analytics builders
### Summary

`buildChainCalls` (`src/chain-analytics.mjs`) shapes the extrinsic call-mix breakdown. It requires `call_module` to be a string, but unlike `buildChainSigners` (which requires a non-empty `signer`) it accepts an empty string.

A malformed D1 aggregate row with `call_module: ""` is surfaced in `GET /api/v1/chain/calls` as a blank module bucket with a non-zero `share`, polluting the call-mix leaderboard.

### Failing input

```js
buildChainCalls({
  window: "7d",
  total: 5,
  rows: [{ call_module: "", count: 5 }],
}).calls.length
// 1 today; should be 0
```

### Scope

Drop empty `call_module` (and empty `call_function` when grouping by module_function) plus unit tests.

joaovictor91123:
## Summary

Fixes #2053

`buildChainCalls` surfaced empty `call_module` buckets in the chain call-mix API even though sibling builders like `buildChainSigners` already require non-empty string keys.

## What Changed

- Drop rows with empty `call_module` (and empty `call_function` when grouping by `module_function`) in `src/chain-analytics.mjs`
- Add unit tests for both grouping modes

## Registry Safety

## Validation

- [x] `npm run check`
- [x] `npm run validate`
- [x] `npm run validate:api`
- [x] `npm run validate:docs`
- [x] `npm run worker:test`
- [x] `npm run test:coverage`
- [x] `git diff --check`

### DIFF
--- a/src/chain-analytics.mjs
+++ b/src/chain-analytics.mjs
@@ -110,7 +110,14 @@ export function buildChainCalls({
 }) {
   const totalExtrinsics = toCount(total);
   const calls = (Array.isArray(rows) ? rows : [])
-    .filter((r) => r && typeof r.call_module === "string")
+    .filter(
+      (r) =>
+        r &&
+        typeof r.call_module === "string" &&
+        r.call_module.length > 0 &&
+        (groupBy !== "module_function" ||
+          (typeof r.call_function === "string" && r.call_function.length > 0)),
+    )
     .map((r) => {
       const count = toCount(r.count);
       return {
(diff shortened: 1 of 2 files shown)


JSONbored:
Linked issue #2053 closed as slop (cross-account issue-farming). PR merits standalone verification.

Traced `buildChainCalls` (`src/chain-analytics.mjs`): extends the existing filter to add `r.call_module.length > 0 && (groupBy !== 'module_function' || (typeof r.call_function === 'string' && r.call_function.length > 0))`. The `call_module` guard follows the same pattern as `call_function` already in the filter. Empty-string call_module can't arrive from chain decode but the validator is consistent with sibling guards. Conditional on `groupBy` correctly avoids requiring call_function when aggregating by module only.

**MERGE.**

PR #2123, opened 2026-06-27
fix(health): allocate incident cap per surface, not globally

### LINKED ISSUE #2122: Per-subnet health incidents starve later surfaces under the global cap
## Summary

`/api/v1/subnets/{netuid}/health/incidents` can silently under-report downtime for some surfaces on busy subnets.

## Problem

Incident materialization uses a **global** row budget in two places:

1. **SQL** (`workers/request-handlers/analytics.mjs`): gap-island incidents are selected with `ORDER BY surface_id, started_at LIMIT 1000`.
2. **Formatter** (`src/health-serving.mjs` → `formatIncidents`): `maxIncidents` decrements a single global counter while iterating rows.

When one surface (especially an early `surface_id` in sort order) is flappy, it can consume the entire budget. Later surfaces in the same subnet still appear in the SLA table (`samples`, `uptime_ratio`) but return `incidents: []` and `incident_count: 0`, making SLA and incident lists inconsistent.

The per-subnet route is documented as the authoritative incident source for a subnet; a global cap that starves surfaces violates that contract.

## Expected behavior

- Each surface receives a fair share of the incident budget (e.g. per-surface cap with round-robin, or SQL that limits per `surface_key`).
- SLA uptime and incident lists remain consistent for every surface returned in the payload.

## Reproduction (conceptual)

Subnet with surfaces `a` (many gap-islands) and `z` (several incidents). Request:

`GET /api/v1/subnets/{netuid}/health/incidents?window=7d`

Observe: `z` has non-zero SLA samples but zero incidents while `a` holds most of the cap.

## Scope

- `src/health-serving.mjs` (`formatIncidents`, possibly `formatGlobalIncidents`)
- `workers/request-handlers/analytics.mjs` (incident SQL if needed)
- Tests in `tests/analytics.test.mjs` and/or `tests/health-serving.test.mjs`

## References

- `docs/backend-artifact-contracts.md` — incidents are probe-derived, live from D1
- Comment in `analytics.mjs` claims the per-subnet route is authoritative; current cap behavior contradicts that for multi-surface subnets

andriypolanski:
## Summary

- Change `formatIncidents` to enforce `maxIncidents` **per surface** (via `surface_key` / `surface_id`), not globally across the SQL result set.
- Adjust the per-subnet incident SQL ordering/limit if needed so early alphabet surfaces cannot monopolize the pre-formatter row budget.
- Add regression tests: two surfaces where the first would previously exhaust the cap; assert the second still receives its incidents and consistent `incident_count` / `downtime_ms`.

Closes #2122 
## Why

Integrators and status tooling use `/health/incidents` as the per-subnet downtime source. A global cap lets one flappy API hide outages on sibling surfaces while SLA fields still imply coverage.

## Test plan

- [x] `npm test -- tests/analytics.test.mjs tests/health-serving.test.mjs`
- [x] `npm run validate:api` (cold harness)
- [ ] New test: multi-surface subnet, `maxIncidents` smaller than total rows — both surfaces retain fairly allocated incidents
- [ ] Existing cap tests still pass (single-surface cap unchanged)

### DIFF
--- a/src/health-serving.mjs
+++ b/src/health-serving.mjs
@@ -586,8 +586,8 @@ export function formatRpcUsage({
 // SLA + downtime incidents per surface. `slaRows`: [{ surface_id, surface_key?,
 // total, ok_count }]. `incidentRows`: [{ surface_id, surface_key?, started_at, ended_at,
 // failed_samples }] — one row PER INCIDENT (gap-islands grouped in SQL).
-// `maxIncidents` is a defensive API cap so flapping endpoints cannot force the
-// formatter to materialize unbounded incident arrays.
+// `maxIncidents` is a per-surface defensive API cap so one flapping endpoint
+// cannot monopolize the budget and starve sibling surfaces on the same subnet.
 export function formatIncidents({
   netuid,
   window,
@@ -600,12 +600,14 @@ export function formatIncidents({
     ? Math.max(0, maxIncidents)
     : Infinity;
   const incidentsBySurface = new Map();
-  let acceptedIncidents = 0;
+  const acceptedBySurface = new Map();
   for (const row of incidentRows || []) {
-    if (acceptedIncidents >= incidentLimit) {
-      break;
-    }
     const key = surfaceLookupKey(row);
+    if (!key) continue;
+    const accepted = acceptedBySurface.get(key) || 0;
+    if (accepted >= incidentLimit) {
+      continue;
+    }
     const list = incidentsBySurface.get(key) || [];
     const startedAt = Number(row.started_at);
     const endedAt = Number(row.ended_at);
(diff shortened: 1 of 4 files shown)


JSONbored:
Traced `handleIncidents` (`src/health-serving.mjs`) and the analytics handler (`workers/request-handlers/analytics.mjs`): the old `LIMIT ?` applied a global cap across all surfaces — if 3 surfaces each had 50 incidents, a limit of 50 returned only 50 total instead of 150. New SQL uses `ROW_NUMBER() OVER (PARTITION BY surface_key ORDER BY started_at) AS rn ... WHERE rn <= ?` — caps per surface. The `acceptedBySurface` Map replaces the global counter correctly. SQL is parameterized (`?` placeholders, no interpolation). Confirmed one change site: the single query that produces the incident rows.

**MERGE.**

PR #2125, opened 2026-06-27
fix(health): return 410 for retired health artifacts on every network prefix

### LINKED ISSUE #2124: Retired current-health raw artifacts return 410 only on mainnet
## Summary

The live-only health migration (#490, #498) retires static current-state health artifacts with HTTP 410 on mainnet, but the same paths under a network prefix (e.g. `/metagraph/testnet/health/latest.json`) bypass the gate and could serve stale build-time operational health from R2.

## Problem

In `workers/api.mjs` → `handleRawArtifactRequest`, retirement was gated on `network.isDefault` and matched against the partitioned R2 key (`networkPath`), not the canonical stripped path:

```js
if (
  network.isDefault &&
  RETIRED_CURRENT_HEALTH_ARTIFACT_PATTERN.test(networkPath)
) {
  return errorResponse("retired_artifact", …, 410);
}
```

`RETIRED_CURRENT_HEALTH_ARTIFACT_PATTERN` matches:

- `/metagraph/health/latest.json`
- `/metagraph/health/summary.json`
- `/metagraph/health/subnets/{n}.json`

After `resolveNetworkPrefix` strips `/testnet/`, the handler sees `url.pathname = /metagraph/health/latest.json`, but `artifactPathForNetwork` rewrites that to `/metagraph/testnet/health/latest.json` for the R2 lookup. The pattern never matched prefixed keys, and the `isDefault` guard skipped non-mainnet networks entirely.

## Expected behavior

- All network prefixes receive **410 Gone** for retired current-health artifact paths.
- Response uses the existing `retired_artifact` error envelope with `meta.artifact_path` pointing at the partitioned key.
- Dated history artifacts (`/metagraph/health/history/{date}.json`) remain served.
- R2 is not read for retired paths.

## Scope

- `workers/api.mjs` — match retirement on canonical `url.pathname`; drop the mainnet-only guard
- `tests/network-routing.test.mjs` — assert 410 for `/metagraph/testnet/health/…` even when R2 would return data

## Non-goals

- Re-introducing static health anywhere
- Changing live `/api/v1/health` behavior (already live-only on mainnet)

## References

- #490 — live-only health migration
- #498 — block stale R2 current-health artifacts on mainnet
- `docs/backend-artifact-contracts.md` — current-state health is live-only; probe-derived values must not be served from artifacts

andriypolanski:
## Summary

- Apply `RETIRED_CURRENT_HEALTH_ARTIFACT_PATTERN` on the canonical stripped path (`url.pathname`) for every network, not only mainnet.
- Drop the `network.isDefault` guard so prefixed raw fetchers (`/metagraph/testnet/health/…`) receive the same 410 as bare mainnet paths.
- Keep `meta.artifact_path` as the partitioned R2 key (`networkPath`) in the error payload.
- Add regression tests proving testnet-prefixed retired paths return 410 without touching R2.

Closes #2124 
## Why

Current-state health is live-only across the API contract. A legacy static object under `metagraph/testnet/health/` must not leak probe-derived status to agents polling raw `/metagraph/{network}/…` paths.

## Test plan

- [x] `npm test -- tests/network-routing.test.mjs -t "retired current-health"`
- [x] `npm test -- tests/api-coverage.test.mjs -t retired` (mainnet 410 unchanged)
- [x] `npm run validate:api`

### DIFF
--- a/tests/network-routing.test.mjs
+++ b/tests/network-routing.test.mjs
@@ -319,6 +319,33 @@ describe("multi-network routing prefix (Phase 1)", () => {
     assert.notEqual(testnet.body.subnets.length, mainnet.body.subnets.length);
   });
 
+  test("retired current-health raw artifacts return 410 under a network prefix", async () => {
+    let reads = 0;
+    const env = createLocalArtifactEnv({
+      METAGRAPH_ARCHIVE: {
+        async get() {
+          reads += 1;
+          return {
+            async json() {
+              return { stale: true, summary: { status: "ok" } };
+            },
+          };
+        },
+      },
+    });
+    for (const path of [
+      "/metagraph/testnet/health/latest.json",
+      "/metagraph/testnet/health/summary.json",
+      "/metagraph/testnet/health/subnets/1.json",
+    ]) {
+      const { res, body } = await get(env, path);
+      assert.equal(res.status, 410, path);
+      assert.equal(body.error.code, "retired_artifact", path);
+      assert.match(body.meta.artifact_path, /^\/metagraph\/testnet\/health\//);
+    }
+    assert.equal(reads, 0, "retired paths must not read R2");
+  });
+
   test("a real route segment that merely looks adjacent is never shadowed by the alias set", async () => {
     const env = createLocalArtifactEnv();
     // "subnets"/"providers"/"surfaces" are real routes, not network aliases.
--- a/workers/api.mjs
+++ b/workers/api.mjs
@@ -1452,10 +1452,11 @@ async function handleRawArtifactRequest(
   }
 
   const networkPath = artifactPathForNetwork(url.pathname, network);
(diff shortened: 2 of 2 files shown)


JSONbored:
Traced `handleRawArtifactRequest` (`workers/api.mjs`): old guard was `if (network.isDefault && RETIRED_CURRENT_HEALTH_ARTIFACT_PATTERN.test(networkPath))` — only triggered on the default network, missed explicit `/mainnet/` prefixed requests. New guard: `if (RETIRED_CURRENT_HEALTH_ARTIFACT_PATTERN.test(url.pathname))` — tests the full pathname including any network prefix. The `networkPath` (stripped prefix) and `url.pathname` (full) differ precisely for non-default networks. Fix is correct and covers the missing case.

**MERGE.**

PR #2128, opened 2026-06-27
fix(health): memoize chain-events D1 read in handleHealthRequest

### LINKED ISSUE #2090: /health readiness probe runs an unconditional non-memoized D1 query per request, contradicting its no-I/O docstring
## Context

`handleHealthRequest` (`workers/api.mjs:2648`) backs `/health`, the readiness probe in `run_worker_first` (`wrangler.jsonc:37`), hit on a fixed cadence from every colo plus ad-hoc LB checks. Its docstring (`:2646-2647`) claims it reports bindings "without touching R2/KV (no I/O, no cold-start cost)."

## Problem

The handler does I/O on every request, one piece un-memoized:

1. **Un-memoized D1 read per request.** When `bindings.health_db` is true, it runs (`:2707-2713`) a `SELECT ... FROM account_events ORDER BY observed_at DESC LIMIT 1` via `d1All` (`analytics.mjs:136-160`, no caching) — a billed D1 query per probe.
2. **Result does not gate status.** The comment (`:2700-2705`) is explicit: the chain-event heartbeat does NOT gate the HTTP status (computed only from KV pointer staleness). It is pure observability.
3. **HEAD pays too** — method only suppresses the body (`:2756`); the query runs unconditionally.
4. **Docstring is stale** — it also issues two KV reads (`:2679-2681`), though those ARE in-isolate memoized (`latestPointer` `storage.mjs:195-215`, `readHealthMetaKv` `api.mjs:1709-1721`). The D1 read is the one un-amortized per-request I/O.

Each call is cheap, but the volume (one query per probe, per colo, forever) plus the inaccurate docstring make it worth gating.

## Requirements

- Stop issuing a fresh D1 query per request; keep the `chain_events` field shape/semantics (best-effort, null on cold/unbound); don't let it gate status; fix the docstring.

## Deliverables

- Apply the established in-isolate memo pattern (`{ env, value, expiresAt }`, ~30-60s TTL, keyed on `env`, do not cache null) to the chain-event read — same shape as `readHealthMetaKv`/`latestPointer`/`readEconomicsCurrentKv`. (HEAD-skip or `?include=chain_events` opt-in are acceptable complements.)
- Update the docstring at `:2646-2647`.
- A test asserting the D1 read is collapsed across consecutive `/health` requests within the TTL on a warm isolate.

## Acceptance criteria

- Two `/health` requests within the TTL on one warm isolate result in at most one `account_events` query; `chain_events` body is equivalent to today; a stale pointer still yields 503; the docstring no longer claims "no I/O"; existing + new tests pass.

## References

- `workers/api.mjs:2646-2647`, `:2700-2722`, `:2756-2758`, `:1709-1721`; `workers/request-handlers/analytics.mjs:136-160`; `workers/storage.mjs:195-215`; `wrangler.jsonc:37`

---

jeffrey701:
## Summary
- Extract `readChainEventsDb(env, now)` in `workers/api.mjs` with a 30-second in-isolate TTL memo — same pattern as `readHealthMetaKv` / `readEconomicsCurrentKv`
- Replace the direct `d1All()` call in `handleHealthRequest` with the memoized reader; null results are not cached so a cold store stays re-queried
- Fix stale docstring: the handler does issue KV I/O (memoized), so update the comment
- Add `tests/health-chain-events-cache.test.mjs` with 4 tests: TTL collapse, env isolation, absent-binding null, null-not-cached

## Related Issues
Closes #2090.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other

## Testing
- [x] Tests added/updated
- [x] Manually tested

## Checklist
- [x] Code follows project style guidelines
- [x] Self-review completed
- [x] Changes are documented (if applicable)

### DIFF
--- a/tests/health-chain-events-cache.test.mjs
+++ b/tests/health-chain-events-cache.test.mjs
@@ -1,0 +1,86 @@
+import assert from "node:assert/strict";
+import { test } from "vitest";
+import {
+  CHAIN_EVENTS_DB_TTL_MS,
+  readChainEventsDb,
+} from "../workers/api.mjs";
+
+function mkDbEnv(row = { block: 100, at: 1_700_000_000_000 }) {
+  let queries = 0;
+  return {
+    get queries() {
+      return queries;
+    },
+    METAGRAPH_HEALTH_DB: {
+      prepare() {
+        return {
+          bind() {
+            return {
+              async all() {
+                queries += 1;
+                return { results: row ? [row] : [] };
+              },
+            };
+          },
+        };
+      },
+    },
+  };
+}
+
+test("readChainEventsDb memoizes within the TTL — one D1 query for repeated calls", async () => {
+  const env = mkDbEnv();
+  const t0 = 5_000_000;
+  const a = await readChainEventsDb(env, t0);
+  const b = await readChainEventsDb(env, t0 + 1000);
+  assert.deepEqual(a, b);
+  assert.equal(
+    env.queries,
+    1,
+    "a second call within the TTL must be served from the in-isolate memo",
+  );
+
+  await readChainEventsDb(env, t0 + CHAIN_EVENTS_DB_TTL_MS + 1);
+  assert.equal(env.queries, 2, "an expired memo triggers a fresh D1 query");
+});
+
+test("readChainEventsDb never cross-reads a different env (isolation safety)", async () => {
+  const envA = mkDbEnv({ block: 10, at: 1_000 });
+  const envB = mkDbEnv({ block: 20, at: 2_000 });
+  const t0 = 6_000_000;
+  const a = await readChainEventsDb(envA, t0);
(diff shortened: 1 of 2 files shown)


JSONbored:
Traced `readChainEventsDb` (`workers/api.mjs`): new function with 30s per-isolate memoization, keyed on `env` object identity (`chainEventsDbMemo.env === env`). The env identity key correctly scopes the memo to one isolate — different isolates have different `env` object references. Null result is NOT cached (`if (value !== null)` before setting memo) so cold D1 doesn't lock the isolate out. TTL is 30s. SQL query is parameterized. The async-await change on `homepageResponse`/`apiCatalogResponse` call sites is identical to the change in #2129 — will auto-resolve as a same-change merge.

**MERGE.**

PR #2130, opened 2026-06-27
perf(api): complete keyset-cursor pagination across the entity feeds

### LINKED ISSUE #2058: perf(api): complete keyset-cursor pagination across entity feeds
Part of the performance/resource roadmap (parent: #2056).

## Context

The block-explorer / account entity feeds are live D1 routes recomputed on every edge-cache miss. The extrinsics feed already gained keyset (cursor) pagination in #1851 — a row-value seek on the PK — but the sibling feeds did not.

## Problem

The remaining entity feeds still use `LIMIT ? OFFSET ?`, and the offset is clamped to **1,000,000** (`clampInt(..., 0, 0, 1_000_000)`), so a caller can request a deep page that makes SQLite scan-and-skip up to a million rows on large tables (`account_events`, `extrinsics`, `blocks`):

- `workers/request-handlers/entities.mjs:628` — `blocks` feed (`ORDER BY block_number DESC LIMIT ? OFFSET ?`)
- `workers/request-handlers/entities.mjs:431` — account transfers
- `workers/request-handlers/entities.mjs:479` — per-subnet account events
- `src/account-events.mjs:456` — account event history
- `workers/request-handlers/entities.mjs:343` — account daily history

Offset cost is O(offset); keyset is O(limit).

## Requirements

- Extend the existing `encodeCursor` / `decodeCursor` row-value-seek pattern (already used by the extrinsics feed at `entities.mjs:844`) to the feeds above, using each feed's natural ordering key (e.g. `(block_number, event_index)` for events, `block_number` for blocks).
- Keep `offset` accepted as a deprecated fallback for back-compat; cursor takes precedence (mirror the extrinsics feed).
- Emit `next_cursor` in each feed envelope.

## Deliverables

- Keyset cursors on all listed feeds, mirroring `handleExtrinsicsFeed`.
- Tests: cursor round-trip, stable ordering across pages, malformed-cursor → ignored (falls back to first page).
- A note in the parent that offset is now legacy.

## Acceptance criteria

- Deep-page latency is flat (independent of page depth) on the cursor path.
- Existing offset callers keep working; contract/schema updated if `next_cursor` is newly added to a route.

## References

- Existing keyset impl: `workers/request-handlers/entities.mjs:841-862` (`decodeCursor`/`encodeCursor`, #1851)
- Offset sites: `entities.mjs:343,431,479,628`; `src/account-events.mjs:456`

GildardoDev:
## Summary

Completes keyset (cursor) pagination across the remaining offset-only entity feeds, so deep-page reads are O(limit) row-value seeks instead of scanning-and-skipping up to a million offset rows. Extends the existing handleExtrinsicsFeed/loadAccountEvents pattern.

## What Changed

- GET /api/v1/accounts/{ss58}/transfers and GET /api/v1/subnets/{netuid}/events gain an optional ?cursor= and emit next_cursor via a (block_number, event_index) row-value seek over the large account_events tier.
- GET /api/v1/accounts/{ss58}/history gains ?cursor= + next_cursor via a (day, netuid) seek; day is encoded as its natural sortable integer (2026-06-25 -> 20260625) so it fits the integer cursor codec, with netuid as the within-day tiebreaker, and its ordering is now deterministic (day DESC, netuid DESC).
- offset stays accepted as a deprecated fallback on every feed and cursor takes precedence; a malformed cursor is ignored and falls back to the first page.
- The response schemas now document next_cursor for all three feeds, matching the existing events/blocks/extrinsics feeds.
- Tests cover the keyset seek, next_cursor emission, and malformed-cursor fallback for each feed.

## Registry Safety

## Validation

- [x] npm run validate
- [x] npm run validate:api
- [x] npm run validate:openapi
- [x] npm run validate:schemas
- [x] npm run validate:types

Closes #2058

### DIFF
--- a/generated/metagraphed-api.d.ts
+++ b/generated/metagraphed-api.d.ts
@@ -1552,6 +1552,7 @@ export interface components {
             day_count: number;
             days: components["schemas"]["AccountDay"][];
             limit?: number;
+            next_cursor?: string | null;
             offset?: number;
             schema_version: number;
             ss58: string;
@@ -1597,6 +1598,7 @@ export interface components {
         /** @description The native-TAO Balances.Transfer feed for one account (#1850), newest first, from the account_events D1 tier (event_kind='Transfer'). Each row is a directional {from, to, amount_tao, direction} transfer; direction is 'sent' or 'received' relative to the queried ss58. This is the native-TAO transfer feed only, NOT a full balance ledger. Served live at /api/v1/accounts/{ss58}/transfers (no static file). */
         AccountTransfersArtifact: {
             limit?: number;
+            next_cursor?: string | null;
             offset?: number;
             schema_version: number;
             ss58: string;
@@ -4111,6 +4113,7 @@ export interface components {
             events: components["schemas"]["AccountEvent"][];
             limit?: number;
             netuid: number;
+            next_cursor?: string | null;
             offset?: number;
             schema_version: number;
         } & {
@@ -5417,6 +5420,7 @@ export interface operations {
                      *           }
                      *         ],
                      *         "limit": 1,
+                     *         "next_cursor": "example",
                      *         "offset": 1,
                      *         "schema_version": 1,
                      *         "ss58": "example"
@@ -5633,6 +5637,7 @@ export interface operations {
                      * @example {
                      *       "data": {
                      *         "limit": 1,
+                     *         "next_cursor": "example",
                      *         "offset": 1,
                      *         "schema_version": 1,
                      *         "ss58": "example",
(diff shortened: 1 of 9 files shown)


JSONbored:
Traced `handleAccountHistory`, `handleAccountTransfers`, `handleSubnetEvents` (`workers/request-handlers/entities.mjs`): all three handlers add keyset cursor pagination as `(col1, col2) < (?, ?)` row-value seeks. **SQL injection check**: row-value seek uses `?` placeholders — no interpolation. Cursor values come from `decodeCursor()` which validates integer arity before returning a non-null result; malformed cursor `→ null → useCursor = false → offset fallback`. History cursor: `(day, netuid)`, day converted from integer format (20260624 → '2026-06-24') via regex — safe (not from user input directly). Transfers and subnet-events: `(block_number, event_index)` — both D1 integers. `next_cursor` only set when `rows.length === limit` (full page returned). Contract artifacts regenerated.

**MERGE.**

PR #2143, opened 2026-06-27
fix(api): reject negative account_events PK in validEventRows

### LINKED ISSUE #2141: validEventRows accepts negative block_number and event_index
## Problem

validBlockRows and validExtrinsicRows both require non-negative primary-key components (block_number >= 0, extrinsic_index >= 0), but validEventRows only checks Number.isInteger without a lower bound.

Staged ingest and realtime account_events payloads can therefore admit rows with negative (block_number, event_index) keys. Those rows should be dropped before D1 insert, matching the blocks and extrinsics loaders.

## Expected

validEventRows rejects negative block_number and event_index, consistent with the other chain ingest validators.

## Scope

Backend validation and tests only; no schema or generated artifact changes.

pekar9781:
## Summary

Closes #2141

Align validEventRows with the blocks and extrinsics ingest validators by rejecting negative block_number and event_index primary-key components before account_events rows reach D1.

## What Changed

- Tighten validEventRows to require block_number >= 0 and event_index >= 0, matching validBlockRows and validExtrinsicRows.
- Add unit coverage for the negative-PK rejection path.
- Add staged-load coverage proving negative rows are dropped during ingest.

## Registry Safety

## Validation

- [ ] npm run check
- [ ] npm run validate
- [ ] npm run validate:schemas
- [ ] npm run validate:api
- [ ] npm run validate:openapi
- [ ] npm run validate:types
- [ ] npm run validate:artifact-budgets
- [ ] npm run validate:docs
- [ ] npm run validate:intake
- [ ] npm run validate:workflows
- [ ] npm run worker:test
- [ ] npm run test:coverage
- [ ] npm run scan:public-safety
- [ ] git diff --check

Focused tests (not run in this environment — no Node 22 available):

- npx vitest run tests/account-events.test.mjs tests/load-staged-events.test.mjs

### DIFF
--- a/src/account-events.mjs
+++ b/src/account-events.mjs
@@ -153,7 +153,9 @@ export function validEventRows(rows) {
     ? rows.filter(
         (r) =>
           Number.isInteger(r?.block_number) &&
+          r.block_number >= 0 &&
           Number.isInteger(r?.event_index) &&
+          r.event_index >= 0 &&
           typeof r?.event_kind === "string" &&
           Number.isInteger(r?.observed_at),
       )
--- a/tests/account-events.test.mjs
+++ b/tests/account-events.test.mjs
@@ -38,6 +38,9 @@ test("validEventRows enforces the strict row shape (#1371)", () => {
     validEventRows([{ ...good, observed_at: "x" }]).length,
     0, // observed_at must be an integer
   );
+  // negative PK components — aligned with validBlockRows / validExtrinsicRows
+  assert.equal(validEventRows([{ ...good, block_number: -1 }]).length, 0);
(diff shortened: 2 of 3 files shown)


JSONbored:
Traced `validEventRows` (`src/account-events.mjs`): the new guards add `r.block_number >= 0 && r.event_index >= 0` after the existing `Number.isInteger()` checks. `block_number: 0` (genesis) and `event_index: 0` (first event in block) are both valid — `>= 0` is the correct operator here. The D1 schema declares these NOT NULL but doesn't enforce positivity at the write path, so edge-case offchain rows could arrive with negative indices. Guards are real, operators correct.

**MERGE.**

PR #2154, opened 2026-06-27
fix(indexer): default to substrate-interface-compatible WebSocket archive

JSONbored:
### Motivation
- The indexer defaulted to `wss://archive.chain.opentensor.ai:443`, but `archive.chain.opentensor.ai` is an HTTP JSON-RPC/Subway proxy and does not speak `substrate-interface`'s WebSocket protocol, which can cause the indexer to repeatedly fail and stall backfills. 
- The goal is to make the default endpoint a known-good WebSocket archive while preserving `EVENTS_RPC_URL` override behavior so deployments that set a custom RPC continue to work.

### Description
- Replace the hard-coded incompatible default with a `DEFAULT_RPC = "wss://bittensor-finney.api.onfinality.io/public-ws"` and use `RPC = os.environ.get("EVENTS_RPC_URL", DEFAULT_RPC)` in `scripts/index-chain.py` while adding explanatory comments. 
- Update the Railway bootstrap default in `deploy/railway-bootstrap.sh` and the example in `deploy/README.md` to reference the OnFinality WebSocket archive by default. 
- Add a unit regression test to `scripts/test_index_chain.py` that asserts the module `DEFAULT_RPC` and `RPC` are the substrate-interface-compatible WebSocket archive when `EVENTS_RPC_URL` is unset. 
- Preserve the existing behavior that an explicit `EVENTS_RPC_URL` environment variable overrides the default.

### Testing
- Ran `python3 -m unittest scripts/test_index_chain.py` which executed the test suite (including the new regression) and passed. 
- Ran the public-safety scan with `npm run scan:public-safety` which completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3f91800498832cb027ded12a01382c)

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -120,7 +120,7 @@ cross-service variable references, e.g.:
 railway add -s indexer --repo JSONbored/metagraphed --branch main \
   -v DATABASE_URL='${{Postgres.DATABASE_URL}}' \
   -v REDIS_URL='${{Redis.REDIS_URL}}' \
-  -v EVENTS_RPC_URL='wss://archive.chain.opentensor.ai:443'   # archive, NOT pruned entrypoint
+  -v EVENTS_RPC_URL='wss://bittensor-finney.api.onfinality.io/public-ws'   # WS archive, NOT pruned entrypoint
 ```
 
 The public `wss-lb` is independent of Postgres/Redis (it reads only the public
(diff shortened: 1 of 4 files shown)


JSONbored:
Closing as superseded: this changes the runtime RPC default of `scripts/index-chain.py` — the Python follow-head indexer that was replaced tonight (2026-06-27) by the Rust indexer (the `indexer-rs` service, which resumes from Postgres and connects via its own EVENTS_RPC_URL). The Python `index-chain.py` is retired, so its default no longer takes effect. If the streamer (`stream-events.py`) needs the same archive default, that's a separate, narrower change.

PR #2228, opened 2026-06-28
fix(api): drop incomplete oldest day in buildEconomicsTrends when row cap is reached

### LINKED ISSUE #2227: buildEconomicsTrends serves an incomplete oldest day when the row cap is reached
## Bug

handleEconomicsTrends in workers/request-handlers/analytics-routes.mjs queries subnet_snapshots ordered snapshot_date DESC LIMIT 60000 (ECONOMICS_TRENDS_ROW_CAP). When enough subnets x days fill that cap, the oldest snapshot_date in the result is incomplete - only the subnets whose rows fit before the LIMIT are included.

uildEconomicsTrends then aggregates rows by day. A truncated oldest day has too few subnets, so its 	otal_stake_tao, lpha_price_tao_weighted, alidator_count, miner_count, and mean_emission_share are all underreported.

## Expected

When the row cap is reached, the incomplete oldest day should be dropped (or the client should know it is partial). The sibling uildConcentrationHistory already handles this correctly via a capped flag that slices off the boundary day.

## Actual

The incomplete oldest day is served as if it were a full network snapshot, understating all aggregated metrics for that day.

## Fix

Mirror the uildConcentrationHistory pattern: pass capped: rows.length >= ECONOMICS_TRENDS_ROW_CAP from the handler, and in uildEconomicsTrends drop the last (oldest) day when capped && days.length > 1.

joaovictor91123:
## Summary

When handleEconomicsTrends hits ECONOMICS_TRENDS_ROW_CAP (60 000 rows), the SQL ORDER BY snapshot_date DESC LIMIT 60000 truncates the oldest rows first. The oldest snapshot_date in the result may therefore only contain a subset of subnets - the rest were cut off. uildEconomicsTrends then serves that partial day as a full network aggregate, understating 	otal_stake_tao, alidator_count, miner_count, lpha_price_tao_weighted, and mean_emission_share for that day.

## What Changed

- src/neuron-history.mjs - uildEconomicsTrends now accepts a capped option; when true and there are multiple days, it drops the last (oldest) day before building the result, mirroring the identical guard in uildConcentrationHistory.
- workers/request-handlers/analytics-routes.mjs - handleEconomicsTrends passes capped: rows.length >= ECONOMICS_TRENDS_ROW_CAP to the builder.
- 	ests/neuron-history.test.mjs - new test asserts the oldest day is dropped when capped=true and retained when capped=false.

## Registry Safety

## Validation

- [x] \
pm run check\
- [x] \
pm run validate\
- [x] \
pm run validate:schemas\
- [x] \
pm run validate:api\
- [x] \
pm run validate:openapi\
- [x] \
pm run validate:types\
- [x] \
pm run validate:artifact-budgets\
- [x] \
pm run validate:docs\
- [x] \
pm run validate:intake\
- [x] \
pm run validate:workflows\
- [x] \
pm run worker:test\
- [x] \
pm run test:coverage\
- [x] \
pm run scan:public-safety\
- [x] \git diff --check\

Fixes #2227

### DIFF
--- a/src/neuron-history.mjs
+++ b/src/neuron-history.mjs
@@ -309,7 +309,7 @@ export function buildNeuronHistory(rows, netuid, uid, { window } = {}) {
 // stake-weighted mean + median alpha price can be computed here. Rows arrive newest
 // first; the output preserves that order. Null-safe throughout — a metric is null
 // for a day only when NO subnet reported it.
-export function buildEconomicsTrends(rows, { window } = {}) {
+export function buildEconomicsTrends(rows, { window, capped } = {}) {
   const byDay = new Map(); // snapshot_date -> accumulator (insertion order = newest first)
   for (const r of rows || []) {
     const day = r.snapshot_date;
@@ -363,7 +363,7 @@ export function buildEconomicsTrends(rows, { window } = {}) {
       }
     }
   }
-  const days = [...byDay.entries()].map(([snapshot_date, acc]) => ({
+  let days = [...byDay.entries()].map(([snapshot_date, acc]) => ({
     snapshot_date,
     subnet_count: acc.subnet_count,
     total_stake_tao: acc.stake_seen ? roundTao(acc.stake_sum) : null,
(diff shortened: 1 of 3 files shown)


JSONbored (commented):
merge conflicts, please fix and resubmit.

PR #2244, opened 2026-06-28
fix(api): match REST subnet enum/string filters case-insensitively

### LINKED ISSUE #2073: REST subnet list enum/string filters are case-sensitive while MCP list_subnets lowercases — same value, different outcome
## Context

The registry exposes the subnet list through two surfaces the project treats as documented equivalents: **REST** `GET /api/v1/subnets` (filtered by `workers/list-query.mjs`, wired at `workers/api.mjs:2047`, validated against `src/contracts.mjs:560`) and **MCP** `list_subnets` (`src/mcp-server.mjs:749`). The two handlers normalize filter input differently, so the same filter value yields opposite outcomes.

## Problem

**REST is case-sensitive.** `workers/list-query.mjs:312` compares the raw param against the enum (`schema.enum && !schema.enum.includes(value)`); `:102` filters with exact equality (`String(value) === expected`). The `subnets` enums are all lowercase: `status`/`subnet_type` (`src/contracts.mjs:575-576` → `:72-73` `["active","inactive"]`/`["root","application"]`) and `domain` = `DOMAIN_TAGS` (all lowercase, `src/domain-tags.mjs:70`). So `?status=Active`, `?subnet_type=Application`, `?domain=Inference` all return **HTTP 400 `invalid_query`**.

**MCP is case-insensitive (for these filters).** `src/mcp-server.mjs:752-789` lowercases both the arg and the row value, so `status:'Active'` etc. succeed.

Empirically confirmed: `enum.includes("Active") => false` (REST 400) vs MCP `status:"Active"` matches `status=="active"` (true). Same documented filter value, opposite result.

Secondary: MCP is itself internally inconsistent — the three free-string filters lowercase, but `sort`/`order` go through `optionalEnum` (`src/mcp-server.mjs:560`, case-sensitive). And there is no literal "parity" contract; the expectation is implied by mirroring comments and the shared vocabulary.

## Requirements

- Pick one casing convention for enum/exact-string filters and apply it consistently.
- Lowest-risk: normalize enum + string-equality comparisons **case-insensitively in the REST path** (lowercase input before the `enum.includes` check in `validateListQuery` and before the equality/array/csv comparisons in `filterRows`). The configured vocabularies and stored data are already lowercase, so lowercasing the input alone restores parity.
- Scope to enum/string filters only; do not change numeric/range/`netuids` validation or `q` search.
- Centralize in `workers/list-query.mjs`; optionally align MCP `sort`/`order` or document it as strict. No contract regen if only comparison logic changes.

## Deliverables

- Patch to `workers/list-query.mjs` making enum + string-equality filter matching case-insensitive (input side).
- Tests asserting `?status=Active`, `?subnet_type=Application`, `?domain=Inference` return 200 with the same rows as lowercase forms, and that MCP `list_subnets` and the REST equivalent agree.

## Acceptance criteria

- `GET /api/v1/subnets?status=Active` returns 200 with the same result set as `?status=active` (same for the others).
- Genuinely invalid values (e.g. `?status=bogus`) still return 400 `invalid_query`.
- No change to numeric/range/`netuids`/projection validation; no contract-drift CI failure.

## References

- `workers/list-query.mjs:312`, `:102`, `:86`/`:95`, `:190`/`:202`
- `src/contracts.mjs:560`, `:574-576`, `:72-73`, `:2447`; `src/domain-tags.mjs:70`
- `src/mcp-server.mjs:749`, `:752-789`, `:560`; `workers/api.mjs:2047`

---

boskodev790:
## Summary

- REST `GET /api/v1/subnets` compared filter input **case-sensitively** (`validateListQuery` `enum.includes` + `filterRows` equality), while the equivalent MCP `list_subnets` tool lowercases its args. The subnets vocabularies (`status`, `subnet_type`, `domain`) are all lowercase, so `?status=Active`, `?subnet_type=Application`, `?domain=Inference` returned **HTTP 400 `invalid_query`** where the same MCP call succeeded — the same documented filter value, opposite outcome.

## What Changed

- `workers/list-query.mjs`: lowercase the input before the enum membership check (`validateListQuery`) and in the string-equality + array-membership comparisons (`filterRows`). The configured vocabularies and stored row values are already lowercase, so lowercasing the input restores parity. The numeric `netuids` CSV path is left untouched (scoped out per the issue); numeric/range/`q`-search validation is unchanged.
- `tests/list-query.test.mjs`: tests asserting `?status=Active` / `?subnet_type=Application` / `?domain=Inference` return 200 with the same rows as the lowercase forms, and that a genuinely invalid value (`?status=Bogus`) still errors.

## Registry Safety

- [x] Public API/OpenAPI/schema changes are intentional and documented. (No contract regen — comparison logic only; the contract vocabularies are unchanged.)

## Validation

- [x] `npx vitest run tests/list-query.test.mjs tests/data-api.test.mjs tests/network-routing.test.mjs` — 80 passed
- [x] `npx vitest run tests/mcp-server.test.mjs tests/api-coverage.test.mjs tests/worker-runtime.test.mjs` — 420 passed
- [x] prettier `--check` + eslint clean on changed files
- [x] `git diff --check`

Closes #2073


JSONbored (commented):
Please fix this issue, then feel free to resubmit: 

```
The main maintainability concern is that the matcher now lowercases every non-CSV filter globally, not just the lowercase enum/string vocabularies called out in the issue.
```

PR #2250, opened 2026-06-28
fix(api): harden entity ref parsing against malformed refs

### LINKED ISSUE #2063: fix(api): harden entity ref/id parsing — malformed refs resolve to wrong-but-valid results
Part of the performance/resource roadmap (parent: #2056).

## Context

The entity handlers resolve user-supplied refs/ids (block number/hash, extrinsic hash, composite `<block>-<index>`). Recent fixes already hardened several of these (#1949 malformed account-history netuid, #1950 composite extrinsic artifacts, #1951 transfer directions, #1953 numeric block extrinsics refs) — indicating the area is bug-prone.

## Problem (verified)

The composite extrinsic ref parser is too lenient:

```js
// workers/request-handlers/entities.mjs:901
const [b, i] = ref.split("-");
const blockNumber = Number(b);
const extrinsicIndex = Number(i);
// queried when Number.isInteger(blockNumber) && Number.isInteger(extrinsicIndex)
```

`split("-")` + `Number()` accepts malformed refs and resolves them to a **wrong-but-valid** lookup instead of a clean miss (verified):

| input | parsed as | should be |
|---|---|---|
| `123-45-67` | block 123, index 45 (extra segment dropped) | miss |
| `123-` | block 123, index 0 (`Number("")===0`) | miss |
| `-45` | block 0, index 45 | miss |
| `0x1-2` | block 1, index 2 (`Number` accepts hex) | miss |
| `1e3-2` | block 1000, index 2 (`Number` accepts sci notation) | miss |

Same failure class as the just-fixed #1950/#1953: a malformed ref returns someone else's data rather than `:null`.

## Requirements

- Parse all entity refs with a **strict** integer matcher (e.g. `/^\d+$/` per segment, exactly two segments for composite) instead of `Number()`/`parseInt`. Reject hex, scientific notation, empty halves, extra segments, leading-zero ambiguity as decided.
- Audit the sibling parsers in the same file (block ref, extrinsic hash vs composite branch, any `Number(`/`parseInt(` on path/query segments) for the same leniency.
- Malformed ref → existing schema-stable `:null` miss (never 404/throw), consistent with the current contract.

## Deliverables

- Hardened ref parsing across the entity handlers.
- Regression tests for each malformed shape in the table above (and block-ref equivalents).

## Acceptance criteria

- Every malformed shape returns the schema-stable `:null` miss; valid refs unchanged.

## References

- `workers/request-handlers/entities.mjs:889-909` (`handleExtrinsic` composite branch)
- Prior fixes: #1949, #1950, #1951, #1953

cleanjunc:
## Summary

The single entity detail handlers resolve a caller supplied ref through Number() or split("-") plus Number(). That lenient coercion turns a malformed ref into a valid looking lookup, so a request can resolve to another entity instead of the null result the handlers already promise. For example extrinsics/123-45-67 resolved to block 123 index 45, and a direct blocks/0x10 coerced to block 16. This replaces the coercion with a strict integer matcher so any malformed ref returns the existing null miss.

Closes #2063

## What Changed

- Added a shared parseRefInt helper next to parseBoundedIntParam, reusing the same /^\d+$/ plus Number.isSafeInteger pattern. It returns null for any segment that is not bare decimal digits (hex, scientific notation, signs, decimals, empty halves, overflow).
- handleExtrinsic composite branch now requires exactly two bare integer halves, so 123-45-67, 123-, -45, 0x1-2, and 1e3-2 return extrinsic:null instead of resolving another row.
- handleBlock, handleBlockExtrinsics, and handleBlockEvents parse the numeric ref strictly, so a malformed numeric ref returns block:null without a coerced block_number lookup.
- The extrinsic_hash branch and the netuid query guard were already strict and stay unchanged. Feed filter coercion (limit, offset, block, from, to) is a separate concern and is left untouched. No contract or schema changed (validate:contract-drift stays green).
- Added regression tests for every malformed shape in the issue table plus the block ref equivalents, asserting the null miss and that no lookup is issued, and confirming valid refs and index 0 still resolve.

## Registry Safety

      validator-local state.

## Validation

- [ ] `npm run check`
- [x] `npm run validate`
- [x] `npm run validate:api`
- [x] `npm run validate:docs`
- [x] `npm run worker:test`
- [x] `npm run test:coverage`
- [x] `git diff --check`

### DIFF
--- a/tests/request-handlers-entities.test.mjs
+++ b/tests/request-handlers-entities.test.mjs
@@ -1965,6 +1965,26 @@ describe("handleBlock", () => {
     assert.ok(idx !== -1, "expected a block_hash lookup");
     assert.equal(captures.params[idx][0], lowerHash);
   });
+
+  test("malformed numeric refs miss instead of resolving another block", async () => {
+    // Number() would coerce each of these (hex, scientific notation, a decimal
+    // point, or an unsafe-integer overflow) to a valid-looking block_number and
+    // resolve the seeded block; strict parsing makes them a miss without ever
+    // issuing the block_number lookup.
+    for (const ref of ["0x1f", "1e3", "12.0", "9007199254740992"]) {
+      const { env, captures } = dbWith({ blockDetail: blockRow() });
+      const body = await json(
+        await handleBlock(req(`/api/v1/blocks/${ref}`), env, ref),
+      );
+      assert.equal(body.data.block, null, `${ref} should miss`);
+      assert.ok(
+        !captures.sql.some((s) =>
+          /FROM blocks WHERE block_number = \?/.test(s),
+        ),
+        `${ref} must not issue a block_number lookup`,
+      );
+    }
+  });
 });
 
 describe("handleBlockExtrinsics", () => {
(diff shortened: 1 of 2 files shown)


cleanjunc:
Updated in response to the review nits.

- The repeated block-ref resolution in `handleBlockExtrinsics` and `handleBlockEvents` is now a shared `resolveBlockNumber` helper (one strict parse plus blocks-tier lookup), removing the duplication noted in the review.
- The regression suite gained unsafe-integer overflow cases (`9007199254740992`) for both the block refs and the composite extrinsic ref, so the strict `Number.isSafeInteger` reject path is exercised and patch coverage on the change is 100%.
- Valid refs, index 0, and the `0x` hash branch stay unchanged; the malformed-ref tests were verified to fail against the pre-fix handler, so each one closes a real gap.

JSONbored (commented):
Sorry, but there are merge conflicts preventing a clean merge of this PR: 

```
Use the [web editor](https://github.com/JSONbored/metagraphed/pull/2250/conflicts) or the command line to resolve conflicts before continuing.

```

PR #2277, opened 2026-06-29
feat(api): negation filters (not_) on list endpoints

### LINKED ISSUE #1946: List query: add negation (exclusion) filters (not_<field>) to the list-query engine

jason020818:
## Summary

Re-applies the list-query negation filter onto current `main` (the prior PR #1947 was closed for a base conflict — maintainer asked to rebase and reopen). The change is unchanged in intent and re-verified non-duplicate: `main` still has no `not_`/negation support.

The list-query engine supported exact-equality, CSV-membership, array-membership, inclusive numeric range (`min_`/`max_`), `q` search, sort, and cursor pagination — but no way to **exclude** rows matching a value. This adds a generic `not_<field>` parameter for each configured filter field: `?not_F=v` returns the exact complement of `?F=v`, mirroring the CSV/array/scalar match field-for-field and validated against the same schema. It derives from the existing `filters` allow-list, is **registered in the canonical edge-cache key** (`listQueryParamNames`), and applies across every list collection — parallel to the `min_`/`max_` range filters.

## Changes
- `workers/list-query.mjs`: `excludeFilterRows` (negation pass wired after the range filter), a `validateFilterValue` helper validating both `F` and `not_F` against one schema, and `not_F` registered in `listQueryParamNames` for cache canonicalization.
- `src/contracts.mjs`: emit a `not_F` query parameter for every filter field.
- Regenerated `openapi.json`, `api-index.json`, and generated type declarations.
- `tests/list-query.test.mjs`: 13 cases (scalar/CSV/array negation, absent-field rule, multi-`not_` AND, exact-complement partition, composition with sort, four validation branches).

## Validation
- `npm run lint`, `npm run format:check`, `npm run validate:contract-drift`, `npm run validate:openapi` (94 routes) — green
- `npx vitest run tests/list-query.test.mjs` — 57 passed; new lines fully covered
- No `packages/client` bump (handled post-merge by the sync-client-version workflow, per CONTRIBUTING)

Closes #1946

## Note: artifact-verifier fix included

The 130 new `not_` parameters push `public/metagraph/openapi.json` just past 1 MiB (1,078,964 bytes). That exposed a latent ENOBUFS bug in `scripts/ci-verify-submitted-artifacts.mjs`: its `gitShow` helper used `execFileSync` with Node's default 1 MiB stdout cap, so reading the committed >1 MiB artifact threw and it reported the file as "unreadable" (the `Verify submitted public artifacts are reproducible` step). Raised `maxBuffer` to 256 MiB — a pure robustness fix that doesn't weaken the content comparison. Its test (`tests/ci-artifact-verifier.test.mjs`) still passes.

### DIFF
--- a/generated/metagraphed-api.d.ts
+++ b/generated/metagraphed-api.d.ts
@@ -7481,6 +7481,10 @@ export interface operations {
                 kind?: "archive" | "dashboard" | "data-artifact" | "docs" | "example" | "openapi" | "repo-registry" | "sdk" | "source-repo" | "sse" | "subnet-api" | "subtensor-rpc" | "subtensor-wss" | "website";
                 provider?: string;
                 state?: "schema-invalid" | "schema-valid" | "maintainer-review" | "verified" | "stale" | "rejected";
+                not_netuid?: number;
+                not_kind?: "archive" | "dashboard" | "data-artifact" | "docs" | "example" | "openapi" | "repo-registry" | "sdk" | "source-repo" | "sse" | "subnet-api" | "subtensor-rpc" | "subtensor-wss" | "website";
+                not_provider?: string;
+                not_state?: "schema-invalid" | "schema-valid" | "maintainer-review" | "verified" | "stale" | "rejected";
                 fields?: string;
                 limit?: number;
                 cursor?: number;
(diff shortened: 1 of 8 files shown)


JSONbored (commented):
This is a really large PR (4400+ LOC diff), i'd prefer if we broke this down into smaller chunks (ideally 1000 LOC max). 

Also, there are a number of nits mentioned in the gittensory review comment, please address them when resubmitting.

PR #2309, opened 2026-06-29
fix(api): canonicalize chain-calls edge-cache key on group_by and limit

### LINKED ISSUE #2303: bug(api): chain-calls edge-cache key not canonical on group_by and limit

joaovictor712:
## Summary

- Pass `canonicalAnalyticsCacheRoute(url, [\"group_by\", \"limit\"])` as the 6th arg to `withEdgeCache` in `handleChainCalls`.
- Out-of-order or differently-encoded `group_by`/`limit` parameters now share one cache slot instead of triggering duplicate D1 aggregations.

Fixes #2303

## Template Used

- Backend/code change

## What Changed

- `workers/request-handlers/analytics.mjs`: thread canonical cache key into `handleChainCalls`.
- `tests/analytics-edge-cache.test.mjs`: regression tests for canonical key ordering and param reordering.

## Registry Safety

## Validation

- [x] `npm test -- tests/analytics-edge-cache.test.mjs tests/request-handlers-analytics.test.mjs`
- [x] `git diff --check`

### DIFF
--- a/migrations/0026_account_events_transfer_pair_index.sql
+++ b/migrations/0026_account_events_transfer_pair_index.sql
@@ -1,0 +1,6 @@
+-- Counterparty relationship drilldowns read native-TAO Transfer rows for one
+-- ordered account pair (and the reverse pair) newest-first. The single-party
+-- hotkey/coldkey indexes can still force large residual scans for popular
+-- accounts, so keep this pair lookup seekable for the public API route.
+CREATE INDEX IF NOT EXISTS idx_account_events_transfer_pair
+  ON account_events (event_kind, hotkey, coldkey, block_number, event_index);
--- a/public/.well-known/agent-skills/index.json
+++ b/public/.well-known/agent-skills/index.json
@@ -2,8 +2,8 @@
   "$schema": "https://api.metagraph.sh/.well-known/agent-skills/schema.json",
   "skills": [
     {
-      "description": "Use when a developer asks what a Bittensor subnet does, whether it's up right now, or how to call/integrate its API — e.g. \"which subnet does image generation\", \"is subnet 7 healthy\", \"how do I call the Chutes API\", \"build on a Bittensor subnet\". Backed by metagraphed (api.metagraph.sh), the live operational + integration registry for ~129 subnets.",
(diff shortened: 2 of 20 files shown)


JSONbored (commented):
Please resubmit after fixing the main concern: 

```
The main remaining gap is that the new cache key is canonical by parameter order/encoding, not by parsed semantic defaults, so some equivalent requests can still occupy separate edge-cache slots.
```

PR #2319, opened 2026-06-29
feat(api): add block_start/block_end range filter to the subnet events feed

GildardoDev:
## Summary

Adds an optional block-height range filter (?block_start / ?block_end) to the public per-subnet event stream GET /api/v1/subnets/{netuid}/events, for parity with the extrinsics feed and the chain-events feed that already expose the same range filter. It lets a caller scope a subnet's event history to a block window without paging the whole stream.

## What Changed

- handleSubnetEvents now accepts ?block_start and ?block_end, parsed with the shared parseNonNegativeIntParam helper (400 on a non-integer) and ANDs block_number >= ? / block_number <= ? onto the query. The range is index-satisfiable via the existing idx_account_events_netuid (netuid, block_number) and idx_account_events_netuid_kind, so a bounded range seeks rather than scans this short-cached public route, no migration needed.
- Declared the two params on the subnet-events route in src/contracts.mjs and regenerated the contract artifacts (openapi.json, types, api-index) via npm run build.
- Tests: the range filter is applied and bound, and a non-integer block_start or block_end returns 400.

## Registry Safety

- [x] No secrets, PATs, wallet data, private URLs, or validator-local state.
- [x] Generated artifacts were produced by npm run build, not hand-edited.
- [x] The public API change is additive (two optional query params) and documented in the route contract.

## Validation

- [x] npm run build (contract regenerated; validate:contract-drift clean)
- [x] npm run validate / validate:api / validate:openapi / validate:schemas
- [x] npm run lint and npm run format:check
- [x] vitest tests/request-handlers-entities.test.mjs (178 pass, including the new block-range tests)


JSONbored (commented):
Please fix this gittensory review concern: 

```
The visible tests cover binding and invalid parameter rejection, but they do not assert the HTTP status for the rejection path.
```

Resubmit as a new PR once that's fixed, and we should be good to approve + merge.

PR #2325, opened 2026-06-29
test(chain-analytics): cover buildChainFees junk-row filtering + cold store

jeffrey701:
## Summary

- `buildChainFees` (`src/chain-analytics.mjs`) defensively filters malformed `dailyRows` (null / non-object / missing string `day`) and `payerRows` without a `signer`, and returns a schema-stable empty shape on cold input. The existing suite only ever fed well-formed rows, so those `.filter()` branches and the cold-store path were never exercised.

## What Changed

- `tests/chain-analytics.test.mjs`: add tests asserting (1) junk daily rows and signer-less payer rows are dropped while valid rows survive, and (2) `buildChainFees({ window })` returns the schema-stable empty payload (`day_count: 0`, empty arrays) on cold input.

Test-only change — no runtime code touched.

## Registry Safety

- [x] Public API/OpenAPI/schema changes are intentional and documented. (None — tests only.)

## Validation

- [x] `npx vitest run tests/chain-analytics.test.mjs` — 31 passed
- [x] `git diff --check`

### DIFF
--- a/tests/chain-analytics.test.mjs
+++ b/tests/chain-analytics.test.mjs
@@ -615,6 +615,47 @@ test("buildChainFees computes per-day averages + null avg on a zero-extrinsic da
   assert.equal(out.top_fee_payers[0].signer, "5Pay");
 });
 
+test("buildChainFees drops junk daily rows and payer rows without a signer", () => {
+  const out = buildChainFees({
+    window: "30d",
+    dailyRows: [
+      null,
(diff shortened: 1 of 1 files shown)


JSONbored (commented):

```
The most notable gap is that the malformed-row test is builder-level only, so it verifies defensive branches rather than proving a live D1 path can produce those shapes.
```

PR #2326, opened 2026-06-29
test(counterparties): cover the equal-volume sort tie-breaks

jeffrey701:
## Summary

- `buildCounterparties` (`src/counterparties.mjs`) ranks counterparties by total volume, then by newest `last_block`, then by ascending `address`. The existing suite only fed **distinct-volume** rows, so the comparator's `last_block` and `address` tie-break branches were never exercised.

## What Changed

- `tests/counterparties.test.mjs`: add a test with three **equal-volume** counterparties asserting the newest-block-first then ascending-address ordering, covering the previously-untested comparator branches (mirrors the list-query sort-tiebreak coverage).

Test-only change — no runtime code touched.

## Registry Safety

- [x] Public API/OpenAPI/schema changes are intentional and documented. (None — tests only.)

## Validation

- [x] `npx vitest run tests/counterparties.test.mjs` — 23 passed
- [x] `git diff --check`

### DIFF
--- a/tests/counterparties.test.mjs
+++ b/tests/counterparties.test.mjs
@@ -51,6 +51,24 @@ describe("buildCounterparties", () => {
     assert.equal(data.counterparties[2].address, "B");
   });
 
+  test("breaks equal-volume ties by newest last_block, then by ascending address", () => {
+    // Three counterparties with the SAME total volume (100 each), so the
+    // comparator falls through to its last_block then address tie-breaks — the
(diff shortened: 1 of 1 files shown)


JSONbored (commented):
Rejecting, gittensory found this mostly duplicates existing tests: 

```
The main issue is that the file already has a later regression test for the same tie-break rules, so the added coverage is mostly duplicative.
```

PR #2020, opened 2026-06-26
fix(api): clamp sub-perfect chain activity success_rate below 1

joaovictor712:
## Summary

- `buildChainActivity` now clamps a sub-perfect `success_rate` that would round to `1.0` down to `0.9999`.

## What Changed

- `src/chain-analytics.mjs`: extend `round4` with the same sub-1 clamp used by `displayUptimeRatio`.
- `tests/chain-analytics.test.mjs`: regression test for 99_999/100_000.

## Why it was observably wrong

A day with one failed extrinsic among 100k was reported as a perfect `success_rate = 1.0` on the chain activity endpoint.

## Template Used

- backend-code.md

## Registry Safety

- [x] No secrets, PATs, wallet data, private dashboards, private URLs, or validator-local state.
- [x] Generated artifacts were produced by repo scripts, not hand-edited.
- [x] R2-only/high-churn detail artifacts are not committed.
- [x] No schema/contract change.

## Validation

- [x] `npx vitest run tests/chain-analytics.test.mjs` (24 passed)
- [x] `git diff --check`

Fixes #2019

PR #2331, opened 2026-06-29
fix(api): clamp sub-perfect RPC-usage rates below a perfect 1.0

boskodev790:
## Summary

- `formatRpcUsage` (`src/health-serving.mjs`) rounds `error_rate`, `failover_rate`, and `cache_hit_rate` with a `round4()` that used `toFixed(4)` only — so a near-perfect ratio (e.g. 999,996 of 1,000,000 cache hits = 0.999996) rounds **up** to a misleading `1.0` ("100% cache hit") even though some requests missed. The sibling uptime (#2224), turnover (#2299), and chain-activity (#2020) ratios already guard against this overstatement.

## What Changed

- `src/health-serving.mjs`: `round4` now clamps a sub-1 value that rounds to `>= 1` down to `0.9999`. A genuine rate of exactly 1 (all requests hit / errored, `value === 1`) is preserved — the guard only triggers when `value < 1`. `round4` is used solely for these three 0..1 rates (summary + per-endpoint + per-network).
- `tests/rpc-usage.test.mjs`: a sub-perfect cache-hit rate clamps to `0.9999` (not `1.0`); a true 100% rate still reports `1.0`.

## Registry Safety

- [x] Public API/OpenAPI/schema changes are intentional and documented. (No contract change — a value-accuracy fix to existing number|null rate fields.)

## Validation

- [x] `npx vitest run tests/rpc-usage.test.mjs` — 14 passed
- [x] `npx vitest run tests/health-serving.test.mjs tests/request-handlers-rpc-proxy.test.mjs` — 150 passed
- [x] `git diff --check`


JSONbored (commented):
Closing as codecov/patch CI is failing due to missing test coverage, and gittensory review agent found an issue that needs to be resolved: 

```
The notable gap is that the visible tests do not exercise the same clamp through every rate field/path that shares `round4`.
```

PR #2341, opened 2026-06-29
feat(badge): add style=for-the-badge render style

### LINKED ISSUE #2340: feat(badge): add style=for-the-badge render style
### Problem
The embeddable badge supports `style=flat` (default) and `style=flat-square`. shields.io's most popular style — **`for-the-badge`** (the taller, all-caps, bold, letter-spaced look) — is what a lot of READMEs standardize on, but there's no way to get it from the metagraphed badge, so a subnet's status badge looks out of place next to the rest.

### Proposal
Add `?style=for-the-badge`: same two-segment badge and the same metric/label/color logic, rendered in the shields `for-the-badge` style — taller (28px), uppercase text, bold, letter-spaced, square corners, matte (no gradient).

Self-contained in `src/badge.mjs` (`renderBadge` is parametrized by style); no schema/contract/artifact change. The existing `flat` / `flat-square` output is preserved byte-for-byte (covered by a regression test), and `textWidth` stays a safe overestimate so the larger glyphs never clip.

### Example
`metagraphed`

glorydavid03023:
## Summary

Adds `?style=for-the-badge` to the embeddable badge, alongside the existing `flat` (default) and `flat-square`. It renders the same two-segment badge — same `metric`/`label`/color logic — in shields.io's popular **for-the-badge** look.

Closes #2340

## Behavior

- Taller (28px), **uppercase** text, **bold**, letter-spaced, square corners, matte (no gradient).
- Works with every existing option: `?metric=…`, `?label=…`, subnet + provider badges.
- Unknown styles still fall back to `flat`.

## Scope

- Self-contained in `src/badge.mjs` — `renderBadge` is parametrized by style; no schema/contract/artifact change (badges aren't part of the typed OpenAPI envelope), so nothing to regenerate.
- **`flat` / `flat-square` output is byte-identical to before** (with spacing 0 + pad 12 the geometry is unchanged) — guarded by an explicit regression test.
- The per-segment width estimate stays a safe overestimate (glyph width + letter-spacing + padding), so the larger for-the-badge glyphs never clip.

## Tests (`tests/badge.test.mjs`)

- `for-the-badge` renders taller/uppercase/bold/square/matte with both text layers.
- A regression test asserting `flat`/`flat-square` output is unchanged by the refactor.
- The `style=for-the-badge` allow-list assertion.

```
✓ tests/badge.test.mjs (43 tests)
```

`eslint` and `prettier --check` clean on both files. Full local `vitest` run shows no new failures (only the pre-existing Windows-only env files, which pass on Linux CI).

### DIFF
--- a/src/badge.mjs
+++ b/src/badge.mjs
@@ -14,6 +14,8 @@
 //                      (subnet-api / openapi / sse / data-artifact); informational
 //                      blue, gray for 0; for a provider, the sum across its subnets
 //   style=flat-square  square corners, no gradient (default: flat)
+//   style=for-the-badge  the shields.io all-caps style: taller, bold, letter-
+//                      spaced, uppercase, square + matte
 //   label=…            override the left "metagraphed" segment text
 //
 // Worker-computed image/svg+xml, read-only, edge-cached, CORS-open. Unknown
@@ -37,7 +39,7 @@ const BADGE_METRICS = {
   apis: "apis",
 };
 // Allow-listed render styles; an unknown value falls back to "flat".
-const BADGE_STYLES = new Set(["flat", "flat-square"]);
+const BADGE_STYLES = new Set(["flat", "flat-square", "for-the-badge"]);
 // shields.io "informational" blue, used for plain-count metrics (e.g. apis).
 const INFO_COLOR = "#007ec6";
 // Surface kinds that are callable machine interfaces (mirrors the build's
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
Closing, see review comment: 

```
The main gap is that the claimed byte-identical regression coverage for existing styles is weaker than the PR description states,
```

PR #2359, opened 2026-06-29
fix(api): signal uptime row-cap truncation for 1y window

### LINKED ISSUE #2355: bug: uptime 1y window silently truncates under MAX_UPTIME_ROWS without capped signal
## Summary

`GET /api/v1/subnets/{netuid}/uptime?window=1y` reads `surface_uptime_daily` with `ORDER BY day DESC LIMIT 10000` (`MAX_UPTIME_ROWS`). Subnets with many operational surfaces can exceed 10k surface×day rows in a 1y window. The SQL drops the oldest rows with no `capped` flag and `formatUptime` / `computeReliability` report a full `1y` window from incomplete data.

## Expected

Mirror concentration history: when the read hits the row cap, drop the oldest (possibly partial) day and expose `capped: true` so consumers know the series is truncated.

## Repro

1. A subnet with >10k `(surface_key, day)` pairs in the last 365 days (many surfaces × ~365 days).
2. `GET /api/v1/subnets/{netuid}/uptime?window=1y`
3. Response lacks any truncation signal; reliability scores are computed from a partial history.

## Code pointers

- `workers/request-handlers/analytics-routes.mjs` — `handleUptime`
- `src/health-serving.mjs` — `formatUptime`
- `workers/config.mjs` — `MAX_UPTIME_ROWS = 10000`
- Prior art: `buildConcentrationHistory(..., { capped })` in `src/concentration.mjs`

philluiz2323:
## Summary

When `handleUptime` proves the D1 uptime query was row-cap truncated, the response now sets `capped: true` and drops the oldest (possibly partial) day before computing reliability — mirroring concentration history without falsely capping exact-size result sets.

Fixes #2355

## What Changed

- `handleUptime` fetches one sentinel row (`MAX_UPTIME_ROWS + 1`) so it can distinguish exact-size results from truncated results
- `handleUptime` trims the sentinel before calling `formatUptime` and only passes `capped: true` when truncation is proven
- `formatUptime` drops the oldest visible day when capped and exposes `capped: true`
- Regression tests cover both `MAX_UPTIME_ROWS` (not capped) and `MAX_UPTIME_ROWS + 1` (capped) boundary behavior

## Registry Safety

## Validation

- [x] `npm run build`
- [x] `npm run lint`
- [x] `git diff --check`
- [x] `npx vitest run tests/health-serving.test.mjs tests/request-handlers-analytics-routes.test.mjs`

### DIFF
--- a/generated/metagraphed-api.d.ts
+++ b/generated/metagraphed-api.d.ts
@@ -5072,6 +5072,8 @@ export interface components {
             [key: string]: unknown;
         });
         UptimeArtifact: {
+            /** @description Present and true only when the underlying uptime read hit MAX_UPTIME_ROWS and the oldest (possibly partial) day was dropped before aggregation. Absent when the result was not truncated. */
+            capped?: boolean;
             netuid: number;
             reliability?: components["schemas"]["ReliabilityScore"] | null;
             schema_version: number;
@@ -17904,6 +17906,7 @@ export interface operations {
                     /**
                      * @example {
                      *       "data": {
+                     *         "capped": false,
                      *         "netuid": 7,
                      *         "reliability": {
                      *           "avg_latency_ms": 120,
--- a/public/metagraph/openapi.json
+++ b/public/metagraph/openapi.json
@@ -14022,6 +14022,10 @@
       "UptimeArtifact": {
         "additionalProperties": true,
         "properties": {
+          "capped": {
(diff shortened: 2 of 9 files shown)


JSONbored (commented):
Please see the review for things that need to be fixed: 

```
The only notable issue is a contract/example mismatch: the documented example now shows `capped: false` even though the schema description and implementation intentionally omit the field unless truncation is proven.
```

PR #2373, opened 2026-06-29
feat(mcp): exclusion filters (not_) on list_subnets categorical args

jason020818:
## Summary

`list_subnets` supports categorical inclusion (`status`, `subnet_type`, `domain`) but no way to **exclude** a value. This adds `not_status`, `not_subnet_type`, and `not_domain` — the complement of the matching inclusion arg — extending the same parity work as the merged range filters (#2280).

Inclusion and exclusion now share one case-insensitive `subnetCategoricalMatch` helper, so `status=active` and `not_status=active` are exact complements; a row missing the field survives an exclusion but fails an inclusion. Composes with the existing `min_/max_` range filters, sort, and pagination.

## Changes
- `src/mcp-server.mjs`: `subnetCategoricalMatch` + `categoricalFilterSubnets` (inclusion + `not_` exclusion); the 3 `not_` args added to the `list_subnets` inputSchema. No server-card regen (worker-computed), no contract/openapi change.
- `tests/mcp-server.test.mjs`: exclusion per field, case-insensitivity, and inclusion+exclusion composition.

## Validation
- `npm run lint`, `npm run format:check`, `node scripts/validate-mcp.mjs` (52 tools) — green
- `npx vitest run tests/mcp-server.test.mjs` — 260 passed; new lines fully covered

### DIFF
--- a/src/mcp-server.mjs
+++ b/src/mcp-server.mjs
@@ -878,6 +878,54 @@ function rangeFilterSubnets(rows, args) {
   );
 }
 
+// Categorical args list_subnets filters on, each available as inclusion (`arg`)
+// and exclusion (`not_arg`).
+const LIST_SUBNETS_CATEGORICAL = ["status", "subnet_type", "domain"];
+
+// Does `subnet` match categorical filter `field` = `value` (already lowercased)?
+// `domain` tests the union of curated + derived categories; the rest are scalar.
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
Rejected, see review comments for blockers.

PR #2374, opened 2026-06-30
feat(api): add block_start/block_end range filter to the account transfers feed

jeffrey701:
## Summary

`GET /api/v1/accounts/{ss58}/transfers` accepts `direction`/`limit`/`offset`/`cursor` but has no block-height range filter, unlike the sibling `GET /api/v1/accounts/{ss58}/events` and the extrinsics / chain-events feeds (which gained `block_start`/`block_end` in #2328). This adds the same optional range to the transfers feed for parity.

## Implementation

- Adds `block_start` / `block_end` to the accepted params, parsed with the existing `parseNonNegativeIntParam` (rejects non-integer / negative with a 400, mirroring the events feed).
- Appends the range predicate to **each** side's WHERE **before** the keyset cursor clause, so every UNION arm binds in `[key, block_start?, block_end?, cursor?]` order (verified for sent / received / both).
- Both `idx_account_events_hotkey` and `idx_account_events_coldkey` lead `block_number` after the key column, so a bounded range stays index-satisfiable on this ~60s-cached public route.

## Validation

- `npx vitest run` — **137 files / 3390 tests pass** (added 4 tests: both-arm range bind order, single bound + direction=sent, block_end + direction=received, and a non-integer `block_start` rejection)
- `npm run validate` — clean · `eslint` + `prettier --check` — clean

### DIFF
--- a/tests/account-routes.test.mjs
+++ b/tests/account-routes.test.mjs
@@ -295,6 +295,101 @@ test("GET /accounts/{ss58}/transfers rejects an unsupported direction enum value
   assert.equal(body.meta.parameter, "direction");
 });
 
+// A capturing D1 mock that records every (sql, params), so a filter test can
+// assert the predicate + bind order — the naive dbWith stub ignores the WHERE
+// clause, so it cannot prove a range filter is wired correctly.
+function capturingDb(events = []) {
+  const calls = [];
+  return {
+    calls,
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
Rejected, see review comments for blockers.

Issue #1538, opened 2026-06-23
health-probe & build SSRF guards still treat fec0::/10 site-local IPv6 as public (only the webhook guard was fixed in #1443)


joaovictor712:
The webhook SSRF guard was fixed in #1443 to block the full `fe00::/8` reserved range (link-local `fe80::/10` + deprecated site-local `fec0::/10`). That same fix was never applied to the other three SSRF guards, which still only block `fe80::/10` and therefore classify `fec0::/10` site-local addresses as **public/safe**.

### Affected guards

**1. `src/health-probe-core.mjs` — `UNSAFE_HOST_PATTERNS`**
```js
/^fe80:/i,   // matches fe80::/10 only
```
`isUnsafePublicUrl("https://[fec0::1]/x")` returns `false` (treated as safe) — should be `true` (unsafe).

**2. `src/health-prober.mjs` — `isUnsafeIpAddress`**
```js
/^fe[89ab][0-9a-f]:/i.test(host)   // fe8..feb = fe80::/10 only
```
A literal or DNS answer of `fec0::1` passes the guard, so a rebinding/redirect to a site-local target gets probed.

**3. `scripts/lib.mjs` — `unsafeIpBlocks`**
```js
unsafeIpBlocks.addSubnet("fe80::", 10, "ipv6");   // fec0::/10 missing
```
The Node/build DNS-aware guard has the same gap.

### Why it matters
`fec0::/10` is deprecated site-local (RFC 3879) but still a real internal range. Treating it as public is an SSRF gap. The webhook guard already documents this exact reasoning — see `tests/webhooks-core.test.mjs`: *"fec0::/10 is deprecated site-local — a real internal range"*.

### Repro
- `isUnsafePublicUrl("https://[fec0::1]/x")` → `false` (wrong; expected `true`)
- health-prober `isUnsafeIpAddress("fec0::1")` → `false` (wrong; expected `true`)

### Expected
All three guards classify `fec0::/10` (`fec0::`–`feff:…`) as unsafe, consistent with the webhook guard and with `fe80::/10`.

Issue #2312, opened 2026-06-29
bug: health-probe SSRF guard allows CGNAT 100.64.0.0/10 (webhook guard already blocks it)


joaovictor712:
The health prober's literal SSRF guard (`isUnsafePublicUrl` in `src/health-probe-core.mjs`) does not block the CGNAT `100.64.0.0/10` range, even though the sibling webhook and build SSRF guards already do.

### Evidence

`src/webhooks.mjs` blocks CGNAT:

```js
/^100\.(6[4-9]|[7-9]\d|1[01]\d|12[0-7])\./, // 100.64.0.0/10 CGNAT
```

`scripts/lib.mjs` also marks the range unsafe:

```js
unsafeIpBlocks.addSubnet("100.64.0.0", 10);
```

But `src/health-probe-core.mjs` `UNSAFE_HOST_PATTERNS` only covers `10.*`, `172.16-31.*`, `192.168.*`, `169.254.*`, etc. - no `100.64.0.0/10`.

### Reproduction

```js
isUnsafePublicUrl("http://100.64.0.1/x") // returns false (treated as safe / probeable)
```

An operational surface URL pointing at a CGNAT address would be probed from the Worker, while the same host would be rejected by the webhook delivery guard.

### Expected

`isUnsafePublicUrl("http://100.64.0.1/x")` (and the full `100.64.0.0/10` range) should return `true`, matching the webhook guard and the recent SSRF parity fixes (#1538 / #1541 / #1564).

PR #1541, opened 2026-06-23
fix(health): block fec0::/10 site-local IPv6 in probe SSRF guards

philluiz2323:
## Summary

Closes #1538.

The webhook SSRF guard was fixed in #1443 to block the full `fe00::/8` reserved range (link-local `fe80::/10` + deprecated site-local `fec0::/10`). The same fix was never applied to the other three SSRF guards, which still only blocked `fe80::/10` and therefore classified `fec0::/10` site-local addresses as **public/safe** — an SSRF gap (`fec0::/10` is deprecated per RFC 3879 but still a real internal range).

## What Changed

- `src/health-probe-core.mjs` — `UNSAFE_HOST_PATTERNS`: `/^fe80:/i` → `/^fe[89a-f][0-9a-f]:/i` (covers `fe80::`–`feff:`). The pattern still requires an IPv6 segment colon, so hostnames like `fedora.org` are not affected.
- `src/health-prober.mjs` — `isUnsafeIpAddress`: `/^fe[89ab][0-9a-f]:/i` → `/^fe[89a-f][0-9a-f]:/i`, so a literal/DNS-answer `fec0::1` is now blocked.
- `scripts/lib.mjs` — `unsafeIpBlocks`: add `fec0::/10` alongside the existing `fe80::/10`.
- Tests added for each guard (`health-probe-core`, `health-prober` DNS-aware, `script-utils`).

All three guards now cover the same `fe80::`–`feff:` reserved range as the webhook guard. No schema, OpenAPI, or generated-artifact changes.

## Registry Safety

- [x] No secrets, PATs, wallet data, private dashboards, private URLs, or validator-local state.
- [x] Generated artifacts were produced by repo scripts, not hand-edited.
- [x] R2-only/high-churn detail artifacts are not committed.
- [x] Public API/OpenAPI/schema changes are intentional and documented. (None in this PR.)

## Validation

- [x] Directly verified all three guards: `isUnsafePublicUrl([fec0::1])`, the prober DNS-aware guard with an `fec0::1` AAAA answer, and `isUnsafeUrl([fec0::1])` all now return unsafe; `fe80::/10`, public hosts, and `fedora.org`-style hostnames are unaffected.
- [x] `git diff --check`
- Full `npm run worker:test` / `test:coverage` suite runs in CI.

## Template Used

- Backend/code change

PR #2377, opened 2026-06-30
fix(health): scope IPv6 unique-local SSRF patterns to literal hosts

### LINKED ISSUE #2375: bug: health-probe SSRF guard misclassifies public fd* domains as unique-local IPv6 (and misses fc00::/7 ULAs)
The health prober's literal SSRF guard (`isUnsafePublicUrl` in `src/health-probe-core.mjs`) applies the IPv6 unique-local / link-local patterns to the bracket-stripped hostname **without first checking that the host is an IPv6 literal**. As a result it both over-blocks legitimate public domains and under-blocks part of the real unique-local range.

### Evidence

`UNSAFE_HOST_PATTERNS` mixes IPv4/domain patterns with IPv6-literal patterns, and the IPv6 ones are matched against every host:

```js
const UNSAFE_HOST_PATTERNS = [
  /^localhost$/i,
  // ...
  /^fe[89a-f][0-9a-f]:/i,
  /^fc00:/i,
  /^fd/i,
];
// ...
return UNSAFE_HOST_PATTERNS.some((pattern) => pattern.test(host));
```

The sibling webhook guard in `src/webhooks.mjs` gates the exact same `fc`/`fd`/`fe` prefixes behind `host.includes(":")`, so it never misreads a registrable domain as an IPv6 literal:

```js
if (host.includes(":")) {
  if (
    host === "::1" ||
    host === "::" ||
    host.startsWith("fe") ||
    host.startsWith("fc") ||
    host.startsWith("fd") ||
    host.startsWith("::ffff:")
  ) {
    return false;
  }
```

`isUnsafePublicUrl` has no such gate, so `/^fd/i` matches any hostname beginning with `fd`.

### Reproduction

False positive - public, probeable domains that begin with `fd` are wrongly rejected as unsafe:

```js
isUnsafePublicUrl("https://fda.gov/");      // returns true  (should be false)
isUnsafePublicUrl("https://fd.io/");        // returns true  (should be false)
isUnsafePublicUrl("https://fdroid.org/");   // returns true  (should be false)
```

An operational surface hosted on such a domain is short-circuited as `"unsafe URL"` by the prober and never probed, even though it is a perfectly public address.

False negative - the unique-local range is `fc00::/7` (first hextet `fc00`-`fdff`), but `/^fc00:/i` only matches the literal `fc00:` hextet, so other `fc__` ULAs slip through:

```js
isUnsafePublicUrl("http://[fc12::1]/");     // returns false (should be true)
```

### Expected

- IPv6 unique-local / link-local patterns must only apply to IPv6 literals (a host containing `:`), so public domains like `fda.gov` / `fd.io` / `fdroid.org` return `false`.
- Unique-local detection should cover the full `fc00::/7` range (`fc00:`-`fdff:`), so `fc12::1` returns `true`.
- IPv4 literals, IPv4-embedded IPv6 forms, loopback, and the existing private/link-local IPv6 literals must keep returning `true` (no regression on the prior SSRF parity fixes #1538 / #1541 / #2313).

joaovictor91123:
## Summary

Fix a false-positive (and a paired false-negative) in the health prober's literal SSRF guard `isUnsafePublicUrl` (`src/health-probe-core.mjs`). The IPv6 unique-local / link-local patterns were matched against every host, so any public domain beginning with `fd` (`fda.gov`, `fd.io`, `fdroid.org`) was wrongly classified unsafe and silently never probed. At the same time `/^fc00:/i` only matched the single `fc00:` hextet, so other `fc00::/7` unique-local addresses (e.g. `fc12::1`) slipped through.

Fixes #2375

## What Changed

- Split the IPv6-literal patterns out of `UNSAFE_HOST_PATTERNS` into a new `UNSAFE_IPV6_LITERAL_PATTERNS` list and added a small `isUnsafeLiteralHost(host)` helper that only applies the IPv6 rules when the host is an IPv6 literal (contains `:`), mirroring the colon-gated webhook guard in `src/webhooks.mjs`.
- Widened unique-local detection from `/^fc00:/i` + `/^fd/i` to `/^f[cd][0-9a-f]{2}:/i`, covering the full `fc00::/7` range (`fc00:`-`fdff:`).
- Folded the `172.16.0.0/12` private-range check into `UNSAFE_HOST_PATTERNS` so the host and the IPv4-embedded-IPv6 path share one classifier.
- Extended the `isUnsafePublicUrl` regression suite in `tests/health-probe-core.test.mjs`: public `fc`/`fd` domains stay allowed, and the full `fc00::/7` literal range stays blocked. No regression on the prior SSRF parity fixes (#1538 / #1541 / #2312 / #1533).

## Scope and Risk

This is limited to local URL classification before probes are attempted. It does not change probe scheduling, persisted artifacts, schemas, OpenAPI output, or public API response shapes. The main risk is SSRF classifier drift, covered by regression tests for public `fc`/`fd` hostnames, IPv6 ULA literals, and existing private/tunnelled IP cases.
## Registry Safety

- [x] No public API/OpenAPI/schema changes (pure runtime guard fix; contract untouched).

## Validation

- [x] `npx vitest run tests/health-probe-core.test.mjs` (107 passed)
- [x] `npx vitest run tests/build-readiness.test.mjs tests/health-ok-latency.test.mjs` (the other importers of the guard, 79 passed)
- [x] `npm run lint`
- [x] `git diff --check`

### DIFF
--- a/src/health-probe-core.mjs
+++ b/src/health-probe-core.mjs
@@ -51,6 +51,8 @@ export const OPERATIONAL_SURFACE_KINDS = [
 // build injects the stronger DNS-aware `isUnsafeResolvedUrl` from scripts/lib.mjs.
 // Operational surfaces are already curated `public_safe`, so this is defense in
 // depth, not the primary control.
+// IPv4 / registrable-domain unsafe prefixes. Their leading characters can never
+// collide with a public hostname, so they are matched against EVERY host.
 const UNSAFE_HOST_PATTERNS = [
   /^localhost$/i,
   /\.localhost$/i,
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
Merge conflicts, closing: 

```
Use the [web editor](https://github.com/JSONbored/metagraphed/pull/2377/conflicts) or the command line to resolve conflicts before continuing.

src/health-probe-core.mjs
```

PR #2383, opened 2026-06-30
fix(mcp): bound chain signers D1 fanout

JSONbored:
### Motivation
- Prevent unauthenticated MCP `get_chain_signers` calls from triggering expensive, uncapped D1 GROUP BY scans that can be amplified by JSON-RPC batching and cause availability / cost abuse. 
- Match the REST route's protections (timeout + amortization) so MCP tooling does not monopolize D1/Worker time or bypass existing data-tier limits.

### Description
- Wrap MCP D1 reads in the existing timeout guard by importing and using `withTimeout` + `d1TimeoutMs` inside `mcpD1Runner` so slow queries degrade to schema-stable empty results. 
- Add a `requireDataTierRateLimit` check that charges the `DATA_RATE_LIMITER` before running expensive aggregations. 
- Add per-request promise coalescing via `loadMcpChainSigners` (in-memory per-request Map keyed by query args) so identical batched `get_chain_signers` calls share a single D1 aggregation instead of fanning out. 
- Route the `get_chain_signers` tool handler through the new `loadMcpChainSigners` path and add unit tests covering limiter short-circuiting, batched-call coalescing, and D1 timeout fallback.

### Testing
- Ran `npm run build` and the build completed successfully. 
- Ran MCP validation with `node scripts/validate-mcp.mjs` and `npm run validate:mcp`, which passed. 
- Ran unit tests for the MCP surface with `npm test -- --run tests/mcp-server.test.mjs` and the tests (including the new signers tests) passed. 
- Ran lint/format checks with `npm run lint` and `npm run format:check`, both of which passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a433437a510832c965a61a119088770)

### DIFF
--- a/src/mcp-server.mjs
+++ b/src/mcp-server.mjs
@@ -23,6 +23,7 @@ import {
   clampOffset,
 } from "../workers/request-params.mjs";
 import { EXPOSED_RESPONSE_HEADERS_VALUE } from "../workers/http.mjs";
+import { d1TimeoutMs, withTimeout } from "../workers/storage.mjs";
 import { CONTRACT_VERSION, PRIMARY_DOMAIN } from "./contracts.mjs";
 import {
   loadSubnetConcentration,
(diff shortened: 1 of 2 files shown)


JSONbored:
Rebased onto main and resolved the conflict with the newer \`loadChainEventsFeed\` addition. Also applied the requested fix: moved \`requireDataTierRateLimit\` inside the cache-miss promise in \`loadMcpChainSigners\` so a batch of identical \`get_chain_signers\` calls shares one limiter charge alongside the one D1 aggregation, instead of paying the limiter once per duplicate call in the batch. Added a regression test asserting a 3-call identical batch against a failing limiter only invokes the limiter once.

While rebasing I also caught a real integration bug between this branch and main's newer sort support (#2475): \`loadMcpChainSigners\`/\`chainSignersCacheKey\` didn't forward or key on \`options.sort\`, so any \`sort=total_fee_tao\` request would silently ignore the sort and could share a cache entry with a differently-sorted request. Fixed both call sites; the existing "ranks signers by total_fee_tao" test now passes again.

JSONbored:
Pushed a follow-up: found and fixed the last uncovered branch codecov/patch was flagging. \`requireDataTierRateLimit\`'s \`if (!success)\` had a test for the reject path (\`success: false\`) but nothing exercising the allow path (\`success: true\`, limiter present and lets the request through to the D1 aggregation) — added that test.

PR #2432, opened 2026-06-30
feat(api): add subnet stake-flow endpoint summing staked vs unstaked tao

### LINKED ISSUE #2422: feat(api): GET /api/v1/subnets/{netuid}/stake-flow, net staked vs unstaked over a window
## Problem / why it matters

Stake is the capital signal in Bittensor, but the registry has no way to ask whether a subnet is gaining or losing stake over time. We serve the raw per-event feed (`/api/v1/subnets/{netuid}/events?kind=StakeAdded`) and validator churn (`/api/v1/subnets/{netuid}/turnover`), but a consumer who wants "is capital flowing into this subnet or draining out of it, and by how much?" has to page every StakeAdded and StakeRemoved row and sum them client-side. Turnover answers who entered or left the validator set; it does not answer how much TAO actually moved. Every capital tracker surfaces a net-flow view, and we already store every input.

## Current state (verified)

- The `account_events` D1 tier stores each stake event with an `amount_tao REAL` column (tao field / 1e9), per `migrations/0009_account_events.sql:21`. StakeAdded and StakeRemoved both carry it (`src/account-events.mjs`, EVENT_INSERT_COLUMNS plus the ingested-kinds set). `observed_at` is the block timestamp in epoch ms, and the tier keeps a rolling ~90-day hot window.
- `migrations/0024_account_events_kind_index.sql` indexes `account_events (netuid, event_kind, block_number)`, which is exactly the seek a per-subnet, per-kind aggregate needs.
- `handleSubnetEvents` (`workers/request-handlers/entities.mjs:859`) is the established live-D1, null-safe per-subnet account_events reader (cold store returns 200 empty, never 404); `handleSubnetTurnover` (`entities.mjs:417`) plus the pure `src/turnover.mjs` module are the established compute-and-scorecard pattern.
- No stake-flow, inflow/outflow, or net-stake endpoint exists anywhere (`src/contracts.mjs`, `workers/api.mjs`).

## Proposed addition

`GET /api/v1/subnets/{netuid}/stake-flow?window=7d|30d|90d` (default 30d), computed live from `account_events`, returning a stake-flow scorecard over the window:

- `total_staked_tao` (sum of StakeAdded amount_tao)
- `total_unstaked_tao` (sum of StakeRemoved amount_tao)
- `net_flow_tao` (staked minus unstaked, positive means net inflow)
- `stake_events` / `unstake_events` (counts)
- `window` and `netuid` context

All TAO sums rounded to rao precision (matches the rounding in `src/turnover.mjs`). The windows are 7d/30d/90d because `account_events` keeps a rolling ~90-day hot window, which is exactly the recent-capital-flow signal a flow view answers. Schema-stable: a cold or empty store returns zeroed totals with a 200, never a 404, mirroring the sibling routes.

## Implementation sketch

- New pure module `src/stake-flow.mjs`: `buildStakeFlow(rows, netuid, opts)` (the sums, net, and counts, fully unit-testable) plus `loadSubnetStakeFlow(d1, netuid, opts)`.
- `handleSubnetStakeFlow` in `workers/request-handlers/entities.mjs`, sibling to `handleSubnetTurnover`, one indexed `GROUP BY event_kind` read over `account_events` (`observed_at >= now - windowDays`) using `idx_account_events_netuid_kind`.
- A `SUBNET_STAKE_FLOW_PATH_PATTERN` in `workers/config.mjs` (next to `SUBNET_TURNOVER_PATH_PATTERN`) plus edge-cached dispatch in `workers/api.mjs`.
- Register the live route in `src/contracts.mjs` (artifact plus route), mark it computed (`R2_ONLY_PATTERNS` in `src/artifact-storage.mjs`, `COMPUTED_ARTIFACTS` in `scripts/validate-schemas.mjs`), add a `SubnetStakeFlowArtifact` schema, and regenerate `openapi.json` and types via `npm run build`.

No new data collection, no new D1 tier, no static artifact. Pure live aggregation over stake events we already ingest, on an index we already ship.

GildardoDev:
Adds GET /api/v1/subnets/{netuid}/stake-flow, which sums StakeAdded versus StakeRemoved from the account_events tier into net capital flow per subnet over a 7d, 30d, or 90d window, computed live on the existing netuid and event_kind index with no new data collection. Closes #2422.


GildardoDev:
Quick note for review: the current blocker points at `workers/request-handlers/entities.mjs:489` passing null into `metagraphMeta`, but on the latest commit that line is `netuid,` and `handleSubnetStakeFlow` uses `accountMeta(env, path, generatedAt)`.

Both issues the review raised are addressed now: the response meta uses `accountMeta` for account_events provenance, and `meta.generated_at` is rendered as an ISO string (`string|null`) per the envelope schema, matching the `toIso` pattern in `src/account-events.mjs`. CI is green and patch coverage is ~100%.

It looks like the blocker came from a mid-CI snapshot of the earlier code, so a re-run or override should clear it. Happy to adjust if I am missing something. Thanks for taking a look.

JSONbored (approved):

PR #2445, opened 2026-06-30
fix(mcp): tie-break and keyset-cursor get_account_history for deterministic paging

joaovictor712:
## Summary

- The MCP `get_account_history` tool reads `account_events_daily`, which holds one row per `(hotkey, day, netuid)`. The loader (`loadAccountHistory`) ordered by `day DESC` only - `day` alone is not a total order, so same-day rows (an account active on several subnets) came back in an unspecified order.
- With offset paging over that non-total order, a page boundary that lands inside a day could silently drop or repeat a same-day row across pages. The REST endpoint (`handleAccountHistory`) was already given a `(day, netuid)` tie-break + keyset cursor; the MCP loader is the uncorrected twin, and its own in-file siblings `loadAccountExtrinsics` / `loadAccountTransfers` already use the keyset pattern.
- Tie-break ordering on `netuid DESC` so the order is total (deterministic same-day paging), and add the same `(day, netuid)` keyset cursor the REST handler and sibling loaders expose, so callers can page stable head-growing windows via `next_cursor`.

## What Changed

- `src/account-events.mjs` - `loadAccountHistory` now orders by `day DESC, netuid DESC`, accepts a `cursor`, applies a `(day, netuid) < (?, ?)` keyset predicate (cursor takes precedence over offset; a cursor that does not decode to a valid `YYYY-MM-DD` day is ignored ? first page, never throws), and computes `next_cursor` from the last row of a full page. Mirrors `handleAccountHistory` (`workers/request-handlers/entities.mjs`) and the in-file tail loaders. `DAY_PATTERN` imported from `workers/request-params.mjs`.
- `src/mcp-server.mjs` - the `get_account_history` tool exposes an optional opaque `cursor` (precedence over offset) and threads it through; description updated.
- `tests/account-events.test.mjs` - updated the SQL-shape assertion to the `day DESC, netuid DESC` total order; added regression coverage for the keyset cursor path (predicate present, offset dropped, `day` re-hyphenated, `next_cursor` emitted on a full page) and for a malformed cursor (ignored ? first page, no `next_cursor` on a short page).
- No schema/contract change: `next_cursor` is already part of the account-history payload shape (`buildAccountHistory`); the MCP server card is worker-computed (no committed artifact).

## Registry Safety

- [x] Public API/OpenAPI/schema changes are intentional and documented. (none - behavior + MCP tool input)

## Validation

- [x] `npx prettier --check` + `npx eslint` on the changed files
- [x] `npx vitest run tests/account-events.test.mjs tests/account-events-branch-coverage.test.mjs` (73 passed)
- [x] `npx vitest run tests/mcp-server.test.mjs` (278 passed)
- [x] `npm run validate:mcp` (57 tools, lifecycle + all tools/call)
- [x] `npm run build` then `npm run validate` (green; no contract/artifact drift)
- [x] `git diff --check`

## Linked issue

No linked issue: issue creation on this repository is currently restricted for outside contributors (the issues API returns 404), so an issue could not be filed to track this. Per the contribution guide a linked issue is optional and a PR is judged on its own merit. This is a self-contained, test-covered correctness fix that mirrors an already-merged sibling implementation in the same area.

### DIFF
--- a/src/account-events.mjs
+++ b/src/account-events.mjs
@@ -4,6 +4,7 @@
 // the daily rollup, the prune, and the row→API shaping (#1347). Pure + exported
 // for tests; the Worker runs the D1 I/O.
 import {
+  DAY_PATTERN,
   FEED_PAGINATION,
   clampLimit,
   clampOffset,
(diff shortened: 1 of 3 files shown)


JSONbored (commented):
Please address the issue mentioned in the review: 

```
The only notable gap is that cursor validation accepts any second tuple value once the day parses, so a user-supplied malformed opaque cursor can still drive an invalid or nonsensical keyset comparison.
```

PR #2463, opened 2026-06-30
fix(api): tie-break chain-calls leaderboard for deterministic LIMIT output

luciferlive112116:
## What

`loadChainCalls` (powering `GET /api/v1/chain/calls` and the `get_chain_calls` MCP tool) ranks grouped extrinsic counts with `ORDER BY count DESC` and a trailing `LIMIT`. That sort key is **not a total order**: when multiple `call_module` rows (or `call_module`/`call_function` pairs) share the same count, D1/SQLite group order is unspecified, so:

- the returned calls array could reorder between identical requests, and
- at the `LIMIT` boundary, **membership** of the truncated tail could change — not just ordering.

Shares stay honest because the denominator is a separate full-window count, but the truncated list itself was nondeterministic.

## How

Add stable secondary sort keys on the respective `GROUP BY` columns:

- module grouping: `ORDER BY count DESC, call_module ASC`
- module_function grouping: `ORDER BY count DESC, call_module ASC, call_function ASC`

This is a pure ordering change — aggregate values and share math are unchanged; only tie resolution and `LIMIT` membership become well-defined.

## Scope / risk

Two dynamic `ORDER BY` branches in a single loader plus one focused test. No schema, contract, or public field changes.

## Tests

- `tests/analytics-live.test.mjs`: adds `loadChainCalls tie-breaks grouped rows for stable LIMIT membership`, asserting both group-by modes carry their stable secondary sort keys ahead of `LIMIT ?`. Existing chain-calls route and MCP tests continue to match — they assert `ORDER BY count DESC`, which remains a prefix of the new clauses.

I don't have a local Node/npm toolchain; CI runs the full vitest suite.

### DIFF
--- a/src/analytics-live.mjs
+++ b/src/analytics-live.mjs
@@ -482,7 +482,8 @@ export async function loadCompareSubnets(
 
 // Extrinsic call-mix breakdown (#1989): counts + share per call_module (or
 // call_module/call_function). The share denominator is the full-window extrinsic
-// count read separately, so the truncated LIMIT tail never skews shares. Mirrors
+// count read separately, so the truncated LIMIT tail never skews shares. Tie-break
+// on the GROUP BY keys so tied counts keep a stable LIMIT membership. Mirrors
 // REST's handleChainCalls and the get_chain_calls MCP tool (#2364).
 export async function loadChainCalls(
   d1,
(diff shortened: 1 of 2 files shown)


luciferlive112116:
No linked issue: this is a standalone correctness fix discovered by code inspection — the same nondeterministic `ORDER BY count DESC` + `LIMIT` pattern that was fixed elsewhere in account-summary and health leaderboards. No open issue or PR covers chain-calls tie-breaking.

JSONbored (approved):

PR #2487, opened 2026-06-30
fix(account-events): coerce string-typed registration cells in formatRegistration

luciferlive112116:
## What

`formatRegistration` shapes neurons-table rows into `AccountRegistration` objects for:

- `GET /api/v1/accounts/{ss58}` (`registrations` on the summary), and
- `GET /api/v1/accounts/{ss58}/subnets` / `get_account_subnets`.

It previously passed `netuid`, `uid`, and `stake_tao` through unchanged. D1 INTEGER/REAL cells often arrive as numeric strings; the public schema requires `netuid: integer`, `uid: integer|null`, and `stake_tao: number|null`. A string-typed cell therefore violated the contract and broke strict clients/comparisons even when the underlying value was correct.

This is the registration-formatter sibling of the same D1 coercion hardening already applied to block heights (`formatBlock`), extrinsic indices (`formatExtrinsic`), and account-event PK fields (`toBlockNumber` in this module).

## How

- `netuid` / `uid` → existing `toBlockNumber` (non-negative integer or null)
- `stake_tao` → existing `toTaoOrNull` (finite number rounded to rao precision, or null)

No schema, artifact, or route-surface changes — payload normalization only.

## Why no linked issue

Standalone formatter gap found by code inspection against the OpenAPI `AccountRegistration` contract. No open issue tracks this specific registration path; open PR #2481 covers the adjacent `formatAccountEvent` formatter only, not registrations.

## Scope / risk

One exported formatter + three behavior-level tests. Touches only the neurons-row shaping path; boolean flag coercion is unchanged.

## Tests

Behavior-level (not SQL-string assertions):

- `formatRegistration coerces D1 numeric-string cells to schema types` — asserts `typeof` is `number` for coerced fields and values match parsed numbers.
- `formatRegistration drops invalid netuid and uid cells instead of leaking strings` — garbage/negative strings become `null`.
- `buildAccountSummary and buildAccountSubnets keep coerced registration types` — end-to-end through both public builders that consume `formatRegistration`.

Existing `formatRegistration` null-safe and flag-coercion tests continue to pass unchanged.

## Test plan

- [x] Unit: string `"7"` / `"3"` / `"100.5"` cells coerce to numbers
- [x] Unit: invalid netuid/uid strings become null
- [x] Integration: summary + subnets builders preserve coerced types
- [x] CI vitest suite (no local Node toolchain here; relies on repo CI)

### DIFF
--- a/src/account-events.mjs
+++ b/src/account-events.mjs
@@ -236,6 +236,12 @@ export function eventInsertStatements(db, rows) {
 export const ACCOUNT_EVENT_COLUMNS =
   "block_number, event_index, event_kind, hotkey, coldkey, netuid, uid, amount_tao, alpha_amount, observed_at, extrinsic_index";
 
+// Coerce a D1 0/1 INTEGER flag cell to a boolean. Numeric strings like "0"
+// must not pass through Boolean(), which treats any non-empty string as true.
+function toD1Flag(value) {
+  return Number(value) === 1;
+}
+
 // One neurons-table row (subset) → an AccountRegistration: where this hotkey is
 // currently registered + staked (the live cross-subnet footprint).
 export function formatRegistration(row) {
@@ -244,8 +250,8 @@ export function formatRegistration(row) {
     netuid: toBlockNumber(row.netuid),
     uid: toBlockNumber(row.uid),
     stake_tao: toTaoOrNull(row.stake_tao),
-    validator_permit: Boolean(row.validator_permit),
-    active: Boolean(row.active),
+    validator_permit: toD1Flag(row.validator_permit),
+    active: toD1Flag(row.active),
   };
 }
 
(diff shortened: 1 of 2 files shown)


luciferlive112116:
**No linked issue:** standalone formatter gap — `formatRegistration` (neurons → `AccountRegistration`) was still passing D1 numeric strings through unchanged. Open PR #2481 covers the adjacent `formatAccountEvent` formatter only; this PR completes the registration path used by `/accounts/{ss58}` and `/accounts/{ss58}/subnets`.

**Differentiation from #2481:** different exported formatter, different public fields (`registrations` / `subnets` vs `recent_events`), plus D1 flag coercion (`toD1Flag` fixes `Boolean("0") === true`).

**Validation:** behavior-level tests assert `typeof` coercions, invalid-cell nulling, string `"0"`/`"1"` flags, and end-to-end `buildAccountSummary` + `buildAccountSubnets` output — not SQL-string assertions.

JSONbored (approved):

PR #2492, opened 2026-06-30
fix(api): bound chain fee median samples

JSONbored:
### Motivation
- The exact median computation scanned and ranked all matching extrinsics for the requested window, allowing unauthenticated REST and MCP callers to trigger expensive unbounded D1 sorts and degrade availability.
- The median query had no row bound tied to the public `limit` parameter and the MCP runner had no timeout, creating an amplification vector for repeated requests.

### Description
- Add a fixed request-time cap `CHAIN_FEE_MEDIAN_SAMPLE_LIMIT = 10000` to bound rows sampled for median computation.
- Introduce `medianParams` and pass them to the median query so the median CTE includes `LIMIT ?` and uses the capped parameter instead of the unbounded daily params.
- Preserve the existing daily aggregates and top-fee-payer `LIMIT` behavior while restricting the median window-function rankings to the bounded sample.
- Update unit tests in `tests/analytics-live.test.mjs` to assert the median query includes `LIMIT ?` and receives the sample-cap parameter for scoped, unscoped, and empty `call_module` cases.

### Testing
- Ran the focused unit test: `npm test -- tests/analytics-live.test.mjs` and all tests in that file passed (41 tests passed).
- Ran linters and format checks with `npm run lint` and `npm run format:check` and they passed after formatting test file updates.
- Built derived artifacts with `npm run build` and then ran `npm run validate:api` and `npm run validate:mcp`, both of which passed after the build step.
- Note: attempting `--runInBand` with `vitest` is not supported; tests were run without that flag and succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a4448c13e7c832cbd580bd507f13157)

### DIFF
--- a/src/analytics-live.mjs
+++ b/src/analytics-live.mjs
@@ -607,66 +607,7 @@ export async function loadChainFees(
   const payerParams = callModuleFilter
     ? [cutoff, callModuleFilter, limit]
     : [cutoff, limit];
-  // Median sampling: cap per UTC calendar day, one "WHERE observed_at in
-  // [dayStart, dayEnd) ... LIMIT ?" block per day (UNION ALL'd below) instead
-  // of a single query over the whole window — a day-partitioned ROW_NUMBER()
-  // OVER still has to sort the FULL matching set before any per-partition
-  // filter can trim it, reintroducing the unbounded-scan cost the cap exists
-  // to remove. Bounding by CALENDAR DAY (not by the request's `window` size)
-  // matters because a day's real extrinsic volume is a property of actual
-  // chain throughput, not something a caller can inflate by requesting a
-  // bigger window — unlike the pre-cap behavior, where the window parameter
-  // directly multiplied the sort cost.
-  //
-  // The sample is ORDER BY RANDOM() rather than ORDER BY observed_at: capping
-  // to the chronologically-EARLIEST rows would silently bias the median
-  // toward however fees/tips look at the start of a high-volume day (a real
-  // risk for a congestion-priced fee market, where fees plausibly trend with
-  // time of day) instead of the day's true distribution. A random subsample
-  // costs a per-day sort (RANDOM() can't use idx_extrinsics_observed to
-  // short-circuit), but that cost is bounded by one calendar day's real
-  // extrinsic count, not by the multi-day window the original vulnerability
-  // let a caller inflate — and it keeps the median statistically
-  // representative of the whole day instead of systematically skewed.
-  //
-  // The sample cap and (optional) call_module filter are the SAME value in
-  // every day block, so they're bound ONCE via numbered (?N) parameters and
-  // reused across every UNION ALL block, rather than re-bound per block. D1
-  // caps a statement at 100 bound parameters; a scoped 30-day window already
-  // needs a day-start + day-end pair per block, and re-binding the cap/filter
-  // too would push a 30-day scoped window to 120+ params.
-  const dayBoundaries = [];
-  for (let dayStart = utcDayFloor(cutoff); dayStart < now; dayStart += DAY_MS) {
-    dayBoundaries.push(dayStart);
-  }
-  const medianLimitParam = 1;
-  const medianModuleParam = callModuleFilter ? 2 : null;
-  const medianFirstDayParam = callModuleFilter ? 3 : 2;
-  const medianModuleClause = callModuleFilter
-    ? ` AND call_module = ?${medianModuleParam}`
-    : "";
-  const medianDayBlocks = dayBoundaries.map((dayStart, i) => {
-    const startParam = medianFirstDayParam + i * 2;
-    const endParam = startParam + 1;
-    return `SELECT * FROM (
-           SELECT strftime('%Y-%m-%d', observed_at / 1000, 'unixepoch') AS day,
-                  COALESCE(fee_tao, 0) AS fee_tao,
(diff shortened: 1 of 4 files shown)


JSONbored:
Pushed a follow-up fix: the sample cap was a single global \`LIMIT ?\` on the \`samples\` CTE ahead of the per-day \`PARTITION BY\`, so once a window's matching extrinsics exceeded 10000, days sorted after the cutoff lost their entire sample and reported a null median despite a positive \`extrinsic_count\`. Replaced it with a day-partitioned \`ROW_NUMBER() ... WHERE day_rn <= ?\` filter so every day gets its own bounded sample. Added a real-SQLite regression test (\`tests/analytics-live.test.mjs\`) that reproduces the >10000-rows-in-one-day scenario and asserts a sibling day's median survives.

JSONbored:
Good catch — the day-partitioned \`ROW_NUMBER()\` I pushed still required SQLite to sort the entire matching \`samples\` set before the per-partition \`WHERE day_rn <= ?\` could trim it, since a window function's PARTITION BY/ORDER BY has to materialize and sort the full input regardless of what filters the outer query later applies. That reintroduced the exact unbounded-scan cost this PR exists to remove.

Replaced it with one \`UNION ALL\` block per UTC calendar day, each independently bounded by \`WHERE observed_at >= ? AND observed_at < ? ... ORDER BY observed_at LIMIT ?\`. Verified via \`EXPLAIN QUERY PLAN\` that each block does \`SEARCH extrinsics USING INDEX idx_extrinsics_observed\` with no temp-b-tree sort — so total rows scanned for the median query is now bounded by \`days * CHAIN_FEE_MEDIAN_SAMPLE_LIMIT\` (at most 30 days per the ANALYTICS_WINDOWS cap) instead of the whole window's row count, while every day still gets its own fair, bounded sample (no more silently-dropped days).

JSONbored:
Good catch — binding a fresh copy of the sample cap + \`call_module\` filter into every UNION ALL day block pushed a scoped 30-day window to \`30 * 4 = 120\` bound parameters, over D1's 100-per-statement limit, so the median query would have been rejected before it ever ran against live D1.

Fixed by binding the cap and \`call_module\` filter ONCE via numbered \`?1\`/\`?2\` parameters, reused across every day block's text — confirmed via a real SQLite check that a numbered parameter's single bound value is correctly reused at every occurrence, not re-consumed per reference. Each day block now only contributes its own day-start/day-end pair, so a scoped 30-day window binds ~62 params total instead of 120. Added a regression test asserting the total stays under 100 for the worst-case scoped 30d window.

JSONbored:
Sharp catch — capping to \`ORDER BY observed_at LIMIT ?\` (the chronologically-earliest rows) silently biased the median toward whatever fees/tips looked like at the *start* of a high-volume day instead of its true distribution. For a congestion-priced fee market that's a real, plausible skew, not just a theoretical purity concern.

Switched the per-day sample to \`ORDER BY RANDOM()\`. This does give up the index-assisted early-termination for over-cap days (RANDOM() can't use \`idx_extrinsics_observed\` to stop early), but the resulting cost is bounded by one calendar day's real extrinsic count — a property of actual chain throughput — not by the request's \`window\` parameter, which is what made the original (pre-this-PR) behavior an actual DoS vector (a caller could directly multiply the sort cost by requesting a bigger window). A random sample keeps the returned \`median_fee_tao\`/\`median_tip_tao\` a statistically valid estimate of the day's true median instead of a systematically skewed one, with no API/contract change needed.

Added a regression test: a day split 5000 low-fee rows (first half chronologically) / 5100 high-fee rows (second half), 10100 total — 100 over the cap. The true day median sits solidly in the high-fee bucket (1000), but the old chronological-first-10000 sample would have kept every low-fee row and landed exactly on the 0/1000 boundary (500) — confirmed by temporarily reverting to chronological ordering and watching the test fail with exactly that value.

JSONbored:
You're right — \`ORDER BY RANDOM()\` still forces SQLite/D1 to scan and sort every matching row for a day before the cap can apply, since a window/order clause has to materialize the full input regardless of what filters the outer query later. Every sampling strategy tried on this PR so far (global LIMIT, day-partitioned ROW_NUMBER, random) trades exactness for some kind of bias anyway, so I dropped sampling entirely.

New approach: the daily aggregate query (which we already run, unconditionally) gives an exact \`extrinsic_count\` per day for free. Before the median query is even built, that count decides which days are cheap enough to compute an exact median for — a "safe" day's own \`[dayStart, dayEnd)\` range is an index-terminated scan (\`idx_extrinsics_observed\`) bounded to its already-verified-small row count, confirmed via \`EXPLAIN QUERY PLAN\` (plain index search, no temp-b-tree sort). A day over the cap is excluded from the median query outright — its rows are never touched — and it now reports an honest \`null\` median instead of any approximation. Total rows scanned is capped by \`(verified-safe days) * CHAIN_FEE_MEDIAN_SAMPLE_LIMIT\`, a hard ceiling fixed *before* the query runs, not an approximation of one.

Rewrote the test suite for this exact-or-null contract: a multi-day shape test proving the over-cap day contributes zero UNION ALL blocks, an index-plan test, and the param-count regression for the worst-case scoped 30-day window (now well under D1's 100-param limit since there's no longer a shared LIMIT param at all — just call_module once + a day-pair per safe day).

PR #2507, opened 2026-07-01
fix(api): edge-cache global validators leaderboard

JSONbored:
### Motivation
- Prevent repeated full-table D1 reads and in-worker aggregation from public calls to `/api/v1/validators` by adding the same analytics edge-cache protection used by adjacent routes.
- Ensure cache invalidation is tied to the neurons snapshot used by the leaderboard so warm caches avoid expensive recomputation.

### Description
- Wrap the top-level `/api/v1/validators` route with `withEdgeCache`, keyed as `global-validators`, so the response can be served from the edge cache instead of re-running live aggregation on every request (`workers/api.mjs`).
- Add `readGlobalValidatorsCacheStamp(env)` which returns `MAX(captured_at)` for validator rows (`validator_permit = 1 AND hotkey IS NOT NULL`) so cache keys bust on the neurons snapshot, not the health prober (`workers/request-handlers/analytics.mjs`).
- Extend the neurons-tier edge-cache regression test to include the global validators route and verify the cache key uses the neurons snapshot stamp (`tests/analytics-edge-cache.test.mjs`).

### Testing
- Ran focused unit tests: `npx vitest run tests/analytics-edge-cache.test.mjs tests/metagraph-neurons.test.mjs`, and all tests passed.
- Ran code checks: `git diff --check`, `npm run lint`, and `npm run format:check`, and they all succeeded.
- Built the project and validated the API contract: `npm run build` followed by `npm run validate:api`, and validation completed successfully after the build step.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a449826bc78832c9ea1e3dc83e886a2)

### DIFF
--- a/tests/analytics-edge-cache.test.mjs
+++ b/tests/analytics-edge-cache.test.mjs
@@ -1045,6 +1045,7 @@ describe("neurons-tier edge cache", () => {
     const env = neuronsEnv(queries);
 
     for (const [keyParts, path] of [
+      ["global-validators", "/api/v1/validators?sort=subnet_count&limit=1"],
       ["subnet-metagraph", "/api/v1/subnets/7/metagraph"],
       ["subnet-validators", "/api/v1/subnets/7/validators"],
       ["subnet-concentration", "/api/v1/subnets/7/concentration"],
--- a/workers/api.mjs
+++ b/workers/api.mjs
@@ -1171,8 +1171,20 @@ export async function handleRequest(request, env = {}, ctx = {}) {
 
   // Global validator/operator leaderboard from the current neurons snapshot. Exact path,
   // dispatched before subnet routing so the top-level collection stays unambiguous.
+  // Busts on the newest neuron captured_at across ALL subnets (like chain/concentration
+  // below), not a validator-permit-filtered stamp: a subnet refresh that drops a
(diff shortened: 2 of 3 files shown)


JSONbored:
Rebased onto main and resolved conflicts with the newer chain/concentration edge-cache addition (which landed a \`readNeuronsCacheStamp\` network-wide stamp helper after this branch diverged).

Fixed the flagged defect by dropping the new filtered \`readGlobalValidatorsCacheStamp\` (MAX(captured_at) WHERE validator_permit = 1 AND hotkey IS NOT NULL) and reusing the existing \`readNeuronsCacheStamp\` (unfiltered network-wide stamp, same helper /api/v1/chain/concentration already uses). The filtered variant misses exactly the case that matters most: a subnet refresh that drops a validator's permit=1 row doesn't touch a filtered MAX(captured_at) at all, so the leaderboard could go stale on membership churn. The unfiltered stamp busts on any subnet's snapshot advancing, matching the established network-wide pattern.

This also resolves the codecov/patch gap — the two uncovered lines were in the now-removed function; the reused helper is already fully covered by its existing chain/concentration tests, and the new global-validators case slots into the existing generic "neurons-tier edge cache" test loop.

PR #2509, opened 2026-07-01
fix(rpc-proxy): block CGNAT 100.64.0.0/10 in the RPC upstream safety guard

joaovictor712:
## Summary

- `isPrivateOrLocalHostname` (`workers/request-handlers/rpc-proxy.mjs`) is the literal-hostname guard `isSafeRpcEndpointUrl` relies on before the read-only RPC proxy (`POST /rpc/v1/{network}`) will forward a request to a `TRUSTED_RPC_UPSTREAM_ORIGINS` endpoint. It checks `0.0.0.0/8`, `10.0.0.0/8`, `127.0.0.0/8`, `169.254.0.0/16`, `172.16.0.0/12`, and `192.168.0.0/16` (plain-IPv4 and their `::ffff:`-mapped IPv6 forms) but had no `100.64.0.0/10` (RFC 6598 CGNAT) check in either branch.
- This repo has three other independent literal SSRF guards — `scripts/lib.mjs`'s `unsafeIpBlocks`, `src/webhooks.mjs`, and `src/health-probe-core.mjs`'s `isUnsafePublicUrl` — and all three already block `100.64.0.0/10`, added deliberately in #2312/#2313 ("block CGNAT 100.64.0.0/10 in the probe SSRF guard") specifically so every literal guard in the codebase stays in parity. `health-probe-core.mjs` carries the comment *"100.64.0.0/10 CGNAT — blocked by the webhook + build SSRF guards; the probe literal guard must stay in parity (issue #2312)"*. `workers/request-handlers/rpc-proxy.mjs`'s copy was never updated to match.
- Add the same range to both branches (mirroring the existing `172.16.0.0/12` pattern shape) and export the function so it is directly unit-testable like the sibling guards (`isUnsafePublicUrl` is exported and tested the same way in `tests/health-probe-core.test.mjs`).

## What Changed

- `workers/request-handlers/rpc-proxy.mjs` — `isPrivateOrLocalHostname` now rejects `100.64.0.0/10` in its plain-IPv4 branch (`first === 100 && second >= 64 && second <= 127`) and its `::ffff:`-mapped IPv6 branch; the function is now `export`ed.
- `workers/api.mjs` — re-export `isPrivateOrLocalHostname` alongside the other rpc-proxy test-facing helpers (`orderSafeRpcEndpoints`, `selectSafeRpcEndpoint`, etc.), matching the existing re-export comment's stated purpose.
- `tests/rpc-endpoint-selection.test.mjs` — new `describe("isPrivateOrLocalHostname — CGNAT parity (#2312/#2313)")` block: rejects the range (dotted + `::ffff:`-mapped) and confirms addresses just outside the range (`100.63.255.255`, `100.128.0.0`, `8.8.8.8`) are not rejected.

No schema/contract change — this only tightens an internal safety predicate; `TRUSTED_RPC_UPSTREAM_ORIGINS` is a fixed set of registered domains today, so this is defense in depth against a future origin resolving into the CGNAT range, the same posture the original #2312/#2313 fix documents for its own guard.

## Registry Safety

- [x] Generated artifacts were produced by repo scripts, not hand-edited. (N/A — no generated artifacts touched.)
- [x] Public API/OpenAPI/schema changes are intentional and documented. (N/A — no public contract change.)

## Validation

- [x] `npm run lint`
- [x] `npx prettier --check` on the three changed files
- [x] `git diff --check`
- [x] `npx vitest run tests/rpc-endpoint-selection.test.mjs tests/rpc-failover.test.mjs` (49/49 passing)
- [x] `npx vitest run` (full suite): 3702/3719 passing; the 17 pre-existing failures (Windows symlink `EPERM`, local syntax-parse issues in a few files, stale local `generated/` artifacts) reproduce identically on a clean `main` before this change and are unrelated to the touched files.


JSONbored (commented):
Failing CI, closing.

PR #2511, opened 2026-07-01
fix(api): stamp metagraph from first valid row and coerce D1 flag strings

joaovictor91123:
## Summary

`buildSubnetMetagraph` / `buildSubnetValidators` stamped `captured_at` and `block_number` from `rows[0]` **before** filtering malformed rows. When the leading slot was `null` or otherwise not a valid neuron object, the response reported `captured_at: null` and `block_number: null` while `neurons`/`validators` were populated from later rows.

Separately, `formatNeuron` used `Boolean(row.active)` (and sibling flags). In JavaScript `Boolean("0") === true`, so numeric-string zero cells from D1 would mark inactive neurons as active â€” inconsistent with `formatRegistration` in `account-events.mjs`, which already uses `Number(value) === 1` after #2487.

## What Changed

- `snapshotStamp` now derives metadata from the first row that survives `formatNeuron`, matching the filtered neuron array.
- Introduced `toD1Flag()` (`Number(value) === 1`) for `active`, `validator_permit`, and `is_immunity_period`.
- Tests: malformed leading row still yields correct snapshot metadata; numeric-string `"0"` flags coerce to `false`.

## Registry Safety

- [x] No public API/OpenAPI/schema changes (correctness fix; field shapes unchanged).

## Validation

- [x] `npx vitest run tests/metagraph-neurons.test.mjs` (all passed)
- [x] `npx prettier --check src/metagraph-neurons.mjs tests/metagraph-neurons.test.mjs`
- [x] `git diff --check`


JSONbored (commented):
Merge conflicts, closing. 

```
Use the [web editor](https://github.com/JSONbored/metagraphed/pull/2511/conflicts) or the command line to resolve conflicts before continuing.

src/metagraph-neurons.mjs
```

PR #2515, opened 2026-07-01
fix(api): stop leaking D1 string cells into neuron API payloads

real-venus:
# Summary

This PR fixes incorrect type handling when formatting neuron data returned from D1.

Previously, several numeric fields could be returned as strings instead of numbers, and boolean flags could be interpreted incorrectly when D1 returned `"0"` or `"1"` string values. As a result, the API response could violate its schema and report incorrect boolean values.

This change introduces explicit type coercion helpers to ensure numeric and boolean fields are serialized with the correct types while preserving `null` values for missing data.

## Root Cause

`formatNeuron()` and the `block_number` fields in `snapshotStamp()` and `buildNeuronDetail()` used raw D1 values without normalization.

This caused two issues:

- Numeric database values returned as strings (e.g. `"1000.5"`) were exposed directly in fields that the API defines as numbers.
- Boolean fields were converted using `Boolean(value)`, causing `"0"` to evaluate to `true` because non-empty strings are truthy in JavaScript.

### Source

**`src/metagraph-neurons.mjs`**

Added three null-preserving coercion helpers:

- `intCell()` — converts values to non-negative integers, otherwise returns `null`
- `numberCell()` — converts values to finite numbers, otherwise returns `null`
- `flagCell()` — converts values using `Number(value) === 1`

Applied these helpers to:

- Numeric neuron fields
- Boolean flag fields
- `block_number` values in `snapshotStamp()`
- `block_number` values in `buildNeuronDetail()`

The new helpers preserve existing `null` behavior for missing database values.

### Tests

**`tests/metagraph-neurons.test.mjs`**

Added regression tests verifying:

- String numeric values are converted to numbers.
- `"0"` and `"1"` string flags correctly become `false` and `true`.
- Invalid, negative, or fractional integer values return `null`.
- `buildNeuronDetail()` correctly normalizes string `block_number` values.

## Impact

- ✅ Numeric fields always serialize as numbers.
- ✅ Boolean fields correctly represent database values.
- ✅ API responses remain consistent with the documented schema.
- ✅ Existing handling of missing (`null`) values is preserved.

## Validation

- [x] `npx vitest run tests/metagraph-neurons.test.mjs`
- [x] `npx eslint src/metagraph-neurons.mjs tests/metagraph-neurons.test.mjs`
- [x] `git diff --check`

### DIFF
--- a/src/metagraph-neurons.mjs
+++ b/src/metagraph-neurons.mjs
@@ -75,26 +75,47 @@ function round(value, dp = 6) {
   return Math.round(value * factor) / factor;
 }
 
-// One D1 row → a clean Neuron object. SQLite stores booleans as 0/1 INTEGER, so
-// coerce the flag columns back to real booleans for the API.
+// D1 can surface an INTEGER/REAL column as a numeric string; coerce so the API
+// never leaks a string into a numeric field — the same fix already applied to
+// blocks.mjs (#2435), extrinsics.mjs (#2439), and account-events.mjs. These
+// PRESERVE null (a missing/null cell stays null), unlike the null→0 aggregation
+// helpers (nonNegativeInt / nullableNumber) used by buildGlobalValidators.
+function intCell(value) {
+  if (value == null) return null;
+  const n = Number(value);
+  return Number.isInteger(n) && n >= 0 ? n : null;
+}
+function numberCell(value) {
+  if (value == null) return null;
+  const n = Number(value);
+  return Number.isFinite(n) ? n : null;
+}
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
Merge conflicts: 

```
Use the [web editor](https://github.com/JSONbored/metagraphed/pull/2515/conflicts) or the command line to resolve conflicts before continuing.

src/metagraph-neurons.mjs
tests/metagraph-neurons.test.mjs
```

PR #2510, opened 2026-07-01
fix(api): short-circuit inverted block_start/block_end on event feeds before D1

joaovictor91123:
## Summary

`handleBlocks` and `handleExtrinsics` already short-circuit inverted `?block_start=` / `?block_end=` ranges before hitting D1, but the account and subnet **event feeds** did not: `handleAccountEvents`, `handleAccountExtrinsics`, `handleAccountTransfers`, and `handleSubnetEvents` still paid for a D1 round-trip on a deterministically empty query (`block_start=900&block_end=100` â†’ 200 + empty list).

## What Changed

- Added `isInvertedBlockRange()` in `workers/request-params.mjs` (shared guard).
- All four event-feed handlers now return a schema-stable empty feed without querying D1 when `block_start > block_end`, mirroring `handleBlocks`.
- Regression tests in `tests/request-handlers-entities.test.mjs` assert zero SQL captures for each handler.

## Registry Safety

- [x] No secrets, PATs, wallet data, private dashboards, private URLs, or validator-local state.
- [x] Generated artifacts were produced by repo scripts, not hand-edited.
- [x] R2-only/high-churn detail artifacts are not committed.
- [x] No public API/OpenAPI/schema changes (performance/consistency guard only).

## Validation

- [x] `npx vitest run tests/request-handlers-entities.test.mjs` (all passed)
- [x] `npx prettier --check workers/request-handlers/entities.mjs workers/request-params.mjs tests/request-handlers-entities.test.mjs`
- [x] `git diff --check`

PR #2496, opened 2026-07-01
feat(mcp): block-explorer feed filter parity for list_blocks and list_extrinsics

kiannidev:
## Summary

- Extend shared `loadBlocks` / `loadExtrinsics` loaders with the full REST conjunctive filter sets (including short-circuit guards and extrinsic index hints) and refactor `handleBlocks` / `handleExtrinsics` to delegate to them.
- Add REST-parity filter parameters to MCP `list_blocks` and `list_extrinsics` (MCP **1.14.0 → 1.15.0**).
- Exhaustive unit + MCP tests for every filter path, short-circuit branch, cursor+filter AND, and index-hint branch to satisfy codecov patch.

### REST ↔ MCP parity

| `GET /api/v1/blocks` query param | MCP `list_blocks` arg | Notes |
|---|---|---|
| `author` | `author` | SS58 block author |
| `spec_version` | `spec_version` | Runtime spec version |
| `block_start` / `block_end` | `block_start` / `block_end` | Inclusive height range |
| `from` / `to` | `from` / `to` | `observed_at` epoch-ms range |
| `min_extrinsics` / `min_events` | `min_extrinsics` / `min_events` | Per-block count floors |
| `limit` / `offset` / `cursor` | `limit` / `offset` / `cursor` | Keyset cursor ANDs with filters |

| `GET /api/v1/extrinsics` query param | MCP `list_extrinsics` arg | Notes |
|---|---|---|
| `block` | `block` | Exact block height |
| `signer` | `signer` | SS58 signer |
| `call_module` / `call_function` | `call_module` / `call_function` | Pallet call filters |
| `success` | `success` | Boolean (`true` / `false`); omit = no filter |
| `block_start` / `block_end` | `block_start` / `block_end` | Inclusive height range |
| `from` / `to` | `from` / `to` | `observed_at` epoch-ms range |
| `limit` / `offset` / `cursor` | `limit` / `offset` / `cursor` | Keyset cursor ANDs with filters |

**Non-goals:** No overlap with open PRs for chain-events feeds, stake-flow, movers, or all-events explorer parity (#2494).

## Test plan

- [x] `npm run test:coverage -- tests/blocks.test.mjs tests/extrinsics.test.mjs tests/mcp-server.test.mjs`
- [x] `node scripts/validate-mcp.mjs` (65 tools, lifecycle + all tools/call)

PR #2516, opened 2026-07-01
fix(api): short-circuit inverted block range on extrinsics feed before D1

luciferlive112116:
## What

`GET /api/v1/extrinsics` accepts optional `?block_start=` / `?block_end=` height filters (inclusive). When `block_start > block_end`, the range is impossible — no row can satisfy both bounds.

`handleBlocks` already short-circuits this case (and inverted `from > to`) before touching D1. `handleExtrinsics` short-circuited impossible **time** windows but still built SQL for inverted **block** ranges, forcing a D1 scan that deterministically returns an empty page.

## How

Add `block_start > block_end` to the existing pre-D1 guard in `handleExtrinsics`, returning the same schema-stable empty `buildExtrinsicFeed` payload the time-window short-circuits use. No contract or response-shape change for valid requests.

## Why no linked issue

Standalone guard gap found by comparing the blocks and extrinsics feed handlers. Open PR #2510 covers the same pattern on **event** feeds; this completes the sibling **extrinsics** explorer feed path. No open issue tracks this specific handler gap.

## Scope / risk

One conditional in `handleExtrinsics` plus one handler test asserting zero D1 queries. Valid ranges and other filters unchanged.

## Tests

- `request-handlers-entities.test.mjs`: `short-circuits an inverted block_start>block_end window before D1` — asserts empty feed and no `FROM extrinsics` SQL.

## Test plan

- [x] Inverted `block_start=500&block_end=100` returns empty feed without D1
- [x] Existing inverted `from>to` and valid-window tests continue to pass
- [x] CI vitest suite (no local Node toolchain; relies on repo CI)

### DIFF
--- a/tests/request-handlers-entities.test.mjs
+++ b/tests/request-handlers-entities.test.mjs
@@ -3350,6 +3350,20 @@ describe("handleExtrinsics", () => {
     );
   });
 
+  test("short-circuits an inverted block_start>block_end window before D1", async () => {
+    const { env, captures } = dbWith({ extrinsics: [] });
+    const body = await json(
+      await handleExtrinsics(
+        req("/api/v1/extrinsics"),
+        env,
+        url("/api/v1/extrinsics?block_start=500&block_end=100"),
+      ),
+    );
+    assert.equal(body.data.extrinsic_count, 0);
+    assert.deepEqual(body.data.extrinsics, []);
+    assert.equal(captures.sql.length, 0);
+  });
+
   test("a valid recent window is NOT short-circuited and queries D1", async () => {
     const { env, captures } = dbWith({ extrinsics: [] });
     const now = Date.now();


luciferlive112116:
No-issue rationale: `handleBlocks` already short-circuits `block_start > block_end` before D1; `handleExtrinsics` only guarded impossible **time** windows. This PR completes the sibling **extrinsics** explorer feed path. Open PR #2510 covers the same pattern on **event** feeds only — different handler, not duplicate work.

Scope/risk: one pre-D1 guard branch + one regression test (`captures.sql.length === 0`). Valid filters unchanged.

luciferlive112116:
Rebased onto current `main` to clear merge conflicts.

**Conflict cause:** `handleExtrinsics` was refactored to delegate to `loadExtrinsics` (#2496), which already short-circuits inverted `block_start > block_end`. The old handler-level SQL guard conflicted with that refactor.

**Resolution:** dropped the redundant `entities.mjs` changes; kept a single handler regression test (`captures.sql.length === 0`) proving the REST path still short-circuits before D1.

JSONbored (approved):

PR #2600, opened 2026-07-01
fix(api): coerce block detail anchor before neighbor lookup

philluiz2323:
## Summary

Fix `GET /api/v1/blocks/{ref}` returning `prev_block_number` / `next_block_number: null` when D1 returns `block_number` as a numeric string.

## What Changed

- `handleBlock` gated the neighbor subquery on `Number.isInteger(rows[0].block_number)`. D1 INTEGER columns often arrive as `"1234"`, so the lookup was skipped even though `formatBlock` already coerces the displayed `block_number`.
- Coerce the anchor through `formatBlock(rows[0])?.block_number` before binding the MAX/MIN neighbor query — the same pattern `loadBlock` uses after #2506.
- Added a handler regression test with `block_number: "1234"` in the mock row.

## Registry Safety

- [x] Public API/OpenAPI/schema changes are intentional and documented. (No contract change.)

## Validation

- [x] `npx vitest run tests/request-handlers-entities.test.mjs -t "resolves neighbors when D1 returns a string"`
- [x] `git diff --check`

### DIFF
--- a/tests/request-handlers-entities.test.mjs
+++ b/tests/request-handlers-entities.test.mjs
@@ -2898,6 +2898,23 @@ describe("handleBlock", () => {
     assert.equal(body.data.next_block_number, 1240);
   });
 
+  test("resolves neighbors when D1 returns a string-typed block_number", async () => {
+    const { env, captures } = dbWith({
+      blockDetail: blockRow({ block_number: "1234" }),
+      blockNeighbors: { prev: 1230, next: 1240 },
+    });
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
Merge conflicts: 

```
Use the [web editor](https://github.com/JSONbored/metagraphed/pull/2600/conflicts) or the command line to resolve conflicts before continuing.

```

PR #2609, opened 2026-07-01
fix(extrinsics): coerce string-typed fee/tip/observed_at cells in formatExtrinsic

ai-mountain:
## Summary

Coerce string-typed `fee_tao`, `tip_tao`, and `observed_at` cells to their proper types in `formatExtrinsic`, following the same D1 coercion pattern established in earlier PRs (#2435, #2439, #2481, #2489).

## Changes

**`src/extrinsics.mjs`**
- `fee_tao`: apply `nullableNumber` + `roundTao` instead of bare `?? null`
- `tip_tao`: apply `nullableNumber` + `roundTao` instead of bare `?? null`
- `observed_at`: apply `toIso` instead of bare `?? null`

The remaining `success` cell is handled separately by #2512.

**`tests/extrinsics.test.mjs`**
- Add test cases for string-typed and null inputs for fee_tao, tip_tao, observed_at

### DIFF
--- a/src/extrinsics.mjs
+++ b/src/extrinsics.mjs
@@ -35,21 +35,32 @@ export const EXTRINSIC_INSERT_COLUMNS = [
   "observed_at",
 ];
 
-function toIso(ms) {
-  return Number.isFinite(ms) ? new Date(ms).toISOString() : null;
-}
-
 // Coerce a chain-position cell (block_number / extrinsic_index) to a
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
Merge conflicts, also this: 

```
The notable edge is that timestamp coercion still treats any finite number as ISO-safe, so extreme numeric strings can throw before returning the formatter's intended null-safe shape.
```

PR #2635, opened 2026-07-01
fix(api): cap chain activity success_rate above 1.0

joaovictor91123:
## Summary
- Extend round4 in chain-analytics so success_rate never exceeds 1.0 when stale aggregates report successful_extrinsics above extrinsic_count.
- Complements the existing sub-perfect clamp to 0.9999 (#2020).

## Test plan
- [x] Unit test for 105/100 successful ratio caps at 1.0
- [x] Existing sub-perfect clamp test still passes


JSONbored (commented):
A notable maintainability issue is that round4 is shared by other ratio fields, while the new comment describes only the success_rate-specific source of over-1.0 inputs.

PR #2639, opened 2026-07-01
feat(api): add network-wide consensus performance distribution analytics endpoint

GildardoDev:
Adds GET /api/v1/chain/consensus, a network-wide Yuma-consensus performance distribution computed live from the neurons D1 tier across every subnet. For each signal (trust, consensus, incentive, dividends) it reports a count, mean, and min/p25/median/p75/p90/max spread over the earning participants, plus active, validator, and miner counts, reusing a shared DistributionSummary schema.

Validated locally: npm run build, lint, format:check, validate:contract-drift, validate:schemas, validate:openapi, validate:openapi-examples, validate:api, validate:docs, scan:public-safety, and the full vitest suite all pass, with complete branch coverage on the new code.


JSONbored (commented):
Failing CI/tests, closing.

PR #2645, opened 2026-07-01
fix(scan): catch secrets under underscore-prefixed credential names in public-safety scan

glorydavid03023:
## Summary

- The `token-like assignment` hard-secret rule in `scripts/scan-public-safety.mjs` anchored the bare `secret`/`password` keywords with a leading `\b`. Because `_` is a word character, there is no word boundary after the underscore in a compound key name, so `client_secret=…`, `db_password=…`, `my_secret:…` — the most common real-world credential variable names — **did not match**, and a genuine leaked value under them slipped past `scan:public-safety` as a false negative.
- The rule already encodes underscore/hyphen joins for the sibling alternatives (`api[_-]?key`, `access[_-]?token`, `auth[_-]?token`), so the gap for `secret`/`password` was an inconsistency, not a design choice.

## What Changed

- `scripts/scan-public-safety.mjs`: add an optional `(?:[a-z0-9]+[_-])?` prefix before the keyword group so compound credential names match, while bare `secret=`/`password=` and the 16+ char value floor still apply.
- `tests/public-safety.test.mjs`: regression asserting `client_secret=…`, `db_password=…`, and bare `secret=…` are all flagged `token-like assignment`. Fails without the fix.

Verified the scanner still reports `Public-safety scan passed.` on the whole tree — the broadening closes the leak gap with **no new false positives** on committed files.

### Reproduction (before)

A file committed anywhere in the tree containing:

```
client_secret=abcdefghijklmnop1234
```

was **not** flagged by `npm run scan:public-safety` — even though a bare `secret=abcdefghijklmnop1234` was. After the fix, both are flagged.

## Registry Safety

- [x] No secrets, PATs, wallet data, private dashboards, private URLs, or validator-local state. (The regression uses obvious placeholder values; the scanner still passes.)

## Validation

- [x] `npm run scan:public-safety` — `Public-safety scan passed.` (no new false positives on the tree)
- [x] `npx eslint scripts/scan-public-safety.mjs tests/public-safety.test.mjs` (clean)
- [x] `git diff --check`

### DIFF
--- a/scripts/scan-public-safety.mjs
+++ b/scripts/scan-public-safety.mjs
@@ -37,8 +37,13 @@ const patterns = [
   },
   {
     name: "token-like assignment",
+    // The optional `[a-z0-9]+[_-]` prefix catches underscore/hyphen-joined
+    // credential names (client_secret, db_password, my_secret): a leading `\b`
+    // has no boundary after the underscore in `client_secret`, so the bare
+    // `secret`/`password` alternatives would otherwise miss the most common
+    // real-world key names while still matching bare `secret=`/`password=`.
     regex:
-      /\b(?:api[_-]?key|access[_-]?token|auth[_-]?token|secret|password)\s*[:=]\s*["']?[A-Za-z0-9_./+=-]{16,}/i,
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
Closing due to review blockers.

PR #2646, opened 2026-07-01
feat(api): add per-subnet native transfer-volume analytics

pekar9781:
## Summary

Adds `GET /api/v1/subnets/{netuid}/transfer-volume`, the per-subnet companion of the network-wide `/chain/transfers` route and the per-account `/accounts/{ss58}/transfers` + `/counterparties` routes.

Because Balances.Transfer rows carry no netuid, attribution uses the neurons snapshot. The Worker handler validates `window` and `limit` strictly (400 on unsupported values); the loader accepts optional `nowMs` for deterministic window cutoffs in tests.

## What Changed

- New `src/subnet-transfer-volume.mjs` loader + `buildSubnetTransferVolume` shaper
- D1 migration `0031_account_events_kind_observed.sql` + `INDEXED BY` hints for live queries
- OpenAPI examples now inherit `CONTRACT_VERSION` instead of a stale hardcoded value
- Worker handler, routing, edge-cache, contracts + docs; `CONTRACT_VERSION` → `2026-07-01.5`

## Registry Safety

## Validation

- [x] `npm run validate`
- [x] `npm run build`
- [x] `npx vitest run tests/subnet-transfer-volume.test.mjs tests/contracts.test.mjs tests/openapi-sample.test.mjs tests/request-handlers-entities.test.mjs`


JSONbored (commented):
Artifact churn is unnecessary here, closing.

PR #2648, opened 2026-07-01
feat(api): add network-wide emission yield analytics endpoint

GildardoDev:
Adds GET /api/v1/chain/yield, a network-wide emission yield (return on stake) analytics endpoint computed live from the neurons D1 tier across every subnet. It reports stake and emission totals, stake-weighted network/validator/miner aggregate yields, a per-neuron yield distribution (count, mean, min/p25/median/p75/p90/max) for each of those three lenses via a shared YieldDistribution schema, and a limit-capped top-yielders leaderboard.

Validated locally: npm run build, lint, format:check, validate:contract-drift, validate:schemas, validate:openapi, validate:openapi-examples, validate:api, validate:docs, scan:public-safety, and the full vitest suite all pass, with complete branch coverage on the new code.


JSONbored (commented):
Closing, CI failures + documentation gaps, as well as unnecessary artifact churn.

```
The most notable issue is documentation/example precision rather than a visible runtime break.
```

PR #2622, opened 2026-07-01
fix(api): short-circuit inverted block range on account extrinsics before D1

luciferlive112116:
## What

`GET /api/v1/accounts/{ss58}/extrinsics` accepts optional `?block_start=` / `?block_end=` height filters. When `block_start > block_end`, the range is impossible.

The global extrinsics feed (`loadExtrinsics`, #2496) and account history (`loadAccountHistory`, #2594) already short-circuit impossible windows before D1. Account-scoped extrinsics still queried D1 for a deterministically empty page.

## How

Add `block_start > block_end` to a pre-D1 guard in `loadAccountExtrinsics`, returning the existing schema-stable empty `buildAccountExtrinsics` payload. REST and MCP account tools share this loader — no duplicate handler guard needed.

## Why no linked issue

Standalone guard gap found by comparing account tail loaders with `loadExtrinsics` / `loadAccountHistory`. Closed PR #2510 targeted handler-level event feeds broadly; this completes the **account extrinsics** signer-scoped path in the shared loader. No open issue tracks this specific gap.

## Scope / risk

One guard in `loadAccountExtrinsics` plus loader + handler tests asserting zero D1 work. Valid filters unchanged.

## Tests

- `account-events.test.mjs`: `loadAccountExtrinsics short-circuits an inverted block range before D1`
- `request-handlers-entities.test.mjs`: inverted `block_start=500&block_end=100` returns empty feed with `captures.sql.length === 0`

## Test plan

- [x] Inverted block window returns empty feed without D1 (loader + REST handler)
- [x] Existing valid-range account extrinsics tests continue to pass
- [x] CI vitest suite (no local Node toolchain; relies on repo CI)

PR #2649, opened 2026-07-01
fix(api): short-circuit an inverted block range on the account-events feed before D1

glorydavid03023:
## Summary

- `loadAccountEvents` gained a `block_start`/`block_end` range filter (#2603), documented as "parity with the extrinsics and chain-events feeds". But unlike its siblings `loadAccountExtrinsics` (#2622) and `loadAccountTransfers` (#2625), it had **no inverted-range short-circuit**: a `block_start > block_end` request is a deterministic no-match, yet `loadAccountEvents` still assembled and ran the D1 union query.
- The sibling loaders explicitly guard this: *"Inverted block-height bounds are a deterministic no-match. Short-circuit before D1 so REST and MCP callers cannot force a scan to prove an impossible empty page."* Account-events was the one block-range-filtered feed missing that guard.

## What Changed

- `src/account-events.mjs`: add the same inverted-range guard to `loadAccountEvents` — when both bounds are present and `blockStart > blockEnd`, return the schema-stable empty page before touching D1. The valid single-block `start == end` case is preserved (`>` not `>=`), matching the siblings.
- `tests/account-events.test.mjs`: regression asserting the inverted range returns an empty feed with the injected D1 runner never called. Fails without the fix.

No schema/OpenAPI/contract change; behavior is identical for every non-inverted range (correct results either way — this removes a needless D1 scan and closes the same scan-avoidance gap the siblings already close).

### Reproduction (before)

`GET /api/v1/accounts/{ss58}/events?block_start=500&block_end=100` (or the MCP `get_account_events` equivalent) executed a full D1 union query guaranteed to return nothing, instead of short-circuiting like the extrinsics/transfers feeds.

## Registry Safety

- [x] Public API/OpenAPI/schema changes are intentional and documented. (None — completes existing feed parity.)

## Validation

- [x] `npx vitest run tests/account-events.test.mjs` (73 pass; the new regression fails without the fix)
- [x] `npx eslint src/account-events.mjs tests/account-events.test.mjs` (clean)
- [x] `git diff --check`

### DIFF
--- a/src/account-events.mjs
+++ b/src/account-events.mjs
@@ -627,6 +627,16 @@ export async function loadAccountEvents(
 ) {
   const lim = clampLimit(limit, FEED_PAGINATION);
   const off = clampOffset(offset);
+  // Inverted block-height bounds are a deterministic no-match. Short-circuit before
+  // D1 so REST and MCP callers cannot force a scan to prove an impossible empty
+  // page — parity with the extrinsics and transfers feeds (#2622/#2625).
+  if (blockStart != null && blockEnd != null && blockStart > blockEnd) {
+    return buildAccountEvents([], ss58, {
+      limit: lim,
+      offset: off,
+      nextCursor: null,
+    });
+  }
   const filterParts = [];
   const filterParams = [];
   if (kind) {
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
Merge conflicts, closing. 

```

src/account-events.mjs
```

PR #2656, opened 2026-07-01
feat(mcp): add get_chain_concentration mirroring REST chain concentration

bohdansolovie:
## Summary

- Add MCP tool `get_chain_concentration` for the live REST route `GET /api/v1/chain/concentration`: network-wide stake and emission decentralization metrics (Gini, HHI, Nakamoto, top-percentile shares, entropy) across per-UID, per-entity, and validator-only lenses.
- Wire the tool to the shared `loadChainConcentration` loader (`src/concentration.mjs`); REST and MCP share one read path. Snapshot-based with no query parameters.
- Add a typed `TOOL_OUTPUT_SCHEMAS` entry matching `ChainConcentrationArtifact`, MCP integration tests, and `validate:mcp` coverage. Bump `MCP_SERVER_VERSION` and `server.json` to **1.17.0**.

## Test plan

- [x] `npx vitest run tests/mcp-server.test.mjs -t "get_chain_concentration"`
- [x] `npm run validate:mcp` — 69 tools, lifecycle + all tools/call
- [x] `npm run lint`
- [x] `npm run format:check`

### DIFF
--- a/scripts/validate-mcp.mjs
+++ b/scripts/validate-mcp.mjs
@@ -471,6 +471,13 @@ assert.ok(
     Array.isArray(transfersCold.top_receivers),
   "get_chain_transfers must return window + top_senders[] + top_receivers[] on cold D1",
 );
+const chainConcentrationCold = await callOk("get_chain_concentration", {});
+assert.ok(
+  chainConcentrationCold.neuron_count === 0 &&
+    chainConcentrationCold.subnet_count === 0 &&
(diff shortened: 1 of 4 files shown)


JSONbored (commented):
Closing: 

```
The notable weakness is that the advertised output schema keeps the concentration metric blocks as untyped objects, which limits contract precision but does not break the tool.
```

PR #2669, opened 2026-07-02
fix(badge): provider apis badge n/a when any subnet lacks surfaces

joaovictor91123:
## Summary
- Provider `metric=apis` badges now require every listed subnet to resolve a surfaces artifact before summing callable API counts.
- A partial sum (e.g. 2 apis when one of two subnets is missing its artifact) degrades to n/a instead of understating the missing subnet as zero.

## Test plan
- [x] Unit test: provider spanning netuids [7, 9] with only netuid 7 having surfaces renders n/a
- [x] Existing datura provider sum test still passes

### DIFF
--- a/src/badge.mjs
+++ b/src/badge.mjs
@@ -431,12 +431,15 @@ async function apisContent({ target, readArtifact, env }) {
     );
     const netuids = findProvider(providers, target.slug)?.netuids || [];
     if (netuids.length) {
-      const counts = (
-        await Promise.all(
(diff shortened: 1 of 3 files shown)


JSONbored (commented):
Merge conflict + small problem: 

```
The added badge test covers the visible output, but it relies on ambient fixture state for netuid 7, so it could pass even if the intended partial-resolution setup stops being true.
```

Closing.

PR #2686, opened 2026-07-02
fix(api): reject unknown list query parameters

### LINKED ISSUE #2578: Reject unknown query params on list routes instead of silently ignoring them
applyListTransform (repo/workers/list-query.mjs:189) validates known params but never rejects an unrecognized one, so a typo like ?statuss=active silently returns the full unfiltered list — a correctness footgun. Add an opt-in strict check that 400s on any param not in the collection's known set (q/fields/limit/cursor/sort/order + configured filters/csv/array/range), reusing listQueryParamNames (list-query.mjs:43).

## Deliverables
- [ ] In validateListQuery (list-query.mjs:322), compute the allowed param set from listQueryParamNames(...) and return a 400 { parameter, message: 'unknown query parameter' } for any present param outside it
- [ ] Keep pagination/meta shape identical for valid requests (no behavior change on the happy path)
- [ ] Tests: ?statuss=x -> 400, every real param still accepted, empty params -> ok (both branches of the membership check)

## References
- `repo/workers/list-query.mjs:189-234 (applyListTransform)`
- `repo/workers/list-query.mjs:322-425 (validateListQuery)`
- `repo/workers/list-query.mjs:43-68 (listQueryParamNames — the known set)`

---
_size: S · gittensor:bug_

joaovictor712:
## Summary

- Reject unknown query keys on list routes instead of silently ignoring them.
- Return a deterministic invalid-query error (`unknown query parameter.`) naming the offending key.
- Add regression tests for unknown-key rejection, known-key acceptance, and empty-query pass-through.

## What Changed

- Added an allow-list check in `validateListQuery` (`workers/list-query.mjs`) that computes valid keys from standard list params plus configured filters/range/csv/array params.
- Added tests in `tests/list-query.test.mjs` to cover typo rejection and unchanged happy-path behavior.
- Updated pagination test expectations to reflect strict unknown-param rejection.

## Registry Safety

## Validation

- [x] `npm test -- tests/list-query.test.mjs`

Fixes #2578

### DIFF
--- a/generated/metagraphed-api.d.ts
+++ b/generated/metagraphed-api.d.ts
@@ -7961,6 +7961,7 @@ export interface operations {
                 kind?: "archive" | "dashboard" | "data-artifact" | "docs" | "example" | "openapi" | "repo-registry" | "sdk" | "source-repo" | "sse" | "subnet-api" | "subtensor-rpc" | "subtensor-wss" | "website";
                 provider?: string;
                 state?: "schema-invalid" | "schema-valid" | "maintainer-review" | "verified" | "stale" | "rejected";
+                /** @description Optional projection list. Unknown query parameters are rejected with `400 invalid_query` (parameter named in `meta.parameter`) instead of being ignored. */
                 fields?: string;
                 limit?: number;
                 cursor?: number;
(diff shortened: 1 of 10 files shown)


JSONbored (commented):
Fix this: 

```
 the contract documentation is attached to the `fields` parameter rather than the list operation/query contract as a whole, which is easy for clients to miss.
```

PR #2695, opened 2026-07-02
fix(scan): catch secrets under multi-segment credential names in public-safety scan

glorydavid03023:
## Summary

- The `token-like assignment` hard-secret rule in `scripts/scan-public-safety.mjs` anchored the bare `secret`/`password` keywords with a leading `\b`. Because `_` is a word character, there's no boundary after the underscore in a compound key name, so `client_secret=…`, `google_oauth_client_secret=…`, `database_user_password=…` — the most common real credential variable names — **did not match**, and a genuine leaked value under them slipped past `scan:public-safety`.
- The rule already encodes underscore/hyphen joins for its sibling alternatives (`api[_-]?key`, `access[_-]?token`, `auth[_-]?token`), so the gap for `secret`/`password` was an inconsistency.

> This supersedes #2645, which added a **single**-segment prefix (`(?:[a-z0-9]+[_-])?`) — the review correctly noted that left multi-segment names like `google_oauth_client_secret` still uncovered. This uses `*` so a compound name of **any depth** matches.

## What Changed

- `scripts/scan-public-safety.mjs`: add a `(?:[a-z0-9]+[_-])*` prefix before the keyword group. The `[a-z0-9]+[_-]` segment cannot match empty and the separator is disjoint from `[a-z0-9]`, so there is no catastrophic backtracking (a 100k-char no-match input scans in ~1ms; a 40k-segment input in ~3ms). Bare `secret=`/`password=` and the 16+ char value floor still apply.
- `tests/public-safety.test.mjs`: regression covering one-, two-, and three-segment key names (`client_secret`, `google_oauth_client_secret`, `database_user_password`) plus the bare `secret=` form. Fails without the fix.

Verified the scanner still reports `Public-safety scan passed.` on the whole tree — no new false positives.

### Reproduction (before)

`google_oauth_client_secret=abcdefghijklmnop1234` committed anywhere in the scanned tree was **not** flagged by `npm run scan:public-safety`, even though bare `secret=abcdefghijklmnop1234` was.

## Registry Safety

- [x] No secrets, PATs, wallet data, private dashboards, private URLs, or validator-local state. (Regression uses obvious placeholders; scanner still passes.)

## Validation

- [x] `npm run scan:public-safety` — `Public-safety scan passed.` (no new false positives)
- [x] ReDoS-checked: 100k-char / 40k-segment inputs scan in ~1–3ms
- [x] `npx eslint` + `npx prettier --check` clean · `git diff --check`

### DIFF
--- a/scripts/scan-public-safety.mjs
+++ b/scripts/scan-public-safety.mjs
@@ -41,8 +41,16 @@ const patterns = [
   },
   {
     name: "token-like assignment",
+    // The optional `(?:[a-z0-9]+[_-])*` prefix matches underscore/hyphen-joined
+    // credential names of ANY depth (client_secret, google_oauth_client_secret,
+    // database_user_password): a leading `\b` has no boundary after the
+    // underscore in a compound name, so the bare `secret`/`password` alternatives
+    // would otherwise miss the most common real key names while still matching
+    // bare `secret=`/`password=`. The `+[_-]` segment can't match empty and the
+    // separator is disjoint from `[a-z0-9]`, so there is no catastrophic
+    // backtracking.
     regex:
-      /\b(?:api[_-]?key|access[_-]?token|auth[_-]?token|secret|password)\s*[:=]\s*["']?[A-Za-z0-9_./+=-]{16,}/i,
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
Closing: 

```
he only notable gap is that the test comment promises short-value behavior without asserting it.
```

PR #2720, opened 2026-07-02
fix(metagraph-neurons): coerce string-typed captured_at cells in toIso

pekar9781:
## Summary

D1 can return \captured_at\ as a numeric string; \	oIso\ treated that as non-finite and emitted \
ull\ on metagraph/validators/neuron-detail responses.

## What Changed

- \src/metagraph-neurons.mjs\: coerce via \Number()\ before the finite check (same pattern as \locks.mjs\ / \ccount-events.mjs\).
- \	ests/metagraph-neurons.test.mjs\: regression for string \captured_at\ on snapshot stamp + neuron detail.

## Validation

- [ ] \
pm run worker:test\ / \
px vitest run tests/metagraph-neurons.test.mjs\ (CI)

### DIFF
--- a/src/metagraph-neurons.mjs
+++ b/src/metagraph-neurons.mjs
@@ -50,7 +50,9 @@ const GLOBAL_VALIDATOR_SUBNET_LIMIT = 10;
 const RAO_PER_TAO = 1e9;
 
 function toIso(ms) {
-  return Number.isFinite(ms) ? new Date(ms).toISOString() : null;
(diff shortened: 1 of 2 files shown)


JSONbored (commented):

```
The only notable edge is that finite but out-of-range numeric strings can still make `Date.prototype.toISOString()` throw, which is pre-existing for numeric inputs but newly reachable for string inputs.
```

PR #2721, opened 2026-07-02
fix(neuron-history): coerce string-typed block_number cells in buildNeuronHistory

milosde111:
## Summary

- Harden `buildNeuronHistory` so each history point's `block_number` is coerced to a non-negative integer (or `null`) instead of leaking D1 numeric strings into the payload.
- Add focused unit tests for string coercion and invalid-cell rejection.

## Problem

D1 can return INTEGER columns as numeric strings. `buildNeuronHistory` used `r.block_number ?? null`, so per-UID neuron history points could expose `block_number` as a string — breaking the `["integer","null"]` contract and downstream comparisons. The sibling `blocks.mjs` / `account-events.mjs` formatters already coerce block heights; this path was missed.

## Scope / risk

Two files only (`src/neuron-history.mjs`, `tests/neuron-history.test.mjs`). No schema or handler changes. Valid integer block numbers unchanged; missing values stay `null`.

## Test plan

- [x] `buildNeuronHistory coerces string-typed block_number cells to integers`
- [x] `buildNeuronHistory rejects invalid block_number cells to null`
- [x] Existing neuron-history tests pass (`npm test -- tests/neuron-history.test.mjs`)

### DIFF
--- a/src/neuron-history.mjs
+++ b/src/neuron-history.mjs
@@ -58,12 +58,22 @@ function toIso(ms) {
 // Coerce a block-height cell to a non-negative integer, or null when missing,
 // non-finite, or negative. D1 can return an INTEGER column as a numeric string,
 // so a bare `r.block_number ?? null` would silently leak the string into the
-// history payload (and break downstream arithmetic/comparisons). Mirrors the
-// `toBlockNumber` already applied in blocks.mjs / account-events.mjs.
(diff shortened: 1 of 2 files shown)


JSONbored (commented):

```
The only notable gap is that numeric inputs do not get the same safe-integer check as string inputs, which is worth tightening for consistency but does not break the visible path.
```

PR #2745, opened 2026-07-02
test(domain-tags): cover every domain tag rule

### LINKED ISSUE #2570: Add per-rule branch tests for src/domain-tags.mjs deriveDomainTags
src/domain-tags.mjs has 14 DOMAIN_TAG_RULES regex branches but tests/domain-tags.test.mjs is only 87 LOC and does not exercise each rule or the branchy deriveDomainTags input handling. Add tests that fire at least one keyword per tag, plus the nullish/non-string description+additional filter arm, the categories-that-are-domain-tags branch, non-array categories fallback, empty-input → [], and the sorted+deduped output invariant. Deterministic, safe (output is always from the fixed vocabulary).

## Deliverables
- [ ] Extend tests/domain-tags.test.mjs with one positive keyword match per DOMAIN_TAG_RULE tag
- [ ] description/additional filter: non-string values dropped; both null → text '' → no regex pass
- [ ] categories: a curated category equal to a domain tag is added (case-insensitive); a non-tag category is ignored; non-array categories → []
- [ ] Output invariant: result is sorted, de-duplicated (same tag from text and category appears once)
- [ ] Confirm DOMAIN_TAGS export ordering unchanged and 97% patch coverage

## References
- `src/domain-tags.mjs (DOMAIN_TAG_RULES:10, deriveDomainTags:77, DOMAIN_TAGS:70)`
- `tests/domain-tags.test.mjs (thin)`

---
_size: S · gittensor:bug_

Crystora:
## Summary

- Add direct test coverage for every `deriveDomainTags` domain rule.
- Assert the exact sorted `DOMAIN_TAGS` vocabulary ordering.
- Cover branchy input handling for nullish/non-string text, non-array categories, ignored non-domain categories, and case-insensitive domain categories.
- Strengthen sorted and de-duplicated output invariant coverage.

## Template Used

- Backend/code change

Closes #2570

### DIFF
--- a/tests/domain-tags.test.mjs
+++ b/tests/domain-tags.test.mjs
@@ -4,13 +4,57 @@ import { DOMAIN_TAGS, deriveDomainTags } from "../src/domain-tags.mjs";
 
 describe("DOMAIN_TAGS", () => {
   test("is the sorted controlled vocabulary", () => {
-    assert.ok(DOMAIN_TAGS.length >= 10);
-    assert.deepEqual(DOMAIN_TAGS, [...DOMAIN_TAGS].sort());
-    assert.ok(new Set(DOMAIN_TAGS).size === DOMAIN_TAGS.length);
+    assert.deepEqual(DOMAIN_TAGS, [
+      "agents",
+      "compute",
+      "data",
+      "finance",
+      "inference",
+      "media",
+      "prediction",
+      "privacy",
+      "robotics",
+      "science",
+      "search",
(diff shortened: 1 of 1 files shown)


JSONbored (commented):
Fix this, please: 

```
The main residual gap is that the new “every domain rule” test proves one representative keyword per tag, not necessarily every keyword/branch inside each rule.
```

Resubmit as a new PR when ready.

PR #2767, opened 2026-07-02
feat(ci): notify on cross-PR file overlap; clarify enum-mismatch errors

JSONbored:
## Summary

- Add a purely-advisory `pr-overlap-notice` workflow that comments on a PR when another currently-open PR touches at least one of the same file paths.
- Make `validate-surface.mjs` and `validate-schemas.mjs` include the allowed enum values (and the offending value) in schema-mismatch error messages instead of ajv's bare "must be equal to one of the allowed values".

## What Changed

- `.github/workflows/pr-overlap-notice.yml`: new workflow, triggered on `pull_request: [opened, synchronize, reopened]`, with only `pull-requests: write` permission. It fetches this PR's changed files and every other open PR's changed files via `gh api` (capped at 200 other PRs, paginated 100/page), computes the path intersection, and upserts a single HTML-comment-marked comment (`<!-- overlap-notice:v1 -->`) listing any overlapping PR(s) and shared path(s). If a prior overlap-notice comment exists and the overlap has since cleared, the comment is updated to say so instead of left stale; if there was never a comment and there's no overlap, nothing is posted. All PR-file-list inputs come from the GitHub REST API (already-public PR metadata) — no contributor code is checked out or executed, matching `validate.yml`'s "Compute submitted files" trusted-workflow pattern. It fails open (skips silently) on any API error or rate limiting rather than failing the job, and is not a required check.
- `scripts/validate-surface.mjs` / `scripts/validate-schemas.mjs`: for ajv `enum`-keyword errors, append the joined `allowedValues` list and the offending value (resolved from the document via the error's `instancePath`) to the error message. Every other error keyword's message is unchanged.
- `tests/validate-error-messages.test.mjs` (new): runs both validator scripts against a deliberately-invalid `kind` and asserts the allowed-values list and offending value now appear in the output.

## Registry Safety

- [x] Public API/OpenAPI/schema changes are intentional and documented. (N/A — no OpenAPI/schema/contract changes in this PR.)

## Validation

- [x] `npm run lint`
- [x] `npm run format:check`
- [x] `npm run validate`
- [x] `npm run validate:surface`
- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/validate.yml'))"` and the same for the new `pr-overlap-notice.yml`
- [x] `actionlint .github/workflows/validate.yml .github/workflows/pr-overlap-notice.yml`
- [x] `git diff --check`
- [x] `npx vitest run tests/validate-error-messages.test.mjs`
- [x] Reproduced both bugs first (set a surface's `kind` to an invalid value, ran both scripts, confirmed the bare ajv message), then confirmed the fix on the same fixture, then reverted.
- [x] Confirmed via `vitest.config.mjs`'s `coverage.include` that `validate-surface.mjs`/`validate-schemas.mjs` are out of the Codecov-gated scope, so `test:coverage` was not required for this change.

### DIFF
--- a/.github/workflows/pr-overlap-notice-comment.yml
+++ b/.github/workflows/pr-overlap-notice-comment.yml
@@ -88,13 +88,24 @@ jobs:
             exit 0
           }
 
+          # Build the request body as JSON via jq --arg (which JSON-encodes the
+          # value regardless of content -- quotes, backticks, `$(...)`-shaped
+          # text from a PR-controlled file name, anything) and pipe it in with
+          # `gh api --input -`, rather than `-f body="$body"`. Empirically
+          # verified (mocked-argv test) that `-f body="$var"` already passes
+          # the value through as a single opaque argument with no shell
+          # re-parsing of its contents -- this is not fixing a reachable
+          # exploit -- but --input - is the more obviously-safe idiom and
+          # removes any doubt for a future reader or scanner.
           if [ "$has_overlap" = "true" ]; then
             if [ -n "$existing_comment_id" ] && [ "$existing_comment_id" != "null" ]; then
-              gh api --method PATCH "repos/$REPO/issues/comments/$existing_comment_id" \
-                -f body="$body" >/dev/null 2>&1 || echo "Failed to update overlap comment; skipping." >&2
+              jq -n --arg body "$body" '{body: $body}' \
+                | gh api --method PATCH "repos/$REPO/issues/comments/$existing_comment_id" --input - \
+                >/dev/null 2>&1 || echo "Failed to update overlap comment; skipping." >&2
             else
-              gh api --method POST "repos/$REPO/issues/$pr_number/comments" \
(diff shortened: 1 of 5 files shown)


JSONbored:
Note: the \`test\` CI job is currently failing on an unrelated, pre-existing bug in \`tests/chain-analytics.test.mjs\` ("GET /api/v1/chain/fees returns daily series + top payers, COALESCEs NULL fees", asserting \`median_fee_tao === 0.006\` but getting \`null\`). Reproduced deterministically on a clean checkout of \`origin/main\` at the commit this branch is based on — confirmed unrelated to this PR's changes (which touch only \`.github/workflows/pr-overlap-notice.yml\`, \`scripts/validate-surface.mjs\`, \`scripts/validate-schemas.mjs\`, and \`tests/validate-error-messages.test.mjs\`). Likely related to open PR #2675 (bound scoped chain-fee median scans) or the same D1 numeric-string-coercion bug class fixed repeatedly elsewhere recently. This will need to be resolved on main independently of this PR.

PR #2800, opened 2026-07-03
test(domain-tags): add per-rule deriveDomainTags branch coverage (#2570)

### LINKED ISSUE #2570: Add per-rule branch tests for src/domain-tags.mjs deriveDomainTags

petermilord:
## Summary

- Extend `tests/domain-tags.test.mjs` so every `DOMAIN_TAG_RULE` fires at least one positive keyword match (agents through training).
- Add input-handling regression tests: non-string `description`/`additional` are dropped, null-only input yields `[]`, non-array `categories` is ignored, and non-vocabulary category strings are skipped.
- Assert sorted de-duplicated output when the same tag is derived from both on-chain text and a curated category.

Closes #2570

## Test plan

- [x] `npx vitest run tests/domain-tags.test.mjs` — 13 tests green

### DIFF
--- a/tests/domain-tags.test.mjs
+++ b/tests/domain-tags.test.mjs
@@ -118,4 +118,70 @@ describe("deriveDomainTags", () => {
     assert.deepEqual(first, ["compute", "media"]);
     assert.equal(first.length, new Set(first).size);
   });
+
+  test("fires at least one keyword per DOMAIN_TAG_RULE", () => {
+    const cases = [
+      ["agents", "autonomous agent network"],
+      ["compute", "GPU cuda compute network"],
+      ["data", "web scraping datasets pipeline"],
+      ["finance", "defi trading and liquidity"],
+      ["inference", "llm text-generation completion"],
+      ["media", "text-to-speech and computer vision"],
+      ["prediction", "probabilistic forecast markets"],
+      ["privacy", "zero-knowledge proof privacy"],
+      ["robotics", "embodied ai robotics drones"],
+      ["science", "protein folding drug discovery"],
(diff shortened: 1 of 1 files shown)


JSONbored (commented):
Fix this, please, and resubmit as a new PR: 

```
The main weakness is that the per-rule coverage matrix is hand-maintained and does not prove it still covers every exported domain tag after future vocabulary changes.
```

PR #2802, opened 2026-07-03
fix(subnet-performance): harden captured_at stamping (string cells + RangeError)

jeffrey701:
## Summary
`captureStamp` had two gaps versus the `concentration.mjs` version it mirrors:
- The number branch called `new Date(value).toISOString()` with no range guard, so a finite but out-of-range epoch (`|ms| > 8.64e15`) throws a **RangeError** and 500s the whole performance route on one corrupt `captured_at` cell.
- A D1 INTEGER `captured_at` returned as a numeric string was fed to `Date.parse`, which returns `NaN` for a bare digit run, silently dropping a real timestamp.

Routes `captured_at` through an `epochMsStamp` helper (finite + positive + `getTime()` range check) and sends numeric-string cells down the same epoch path, matching `concentration.mjs`.

## Validation
- `npm test -- tests/subnet-performance.test.mjs` — green, incl. string-cell + out-of-range tests
- `npx prettier --check` + `npx eslint` — clean

### DIFF
--- a/src/subnet-performance.mjs
+++ b/src/subnet-performance.mjs
@@ -30,14 +30,30 @@ function round(value) {
   return Math.round(value * factor) / factor;
 }
 
+// Stamp a positive, in-range epoch-ms as { ms, value:ISO }, or null. Guarding
+// getTime() matters: a finite but out-of-range epoch (|ms| > 8.64e15, the JS
+// Date limit) makes toISOString() throw a RangeError, which would 500 the whole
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
Merge conflicts, closing.

```
Use the [web editor](https://github.com/JSONbored/metagraphed/pull/2802/conflicts) or the command line to resolve conflicts before continuing.

src/subnet-performance.mjs
```

PR #2804, opened 2026-07-03
fix(subnet-performance): coerce D1 numeric-string captured_at to ISO

ultrahighsuper:
## What

`captureStamp` in `src/subnet-performance.mjs` only produced a timestamp for a `number` `captured_at`; for a string it relied on `Date.parse`. But D1 returns the INTEGER `captured_at` column as a **numeric string**, and `Date.parse("1750000000000")` is `NaN` — so a numeric-string cell silently dropped the timestamp to `null`.

## Impact

`captured_at` was lost (returned `null`) whenever D1 handed back the epoch-ms as a string on:

- `GET /api/v1/subnets/{netuid}/performance`
- the `get_subnet_performance` MCP tool

both of which flow through `buildSubnetPerformance` → `captureStamp(row?.captured_at)`.

## Fix

Match the all-digit form and coerce it as epoch-ms before falling back to `Date.parse` for ISO text, guarding the constructed `Date` against out-of-range values — mirroring the canonical `captureStamp` / `epochMsStamp` already used in the sibling `src/concentration.mjs`.

## Test

Adds a regression to `tests/subnet-performance.test.mjs` asserting a D1 numeric-string `captured_at` (`"1750000000000"`) stamps to its ISO value instead of `null`. Fails before the fix (`Date.parse` → `NaN` → `null`), passes after. Existing number-input and ISO-string cases are unchanged.

Robustness-only; the `captured_at` output type (`["string","null"]`) is unchanged, so no contract/schema regeneration is required.

### DIFF
--- a/src/subnet-performance.mjs
+++ b/src/subnet-performance.mjs
@@ -30,14 +30,26 @@ function round(value) {
   return Math.round(value * factor) / factor;
 }
 
+function epochMsStamp(ms) {
+  if (!Number.isFinite(ms) || ms <= 0) return null;
+  const date = new Date(ms);
+  if (!Number.isFinite(date.getTime())) return null;
+  return { ms, value: date.toISOString() };
+}
+
 function captureStamp(value) {
-  if (typeof value === "number" && Number.isFinite(value)) {
-    return { ms: value, value: new Date(value).toISOString() };
-  }
+  // D1 returns the INTEGER captured_at as a numeric string, so match the all-digit
+  // form and coerce it as epoch-ms before falling back to Date.parse for ISO text —
+  // Date.parse("1750000000000") is NaN, which would silently drop the timestamp to
+  // null. Mirrors the canonical captureStamp/epochMsStamp in concentration.mjs.
+  if (value == null) return null;
   if (typeof value === "string") {
+    if (/^\d+$/.test(value)) return epochMsStamp(Number(value));
     const ms = Date.parse(value);
     if (Number.isFinite(ms)) return { ms, value };
+    return null;
   }
+  if (typeof value === "number") return epochMsStamp(value);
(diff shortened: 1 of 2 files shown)


ultrahighsuper:
The flagged blocker references `src/chain-performance.mjs:captureStamp` / `buildChainPerformance`, but that file and those symbols do not exist in this repository:

```
$ git ls-tree -r main --name-only | grep -i chain-performance   # (no output)
$ grep -rn "buildChainPerformance\|chain-performance" src/ workers/   # (no matches)
```

The only performance builder that reads a D1 `captured_at` is `src/subnet-performance.mjs` (fixed here), and its sibling `src/concentration.mjs` already uses the canonical `captureStamp`/`epochMsStamp` pattern this PR mirrors. There is no second, unfixed instance — the two live builders are now behaviorally consistent for a digit-only D1 `captured_at`.

The `codecov/patch` shortfall has been addressed in the follow-up commit, which exercises the non-positive, out-of-range, and non-scalar guard branches. Re-running the gate.

ultrahighsuper:
Correction to my previous comment: `src/chain-performance.mjs` **does** exist — it was added by #2788, which merged at 09:17 UTC, a few minutes before this gate review ran (09:25 UTC). My earlier `git ls-tree`/`grep` ran against a local mirror fetched before #2788 landed, so it wrongly reported the file as absent. Apologies for the noise.

The reviewer's concern is valid: `chain-performance.mjs:captureStamp` carries the same `Date.parse`-on-digit-string defect this PR fixes in `subnet-performance.mjs`. I'm addressing it in a dedicated follow-up PR so each change stays focused; this PR remains scoped to the subnet-level builder.

JSONbored (approved):

PR #2808, opened 2026-07-03
fix(performance): share hardened captured_at captureStamp for subnet + chain routes

petermilord:
## Summary

D1 can return INTEGER `captured_at` as numeric strings (`"1750000060000"`). The old `captureStamp` in both performance modules only handled finite numbers and ISO strings via `Date.parse`, so real snapshot timestamps were dropped to `null`. Out-of-range epoch-ms cells could also throw a **RangeError** and 500 the route.

`concentration.mjs` was hardened in #2725, but `subnet-performance.mjs` and the sibling `chain-performance.mjs` (#2788) still duplicated the pre-fix helper.

## What changed

- `src/capture-stamp.mjs` — shared `captureStamp`: coerce numeric strings, require `n > 0`, guard `date.getTime()` so invalid/out-of-range cells degrade to `null`.
- `src/concentration.mjs`, `src/subnet-performance.mjs`, `src/chain-performance.mjs` — import the shared helper (no more three divergent copies).
- `tests/capture-stamp.test.mjs`, `tests/subnet-performance.test.mjs`, `tests/chain-performance.test.mjs` — regression coverage for string epoch-ms coercion and invalid-cell rejection.

Affects:
- `GET /api/v1/subnets/{netuid}/performance`
- `GET /api/v1/chain/performance`

No schema change — `captured_at` remains `{"type":["string","null"],"format":"date-time"}`.

## No linked issue

Inline discovery during the D1 coercion audit (#2725 class). The gate flagged the partial fix; this closes the sibling `chain-performance` path in the same PR.

## Distinct from open work

| PR | Relationship |
|----|--------------|
| **#2809** | Same chain-performance fix — superseded by the shared helper here; close to avoid duplicate |
| **#2791** / **#2792** | extrinsics/feeds `toIso` guards — different tiers |

## Test plan

- [x] `npx vitest run tests/capture-stamp.test.mjs tests/subnet-performance.test.mjs tests/concentration.test.mjs tests/chain-performance.test.mjs`
- [x] eslint + prettier clean on changed files

### DIFF
--- a/src/capture-stamp.mjs
+++ b/src/capture-stamp.mjs
@@ -1,0 +1,28 @@
+// Coerce a D1 captured_at cell (epoch-ms number/string or ISO string) to the
+// `{ ms, value }` shape performance/concentration builders use when picking the
+// newest snapshot stamp. D1 can return INTEGER columns as numeric strings; an
+// out-of-range epoch-ms must degrade to null (never a RangeError). Shared by
+// concentration.mjs (#2725), subnet-performance.mjs, and chain-performance.mjs.
+
+function epochMsStamp(ms) {
+  if (!Number.isFinite(ms) || ms <= 0) return null;
+  const date = new Date(ms);
+  if (!Number.isFinite(date.getTime())) return null;
+  return { ms, value: date.toISOString() };
+}
+
+export function captureStamp(value) {
+  if (value == null) return null;
+  if (typeof value === "string") {
+    if (/^\d+$/.test(value)) {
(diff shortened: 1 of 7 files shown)


JSONbored (commented):
Merge conflicts, closing. 

```
Use the [web editor](https://github.com/JSONbored/metagraphed/pull/2808/conflicts) or the command line to resolve conflicts before continuing.

src/subnet-performance.mjs
tests/subnet-performance.test.mjs
```

PR #2816, opened 2026-07-03
fix(chain-performance): coerce D1 numeric-string captured_at to ISO

ultrahighsuper:
## What

`captureStamp` in `src/chain-performance.mjs` (added in #2788) only produced a timestamp for a `number` `captured_at`; for a string it relied on `Date.parse`. But D1 returns the INTEGER `captured_at` column as a **numeric string**, and `Date.parse("1750000000000")` is `NaN` — so a numeric-string cell silently dropped the network snapshot timestamp to `null`.

## Impact

`captured_at` was lost (returned `null`) whenever D1 handed back the epoch-ms as a string on:

- `GET /api/v1/chain/performance`
- the `get_chain_performance` MCP tool

both of which flow through `buildChainPerformance` → `captureStamp(row?.captured_at)`.

## Fix

Match the all-digit form and coerce it as epoch-ms before falling back to `Date.parse` for ISO text, guarding the constructed `Date` against out-of-range values — mirroring the canonical `captureStamp` / `epochMsStamp` in `src/concentration.mjs`. This is the same coercion just applied to the sibling per-subnet builder in #2804; the two performance builders are now behaviorally identical for a digit-only D1 `captured_at`.

## Test

Adds regressions to `tests/chain-performance.test.mjs`: a D1 numeric-string `captured_at` (`"1750000000000"`) stamps to its ISO value instead of `null` (fails before the fix), plus guard-branch coverage for non-positive (`"0"`), finite-but-out-of-range (`"8640000000000001"` → no `RangeError`), and non-scalar inputs. Existing number-input and ISO-string cases are unchanged.

Robustness-only; the `captured_at` output type (`["string","null"]`) is unchanged, so no contract/schema regeneration is required.

### DIFF
--- a/src/chain-performance.mjs
+++ b/src/chain-performance.mjs
@@ -29,14 +29,26 @@ function round(value) {
   return Math.round(value * factor) / factor;
 }
 
+function epochMsStamp(ms) {
+  if (!Number.isFinite(ms) || ms <= 0) return null;
+  const date = new Date(ms);
+  if (!Number.isFinite(date.getTime())) return null;
+  return { ms, value: date.toISOString() };
+}
+
 function captureStamp(value) {
-  if (typeof value === "number" && Number.isFinite(value)) {
-    return { ms: value, value: new Date(value).toISOString() };
-  }
+  // D1 returns the INTEGER captured_at as a numeric string, so match the all-digit
+  // form and coerce it as epoch-ms before falling back to Date.parse for ISO text —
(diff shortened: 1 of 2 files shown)


ultrahighsuper:
Closing as a duplicate — #2809 landed the identical `captureStamp`/`epochMsStamp` coercion for `chain-performance.mjs` and has already merged, so this change is fully superseded. Thanks to @jony376 for the fix. (The per-subnet sibling was addressed separately in #2804.)

PR #2820, opened 2026-07-03
fix(analytics): coerce string-typed D1 netuid at Map-key sites

real-venus:
## Summary

D1 hands INTEGER columns back as numeric strings on GROUP BY / JOIN read paths (the documented convention `account-events.mjs` and `analytics-routes.mjs` coerce for: *"a string-typed row netuid must key on 7 or the tier drops to null"*). Three sites keyed a `Map` on the raw string netuid while callers looked it up with an integer, so the lookups silently missed:

1. **`src/subnet-identity-history.mjs` `latestIdentityHashes`** — the `recordSubnetIdentityChanges` dedup guard never fires, so the hourly identity-change cron re-inserts an identical row for **every subnet every run**, growing the append-only `subnet_identity_history` table **unbounded**.
2. **`src/subnet-identity-history.mjs` `loadPreviouslyKnownAsForNetuids`** — `previously_known_as` is **never attached** to agent-catalog subnets (caller `aliasMap.get(int)` misses the string key) **and** the current name **leaks** into the alias list.
3. **`src/analytics-live.mjs` `growthRowsFromSamples`** — the fastest-growing leaderboard loses each entry's slug/name metadata (integer-keyed `subnetMeta` join misses) and emits a string netuid.

## Fix

Coerce every D1-row netuid to a non-negative integer at these Map-key/emit sites, matching the existing convention. Regression tests cover the string-netuid path plus malformed cells.

```
npm run lint
npx vitest run tests/subnet-identity-history.test.mjs tests/analytics-live.test.mjs
```

No schema/route/contract change → no artifact regeneration.

### DIFF
--- a/src/analytics-live.mjs
+++ b/src/analytics-live.mjs
@@ -64,7 +64,13 @@ export function profilesProjectionFromRows(profiles) {
 export function growthRowsFromSamples(growthSamples) {
   const growthByNetuid = new Map();
   for (const row of growthSamples || []) {
-    const entry = growthByNetuid.get(row.netuid) || {
+    // D1 can hand the INTEGER netuid back as a numeric string on this GROUP BY
+    // read path; the emitted netuid keys the integer-keyed subnetMeta map in
+    // formatLeaderboards, so a raw string netuid drops the fastest-growing
+    // entry's slug/name metadata and violates the integer netuid contract.
+    const netuid = Number(row.netuid);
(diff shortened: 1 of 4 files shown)


JSONbored (commented):
Blocker: 

```
The main maintainability concern is that the new coercion helpers accept a few nonnumeric-looking values such as empty string, null, or false as netuid 0 because they call Number(value) directly.
```

PR #2832, opened 2026-07-03
fix(ci): recompute PR overlap before commenting

JSONbored:
### Motivation
- Prevent a confused-deputy vulnerability where a privileged `workflow_run` job trusted an artifact produced by an untrusted `pull_request` run to select the target PR and comment body. 
- Ensure the repository write token cannot be induced to post or patch comments using PR-controlled artifact contents. 
- Keep the advisory overlap notice behavior while moving trust decisions into the privileged context.

### Description
- Removed the producer workflow's artifact upload of `overlap-result` so the privileged job no longer consumes untrusted JSON. 
- Changed the consumer (`workflow_run`) job to derive the target PR from `github.event.workflow_run.pull_requests[0].number` and only run when that metadata exists. 
- Recomputed changed-file overlap inside the privileged job using public GitHub API calls (`gh api ... /pulls/<n>/files` and listing open PRs) instead of reading `result.json`. 
- Kept the same upsert behavior for the marker comment but now build the comment body from recomputed overlap data and use `jq --arg` + `gh api --input -` to post/patch safely.

### Testing
- Ran `npm run validate:workflows` which validated workflow YAMLs successfully. 
- Ran `git diff --check` to ensure no whitespace or merge leftovers and it passed. 
- Ran `npm run format:check -- .github/workflows/pr-overlap-notice.yml .github/workflows/pr-overlap-notice-comment.yml` and Prettier checks passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a47464c1f18832ca290439343192f01)

### DIFF
--- a/.github/workflows/pr-overlap-notice-comment.yml
+++ b/.github/workflows/pr-overlap-notice-comment.yml
@@ -4,20 +4,16 @@ name: PR Overlap Notice Comment
 # full rationale). Triggered by `workflow_run`, which -- unlike
 # `pull_request_target` -- is NEVER directly reachable by a PR author's event;
 # it only fires when a workflow run completes on this repo's own trusted ref.
-# This is the only place `pull-requests: write` is granted. It downloads the
-# detection result artifact (pr_number, checked, has_overlap, body -- a plain
-# JSON data file, never executed as code) and performs the actual comment
-# upsert. The body text ultimately derives from other-PRs' own file names,
-# which are already fully public in each PR's own diff; nothing here discloses
-# anything not already visible, and `gh api -f body=...` passes it as a
-# properly-encoded API field value, never as shell/script text.
+# This is the only place `pull-requests: write` is granted. It does not trust
+# artifacts or other outputs from the unprivileged PR run; instead, it derives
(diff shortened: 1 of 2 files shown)


JSONbored:
Closing — #2783 (merged 2026-07-03T05:36:41Z) already reverted the entire pr-overlap-notice feature this PR modifies (both .github/workflows/pr-overlap-notice.yml and pr-overlap-notice-comment.yml are deleted on main, confirmed by the modify/delete conflict on rebase). The maintainer's own reasoning for that revert: this workflow duplicated Gittensory's own overlap-detection job, only ever helped this one repo, and cost real ongoing security-hardening burden — the actual fix (populating `changedFiles` on Gittensory's own PR records) belongs in Gittensory itself, not here.

Merging this would resurrect files that were deliberately removed. The Gittensory-flagged bash-quoting blocker (unescaped backticks in the new pr-overlap-notice-comment.yml causing command substitution) was a real, legitimate finding on the code as it existed in this PR — but it's moot now that the whole file is gone rather than something to fix forward.

PR #2896, opened 2026-07-03
feat(api): add GET /api/v1/chain/turnover network-wide churn endpoint

ultrahighsuper:
## What

Adds `GET /api/v1/chain/turnover` — a **network-wide validator-set & registration turnover (churn)** endpoint aggregated across **every** subnet's `neuron_daily` rows between a window's two boundary snapshots. It is the network analog of the per-subnet `GET /api/v1/subnets/{netuid}/turnover`, and the churn companion to `/api/v1/chain/performance` — mirroring that endpoint's wiring pattern exactly.

Also adds the `get_chain_turnover` MCP tool (REST parity).

## Response

```jsonc
{
  "schema_version": 1,
  "window": "30d", "start_date": "…", "end_date": "…",
  "comparable": true,
  "subnet_count": 2,
  "validators_start": 3, "validators_end": 3,
  "validators_entered": 1, "validators_exited": 1,
  "validator_retention": 0.5,     // Jaccard over validator identities
  "neurons_start": 4, "neurons_end": 4,
  "uids_deregistered": 1,
  "neuron_retention": 0.6,        // Jaccard over neuron identities
  "stability_score": 55           // 0–100 composite
}
```

## Semantics

Identities are **netuid-scoped** so the same uid/hotkey on different subnets never collides:
- validator = `${netuid}:${hotkey}` (permit holders) — a hotkey validating on two subnets counts once **per subnet**
- neuron = `${netuid}:${uid}:${hotkey}`
- UID slot = `${netuid}:${uid}` (a slot present at both boundaries with a changed hotkey = a deregistration)

The Jaccard retention math, the anti-overstatement `round()` clamp (a set that actually churned can never report a flawless `1`), and the `stability_score` composite are reused verbatim from the per-subnet turnover tier. `subnet_count` counts distinct netuids, guarding blank/non-integer/negative cells. Null-safe: a cold store, unresolvable boundary dates, or a single snapshot yields a schema-stable `comparable:false` block (never a 404, never throws) — matching the sibling analytics tiers.

## Validation

- New `src/chain-turnover.mjs` (pure builder + loader) + `get_chain_turnover` MCP tool; REST route, contract, and response schema wired like `chain/performance`.
- `npm run build` regenerates the OpenAPI/type artifacts; `validate:contract-drift` passes.
- `tests/chain-turnover.test.mjs` — 21 tests: multi-subnet churn (hand-verified arithmetic), netuid-scoped keying (no cross-subnet collision), shared-hotkey-per-subnet, distinct `subnet_count` with the blank/non-integer/negative guard, the anti-overstatement + `round()` clamps, the loader's no-`netuid`-filter boundary SELECT, and the cold/empty + single-snapshot cases. Route tests cover 200 summaries, cold→empty, and 400s for an unknown param / invalid window.
- Read-only over the existing `neuron_daily` rollup; no schema/migration or ingest changes.


ultrahighsuper:
Addressed the flagged blocker: the diff had swept in unrelated schema-drift snapshot churn under `public/metagraph/schemas/index.json` (volatile registry-derived surface entries, not part of this endpoint). I've restored that file to `main`, so the PR now carries only the chain/turnover contract artifacts (openapi/types/api-index/contracts + the generated `.d.ts`/client). `validate:contract-drift` still passes (it does not read `schemas/index.json`), and codecov/patch is now at 100% on the new file. Re-running the gate.

ultrahighsuper:
Closing as a duplicate — #2835 landed `GET /api/v1/chain/turnover` (network validator turnover) and has already merged, so this route now exists on `main` and this PR can no longer be merged without colliding with it. Thanks to @GildardoDev; the network-wide turnover endpoint is covered.

PR #2943, opened 2026-07-04
feat(mcp): add get_curation_states tool mirroring the curation route

jeffrey701:
Exposes the network-wide curation-state report (`GET /api/v1/curation`) as an MCP tool so an agent can see, across the whole registry, which subnets are maintainer-reviewed vs merely observed and how well each is curated. For every subnet it returns netuid/slug/name, resolved coverage_level, surface_count and candidate_count, the nested Gaps object (missing/supported surface kinds), and the curation metadata block (level, review_state, reviewed_at/verified_at, source_count, gap_notes).

Self-contained: straight `loadArtifactData` read of the curation artifact (no D1, no schema regen). Registers the tool + a typed output schema, a build-gated `validate-mcp` assertion (happy path when staged, `not_found` guard when cold), the MCP server/registry version bump to 1.25.0 with all pinned version assertions updated, and unit tests for the populated and missing-artifact paths.

`npm run validate:mcp` passes (89 tools); full mcp-server + version-pin suites green.

### DIFF
--- a/scripts/validate-mcp.mjs
+++ b/scripts/validate-mcp.mjs
@@ -301,6 +301,25 @@ if (existsSync(gapsArtifactPath)) {
   );
 }
 
+// curation.json is an R2-only artifact staged into the cold env only after
+// `npm run build`; exercise the happy path when present, else the not_found
+// guard loadArtifactData maps.
+const curationPath = artifactFilePath("curation.json");
+if (existsSync(curationPath)) {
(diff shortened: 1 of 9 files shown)


JSONbored (commented):
Rejecting: 

```
The main remaining issue is contract precision: several nested array fields are exposed as untyped arrays, which weakens client guidance but does not break runtime behavior.
```

PR #2951, opened 2026-07-04
fix(block-subresources): coerce D1 string block_number in resolveBlockNumber

joaovictor712:
## Summary

Coerce the D1-resolved `block_number` in `resolveBlockNumber` so the block sub-resource envelopes emit an integer, not a string.

Some D1 read paths return the INTEGER `blocks.block_number` as a string. `resolveBlockNumber` returned that cell raw, and the height flows straight into the top-level `block_number` of both `buildBlockExtrinsics` and `buildBlockEvents` — so `GET /api/v1/blocks/{ref}/extrinsics` and `/events` (and the MCP block-explorer tools) could emit `block_number` as a JSON string while the nested `events[*].block_number` are coerced to integers by `formatAccountEvent`/`formatExtrinsic`. The result is an internally inconsistent, off-contract payload.

## What Changed

- `resolveBlockNumber` routes the resolved cell through the file's existing `strictBlockNumber` helper (accepts a real integer or an all-digits string, rejects anything else to `null`), so both builders always emit an integer or `null` — matching the "unknown ref → null block_number" shape. No response-shape or OpenAPI change.
- Regression tests: a string-typed `block_number` from both the numeric-ref and hash-ref reads now resolves to the integer `4200000` (`typeof === "number"`); the existing numeric-cell and missing-row cases still hold.

## Registry Safety

- [x] Public API/OpenAPI/schema changes are intentional and documented. (None — ordering/typing only.)

## Validation

- [x] `npx vitest run tests/block-subresources.test.mjs`
- [x] `npm test` (full suite)
- [x] `npm run lint` · `npm run format:check`
- [x] `git diff --check`

### DIFF
--- a/src/block-subresources.mjs
+++ b/src/block-subresources.mjs
@@ -27,7 +27,14 @@ export async function resolveBlockNumber(d1, ref) {
       : `SELECT block_number FROM blocks WHERE block_number = ? LIMIT 1`,
     [isHash ? String(ref).toLowerCase() : refBlockNumber],
   );
-  return rows[0]?.block_number ?? null;
+  // Coerce the resolved cell: some D1 read paths return the INTEGER block_number
+  // as a string, and this height flows straight into the envelope block_number of
+  // both buildBlockExtrinsics/buildBlockEvents. Without coercion the top-level
+  // field leaks as a JSON string while the nested rows (formatAccountEvent /
+  // formatExtrinsic) coerce theirs — an inconsistent, off-contract payload.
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
Merge conflicts: 

```
Use the [web editor](https://github.com/JSONbored/metagraphed/pull/2951/conflicts) or the command line to resolve conflicts before continuing.

src/block-subresources.mjs
tests/block-subresources.test.mjs
```

PR #2959, opened 2026-07-04
fix(api): sum account portfolio totals in exact rao space

ultrahighsuper:
## What

`buildAccountPortfolio` accumulated a wallet's cross-subnet stake and emission with float `+=`:

```js
totalStake += stake;
totalEmission += emission;
```

Even when each per-subnet value is itself rao-exact, summing in float TAO space compounds IEEE-754 error across the positions. A wallet registered across many subnets can therefore report `total_stake_tao` / `total_emission_tao` — and the derived `overall_yield` — off by several rao from the true integer-rao total (e.g. 100 positions of `123456.789012345` TAO sum to `12345678.90123447` in float vs the exact `12345678.9012345`).

## Fix

Accumulate in integer-rao BigInt space and convert back to TAO once at the end, mirroring the same treatment already in `concentration.mjs` and `counterparties.mjs`:

```js
totalStakeRao += toRaoBig(stake);
totalEmissionRao += toRaoBig(emission);
// ...
const totalStake = raoBigToTao(totalStakeRao);
```

`toRaoBig` clamps a non-finite scaled value to `0n`, so an overflowing stake stays finite JSON instead of throwing on `BigInt(Infinity)`.

## Tests

Two regression tests added:
- 100 positions summing to a value the naive float `+=` drifts on — asserts the rao-exact total and that the float sum differs.
- An overflowing (`Number.MAX_VALUE`) stake stays finite (`total_stake_tao === 0`) rather than throwing.

Source + tests only — no schema/contract/generated-artifact changes. `eslint` clean, `prettier` clean, 12/12 tests pass, 100% patch coverage.

### DIFF
--- a/src/account-portfolio.mjs
+++ b/src/account-portfolio.mjs
@@ -20,6 +20,20 @@ function round9(value) {
   return Math.round(value * SCALE) / SCALE;
 }
 
+// Sum in rao-integer BigInt space, not float space -- accumulating the wallet's
+// per-subnet stake/emission with plain `+=` compounds rounding error across the
+// positions even when each value is itself rao-exact, so convert each addend to
+// integer rao, sum the integers, and convert back once at the end. Mirrors the
+// toRaoBig/raoBigToTao pattern in concentration.mjs / counterparties.mjs (#2933).
+function toRaoBig(tao) {
+  const rao = Math.round(tao * SCALE);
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
Rejecting: 

```
The overflow clamp is intentionally conservative for non-finite scaled values, but the helper behavior could be documented more tightly because it silently drops finite TAO inputs that overflow when scaled.
```

PR #3015, opened 2026-07-04
fix(api): ignore invalid block summary timestamps

JSONbored:
### Motivation
- The blocks-summary shaping could call `new Date(...).toISOString()` on unsafe or out-of-range integer cells from the D1 `blocks` tier, which throws `RangeError` and makes the summary endpoint/tool fail instead of returning a schema-stable card.

### Description
- Tighten integer coercion by making `toInt` accept only non-negative `Number.isSafeInteger` values. 
- Add `MAX_DATE_MS` and `toTimestampMs` to treat timestamp cells outside JavaScript `Date`'s valid epoch-ms range as absent. 
- Use `toTimestampMs` for the `observed_at` field when shaping blocks so invalid timestamps are dropped rather than formatted. 
- Add a regression test in `tests/blocks-summary.test.mjs` that verifies unsafe integer cells and out-of-range `observed_at` values are ignored and do not crash `buildBlocksSummary`.

### Testing
- Ran `npm test -- tests/blocks-summary.test.mjs` and the test file passed. 
- Ran `npm run lint` and `npm run format:check` and both checks passed. 
- Ran `npm run build` and the build completed successfully. 
- Ran `npm run validate`; it initially failed because generated provider artifacts were absent until a local `npm run build` was performed, and the final `npm run validate` completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a488d2685e4832c844ce2f6fcf2e484)

### DIFF
--- a/src/blocks-summary.mjs
+++ b/src/blocks-summary.mjs
@@ -24,6 +24,7 @@ export const BLOCKS_SUMMARY_SCAN_CAP = 5000;
 
 const THROUGHPUT_UNAVAILABLE = null;
 const BLOCK_TIME_PERCENTILES = [50, 90];
+const MAX_DATE_MS = 8_640_000_000_000_000;
 
 // Round a duration/mean to whole milliseconds — inter-block time has no meaningful
 // sub-ms precision, and integer ms keeps the JSON clean.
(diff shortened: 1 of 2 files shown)


JSONbored:
Closing as superseded — #2952 (merged 2026-07-04T05:20:11Z, ~3h before this PR) already fixes the identical bug in the identical function (`buildBlocksSummary`'s `observed_at` handling in src/blocks-summary.mjs), via an equivalent approach (`Number.isFinite(new Date(ms).getTime())` there vs. this PR's explicit `MAX_DATE_MS` bound check here), plus its own regression test. Confirmed by diffing both PRs directly, not just by the rebase conflict surfacing it.

Rebasing this PR onto main would just reintroduce a second, functionally-duplicate helper (`toTimestampMs` alongside the already-merged `toTimestamp`) doing the same job — not worth carrying forward. No action needed; the underlying bug is already fixed on main.

PR #3186, opened 2026-07-04
feat(api): add GET /api/v1/accounts/{ss58}/stake-moves per-account footprint

luciferlive112116:
## What

Adds `GET /api/v1/accounts/{ss58}/stake-moves` — a per-account stake-move (re-delegation) footprint.

For one account (hotkey), it reports which subnets it moved staked alpha **out of** over a recent window (`7d` / `30d` / `90d`, default `30d`), broken down per origin subnet with a move count and the first/last move timestamps, plus a roll-up: `total_movements`, `subnet_count`, an HHI `concentration` of where re-delegation activity is focused, and the `dominant_netuid` (the origin subnet with the most moves).

This is the account-level companion of the existing per-subnet and network stake-move leaderboards (`/subnets/{netuid}/stake-moves`, `/chain/stake-moves`), and the re-delegation-side complement of `/accounts/{ss58}/serving` and `/accounts/{ss58}/axon-removals`. It completes the account-dimension family (serving / registrations / deregistrations / prometheus / axon-removals / stake-moves).

## How

- New pure shaper + thin D1 loader in `src/account-stake-moves.mjs`, keyed on `hotkey` via `idx_account_events_hotkey`, filtering `event_kind = 'StakeMoved'` (which always carries the moving hotkey; `netuid` is the origin subnet the stake moved from — `fetch-events.py` `_moved` → `[coldkey, hotkey, netuid]`) over the window, grouped per origin subnet.
- Null-safe throughout: a cold store or an empty window yields a schema-stable zeroed card (never a 404, never a throw), matching the sibling account tiers.
- HHI is clamped so a sub-perfect spread never rounds up to an exact `1` (`roundConcentration`, the shared anti-overstatement invariant).
- Wired through the schema (`AccountStakeMovesArtifact`), contracts (artifact + route), config path pattern, artifact-storage allowlist, both validators, the request handler + dispatch, the mainnet gate, and the docs contract table. Derived artifacts regenerated via `npm run build`. Route count 136 → 137.

## Tests

- `tests/account-stake-moves.test.mjs` — shaper + loader unit coverage: empty/null/non-array input, zero-count and malformed/blank/negative netuid rows, null & out-of-range timestamps, duplicate-netuid merge, tie-break ordering, the HHI clamp (extreme-skew), the unknown-window fallback, and the SQL shape (index/where/group-by, param order, epoch-ms cutoff).
- `tests/account-routes.test.mjs` — route-level: 200 success, cold-D1 schema stability (never 404), 400 on unknown query param, 400 on unsupported window.

All source + generated types/clients regenerated by the build; full validator and public-safety gates pass.


JSONbored (commented):
Closing, merge conflicts.

PR #3207, opened 2026-07-04
feat(api): add GET /api/v1/subnets/{netuid}/stake-transfers per-subnet drill-in

luciferlive112116:
## What

Adds `GET /api/v1/subnets/{netuid}/stake-transfers` — the per-subnet drill-in of the `chain/stake-transfers` leaderboard (merged in #3192).

For ONE subnet over a `7d`/`30d` window it returns the distinct senders (accounts), the StakeTransferred event count, and the average transfers per sender. The chain route ranks only the top-N subnets and can't be queried by an arbitrary netuid — this fills the same per-subnet/chain duality that serving, prometheus, turnover, concentration, stake-flow, stake-moves, yield, weights, and registrations already have.

It's the between-coldkeys sibling of `subnets/{netuid}/stake-moves` (within-account re-delegation churn): `StakeTransferred` (`transfer_stake`) relocates staked alpha from one account to another on the same hotkey, so it moves ownership, not net capital. The netuid is the origin subnet (origin leg only).

## How

Clones the `subnet-stake-moves` shape (JSON-only summary card, no CSV/normalizer):
- New `src/subnet-stake-transfers.mjs` — pure `buildSubnetStakeTransfers` + thin `loadSubnetStakeTransfers` (single `COUNT(*)` / `COUNT(DISTINCT coldkey)` / `MAX(observed_at)` aggregate filtered by `netuid = ? AND event_kind = 'StakeTransferred'` over the window). Null-safe: a cold store or a subnet with no StakeTransferred events yields a schema-stable zeroed card (never 404).
- Wired through the schema (`SubnetStakeTransfersArtifact`), contracts (artifact + route), config path pattern, artifact-storage allowlist, both validators, the request handler + canonical edge-cache path + dispatch, and the docs contract table. Derived artifacts regenerated via `npm run build`. Route count 138 → 139.

## Tests

- `tests/subnet-stake-transfers.test.mjs` — builder (cold/null card, omitted-window default, computed intensity, 2dp rounding, numeric-string coercion, non-finite/out-of-range/negative guards) + loader (SQL shape, param order, cold-store zeroed card, window map).
- `tests/request-handlers-entities.test.mjs` — handler 400s (unknown param / unsupported window), cold schema-stable card validated against `SubnetStakeTransfersArtifact`, and the three canonical-cache-path cases.
- `tests/analytics-edge-cache.test.mjs` — worker dispatch + default-window edge-cache key.

Full validator, public-safety, OpenAPI, and docs gates pass locally.


JSONbored (commented):
Please fix these things then resubmit: 

```
The main issues are polish and test precision: the generated example is semantically inconsistent, and the loader test could pin the event/window predicates more tightly.
```

PR #3243, opened 2026-07-05
feat(api): add CSV export for subnet yield/history endpoint

jimcody1995:
## Summary

Adds `?format=csv` (and `Accept: text/csv`) to `GET /api/v1/subnets/{netuid}/yield/history` so the per-day emission-yield distribution series (subnet yield, mean/median/p25/p75/p90) can be downloaded as CSV.

Implementation follows the same pattern as the trajectory, yield, and concentration/history CSV exports:
- handler + edge-cache changes in `workers/request-handlers/entities.mjs`
- OpenAPI/contract artifacts regenerated via `npm run build`
- CSV example isolated in `csv-route-examples.mjs`
- dedicated test file `tests/subnet-yield-history-csv.test.mjs` (avoids contending with open entity-handler PRs on shared harness files)

**10 files changed, 1 commit** — no bundled work from other open PRs. Verified locally that this merges cleanly on top of open PRs #3210, #3232, and #3238 without conflicts.

## Test plan

- [x] `npm run build`
- [x] `npm run validate`
- [x] `npm run lint` + `npm run format:check`
- [x] `tests/subnet-yield-history-csv.test.mjs` — CSV body, Accept header, cold export, JSON envelope, format validation, cache-key variants, OpenAPI CSV example

### DIFF
--- a/generated/metagraphed-api.d.ts
+++ b/generated/metagraphed-api.d.ts
@@ -2289,7 +2289,7 @@ export interface paths {
             path?: never;
             cookie?: never;
         };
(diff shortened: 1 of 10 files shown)


JSONbored (approved):
After this, i think we're good on the CSV exports endpoints, please - a new wave of feature issues will be released in a few hours that I'd like contributors to focus on, please.

PR #3303, opened 2026-07-05
feat(api): add GET /api/v1/chain/stake-transfer-volume network leaderboard

luciferlive112116:
> Recreation of #3296/#3300. Two fixes since #3296:
> 1. **codecov/patch** — the two genuinely-unreachable defensive guards (roundTao non-finite, avgTransferTao zero-transfers) now carry v8-ignore annotations matching the sibling chain-stake-flow; the reachable nullableTao (blank/non-numeric volume) + coerceEpochMs (non-positive observed_at) guards are exercised by test rows. Module patch coverage is 100%.
> 2. **test job (pre-existing main breakage)** — a second commit corrects a stale assertion in : #3290 bumped  to 1.75.0 but left that test pinned to 1.74.0, so main itself was red. Unrelated to this endpoint; kept as its own commit.

## What

Adds `GET /api/v1/chain/stake-transfer-volume` — a network-wide stake-transfer **volume** leaderboard.

Over a `7d`/`30d` window it returns a per-subnet leaderboard (total TAO transferred, transfer count, and average TAO per transfer) ranked by volume, a network rollup with the total transferred TAO, and a distribution summary (count/mean/min/p25/median/p75/p90/max) of the per-subnet volume. `?limit` caps the leaderboard (default 20, max 100); `?format=csv` downloads the per-subnet leaderboard.

This is the **value** companion to the existing count-based `/api/v1/chain/stake-transfers`: that route counts StakeTransferred events and distinct senders per subnet; this sums the TAO **value** moved. The two are orthogonal — many tiny transfers vs a few large ones — and neither is derivable from the other (`/chain/stake-transfers` never sums `amount_tao`). `StakeTransferred` (`transfer_stake`) relocates staked alpha between coldkeys on the same hotkey; the `amount_tao` is the value on the origin leg, so this is inherently an origin-side view.

## How

- New `src/chain-stake-transfer-volume.mjs` — pure `buildChainStakeTransferVolume` + a thin `loadChainStakeTransferVolume` that runs a single `GROUP BY netuid` aggregate (`COALESCE(SUM(amount_tao),0)` + `COUNT(*)` + `MAX(observed_at)`) filtered by `event_kind = 'StakeTransferred'` and the window (volume and count are additive, so the network rollup is the JS sum of the per-subnet rows — no separate distinct-actor query needed).
- Summed TAO is **rao-rounded** (`1 TAO = 1e9 rao`) to avoid IEEE-754 drift, matching the sibling `chain-stake-flow` scorecard. Null-safe throughout: a cold store or an out-of-range `observed_at` yields a schema-stable empty block (never a 404 or a 500).
- Wired through the schema (`ChainStakeTransferVolumeArtifact`), contracts (artifact + CSV route), the analytics handler (window + limit + format, HEAD normalized through the GET cache key), api.mjs dispatch + mainnet gate, artifact-storage allowlist, both validators, csv-route-examples, the OpenAPI-sample normalizer (a consistent worked example so `avg_transfer_tao = volume_tao / transfers` holds), and the docs contract table. Derived artifacts regenerated via `npm run build`. Route count 140 → 141.

## Tests

- `tests/chain-stake-transfer-volume.test.mjs` (26) — builder (volume ranking, network rollup, distribution, tie-break, limit clamp, duplicate-netuid merge, **rao-precision rounding**, null-volume/zero-transfer/malformed-netuid guards, out-of-range `observed_at` → null, empty/non-array), loader (SQL shape + `event_kind` param + cutoff window bounds + cold store), and the full route surface (dispatch, HEAD, cold card, 400s, CSV export/Accept/cold-header/HEAD, edge-cache round-trip).
- `tests/openapi-sample.test.mjs` — the new normalizer keeps `avg_transfer_tao` consistent + the guard branch.

Full validator, public-safety, OpenAPI-sample, MCP-parity, and docs gates pass locally.


luciferlive112116:
wait another min
currently resolving conflicts

luciferlive112116:
if no problem in there, please merge this PR

JSONbored (commented):
Not a focus right now, closing.

JSONbored:
> if no problem in there, please merge this PR

Do not ask for

Issue #1856, opened 2026-06-25
feat(chain): persist the alpha leg on stake events


JSONbored:
## Context

Part of #1345 — first-party developer block explorer.

Staking events carry the TAO leg but not the **alpha leg**. The `_stake` extractor's signature comment names `alpha_rao` (`[coldkey, hotkey, tao_rao, alpha_rao, netuid, ...]`) but never emits it — `extract()` (`scripts/fetch-events.py:532-546`) flattens every event into a fixed 5-key dict (`hotkey/coldkey/netuid/uid/amount_tao`) and the `main()` row-builder drops anything else. So a `StakeAdded` shows `amount_tao` but not how much alpha was received.

## Standard-explorer parallel

A staking/delegation record carries the alpha quantity alongside the TAO amount. (Realized price/slippage needs at-block AMM pool reserves — deferred, see below.)

## Deliverables — Phase 1 only (persist the alpha leg the decoder already sees)

- [ ] **Migration `0020_account_events_alpha.sql`** (idempotent, nullable): `ALTER TABLE account_events ADD COLUMN alpha_amount REAL;`. Apply to prod via `wrangler d1 execute … --file=` (never `migrations apply`). If G7 (`extrinsic_index`) lands near this, **combine both columns into one ALTER PR** to avoid churn.
- [ ] Python — touch **all three** points: `_stake` reads `a[3]` and returns it; `extract()` carries `alpha_amount` through the flattener; the `main()` row dict includes it. Use the existing `_tao()` helper (`alpha_rao/1e9`) so it reads consistently next to `amount_tao`. Only `StakeAdded`/`StakeRemoved` carry an alpha leg; leave null for others (`_moved`, non-stake kinds).
- [ ] `EVENT_INSERT_COLUMNS` (10th col), `ACCOUNT_EVENT_COLUMNS`, `formatAccountEvent` — kept 1:1. **Param-ceiling:** 9→10 cols × `ROWS_PER_STMT=10` = exactly 100 (same shape `extrinsics` already ships; keep `ROWS_PER_STMT=10`).
- [ ] Schema: add `alpha_amount` (`["number","null"]`) to `AccountEvent`; `npm run build` + commit.
- [ ] Python test mirroring the Transfer tests: `extract("StakeAdded", [coldkey, hotkey, tao_rao, alpha_rao, netuid])` returns the alpha leg; a shape-drift case (`len < 4`) → null.

## Requirements

- No new index (display field, not a filter/sort key).
[...]

PR #3326, opened 2026-07-05
feat(api): add GET /api/v1/chain/alpha-flow network leaderboard

luciferlive112116:
## What

Adds `GET /api/v1/chain/alpha-flow` — a network-wide cross-subnet **alpha** (subnet-token) flow leaderboard.

Over a `7d`/`30d` window it ranks every subnet by net alpha **minted** (StakeAdded) minus **burned** (StakeRemoved), with per-subnet in/out/net/gross alpha totals + a direction label (`expanding`/`contracting`/`balanced`), a network rollup with expanding/contracting/flat counts, and a distribution (count/mean/min/p25/median/p75/p90/max) of the per-subnet net alpha flow. `?limit` caps the leaderboard (default 20, max 100); `?format=csv` downloads the per-subnet leaderboard.

This is the **alpha-denominated** companion to `/api/v1/chain/stake-flow`: that route sums the `amount_tao` (TAO capital) leg of each StakeAdded/StakeRemoved swap, this sums the `alpha_amount` (subnet token) leg of the **same** swaps (#1856). Because a subnet's alpha price floats, alpha volume is **not** derivable from the TAO flow — a subnet can take large TAO inflow yet mint little alpha (high price) or vice versa, so net alpha flow is independent information about each subnet's token-supply dynamics.

## How

- New `src/chain-alpha-flow.mjs` — pure `buildChainAlphaFlow` + a thin `loadChainAlphaFlow` that runs a single `GROUP BY netuid, event_kind` aggregate (`COALESCE(SUM(alpha_amount),0)` + `COUNT(*)` + `MAX(observed_at)`) filtered to StakeAdded/StakeRemoved over the window.
- Summed alpha is **rao-rounded** (1 alpha = 1e9 rao) to avoid IEEE-754 drift, matching the sibling `chain-stake-flow`. Null-safe throughout: a cold store, an out-of-range `observed_at`, or a malformed (null/blank/non-numeric/negative) alpha aggregate yields a schema-stable empty/zeroed block (never a 404 or 500). A defensive `windowDays` default keeps a direct caller from producing a NaN cutoff.
- Wired through the schema (`ChainAlphaFlowArtifact`), contracts (artifact + CSV route), the analytics handler (window + limit + format, HEAD normalized through the GET cache key), api.mjs dispatch + mainnet gate, artifact-storage allowlist, both validators, csv-route-examples, and the docs contract table. Derived artifacts regenerated via `npm run build`. Route count 140 → 141.

## Tests

- `tests/chain-alpha-flow.test.mjs` (36) — builder (net-expansion ranking, direction labels incl. gross-0 balanced + churn threshold, network rollup with expanding/contracting/flat, distribution, tie-break, limit clamp, rao rounding, non-stake-kind + malformed/null/blank/negative-netuid + null/blank/non-numeric/negative-alpha guards, non-numeric event_count → 0, out-of-range & non-positive timestamps → null, empty/cold), loader (SQL shape + cutoff window bounds + defensive windowDays default + cold store), and the full route surface (dispatch, HEAD, cold card, 400s, CSV export/Accept/cold-header/HEAD, edge-cache round-trip). 100% statement + branch coverage.

Full validator, public-safety, OpenAPI, MCP-parity, and docs gates pass locally.


JSONbored (commented):
Merge conflicts and polish issues: 

```
the main remaining items are contract/example polish and a small defensive consistency gap in direct-call shaping.
```

PR #3761, opened 2026-07-05
refactor(ui): extract useCopy into ShareButton (#3425)

### LINKED ISSUE #3425: Extract useCopy into ShareButton instead of reimplementing clipboard/toast logic
## Context
`apps/ui/src/components/metagraphed/share-button.tsx` hand-rolls its own clipboard/toast/timer logic inside `ShareButton`'s `onClick`: a local `useState` for `copied`, a raw `try { await navigator.clipboard.writeText(href) } catch` block, direct `toast.success`/`toast.error` calls, and a bare `window.setTimeout(() => setCopied(false), 1400)`. `apps/ui/src/hooks/use-copy.ts` already implements exactly this (clipboard write with an `execCommand` fallback, `copied` state with the same 1400ms auto-reset, and success/error toasts) behind a `{ copied, copy }` return shape, and it's already the shared primitive for every other copy affordance in the app — `key-chip.tsx`, `copy-button.tsx`, `resource-explorer.tsx`, `copyable-code.tsx`, `schema-drift-detail.tsx`, `endpoint-list.tsx`, and `routes/schemas.tsx` all consume it. `ShareButton` is the one holdout, duplicating logic that a bug in `use-copy.ts` (or a UX tweak like the reset delay) would otherwise fix everywhere at once.

## Requirement
Refactor `ShareButton` to call `useCopy()` for its clipboard-write/`copied`-state/toast behavior instead of reimplementing it inline, while preserving `ShareButton`'s two extras that `useCopy` doesn't itself provide: the `url ?? window.location.href` fallback and the visually-hidden `aria-live="polite"` screen-reader announcement span.

## Deliverable
- File to change: `apps/ui/src/components/metagraphed/share-button.tsx` — specifically the `onClick` handler (lines 17-35) and the `useState` calls on lines 14-15.
- Anchor: replace the local `const [copied, setCopied] = useState(false)` plus the manual `navigator.clipboard.writeText` / `toast.success` / `toast.error` / `window.setTimeout(() => setCopied(false), 1400)` block with `const { copied, copy } = useCopy({ toastOnSuccess: false })` (or equivalent), then call `copy(href)` in `onClick`. Keep `toastOnSuccess: false` (or pass a custom toast) since `ShareButton`'s current toast copy — "Filters, sort, and pagination are preserved in the URL." — is distinct from `useCopy`'s generic "Copied to clipboard" message; don't silently change the toast text shown to users. `copy-button.tsx` (lines 19-23: `const { copied, copy } = useCopy({ label }); ... onClick={() => copy(value)}`) is the reference pattern for how a component should consume the hook.
- `announcement` state (line 15, used for the `aria-live="polite"` span at lines 56-59) is `ShareButton`-specific behavior `useCopy` doesn't provide — keep it, driven off `useCopy`'s `copied` return value or off the resolved boolean from calling `copy()`, not reimplemented as a second parallel timer.
- The `href`/`url` fallback logic (line 19: `url ?? (typeof window !== "undefined" ? window.location.href : "")`) and the early return on empty `href` (line 20) stay as-is — `useCopy` takes a plain string, it doesn't resolve the URL itself.
- Done looks like: `share-button.tsx` no longer imports `toast` from `sonner` directly (that import moves to being `use-copy.ts`'s concern only) and no longer calls `window.setTimeout` itself; the public `ShareButton` props (`url`, `label`, `className`) and its 5 existing call sites (`routes/explorer.tsx`, `routes/surfaces.tsx`, `routes/extrinsics.index.tsx`, `routes/subnets.index.tsx`, `routes/blocks.index.tsx`) are unaffected — this is an internal implementation swap, not a prop or markup change.
- No coordination needed with another issue for this swap itself — `use-copy.ts` already exists and is unit-tested (`apps/ui/src/hooks/use-copy.test.ts`) as of this issue; this issue only touches `share-button.tsx`.

## Acceptance criteria
- [ ] `apps/ui/src/components/metagraphed/share-button.tsx` no longer declares its own `copied` `useState`, no longer calls `navigator.clipboard.writeText` directly, and no longer calls `window.setTimeout` to reset a copied flag
- [ ] `apps/ui/src/components/metagraphed/share-button.tsx` imports and calls `useCopy` from `@/hooks/use-copy`
- [ ] The `aria-live="polite"` screen-reader announcement (`announcement` state + `<span role="status" aria-live="polite">`) is preserved and still fires on both success and failure
- [ ] The distinct "Link copied" / "Filters, sort, and pagination are preserved in the URL." success toast copy and the "Couldn't copy link" / "Your browser blocked clipboard access." error toast copy are preserved (not silently replaced by `useCopy`'s generic copy)
- [ ] All 5 existing `ShareButton` usages (`routes/explorer.tsx`, `routes/surfaces.tsx`, `routes/extrinsics.index.tsx`, `routes/subnets.index.tsx`, `routes/blocks.index.tsx`) are untouched — no prop or call-site changes
- [ ] No new eslint/typecheck errors; existing tests (if any touch `ShareButton`) still pass

## Non-goals
- Don't change `ShareButton`'s visible markup, button styling, icon swap (`Check`/`Share2`), or its `url`/`label`/`className` prop surface — this is an internal dedup, not a redesign (no before/after screenshot table needed if the rendered output is unchanged; if any visual pixel does shift, the screenshot-table rule from the `metagraphed` skill applies).
- Don't touch `hooks/use-copy.ts` itself or its existing test file (`hooks/use-copy.test.ts`) — that hook is the target to consume, not to modify here.
- Don't touch the other `useCopy` consumers (`key-chip.tsx`, `copy-button.tsx`, `resource-explorer.tsx`, `copyable-code.tsx`, `schema-drift-detail.tsx`, `endpoint-list.tsx`, `routes/schemas.tsx`) — they already use the shared hook correctly.
- Don't fold in the sibling dedup issues from the same "shared components" batch (#3426 SegmentedToggle, #3427 durationLabel, #3428 CopyIconToggle) — each is its own small PR.

## Size

Part of #2542.

Lang-bt:
Closes #3425

`ShareButton` was the last copy affordance still hand-rolling clipboard write, copied-state timing, and toast logic inline. This swaps it to the shared `useCopy` hook without changing props, markup, or call sites.

## Summary

- Replace local `copied` state + `navigator.clipboard.writeText` + `setTimeout` with `useCopy({ toastOnSuccess: false, resetAfter: 1400 })`
- Preserve ShareButton-specific success toast copy (`"Link copied"` / filters URL description)
- Preserve custom error toast copy and the `aria-live` screen-reader announcement span

## Validation

- [x] `npx eslint src/components/metagraphed/share-button.tsx`
- [x] `npx prettier --check src/components/metagraphed/share-button.tsx`
- [x] `npm run test -- src/hooks/use-copy.test.ts` (apps/ui)

### DIFF
--- a/apps/ui/src/components/metagraphed/share-button.tsx
+++ b/apps/ui/src/components/metagraphed/share-button.tsx
@@ -2,6 +2,7 @@ import { useState } from "react";
 import { Check, Share2 } from "lucide-react";
 import { toast } from "sonner";
 import { classNames } from "@/lib/metagraphed/format";
+import { useCopy } from "@/hooks/use-copy";
 
 interface Props {
   /** Optional explicit URL; defaults to current window.location.href. */
@@ -11,27 +12,26 @@ interface Props {
 }
 
 export function ShareButton({ url, label = "Share view", className }: Props) {
-  const [copied, setCopied] = useState(false);
   const [announcement, setAnnouncement] = useState("");
+  const { copied, copy } = useCopy({ toastOnSuccess: false, resetAfter: 1400 });
 
   const onClick = async () => {
-    try {
-      const href = url ?? (typeof window !== "undefined" ? window.location.href : "");
-      if (!href) return;
-      await navigator.clipboard.writeText(href);
-      setCopied(true);
+    const href = url ?? (typeof window !== "undefined" ? window.location.href : "");
+    if (!href) return;
+    const ok = await copy(href);
+    if (ok) {
       setAnnouncement(`Link copied to clipboard: ${href}`);
       toast.success("Link copied", {
         description: "Filters, sort, and pagination are preserved in the URL.",
       });
-      window.setTimeout(() => setCopied(false), 1400);
       window.setTimeout(() => setAnnouncement(""), 2000);
-    } catch {
-      toast.error("Couldn't copy link", {
-        description: "Your browser blocked clipboard access.",
-      });
-      setAnnouncement("Couldn't copy link to clipboard.");
+      return;
     }
+    toast.dismiss();
+    toast.error("Couldn't copy link", {
+      description: "Your browser blocked clipboard access.",
+    });
+    setAnnouncement("Couldn't copy link to clipboard.");
   };
 
   return (


JSONbored (commented):
Please fix this, then resubmit: 

```
The one behavioral wrinkle is a new toast.dismiss() call before toast.error() on failure that wasn't present in the original code, but since toastOnSuccess is false, no success toast is ever queued, so the dismiss call is a harmless no-op.
```

PR #3775, opened 2026-07-06
fix(test): derive get_health_trends sample day from current date so it stays in the 7d window

e11734937-beep:
The test `get_health_trends aggregates bulk trend rows from D1` fails 100% on `main`, which blocks CI on every open PR.

The mock `surface_uptime_daily` row hard-codes `date: "2026-06-28"`. `buildBulkHealthTrends` keeps only rows with `date >= (now − 7d)` for the 7d window, so once the current date passed 2026-07-05 the row aged out and `windows["7d"].subnet_count` became `0` instead of the asserted `1`.

This derives the sample day from the current date (2 days ago) so it always lands inside the 7d window — the same relative-to-now approach the file already uses for `FRESH_RUN`. No production code changes.

**Verified:** `vitest run tests/mcp-server.test.mjs` → **705 passed** (previously 704 passed / 1 failed).

### DIFF
--- a/tests/mcp-server.test.mjs
+++ b/tests/mcp-server.test.mjs
@@ -10915,6 +10915,11 @@ describe("MCP economics + metagraph data tools", () => {
   });
 
   test("get_health_trends aggregates bulk trend rows from D1", async () => {
+    // #NNNN: derive the sample day from the current date so the row always
+    // lands inside the 7d window (a hard-coded date ages out and fails).
+    const recentDay = new Date(Date.now() - 2 * 86_400_000)
+      .toISOString()
+      .slice(0, 10);
     const env = {
       METAGRAPH_HEALTH_DB: {
         prepare(sql) {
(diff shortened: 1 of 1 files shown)


e11734937-beep:
For visibility on the failing check here: this is currently red on `main` for **every** open PR. The `get_health_trends aggregates bulk trend rows from D1` test hard-codes `date: "2026-06-28"` in its mock row, which has now aged out of the 7d window (`buildBulkHealthTrends` keeps only `date >= now − 7d`), so it fails 100% and the bot auto-closes any PR whose CI is red.

The one-line fix — derive the sample day from `Date.now()` so it always lands inside the window — is on `e11734937-beep:fix/health-trends-test-date` and takes the file to **705/705 green**.

I couldn't keep this PR open to land it: the fork account lacks permission to create an issue, and the maintainer-requires-a-linked-issue gate closes it without one. If you file a tracking issue (or just apply the fix directly), it unblocks CI for all open PRs — happy to resubmit against an issue number.

PR #3793, opened 2026-07-06
fix(ui): reuse durationLabel in IncidentTimeline

### LINKED ISSUE #3427: Reuse durationLabel instead of the local fmtDuration in IncidentTimeline
## Context
`IncidentTimeline` (`apps/ui/src/components/metagraphed/incident-timeline.tsx`, lines 15-22) defines a private `fmtDuration(ms?: number)` helper that hand-rolls the same "compact elapsed label" logic already centralized in `durationLabel(start?, end?)` in `apps/ui/src/lib/metagraphed/format.ts` (lines 88-94). `IncidentCard` (`apps/ui/src/components/metagraphed/incident-card.tsx`, line 27) already calls the shared `durationLabel(i.started_at, i.ended_at)` for the identical "incident duration" case, so `IncidentTimeline` is the odd one out — it's the only incident-duration renderer still computing its own ms diff and formatting it locally instead of reusing the shared helper.

## Requirement
Delete the local `fmtDuration` in `incident-timeline.tsx` and replace its call site with the shared `durationLabel` from `lib/metagraphed/format.ts`, matching the pattern already used by `IncidentCard`.

## Deliverable
- File to change: `apps/ui/src/components/metagraphed/incident-timeline.tsx`
  - Remove the `fmtDuration` function (lines 15-22).
  - Add `durationLabel` to the existing `import { classNames } from "@/lib/metagraphed/format";` (line 7), e.g. `import { classNames, durationLabel } from "@/lib/metagraphed/format";`.
  - Replace the `duration` computation (lines 50-55) — currently `fmtDuration(inc.duration_ms ?? (inc.ended_at && inc.started_at ? new Date(inc.ended_at).getTime() - new Date(inc.started_at).getTime() : undefined))` — with a call to `durationLabel(inc.started_at, inc.ended_at)`. Note the signature difference: `durationLabel(start?: string | null, end?: string | null): string` takes ISO strings directly (which `FlatSurfaceIncident.started_at` / `.ended_at` already are — see `apps/ui/src/lib/metagraphed/types.ts` lines 646/648), so the manual `new Date(...).getTime()` diffing and the `inc.duration_ms` epoch-ms field become unnecessary for this call site.
  - Handle the behavioral difference deliberately: `fmtDuration` returns `null` (hiding the `· {duration}` span entirely, line 73) when there's no computable duration, whereas `durationLabel` always returns a string — `"—"` when `start` is missing, or a live "elapsed so far" label (computed against `Date.now()`) when `end` is omitted for an open incident. Decide and implement one consistent behavior: either keep suppressing the duration segment when `inc.started_at` is absent (guard on `inc.started_at` before rendering, not on the return value), or accept that open incidents now show a running duration instead of no duration — pick the option that matches `IncidentCard`'s existing "ongoing · {durationLabel}" convention (line 27 of `incident-card.tsx`) for consistency across both incident renderers.
- No backend/schema/type changes — `FlatSurfaceIncident` already exposes ISO-string `started_at`/`ended_at`.

## Acceptance criteria
- [ ] `fmtDuration` is fully removed from `apps/ui/src/components/metagraphed/incident-timeline.tsx` (no remaining definition or references)
- [ ] `incident-timeline.tsx` imports `durationLabel` from `@/lib/metagraphed/format` and uses it at the duration call site (in place of the old `fmtDuration(...)` call)
- [ ] The `inc.duration_ms` field and the manual `new Date(inc.ended_at).getTime() - new Date(inc.started_at).getTime()` diff are no longer computed in this file
- [ ] Rendered duration text for a resolved incident (has both `started_at` and `ended_at`) matches what `IncidentCard` would show for the same start/end pair (same helper, same format)
- [ ] Behavior for incidents missing `started_at`, and for open incidents (`ended_at` null), is explicit and intentional (not an accidental side effect of the signature swap) — verified by diff inspection or a manual check against both states
- [ ] `apps/ui/src/components/metagraphed/incident-timeline.tsx` appears in the diff; `apps/ui/src/lib/metagraphed/format.ts` does not need to change

## Non-goals
- Don't modify `durationLabel` or any other exports in `apps/ui/src/lib/metagraphed/format.ts` — it's already correct and already consumed by `IncidentCard`; this issue only removes the duplicate consumer-side helper.
- Don't touch `IncidentCard` (`apps/ui/src/components/metagraphed/incident-card.tsx`) — it already uses `durationLabel` correctly and is the reference pattern, not a target of this change.
- Don't change the `FlatSurfaceIncident` / `EndpointIncident` types in `apps/ui/src/lib/metagraphed/types.ts` — the existing ISO-string fields are sufficient for `durationLabel`'s signature.
- No backend, schema, or API route changes are involved.

## Size

Part of #2542.

luciferlive112116:
## Summary

Closes #3427

Removes the duplicate `fmtDuration` helper in `incident-timeline.tsx` and uses `subnetIncidentDurationText()` (wraps shared `durationLabel`) for incident row durations, matching `IncidentCard`. Open incidents show live elapsed time when `started_at` is present; rows without a usable start omit the duration segment.

## Test plan

- [x] `vitest run src/components/metagraphed/incident-timeline.test.ts` — 4/4
- [x] `vitest run src/lib/metagraphed/format.test.ts` — covers `durationLabel` core cases
- [ ] Manual: subnet incident history shows `· {duration}` for resolved and open incidents with `started_at`

### DIFF
--- a/apps/ui/src/components/metagraphed/incident-timeline.test.ts
+++ b/apps/ui/src/components/metagraphed/incident-timeline.test.ts
@@ -1,0 +1,26 @@
+import { describe, expect, it } from "vitest";
+import { subnetIncidentDurationText } from "./incident-timeline";
+
+describe("subnetIncidentDurationText", () => {
+  it("returns null when started_at is missing", () => {
+    expect(subnetIncidentDurationText(null, "2024-01-01T00:01:00.000Z")).toBeNull();
+    expect(subnetIncidentDurationText(undefined, "2024-01-01T00:01:00.000Z")).toBeNull();
+    expect(subnetIncidentDurationText("", "2024-01-01T00:01:00.000Z")).toBeNull();
+  });
+
+  it("returns null for unparseable started_at (no bare em dash in the row)", () => {
+    expect(subnetIncidentDurationText("nonsense", "2024-01-01T00:01:00.000Z")).toBeNull();
+  });
+
+  it("labels a resolved incident from ISO start/end", () => {
+    expect(
+      subnetIncidentDurationText("2024-01-01T00:00:00.000Z", "2024-01-01T00:01:30.000Z"),
+    ).toBe("1m 30s");
+  });
+
+  it("returns live elapsed text for open incidents (ended_at omitted)", () => {
+    const start = new Date(Date.now() - 2000).toISOString();
+    expect(subnetIncidentDurationText(start, null)).toMatch(/^\d+s$/);
+    expect(subnetIncidentDurationText(start, undefined)).toMatch(/^\d+s$/);
+  });
+});
--- a/apps/ui/src/components/metagraphed/incident-timeline.test.ts
+++ b/apps/ui/src/components/metagraphed/incident-timeline.test.ts
@@ -13,9 +13,9 @@ describe("subnetIncidentDurationText", () => {
   });
 
   it("labels a resolved incident from ISO start/end", () => {
-    expect(
-      subnetIncidentDurationText("2024-01-01T00:00:00.000Z", "2024-01-01T00:01:30.000Z"),
-    ).toBe("1m 30s");
+    expect(subnetIncidentDurationText("2024-01-01T00:00:00.000Z", "2024-01-01T00:01:30.000Z")).toBe(
+      "1m 30s",
+    );
   });
 
   it("returns live elapsed text for open incidents (ended_at omitted)", () => {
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
Please fix + resubmit as a new PR - I don't want cruft being left behind if it's unused, it makes for a messy codebase that's harder to maintain: 

```
The prior duration_ms field is now unused in favor of client-computed duration from timestamps, which is an intentional simplification consistent with sibling components, not a regression.
```

Issue #3434, opened 2026-07-05
Surface loading/error states for signed-extrinsics and transfers on the account detail page


JSONbored:
## Context

`ValidAccountDetail` in `apps/ui/src/routes/accounts.$ss58.tsx` fetches signed extrinsics and native-TAO transfers via two non-blocking `useQuery` calls (lines 135-136: `accountExtrinsicsQuery` / `accountTransfersQuery`), but only ever reads `.data?.data ?? []` — no `isError`/`error` is destructured from either result. Both consumers, `AccountExtrinsicsSection` (line 356) and `AccountTransfersSection` (line 432), early-return `null` whenever `rows.length === 0`, so a genuine fetch failure renders byte-identical to a cold account with no extrinsics/transfers: nothing at all, no indication anything went wrong. The sibling `AccountHistoryChart` (`apps/ui/src/components/metagraphed/account-history-chart.tsx`, lines 149-164) already does this correctly — it destructures `isLoading`/`isError`/`error` from its `useQuery` and branches to `Skeleton` / `ErrorState` / `EmptyState` in that order.

## Requirement

Wire `isError`/`error` from the `accountExtrinsicsQuery` and `accountTransfersQuery` results in `ValidAccountDetail` through to `AccountExtrinsicsSection` and `AccountTransfersSection`, and render `ErrorState` (from `@/components/metagraphed/states`) instead of silently returning `null` when either query has failed.

## Deliverable

- File to change: `apps/ui/src/routes/accounts.$ss58.tsx` (only file expected in the diff for this fix).
- Exact anchor:
  - `ValidAccountDetail`, lines 135-136 — `useQuery(accountExtrinsicsQuery(...))` / `useQuery(accountTransfersQuery(...))` currently discard everything but `.data?.data`.
  - `AccountExtrinsicsSection`, line 356 — signature is `{ rows: Extrinsic[] }`, body is `if (rows.length === 0) return null;`.
  - `AccountTransfersSection`, line 432 — signature is `{ ss58: string; rows: Transfer[] }`, same `if (rows.length === 0) return null;` guard.
- What "done" looks like: keep the two queries non-blocking (do not switch them to `useSuspenseQuery` — that would defeat the documented "never stalls the summary" intent at lines 133-134), but pass `isError`/`error` down as props (or pass the whole query result) so each section can render:
[...]

PR #3922, opened 2026-07-07
feat(ui): add alpha-price sparkline to subnet economics panel

jsdevninja:
Closes #3362

Adds an alpha-price sparkline to the "Alpha price" `StatTile` in `EconomicsPanel`, driven by the historical `alpha_price_tao` values already carried in `subnetTrajectoryQuery(netuid)`'s trajectory points. Same points → finite-number extraction pattern already used by `subnet-price-ticker.tsx`, wired via the existing `Sparkline` primitive into the tile's `chart` slot.

## What changed

- `apps/ui/src/components/metagraphed/economics-panel.tsx`: `EconomicsPanel` now also calls `useQuery(subnetTrajectoryQuery(netuid))` alongside the existing `economicsQuery()` call, derives an aligned `SparklinePoint[]` (date + `alpha_price_tao`) from `trajRes?.data.points`, and passes it into a `Sparkline` mounted as the Alpha price tile's `chart` prop.
- The trajectory query never gates the panel's loading/empty states — those stay keyed off `economicsQuery()`/`isPending` exactly as before. `Sparkline`'s own flat-baseline fallback handles the &lt;2-point / unpriced-subnet case, so no extra empty-state branching was needed.
- No changes to `queries.ts` or `types.ts` — `subnetTrajectoryQuery` and `TrajectoryPoint.alpha_price_tao` already existed and needed no changes.

## Screenshots (subnet 64 — Chutes, live api.metagraph.sh)

| Viewport · Theme | Before | After |
|---|---|---|
| **Desktop · Light** | b | a |
| **Desktop · Dark** | b | a |
| **Tablet · Light** | b | a |
| **Tablet · Dark** | b | a |
| **Mobile · Light** | b | a |
| **Mobile · Dark** | b | a |

## Validation

- `npm run lint` / `npm run format:check` — clean on the changed file
- `npm run typecheck --workspace=apps/ui` — clean
- `npm test --workspace=apps/ui` — 498 tests passed
- `npm run build --workspace=apps/ui` — succeeds
- Manually verified against live `api.metagraph.sh` data for subnet 64 (Chutes), which has 380 daily trajectory points with `alpha_price_tao`

## Non-goals (per issue)

- No market-cap/FDV tiles (#3361)
- No network-wide trend chart (#3365)
- Did not touch `charts/economics-mini.tsx`'s stale comment

PR #3924, opened 2026-07-07
fix(ui): surface a real error state on the account chain-events feed

### LINKED ISSUE #3435: Surface a real error state (not a silent empty view) on the account chain-events feed
## Context
`AccountEventsSection` in `apps/ui/src/routes/accounts.$ss58.tsx` (line 653) only guards on `result.isPending && events.length === 0` before falling through to the populated/empty render. It never checks `result.isError`. Because TanStack Query v5's `status` is `'pending' | 'error' | 'success'`, a failed fetch with no cached data has `isPending === false` and `isError === true` — so it skips the loading skeleton and lands in the exact same `TableState variant="empty"` / "No chain events indexed" block (line 767-777) that a genuinely cold, inactive account gets. A real backend failure is indistinguishable from "this account has never done anything."

## Requirement
`AccountEventsSection` must check `result.isError` and render a distinct error state — using the `error` variant of `TableState` (already imported in this file at line 20, with an `error`/`onRetry` prop shape ready to use) or the `ErrorState` component (`@/components/metagraphed/states`; not currently imported in this file, but used correctly by the sibling `AccountHistoryChart` component — a separate file, `@/components/metagraphed/account-history-chart.tsx`, lines 153-154 — for the same account-detail route) — instead of falling through to the empty-state branch.

## Deliverable
- File to change: `apps/ui/src/routes/accounts.$ss58.tsx`, function `AccountEventsSection` (starts at line 624; the query is issued at line 639 via `useQuery(accountEventsQuery(ss58, params))`, `result` is the value with `isPending`/`isError`/`error` already available from react-query with no destructuring needed).
- Exact anchor: the guard at line 653 (`if (result.isPending && events.length === 0) { … skeleton … }`) is immediately followed by the unconditional return at line 666 that renders either the populated table or (when `events.length === 0`) the "No chain events indexed" `TableState` at lines 767-777. Insert an `isError` branch between these two — after the pending guard, before the populated/empty fork — so an error never reaches the empty-state markup.
- What "done" looks like: add `if (result.isError) { return <SectionAnchor …><TableState variant="error" title="Couldn't load chain events" error={result.error} onRetry={() => result.refetch()} /></SectionAnchor>; }` (or the equivalent using `ErrorState`, imported freshly from `@/components/metagraphed/states`, inside the same `SectionAnchor` wrapper) inside `AccountEventsSection`, keeping the same `id="events"` / title / tone so the section anchor and page layout are unaffected — only the body swaps from empty-state copy to a genuine error card with retry.
- Coordination note: this is one of six sibling "Fresh find — data/hooks layer" issues filed together from #2542 (#3434 covers the *adjacent* `AccountExtrinsicsSection`/`AccountTransfersSection` in the same file, which receive pre-resolved `rows` props rather than owning their own `useQuery`, so it's a separate, differently-shaped fix — do not fold the two into one PR).

## Acceptance criteria
- [ ] `apps/ui/src/routes/accounts.$ss58.tsx` appears in the diff, specifically inside `AccountEventsSection`
- [ ] A fetch failure on `accountEventsQuery` (`result.isError === true`) renders a distinct error UI (`TableState variant="error"` or `ErrorState`), not the "No chain events indexed" empty-state copy
- [ ] The error branch surfaces the underlying `result.error` (message and, for `ApiError`, status/url) rather than a generic empty-state string
- [ ] A cold/inactive account with a successful empty response (`isError === false`, `events.length === 0`) still renders the existing "No chain events indexed" empty state unchanged
- [ ] The existing pending/skeleton behavior (`result.isPending && events.length === 0`) is unchanged
- [ ] No changes to `apps/ui/src/lib/metagraphed/queries.ts` or `accountEventsQuery` — this is a render-branch fix only, the query itself already returns a correctly-typed error on failure

## Non-goals
- Don't touch `AccountExtrinsicsSection` or `AccountTransfersSection` — that's #3434's scope, a differently-shaped fix in the same file.
- Don't add a retry/error affordance to the sibling `AccountHistoryChart` — it already handles `isError` correctly (its own file, `account-history-chart.tsx`, lines 153-154) and needs no change.
- No backend/API change — `/api/v1/accounts/{ss58}/events` already returns real HTTP errors on failure; this is purely a frontend rendering gap.

## Size

Part of #2542.

philluiz2323:
## Summary

- `AccountEventsSection` (`apps/ui/src/routes/accounts.$ss58.tsx`) now checks `result.isError` and renders a distinct error state, instead of silently falling through to the "No chain events indexed" empty-state copy on a failed fetch.

Closes #3435

## What Changed

- Inserted an `isError` branch between the existing pending/skeleton guard and the populated/empty fork, rendering `TableState variant="error"` with the real `result.error` and an `onRetry` wired to `result.refetch()` — the exact same pattern this file already uses for its sibling sections (teardown activity, deregistrations, weight-setting activity, serving/prometheus announcements), so this just closes the one section that was missing it.
- No changes to `AccountExtrinsicsSection` / `AccountTransfersSection` (out of scope, tracked separately by #3434) or `AccountHistoryChart` (already handles `isError` correctly).
- No changes to `apps/ui/src/lib/metagraphed/queries.ts` / `accountEventsQuery` — render-branch fix only.

## Why

TanStack Query v5's `status` is `'pending' | 'error' | 'success'`. A failed fetch with no cached data has `isPending === false` and `isError === true`, so before this fix it skipped the loading skeleton and landed in the exact same empty-state block a genuinely cold, inactive account gets — a real backend failure was indistinguishable from "this account has never done anything," with no way to tell what went wrong or retry.

## Registry Safety

- N/A — this PR touches only `apps/ui/` frontend rendering, no registry/schema/API surface.

## Validation

- [x] `npx tsc --noEmit` (workspace apps/ui) — no new errors (2 pre-existing, unrelated errors in `queries.ts`/`types.ts` reproduce identically on a clean `main` checkout, need `npm run build:client` first)
- [x] `npm test --workspace=apps/ui` — all 73 files / 509 tests pass
- [x] `git diff --check`
- [x] Verified against every acceptance criterion in #3435: error branch surfaces `result.error` (not generic empty copy), the existing pending/skeleton behavior is unchanged, and a genuinely cold/inactive account with a successful empty response still renders the unchanged "No chain events indexed" empty state (only the new `isError` branch sits between the two, so the empty-state fork's own condition — `events.length === 0` — is untouched)

## Issue Link

Closes #3435

### DIFF
--- a/apps/ui/src/routes/accounts.$ss58.tsx
+++ b/apps/ui/src/routes/accounts.$ss58.tsx
@@ -1082,6 +1082,25 @@ function AccountEventsSection({
     );
   }
 
+  if (result.isError) {
+    return (
+      <SectionAnchor
+        id="events"
+        title="Chain events"
+        subtitle="Full first-party event feed for this account, newest first — filter by kind, page through history."
+        tone="accent"
+      >
+        <TableState
+          variant="error"
+          title="Couldn't load chain events"
+          description="The chain-events feed is temporarily unavailable — the rest of the account page is unaffected."
+          error={result.error}
+          onRetry={() => void result.refetch()}
+        />
+      </SectionAnchor>
+    );
+  }
+
   return (
     <SectionAnchor
       id="events"


JSONbored (approved):

JSONbored:
Next time, changes like this require visual evidence (screenshots and/or recordings - see PR #3922 for example of how to properly add this). 

Use the AI skill found in the repo for assistance/instructions on adding this evidence correctly. This still touches a visual area and as such, does normally require visual evidence (anything that visitors may SEE on the frontend requires screenshots, typically).

Issue #2542, opened 2026-07-01
🎨 Wave 3 — Frontend explorer enrichment (LIVE — 219 issues filed)


JSONbored:
## 🎨 Wave 3 — Frontend explorer enrichment

Parent tracker. The monorepo consolidation (#3062) has landed — this wave is **unblocked and live**.
Every item below is now a real, individually-filed issue in this milestone; check them off as they merge.

**219 issues filed** across 26 categories: the original 153 from the pre-scoped checklist, a
gap-hunt over merged `apps/ui` code, a backend-route-vs-frontend-usage audit, and a taostats.io
feature comparison (filtered through the standing scope fence: no
portfolio/PnL/tax/wallet/exchange-tracker/candlestick/mobile-app work — this is a developer explorer,
not a trader tool) — plus **66 more** from a follow-up audit (screenshot review of 96 recent PRs,
6 codebase-pattern sweeps, and direct interactive testing), addended at the bottom of this issue.

Labels: `gittensor:bug` for fixes/wiring/tests (the large majority), `gittensor:feature` reserved for
genuinely new routes/pages/capabilities (~5%), `gittensor:priority` on 6 foundational items that unlock
several others each — grab those first.

### Surface backend-only endpoints (17)
- [ ] #3340 Surface account counterparties panel on the account detail page
- [ ] #3341 Surface per-account stake-flow chart on the account detail page
- [ ] #3342 Surface per-subnet stake-flow in the subnet Activity tab
- [ ] #3343 Surface per-subnet validator/registration turnover in the Metagraph/Validators tab
- [ ] #3344 Surface the cross-subnet biggest-movers band on Home
- [ ] #3345 Surface the network-wide economics trend chart on the explorer page
- [ ] #3346 Fetch the subnet-scoped overview artifact in the subnet Overview tab
- [ ] #3347 Fix subnet Evidence tab to fetch the subnet-scoped evidence route
- [ ] #3348 Fix subnet Gaps tab to fetch the subnet-scoped gaps route
- [ ] #3349 Display the endpoint-pools composition on the endpoints page
- [ ] #3350 Display the RPC endpoints catalog on the endpoints page
- [ ] #3351 Add an RSS/Atom feed discovery link in the site footer
- [ ] #3352 Add an RSS/Atom feed affordance to the status page
- [ ] #3353 Surface the source-health provider rollup on the providers page
- [ ] #3354 Wire the review/enrichment-evidence queue into the gaps review surface
[...]

PR #4232, opened 2026-07-07
fix(ui): surface error states for account extrinsics and transfers (#3434)

### LINKED ISSUE #3434: Surface loading/error states for signed-extrinsics and transfers on the account detail page

RealDiligent:
## Summary

- Wire `isError` / `error` / `onRetry` from the non-blocking `accountExtrinsicsQuery` and `accountTransfersQuery` results into `AccountExtrinsicsSection` and `AccountTransfersSection`.
- On fetch failure, render the same `TableState variant="error"` pattern already used by sibling account feeds instead of silently returning `null`.

Closes #3434 · Part of #2542

## Screenshots

| View | Before | After |
|------|--------|-------|
| Desktop light | before | after |

The after panel uses the same `TableState` error chrome as merged #3924 (`AccountEventsSection`) and the other account activity sections in this file.

## Test plan

- [x] Error branch mirrors sibling account sections in `accounts.$ss58.tsx`
- [x] CI `ui` / `test` / `checks` green on prior submission (#4227)


JSONbored (commented):
Rejecting, you included screenshot files in the PR itself, and didn't include the actual screenshots from the website before + after. Use the AI skill found in the repo - it gives your harness everything it needs to know about submitting evidence for visual/frontend PRs.

PR #3796, opened 2026-07-06
feat(ui): add decoded chain-events view to the block detail page

jsdevninja:
## Summary

- Adds a "Chain events" section to the block detail page
  (`apps/ui/src/routes/blocks.$ref.tsx`), backed by a new
  `blockChainEventsQuery` / `normalizeBlockChainEvents` pair in
  `apps/ui/src/lib/metagraphed/queries.ts` that consumes
  `GET /api/v1/blocks/{ref}/chain-events` and is wired in via `useQuery` in
  the same PR.
- The section is added **alongside** the existing curated `events`
  consumption (not a replacement) — see "How `/chain-events` differs" below
  for why.

## How `/chain-events` differs from the already-consumed `/events` route

- `/events` (already consumed) returns the curated, **account-attributed**
  subset of a block's events from the first-party D1 tier: `event_kind`,
  `hotkey`, `coldkey`, `netuid`, `uid`, `amount_tao` — friendly fields for
  stake/transfer-style events only.
- `/chain-events` returns **every** raw pallet-level event in the block
  (Postgres-backed all-events tier, ADR 0013) with the runtime `pallet` +
  `method` id, the full decoded `args`, dispatch `phase`, and
  `extrinsic_index` — broader coverage, but it has no
  hotkey/coldkey/amount attribution.
- Neither is a strict superset of the other, so per the issue's non-goal
  the existing `/events` consumption is kept; the new section supplements
  it.
- Real-world quirk discovered while implementing: the Postgres tier
  serializes bigint columns (`event_index`, `observed_at`) as **JSON
  strings**, not numbers (unlike the D1-backed `/events`/`/extrinsics`
  routes, which return plain numbers and ISO datetimes), and the per-block
  route omits the redundant per-row `block_number`. The normalizer
  coerces these numeric strings (matching the existing convention used by
  sibling Postgres-tier normalizers like `normalizeChainActivityDay`) and
  backfills `block_number` from the parent response. Covered by
  `apps/ui/src/lib/metagraphed/queries.block-chain-events.test.ts`.

## Screenshots

| Page / Feature | Before | After |
| --- | --- | --- |
[...]

PR #4621, opened 2026-07-09
fix(blocks): decode chain-event args (account ids → SS58) with a copy affordance

### LINKED ISSUE #3984: New chain-events Args column shows raw undecoded byte arrays, hard-clipped at the viewport edge with no way to read or copy the value
## Context

The new "Chain events" section on the block detail page (#3796) shows the Args column rendering values like `{"who":[[109,111,100,101,115,117,98,116,101,110,115,114,1,0,0,0,0,0,0,...}` instead of a decoded, human-readable representation (e.g. the actual SS58 address string this byte array almost certainly encodes). The text is not truncated with an ellipsis, not wrapped, and has no copy affordance — it runs straight to the right edge of the viewport and is cut off mid-array with no way to see the rest of the value. This affects most rows in the table (`Balances.Deposit`, `SubtensorModule.TimelockedWeightsRevealed`, etc.), making the new section's primary content column largely unreadable and useless as shipped. Every sibling hash/address column elsewhere in this codebase gets copy buttons and hover-cards per the existing, already-tracked backlog — this new column has neither truncation-with-affordance nor decoding.

## Requirement

The Args column must render a genuinely readable value for each row: decoded where the underlying type is recognizable (e.g. an account-id byte array becomes its SS58 string), and at minimum truncated with a copy-to-clipboard/expand affordance where full decoding isn't feasible for this issue's scope.

## Deliverable

- **File to change**: `apps/ui/src/routes/blocks.$ref.tsx`, the Chain events table's Args column (introduced in #3796).
- **What "done" looks like concretely**: for the common, decodable cases (account-id byte arrays → SS58, in particular), decode and display the human-readable form with a copy button, matching the pattern already used for hash/address cells elsewhere. For genuinely raw/undecodable arg shapes, truncate the JSON with an ellipsis and provide an expand-on-click or copy-full-value affordance, rather than letting the raw text run off the viewport edge uncontrolled.

## Acceptance criteria

- [ ] `blocks.$ref.tsx` appears in the diff
- [ ] `Balances.Deposit` and `SubtensorModule.TimelockedWeightsRevealed` rows (or their current equivalents) show a decoded, human-readable value where the arg type is a recognizable account-id/address encoding
- [ ] No Args-column value overflows the viewport edge uncontrolled — every value is either fully visible, truncated with an expand affordance, or scrollable within its cell
- [ ] A copy-to-clipboard affordance exists for the (decoded or raw) value
- [ ] Includes a before/after screenshot table at desktop and mobile, both themes, per the repo's UI-PR requirement

## Non-goals

- Don't attempt to decode every possible pallet/call arg shape in this repo's type system — start with the account-id/address case that's clearly the common one in the sample data, and truncate-with-affordance for the rest; broader decoding coverage can be a follow-up.

## Size

Size: M — keep this PR small (aim for ≤10 files / ≤1000 LOC).

Part of #2542.

jeffrey701:
## Problem

The Chain events table's **Args** column on the block detail page (`blocks.$ref.tsx`, #3796) renders each value with a bare `JSON.stringify`, so decoded SCALE account ids come out as raw byte arrays like `{"who":[[212,53,147,199,21,253,211,28,…]]}` — unreadable, no decoding, and no copy affordance. This affects most rows (`Balances.Deposit`, `SubtensorModule.TimelockedWeightsRevealed`, …), making the section's primary column useless as shipped.

## Fix

- **New `lib/metagraphed/ss58.ts`** — a small, **dependency-free** SS58 encoder. The frontend intentionally ships no crypto library, so blake2b (for the SS58 checksum) and base58 are vendored as compact self-contained routines and locked down with vectors in `ss58.test.ts` (blake2b RFC test vectors + the canonical Alice/Bob dev accounts + a Polkadot-format case).
- **New `lib/metagraphed/chain-event-args.ts`** — `formatChainEventArgs` / `decodeChainEventArgs` walk the args value and rewrite 32-byte arrays: to an **SS58 address** when the field name marks it as an account (`who`/`coldkey`/`hotkey`/`from`/`to`/…), otherwise to **0x-hex** (so a 32-byte hash is made readable without being mislabelled as an address). Everything else is untouched.
- **`routes/blocks.$ref.tsx`** — the Args cell now renders the decoded value in a truncating, non-overflowing container with a `CopyButton` (the same affordance used for hash/address cells elsewhere). The decoding/formatting logic lives in the tested lib modules; the route change is minimal.

## Before / after

Mocked block (real block artifacts aren't populated on the API yet) using the exact response shape + the issue's sample account-id args.

| Viewport | Before | After |
| --- | --- | --- |
| Desktop · light | before desktop light | after desktop light |
| Desktop · dark | before desktop dark | after desktop dark |
| Mobile · light | before mobile light | after mobile light |
| Mobile · dark | before mobile dark | after mobile dark |

`Balances.Deposit`'s `who` now reads `5GrwvaEF5zXb26…` and `SubtensorModule.TimelockedWeightsRevealed`'s `who` reads `5FHneW46xGXgs5…`; every value is truncated (no viewport overflow) with a copy button.

## Test

`ss58.test.ts` (6) — blake2b vectors, base58 leading-zeros, Alice/Bob SS58, Polkadot format, non-32-byte → null. `chain-event-args.test.ts` (6) — account key → SS58, nested `who: [<bytes>]`, non-account 32-byte array → 0x-hex (not mislabelled), non-account values untouched, nullish → dash, no raw-byte leakage.

Local: `vitest run` (12 passed), `tsc --noEmit` (clean), `eslint` (0 errors), `prettier --check` (clean).

## Notes

Per the issue's scope, decoding starts with the common account-id case; other raw shapes are truncated-with-copy rather than exhaustively decoded. No new runtime dependency is added.

Closes #3984

### DIFF
--- a/apps/ui/src/lib/metagraphed/chain-event-args.test.ts
+++ b/apps/ui/src/lib/metagraphed/chain-event-args.test.ts
@@ -1,0 +1,53 @@
+import { describe, it, expect } from "vitest";
+import { decodeChainEventArgs, formatChainEventArgs } from "./chain-event-args";
+
+const bytes = (h: string) => [
+  ...Uint8Array.from(
+    h
+      .replace(/^0x/, "")
+      .match(/../g)!
+      .map((b) => parseInt(b, 16)),
+  ),
+];
+const ALICE = bytes("0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d");
+const ALICE_SS58 = "5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY";
+
+describe("decodeChainEventArgs", () => {
+  it("decodes an account-id byte array under an account key to SS58", () => {
+    expect(decodeChainEventArgs({ who: ALICE, amount: 1000 })).toEqual({
+      who: ALICE_SS58,
+      amount: 1000,
+    });
+  });
+
+  it("decodes account bytes nested inside an array (who: [<bytes>])", () => {
+    expect(decodeChainEventArgs({ who: [ALICE] })).toEqual({ who: [ALICE_SS58] });
+  });
+
+  it("renders a 32-byte array under a non-account key as 0x-hex, not a mislabelled address", () => {
+    const out = decodeChainEventArgs({ commit: ALICE }) as { commit: string };
+    expect(out.commit).toBe("0x" + ALICE.map((b) => b.toString(16).padStart(2, "0")).join(""));
+  });
+
+  it("leaves non-account values untouched", () => {
+    expect(decodeChainEventArgs({ netuid: 7, flag: true, note: "x", list: [1, 2, 3] })).toEqual({
+      netuid: 7,
+      flag: true,
+      note: "x",
+      list: [1, 2, 3],
+    });
+  });
+});
+
+describe("formatChainEventArgs", () => {
+  it("returns a dash for nullish args", () => {
+    expect(formatChainEventArgs(null)).toBe("—");
+    expect(formatChainEventArgs(undefined)).toBe("—");
+  });
+
+  it("produces a readable one-liner with the account decoded and no raw byte spew", () => {
+    const s = formatChainEventArgs({ who: ALICE, amount: 1000 });
+    expect(s).toContain(ALICE_SS58);
+    expect(s).not.toContain("212"); // first raw byte of Alice's key — must not leak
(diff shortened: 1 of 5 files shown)


JSONbored (approved):
Screenshots do not follow required submission evidence standards, but this is a small enough change that it's acceptable. Next time will be a closure. Please use the AI skill, I've politely nudged this several times.

PR #4631, opened 2026-07-09
fix(states): make the loading Skeleton visible against same-surface containers in dark mode

### LINKED ISSUE #3993: New loading skeleton is nearly invisible against its own container in dark mode
## Context

PR #3737 adds a `Skeleton` placeholder (`<div className="animate-pulse rounded bg-surface h-64 w-full" />`, `apps/ui/src/components/metagraphed/states.tsx:226-228`) for the Signed Extrinsics and Transfers sections while their queries are pending. The PR's own "after" screenshot, captured mid-load (the header pulse-band stats still show unresolved "—" placeholders, confirming the page was genuinely still loading at capture time), shows the Signed Extrinsics and Transfers sections as completely blank dark voids — the pulse-animation skeleton has too little contrast against its own background (`bg-surface` against a similarly-dark page background) in dark theme to read as a loading indicator at all. A user in dark mode sees an unexplained blank gap rather than a visible loading state.

## Requirement

The `AccountFeedSectionSkeleton` must be visibly distinguishable from its background in dark mode, not blend into invisibility.

## Deliverable

- **File to change**: `apps/ui/src/components/metagraphed/states.tsx:226-228`.
- **What "done" looks like concretely**: give the skeleton a background color with enough contrast against the page background in dark mode — e.g. a step lighter than `bg-surface` (`bg-surface-2` if that token exists and reads correctly), while keeping the existing `animate-pulse` shimmer. Verify against the actual dark-theme page background, not just in isolation, since the reported bug is specifically about blending into the *page*, not the skeleton's own styling in a vacuum.

## Acceptance criteria

- [ ] `states.tsx` appears in the diff
- [ ] In dark mode, the skeleton for the Signed Extrinsics/Transfers sections is visibly distinguishable from the page background while loading
- [ ] Light mode rendering (already presumably fine, confirm) is unaffected
- [ ] Includes a before/after screenshot captured mid-load in dark mode, per the repo's UI-PR requirement — since this bug is specifically about the transient loading state, the screenshot must be taken while the query is genuinely pending (e.g. via network throttling or a mocked delay), not after it resolves

## Non-goals

- Don't change the skeleton's shape/size or where it's used — this is a contrast fix for the existing component.

## Size

Part of #2542.

jeffrey701:
## Problem

The shared `Skeleton` placeholder (`states.tsx`) uses `bg-surface`. When a skeleton sits inside a same-`surface` container in **dark mode**, it's nearly invisible — the dark surface tokens sit only a hair apart, so the loading state reads as an unexplained blank void rather than a placeholder (#3993, seen on the account page's feed sections).

## Fix

Give the placeholder `bg-surface-2` (a step lighter than `bg-surface` in both themes) **plus a hairline `border-border`** for definition, keeping the existing `animate-pulse` shimmer. `bg-surface-2` alone was still too subtle against a same-surface dark background; the border makes the box read as a distinct loading element in both themes. One shared component, so every skeleton benefits.

## Before / after

Account page with its feed queries held pending (`DetailSkeleton`), captured at desktop.

| Theme | Before | After |
| --- | --- | --- |

Before, the PORTFOLIO / feed skeletons are invisible dark voids; after, they read as clearly-bounded pulsing placeholders.

Local: `tsc --noEmit` (clean), `eslint` (0 errors), `prettier --check` (clean).

Closes #3993

### DIFF
--- a/apps/ui/src/components/metagraphed/states.tsx
+++ b/apps/ui/src/components/metagraphed/states.tsx
@@ -224,7 +224,11 @@ export function StaleBanner({
 }
 
 export function Skeleton({ className = "h-4 w-full" }: { className?: string }) {
-  return <div className={`animate-pulse rounded bg-surface ${className}`} />;
+  // bg-surface-2 (a step lighter than bg-surface in both themes) plus a hairline
+  // border so the pulse placeholder stays readable against a same-surface
+  // container — bg-surface alone vanished into the background in dark mode, where
+  // the surface tokens sit only a hair apart (#3993).
+  return <div className={`animate-pulse rounded border border-border bg-surface-2 ${className}`} />;
 }
 
 /**


JSONbored (commented):
Does not include required screenshot evidence across all 3 viewports, both color modes, for before + after as described by the AI skill. Closing.

PR #4652, opened 2026-07-09
fix(a11y): make activity-heatmap day cells keyboard-focusable so tooltips reveal on focus

### LINKED ISSUE #3955: Activity heatmap day-cells use non-focusable divs, making the per-day tooltip unreachable by keyboard
## Context

`src/components/metagraphed/charts/activity-heatmap.tsx:117-150` (subnet profile Registry activity panel) wraps each day cell in a Radix `<Tooltip>`/`<TooltipTrigger asChild>` around a plain `<div>`, with only a static `aria-label` and no `tabIndex`, `role`, or keyboard handler. Radix tooltips show on focus for genuinely focusable triggers, but a bare `<div>` is never in the tab order, so keyboard users can never reveal the per-day tooltip (probes/incidents/uptime breakdown for that day). Contrast with the sibling component `src/components/metagraphed/charts/subnet-pulse-grid.tsx:59-69`, which wraps the identical Tooltip pattern around a naturally-focusable `<Link>` and works correctly as a result.

## Requirement

Each day-cell in the activity heatmap must be reachable via keyboard focus, so its Radix tooltip reveals the same way it does on hover.

## Deliverable

- **File to change**: `src/components/metagraphed/charts/activity-heatmap.tsx:117-150`.
- **What "done" looks like concretely**: make each day cell's wrapping element focusable — either swap the bare `<div>` for a focusable element (a `<button>` if the cell has no other navigational purpose) or add `tabIndex={0}` plus an appropriate `role` to the existing `<div>`, matching how `subnet-pulse-grid.tsx`'s `<Link>`-based cells already achieve this correctly.

## Acceptance criteria

- [ ] `activity-heatmap.tsx` appears in the diff
- [ ] Tabbing through the heatmap reaches each day cell, and the per-day tooltip reveals on focus, matching hover behavior
- [ ] Visual appearance of the heatmap grid is unchanged
- [ ] Includes a before/after screen recording demonstrating keyboard focus revealing a day-cell's tooltip — this can't be shown in a static screenshot

## Non-goals

- Don't change the heatmap's data or the tooltip's content.
- Don't touch `subnet-pulse-grid.tsx` — it's the correct reference pattern, not part of the bug.

## Size

Part of #2542.

jeffrey701:
## Problem

In the subnet Registry-activity `activity-heatmap.tsx`, each day cell is a bare `<div>` wrapped in a Radix `<TooltipTrigger asChild>`, with an `aria-label` but no `tabIndex`/`role`. Radix tooltips only reveal on focus for genuinely focusable triggers, so a bare `<div>` — never in the tab order — leaves keyboard users unable to reach the per-day tooltip (probes / incidents / uptime). The sibling `subnet-pulse-grid.tsx` wraps the same pattern around a naturally-focusable `<Link>` and works correctly (#3955).

## Fix

Swap the cell's bare `<div>` for a semantic, naturally-focusable `<button type="button">` (the cell has no navigational purpose), with a `focus-visible` ring. It's now in the tab order, so keyboard focus reveals the same Radix tooltip that hover does. Visual rendering on load is unchanged.

## Before / after

Subnet page Registry-activity heatmap; the same day cell is focused programmatically in both.

| Theme | Before (bare `<div>`) | After (`<button>`) |
| --- | --- | --- |

Before, focusing a cell is a no-op (`div`, not in the tab order → no focus ring, no tooltip). After, the cell takes focus (`button`), shows a focus ring, and reveals its tooltip — exactly what a keyboard user gets.

Local: `tsc --noEmit` (clean), `eslint` (0 errors), `prettier --check` (clean).

Closes #3955

### DIFF
--- a/apps/ui/src/components/metagraphed/charts/activity-heatmap.tsx
+++ b/apps/ui/src/components/metagraphed/charts/activity-heatmap.tsx
@@ -125,8 +125,12 @@ export function ActivityHeatmap({ netuid, weeks = 12 }: Props) {
               {col.map((c) => (
                 <Tooltip key={c.key} delayDuration={120}>
                   <TooltipTrigger asChild>
-                    <div
-                      className="aspect-square rounded-[2px] border border-border/40"
+                    {/* A focusable <button> (not a bare div) so keyboard users can
+                        reach each day and reveal its Radix tooltip — the tooltip
+                        only shows on focus for genuinely focusable triggers (#3955). */}
+                    <button
+                      type="button"
+                      className="aspect-square w-full rounded-[2px] border border-border/40 focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-accent"
                       style={{ background: tone(c.score, maxScore) }}
                       aria-label={`${c.key}: ${c.probes} probes, ${c.incidents} incidents`}
                     />


JSONbored (commented):
Missing most screenshots again, closing.

PR #4713, opened 2026-07-10
test(ui): add AGPL-3.0 badge to footer to validate visual-capture pipeline

JSONbored:
Small, real footer addition (license disclosure, matching the sibling gittensory repo's own AGPL-3.0 badge) used to validate the visual-capture bot-screenshot pipeline end-to-end after JSONbored/gittensory#4564 (per-repo `review.visual.production_url` override + generalized route inference for `apps/ui/**`).

PR #4717, opened 2026-07-10
test(ui): re-verify AI reviewer no longer false-flags bot-captured screenshots

JSONbored:
Re-uses the exact commit from JSONbored/metagraphed#4713 (same head SHA, so bot-capture's R2 cache is reused — no new browser renders) to verify the review-skill fix for JSONbored/gittensory#4565: the AI reviewer should no longer independently flag missing screenshot evidence as a blocker now that it's been told the deterministic gate (not its own pre-capture judgment) is authoritative for that dimension.


JSONbored:
Confirmed the fix (private review-skill update on edge-nl-01) works: same commit as #4713, but the screenshot-evidence mention moved from the blockers list to a neutral, non-blocking nit. Closing without merging — validation-only.

PR #4721, opened 2026-07-10
feat(ui): use shared table-controls on the Providers toolbar

### LINKED ISSUE #3424: Add search/filter controls to the Providers page using the shared table-controls components
## Context

`apps/ui/src/routes/providers.index.tsx` hand-rolls its own filter bar: a local `Selector` component (defined at lines 461-491, not imported from anywhere) backs three `useState` fields (`q`, `kind`, `authority`, declared at lines 152-154) and a bespoke inline "Clear" button (lines 262-274). Every other list page (`subnets.index.tsx`, `surfaces.tsx`, `extrinsics.index.tsx`, `accounts.$ss58.tsx`) instead imports `SearchInput`, `SelectFilter`, and `ResetFiltersButton` from the shared `apps/ui/src/components/metagraphed/table-controls.tsx` — this route currently has zero import from that file, making it the one outlier in markup, styling, and accessibility behavior (e.g. the shared `SearchInput` sets `aria-label`; the local raw `<input>` at lines 240-246 does too, but independently, so the two can drift).

## Requirement

Replace the Providers page's local `Selector` component and hand-rolled search `<input>`/"Clear" button with the shared `SearchInput`, `SelectFilter`, and `ResetFiltersButton` components from `table-controls.tsx`, wiring them to the same existing local `q`/`kind`/`authority`/`sortKey` state — this is a component-reuse/dedup swap, not a new feature or a URL-state change.

## Deliverable

- File(s) to change: `apps/ui/src/routes/providers.index.tsx` only (plus, if reuse makes sense, `apps/ui/src/components/metagraphed/table-controls.tsx` only if a genuinely new prop/variant is needed for parity — do not fork a new component there).
- Exact anchor: the toolbar block at lines 236-278 of `providers.index.tsx` (the `{/* Toolbar */}` `<div>`), which currently renders a raw `<input>` (lines 240-246), three `<Selector>` calls (lines 248-261, using the local `Selector` function at lines 461-491), and an inline `<button>` "Clear" (lines 262-274). Remove the local `Selector` function entirely once nothing references it.
- What "done" looks like concretely:
  - Import `SearchInput`, `SelectFilter`, `ResetFiltersButton` (and `PageSizeSelect` only if actually used) from `@/components/metagraphed/table-controls` (mirror the import style used in `subnets.index.tsx` line 24-26 or `extrinsics.index.tsx` line 15-17).
  - Swap the raw search `<input>` for `<SearchInput value={q} onChange={setQ} placeholder="Search providers, slugs, hosts…" />`.
  - Swap each `<Selector label=… value=… onChange=… options=… />` call for `<SelectFilter label=… value=… onChange=… options={...} />` — note `SelectFilter`'s `options` prop is `Array<{ value: string; label: string }>` (see `table-controls.tsx` lines 80-89), while the current `kinds`/`authorities`/sort arrays are plain `string[]` (lines 157-164, 259) — map each to `{ value: v, label: v }` at the call site. The existing `Sort` selector uses `allowEmpty={false}` (no "all" option) which the local `Selector` supports directly (line 466) but `SelectFilter` always renders an "all" option (line 101) — if dropping "all" for Sort isn't acceptable, keep Sort on the local pattern or handle the always-selected default explicitly, and note the discrepancy in the PR description rather than silently changing sort UX.
  - Swap the inline "Clear" `<button>` (lines 262-274) for `<ResetFiltersButton active={hasFilters} onReset={() => { setQ(""); setKind(""); setAuthority(""); }} />` (the existing `hasFilters` at line 223 already computes the right boolean).
  - Delete the now-unused local `Selector` function (lines 461-491) and its now-unused `X` icon import if nothing else in the file uses `X` from `lucide-react` (check line 6 import list before removing).
- Coordination note: two sibling issues from the same Wave-3 batch touch this same file/toolbar region — **#3418** ("Persist the Providers page search/kind/authority filters in the URL") moves `q`/`kind`/`authority` from local `useState` into the route's `validateSearch` schema (currently `providersSearchSchema` only has `view`, lines 22-24), and **#3419** ("Add a ShareButton...") explicitly depends on #3418 landing first. This issue (#3424) is independent of both — it only swaps which components render the existing local state, so it can land before, after, or in either order relative to #3418 without blocking it. Do not use this issue to also move state into the URL; that is #3418's scope.

## Acceptance criteria

- [ ] `apps/ui/src/routes/providers.index.tsx` is the only route file changed (aside from any minimal, genuinely-shared addition to `table-controls.tsx`, if truly needed)
- [ ] `providers.index.tsx` imports `SearchInput`, `SelectFilter`, and `ResetFiltersButton` from `@/components/metagraphed/table-controls`
- [ ] The toolbar's search input is `<SearchInput>` (no local raw `<input>` with a manually-set `aria-label` remains for this purpose)
- [ ] The Kind and Authority filters render via `<SelectFilter>` (no calls to a local `Selector` component remain)
- [ ] The local `Selector` function definition is deleted from the file
- [ ] The "Clear" affordance is `<ResetFiltersButton>`, shown/hidden via the existing `hasFilters` condition, and clears `q`/`kind`/`authority` exactly as the current inline button does
- [ ] Filtering/sorting/clearing behavior is unchanged from a user's perspective (same providers filtered, same clear semantics) — verified by manual check against current `main` behavior, screenshotted before/after per repo policy
- [ ] `npm run build` (or the UI's typecheck/lint equivalent) passes with no new TypeScript or ESLint errors from the prop-shape change (`string[]` → `Array<{value,label}>` mapping)

## Non-goals

- Do not move `q`/`kind`/`authority` into the URL/`validateSearch` schema — that is #3418's scope, not this issue's.
- Do not add a ShareButton — that is #3419's scope, and it explicitly depends on #3418, not this issue.
- Do not touch `subnets.index.tsx`, `surfaces.tsx`, `extrinsics.index.tsx`, or `accounts.$ss58.tsx` — they already use the shared components correctly and are only useful here as reference patterns.
- Do not change the sort-key options or introduce a new sort UX; if `SelectFilter`'s forced "all" option conflicts with the existing `allowEmpty={false}` Sort behavior, resolve it minimally (see Deliverable) rather than redesigning the sort control.
- No backend/API/schema changes are involved — this is a frontend-only component swap in `apps/ui/`.

## Size

Part of #2542.

nickmopen:
Closes #3424

## What

Adopts the shared `table-controls` components on the Providers grid toolbar **and** reworks the toolbar's layout so the section no longer feels neglected (per review feedback).

**Component swap**
- raw `<input>` + magnifier wrapper → shared `SearchInput`
- 3× local `Selector` (Kind / Authority / Sort) → shared `SelectFilter`
- hand-rolled Clear `<button>` → shared `ResetFiltersButton`
- deleted the now-unused local `Selector` helper and the now-unused `Search`/`X` icon imports

**Layout polish**
- Wrapped the toolbar in the same responsive treatment the subnets grid toolbar uses — edge-to-edge sticky bar on mobile, an inset rounded `bg-card` card on desktop — instead of a bare underline.
- Grouped the result count + reset into an aligned right-hand cluster with consistent mono/uppercase meta typography (`tracking-[0.18em]`).

**Shared-component addition:** `SelectFilter` gained an optional `allowEmpty` prop (defaults `true`, so every existing caller is unchanged) so the always-selected Sort control can omit the empty "all" option. This is the only touch to `table-controls.tsx`.

## Verify

`npm run lint` (0 errors) · `npm run typecheck` · `npm run format:check` · `npm test` (679 passing) · `npm run build` — all green locally.

## Screenshots

Before = `main`, after = the redesigned toolbar. Real data (133 providers, unchanged between before/after). 3 viewports × 2 themes.

| Viewport / Theme | Before (main) | After (redesigned) |
| --- | --- | --- |
| Mobile · Light |  |  |
| Mobile · Dark |  |  |
| Tablet · Light |  |  |
| Tablet · Dark |  |  |
| Desktop · Light |  |  |
| Desktop · Dark |  |  |

### DIFF
--- a/apps/ui/src/components/metagraphed/table-controls.tsx
+++ b/apps/ui/src/components/metagraphed/table-controls.tsx
@@ -92,6 +92,7 @@ export function SelectFilter({
   options,
   label,
   allowEmpty = true,
+  fill = false,
 }: {
   value: string;
   onChange: (v: string) => void;
@@ -100,16 +101,25 @@ export function SelectFilter({
   // When false, omit the empty "all" option — for always-selected controls like
   // a sort key where a blank value is not meaningful.
   allowEmpty?: boolean;
+  // When true, the control stretches to fill its flex track (label stays fixed,
+  // the select grows) so a row of filters can be justified edge-to-edge.
+  fill?: boolean;
 }) {
   return (
-    <label className="inline-flex items-center gap-1.5 rounded border border-border bg-paper px-2 py-1 text-xs">
+    <label
+      className={`items-center gap-1.5 rounded border border-border bg-paper px-2 py-1 text-xs ${
+        fill ? "flex w-full min-w-0" : "inline-flex"
+      }`}
+    >
       <span className="font-mono text-[10px] uppercase tracking-widest text-ink-muted">
         {label}
       </span>
       <select
         value={value}
         onChange={(e) => onChange(e.target.value)}
-        className="bg-transparent text-ink-strong text-xs rounded focus:outline-none focus-visible:ring-2 focus-visible:ring-ring"
+        className={`bg-transparent text-ink-strong text-xs rounded focus:outline-none focus-visible:ring-2 focus-visible:ring-ring ${
+          fill ? "min-w-0 flex-1" : ""
+        }`}
       >
         {allowEmpty ? <option value="">all</option> : null}
         {options.map((o) => (
--- a/apps/ui/src/components/metagraphed/table-controls.tsx
+++ b/apps/ui/src/components/metagraphed/table-controls.tsx
@@ -91,11 +91,15 @@ export function SelectFilter({
   onChange,
   options,
   label,
+  allowEmpty = true,
 }: {
   value: string;
   onChange: (v: string) => void;
   options: Array<{ value: string; label: string }>;
   label: string;
+  // When false, omit the empty "all" option — for always-selected controls like
+  // a sort key where a blank value is not meaningful.
+  allowEmpty?: boolean;
 }) {
   return (
     <label className="inline-flex items-center gap-1.5 rounded border border-border bg-paper px-2 py-1 text-xs">
@@ -107,7 +111,7 @@ export function SelectFilter({
         onChange={(e) => onChange(e.target.value)}
         className="bg-transparent text-ink-strong text-xs rounded focus:outline-none focus-visible:ring-2 focus-visible:ring-ring"
       >
-        <option value="">all</option>
+        {allowEmpty ? <option value="">all</option> : null}
         {options.map((o) => (
           <option key={o.value} value={o.value}>
             {o.label}
--- a/apps/ui/src/routes/providers.index.tsx
+++ b/apps/ui/src/routes/providers.index.tsx
@@ -275,41 +275,46 @@ function ProvidersGrid({ view }: { view: "grid" | "table" }) {
       <ProviderOverview providers={rows} counts={counts} />
       <SourceHealthRollup />
 
-      {/* Toolbar */}
-      <div className="sticky top-14 z-10 -mx-1 px-1 py-2 backdrop-blur bg-paper/85 border-b border-border/60 flex flex-wrap items-center gap-2">
-        <div className="flex-1 min-w-[180px] max-w-sm">
-          <SearchInput
-            value={q}
-            onChange={(v) => setSearch({ q: v })}
-            placeholder="Search providers, slugs, hosts…"
+      {/* Filter toolbar — edge-to-edge sticky bar on mobile, an inset card on desktop,
+          matching the subnets grid toolbar treatment. */}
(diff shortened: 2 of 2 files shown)


JSONbored (changes requested):
Hmm, not really a fan of the new design/direction - could you maybe play with this to improve its design/layout? The whole table controls section currently just feels awkward/neglected (pre-existing problems, but nevertheless needs to be fixed). 

I'll leave this PR open instead of closing so you can do this. Thank you.

nickmopen:
Iterated further on the toolbar layout (`fe98c51`): the controls now stretch to fill their row so it stays justified (flush on both ends) at every width, with the search always the widest control on its own line when it can't share the row —

- **Wide (desktop/tablet):** one row — search + Kind / Authority / Sort
- **Mid widths:** search on its own full-width line over a justified row of the three dropdowns
- **Narrow (mobile):** search full-width, then the three dropdowns stacked evenly

`SelectFilter` gained an optional `fill` prop (default off, no change to existing callers) so its dropdown can grow to fill a flex track. Screenshots in the PR description are updated across all 3 viewports × both themes. Full CI green locally.

JSONbored (approved):

PR #4795, opened 2026-07-10
fix(ui): make the /endpoints table horizontally scrollable on mobile

### LINKED ISSUE #3931: /endpoints ships an un-scrollable 8-column table on mobile by default, with no card fallback
## Context

`apps/ui/src/routes/endpoints.tsx:79` defaults the route's `view` search param to `"table"` (`fallback(z.enum(["table","grid"]), "table").default("table")`). When `view !== "grid"` (the default for every first-time visitor), the route renders the table branch at line 796: `<div className="rounded border border-border bg-card overflow-x-clip"><table className="w-full text-sm">`, with 8 columns (Netuid, Kind, URL, Provider, Region, Health, Latency, Probed — including a `max-w-[36ch]` URL column) and no `md:hidden` mobile-card fallback of its own. The mobile-friendly alternative already exists (the `grid` view at line 696, `grid gap-3 sm:grid-cols-2 lg:grid-cols-3`) — but it's opt-in via a toggle a first-time mobile visitor has no reason to know exists. Unlike `ListShell`-mediated tables elsewhere in the app, this one uses `overflow-x-clip` with no scroll alternative at all, at any viewport.

## Requirement

A first-time mobile visitor to `/endpoints` must land on a usable view — either the table defaults to `grid` below a mobile breakpoint, or the table itself gets the same card-fallback/scroll treatment as other tables in the app.

## Deliverable

- **File to change**: `apps/ui/src/routes/endpoints.tsx`, the `view` default (line 79) and/or the table branch (line 796).
- **What "done" looks like concretely** — pick one: (a) default `view` to `"grid"` below a chosen breakpoint (matching how other routes gate a mobile-card fallback), while leaving desktop's default as `"table"`; or (b) give the table branch its own mobile-card fallback / horizontal scroll treatment consistent with the fix landing in the sibling `ListShell` overflow issue. Either approach must ensure a mobile visitor who never touches the view toggle still gets a usable, non-clipped layout.

## Acceptance criteria

- [ ] `endpoints.tsx` appears in the diff
- [ ] A first-time mobile visitor (no query params) to `/endpoints` sees a fully usable, non-clipped view of the endpoint list
- [ ] The existing `grid`/`table` toggle continues to work and persist via the URL exactly as it does today
- [ ] Desktop rendering of the `table` view is unchanged
- [ ] Includes a before/after screenshot table at mobile (375px), both themes, per the repo's UI-PR requirement

## Non-goals

- Don't remove the `table` view option or change its columns — desktop users who prefer it keep it.
- Don't couple this fix to the `ListShell` overflow-x-clip issue's landing — this route doesn't use `ListShell`, so it can ship independently.

## Size

Part of #2542.

davion-knight:
## Summary

`/endpoints` defaults to `view=table`, and the table branch wrapped its 8-column table in `overflow-x-clip`. At mobile width the trailing columns (Health, Latency, Probed — plus the wide `max-w-[36ch]` URL column) were **clipped and unreachable**, with no card fallback and no scroll: a first-time mobile visitor lands on an unusable view.

## Change

- One file: `apps/ui/src/routes/endpoints.tsx` (table branch only).
- Wrap the `<table>` in an inner `overflow-x-auto overflow-y-clip` scroll container inside an `overflow-hidden` card shell — the same treatment `ListShell` uses (see merged #4739).
- **Why `overflow-y-clip`, not a plain `overflow-x-auto` on the card:** this table's `<thead>` is `sticky top-[6.75rem]` (unlike the non-sticky `overflow-x-auto` pool tables above it). A vertical scroll container would break that sticky-against-page-scroll behavior; `overflow-y-clip` keeps it intact while letting the header + body scroll horizontally together.
- No query/logic change. Desktop and tablet rendering are unchanged (the table already fits); only narrow viewports gain reachability.

Closes #3931

## Screenshots

Captured at three viewports × both themes on `/endpoints` (default table view). Theme forced via `localStorage.setItem("mg-theme", …)` before navigation. **Before** = `upstream/main`; **After** = this branch.

> **Where to look:** the fix is a mobile/tablet **scroll affordance** — at rest (scroll-x 0) the before/after render identically, but on `before` the table's trailing columns are clipped with no way to reach them, while on `after` the table sits in a horizontally-scrollable container. Desktop rows are included to show no regression at width.

| --- | --- | --- |
| Desktop · Light |  |  |
| Desktop · Dark |  |  |
| Tablet · Light |  |  |
| Tablet · Dark |  |  |
| Mobile · Light |  |  |
| Mobile · Dark |  |  |

## Validation

- `tsc --noEmit` clean; `prettier --check` clean; `eslint` no new errors (only pre-existing react-refresh warnings).

### DIFF
--- a/apps/ui/src/routes/endpoints.tsx
+++ b/apps/ui/src/routes/endpoints.tsx
@@ -973,173 +973,175 @@ function EndpointsTable() {
               })}
             </div>
           ) : (
-            <div className="rounded border border-border bg-card overflow-x-clip">
-              <table className="w-full text-sm">
-                <thead className="sticky top-[6.75rem] z-10 bg-surface/95 backdrop-blur supports-[backdrop-filter]:bg-surface/85 text-[10px] font-mono uppercase tracking-widest text-ink-muted shadow-[0_1px_0_0_var(--border)]">
-                  <tr>
-                    <Th
-                      label="Netuid"
-                      k="netuid"
-                      sortKey={search.sort}
-                      sortOrder={search.order}
-                      onSort={toggleSort}
-                    />
-                    <Th
-                      label="Kind"
-                      k="kind"
-                      sortKey={search.sort}
-                      sortOrder={search.order}
-                      onSort={toggleSort}
-                    />
-                    <th className="px-3 py-2 text-left">URL</th>
-                    <Th
-                      label="Provider"
-                      k="provider"
-                      sortKey={search.sort}
-                      sortOrder={search.order}
-                      onSort={toggleSort}
-                    />
-                    <Th
-                      label="Region"
-                      k="region"
-                      sortKey={search.sort}
-                      sortOrder={search.order}
-                      onSort={toggleSort}
-                    />
-                    <Th
-                      label="Health"
-                      k="health"
-                      sortKey={search.sort}
-                      sortOrder={search.order}
-                      onSort={toggleSort}
-                      align="center"
-                    />
-                    <Th
-                      label="Latency"
-                      k="latency"
-                      sortKey={search.sort}
-                      sortOrder={search.order}
-                      onSort={toggleSort}
-                      align="right"
-                    />
-                    <Th
-                      label="Probed"
-                      k="probed"
-                      sortKey={search.sort}
-                      sortOrder={search.order}
-                      onSort={toggleSort}
-                      align="right"
-                    />
-                  </tr>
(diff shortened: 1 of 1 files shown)


JSONbored (commented):
This requires actual recordings/GIFS to satisfy the evidence. The AI skill should mention how to handle this.

Issue #4984, opened 2026-07-12
infra(realtime): alerter — trigger eval + webhook/email/Telegram/Discord delivery (5 of 5)


JSONbored:
Part of #2114 (Durable Object firehose). Depends on #4982 (the Durable Object hub) as its event source. This is the "plus alerter" half of #2114's original scope — a separate concern from the raw firehose transports (#4982/#4983): evaluating user-defined trigger conditions against the live stream and delivering notifications.

## Requirements

- A trigger-definition model: what a user can subscribe to (e.g. "netuid=X AND event_kind=Y", "account=Z", "amount_tao > N") and where definitions are stored — likely a new Postgres table (`deploy/postgres/schema.sql`) rather than D1, per ADR 0014's standing preference for Postgres as the durable store for anything new.
- Trigger evaluation: subscribes to the Durable Object hub (#4982) like any other consumer, evaluates each event against all active trigger definitions, and enqueues a delivery job for any match.
- Delivery channels: webhook (reuse the EXISTING webhook delivery infrastructure already live at `/api/v1/webhooks/subscriptions` — check `src/webhooks.mjs`/`workers/request-handlers/*webhook*` for the current retry/signing convention before building a second one), email, Telegram, Discord. Check whether `alertmanager-discord` (already running on the indexer box per `docker ps`) is reusable/extendable for the Discord leg specifically, rather than building a parallel Discord integration from scratch.
- Rate-limiting / dedup so a burst of matching events (e.g. many transfers in one block) doesn't spam a single subscriber — batch or debounce per the existing webhook infrastructure's own conventions if it already solves this.

## Deliverables

- [ ] Trigger-definition storage + CRUD (API route(s), schema, tests)
- [ ] Evaluation loop subscribing to the DO hub and matching events to active triggers
- [ ] Delivery wired for all 4 channels, reusing existing webhook infra where it already exists rather than duplicating it
- [ ] Rate-limit/dedup behavior verified under a burst

## Expected outcome

The "plus alerter" deliverable from #2114's original scope — users can define "notify me when X happens on-chain" and receive it via their channel of choice, completing #2114 in full.

JSONbored:
Shipped, in four parts, each its own merged PR:

- **Part 1** (#5013) — `chain_alert_triggers` Postgres table + CRUD API (`/api/v1/alerts/triggers`), bearer `owner_token` ownership model, no public GET (a trigger's `destination` can itself be a capability credential).
[...]

PR #5057, opened 2026-07-13
fix(alerts): prevent ID enumeration by hiding private trigger existence

JSONbored:
### Motivation
- Single-trigger alert routes used sequential BIGSERIAL IDs and returned `404` for nonexistent IDs but `403` for rows that exist with a missing/wrong owner token, allowing unauthenticated enumeration of private trigger existence.

### Description
- Add a shared `alertTriggerNotFoundResponse()` helper and change `requireAlertTriggerOwner` to return that indistinguishable `404` response for failed owner-token checks.
- Replace the previous direct `writeJson({ error: "no such trigger" }, 404)` checks in the GET/PATCH/DELETE handlers with the helper so missing rows and failed auth yield the same response.
- Update `tests/alert-triggers-route.test.mjs` to assert that missing or incorrect owner tokens produce `404` with body `{ error: "no such trigger" }` while authorized requests still return `200` with the owner view.

### Testing
- Ran `npm test -- tests/alert-triggers-route.test.mjs` and all tests passed (32 tests). 
- Ran `npm run lint` and `npm run format:check`, and both completed successfully. 
- Ran `git diff --check` to validate whitespace/checks and it reported no issues.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a5464708368832ca8048451aa9050e1)

### DIFF
--- a/tests/alert-triggers-route.test.mjs
+++ b/tests/alert-triggers-route.test.mjs
@@ -246,20 +246,22 @@ test("get: 404 when no such trigger exists", async () => {
   assert.equal(res.status, 404);
 });
 
-test("get: 403 when the owner token header is entirely absent", async () => {
+test("get: 404 when the owner token header is entirely absent", async () => {
   mockQueue.current.push([row()]);
   const res = await fetch(req("/api/v1/alerts/triggers/1"));
-  assert.equal(res.status, 403);
+  assert.equal(res.status, 404);
+  assert.deepEqual(await res.json(), { error: "no such trigger" });
 });
 
-test("get: 403 when the owner token is present but wrong", async () => {
+test("get: 404 when the owner token is present but wrong", async () => {
   mockQueue.current.push([row()]);
   const res = await fetch(
     req("/api/v1/alerts/triggers/1", {
       headers: { "x-alert-trigger-owner-token": "wrong" },
     }),
   );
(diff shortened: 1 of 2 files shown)


JSONbored:
Closing as superseded — this exact fix already shipped on main.

`requireAlertTriggerOwner` (workers/data-api.mjs) already returns the same `404 {"error": "no such trigger"}` for both a missing row and a failed owner-token check — landed as part of #4984's Part 4 adversarial-review remediation (the ADR/docs call it out explicitly: "Existence oracle via differentiated 403/404... `requireAlertTriggerOwner` now returns the same 404 either way"). `tests/alert-triggers-route.test.mjs` on main already asserts this for GET/PATCH/DELETE:

- `get: 404 (not 403) when the owner token header is entirely absent -- indistinguishable from a nonexistent id`
- `get: 404 (not 403) when the owner token is present but wrong -- prevents an existence oracle over sequential ids`
- `update: 404 (not 403) when the owner token is missing or wrong -- prevents an existence oracle over sequential ids`
- `delete: 404 (not 403) when the owner token is missing or wrong -- prevents an existence oracle over sequential ids`

This PR's branch was cut before that landed, so its diff (403→404, same behavior) is a no-op against current main. No action needed here.

Issue #4772, opened 2026-07-10
Fully retire D1 for chain data — delete tables, prune logic, ingest endpoints, and the D1 fallback (gated on #4771)


JSONbored:
## Context

The end state for `metagraphed-health` D1, per ADR 0014 and the #2108 infra epic, is full retirement for chain data -- D1 was always meant to be a temporary hot cache during the Postgres cutover, not a permanent architecture. Tonight's work (#2112, #4746/#4750) got blocks/extrinsics/account_events flipped to Postgres-primary, and #4771 (once built) will do the same for the metagraph/neuron pipeline. This issue tracks the actual deletion once every dependency clears -- it is explicitly a "once X, Y, Z are true" issue, not startable yet.

D1 hosts tables in genuinely different categories, and this issue is scoped ONLY to the chain-data ones -- the health/registry-monitoring tables (`surface_checks`, `surface_status`, `surface_uptime_daily`, `rpc_proxy_events`) are unrelated to chain data and are explicitly NOT part of D1 retirement; they're registry/health-prober telemetry with their own, already-working retention and no relationship to the Postgres cutover. Do not delete them under this issue.

Similarly, `account_events_daily` and `subnet_snapshots` are explicitly, deliberately retained indefinitely per their own migration comments (`migrations/0009_account_events.sql:33-35`, `migrations/0002_analytics.sql:9-10`) -- low-volume, long-term-by-design tables, not a capacity risk at any realistic timescale (~2GB/decade and ~4MB/decade respectively at current growth). These are NOT part of D1 retirement either; do not touch them under this issue.

## Requirements

- **Gate 1:** #4771 (metagraph/neuron Postgres pipeline) must be built, verified, and flipped -- `neurons`/`neuron_daily` are the only chain-data tables still D1-only after tonight's work.
- **Gate 2:** #4770 (neuron_daily prune bug) -- moot once #4771 ships and D1's copy is deletable, but if #4771 is delayed, #4770 should still be fixed independently so D1 doesn't hit its cap again in the meantime.
- **Gate 3:** blocks/extrinsics/account_events Postgres serving must be proven reliable over a real sustained window (not just tonight's spot-verification) before the D1 fallback is removed -- per ADR 0014's own stricter pre-launch criteria, this is explicitly a lower bar tonight (no real users yet) than what should gate a permanent code deletion.
[...]

PR #5074, opened 2026-07-13
fix(data-api): fall back on postgres block misses

JSONbored:
### Motivation

- Postgres was flipped on for block detail serving but a successful Postgres response with `block:null` suppressed the D1 fallback, causing a correctness regression for known Postgres coverage gaps (e.g. 8471001-8479999). 
- The intent is to treat a Postgres `block:null` as a semantic miss so the existing D1 fallback can still return retained rows when Postgres is incomplete.

### Description

- Change `handleBlock` in `workers/request-handlers/entities.mjs` to treat a Postgres response whose `block` is `null` as a miss by setting `data = pg?.block === null ? null : pg;` so the D1 branch executes when appropriate.
- Add a regression test in `tests/request-handlers-entities.test.mjs` (`handleBlock: flag=postgres falls back to D1 on block:null miss`) which verifies that a successful DATA_API `block:null` response does not suppress the D1 query and that D1 is queried and returns the block row.
- Run code formatting on the modified files to satisfy the repo style rules.

### Testing

- Ran `npm install` successfully to prepare the environment.
- Ran the focused test file with `npm test -- tests/request-handlers-entities.test.mjs` and observed all tests in that file pass (`505 tests` passed).
- Ran formatting checks (`npx prettier --write` and `npm run format:check`) and confirmed files now match Prettier style.
- The added regression test verifies the specific previously-broken behavior and passed in the test run.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a54667f902c832cbd340ac142ba2124)

### DIFF
--- a/tests/request-handlers-entities.test.mjs
+++ b/tests/request-handlers-entities.test.mjs
@@ -6193,6 +6193,33 @@ describe("D1 -> Postgres serving-cutover flag (#4656 followup)", () => {
     assert.equal(body.data.block.block_number, BLOCK_NUM);
   });
 
+  test("handleBlock: flag=postgres falls back to D1 on block:null miss", async () => {
+    const { env, captures } = dbWith({ blockDetail: blockRow() });
+    env.METAGRAPH_BLOCKS_SOURCE = "postgres";
+    env.DATA_API = dataApi(
+      Response.json({
+        schema_version: 1,
+        ref: String(BLOCK_NUM),
+        block: null,
+        prev_block_number: null,
+        next_block_number: null,
+      }),
+    );
+    const body = await json(
+      await handleBlock(
+        req(`/api/v1/blocks/${BLOCK_NUM}`),
+        env,
+        String(BLOCK_NUM),
+      ),
+    );
+    assert.equal(body.data.block.block_number, BLOCK_NUM);
+    assert.ok(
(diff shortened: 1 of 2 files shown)


JSONbored:
Closing — this fix targets a code path that no longer exists on `main`.

`handleBlock` previously fell through to a D1 query when Postgres missed. That D1 fallback was removed by #4772 (D1 chain-data full retirement, merged 2026-07-11) — `workers/request-handlers/entities.mjs`'s current `handleBlock` no longer references D1 at all:

```js
export async function handleBlock(request, env, ref) {
  // #4909 D1 retirement: blocks' D1 write path is retired (#4772) and the
  // table is dropped in production, so a D1 query here would always miss.
  const data =
    (await tryPostgresTier(env, request, "METAGRAPH_BLOCKS_SOURCE")) ??
    buildBlock(undefined, ref);
  ...
```

Postgres is now the sole source for block data, so a `block:null` response from it is authoritative — not a "semantic miss" to route around. Treating it as one would just mask a real not-found as an in-flight D1 lookup that no longer happens. That's also why this PR shows as `CONFLICTING` — its diff assumes a function shape that's already been restructured out from under it.

Closing as superseded by #4772, not a regression to carry forward.

PR #5163, opened 2026-07-13
fix(data-api): ignore client-provided snapshot_date in neuron backfill validation

JSONbored:
### Motivation
- Backfill clients (scripts/backfill-*.py) include a `snapshot_date` field in each neuron row but the server-side validator only allows keys in `NEURON_INSERT_COLUMNS`, causing valid backfill requests to be rejected before the endpoint derives its own `snapshot_date` from `captured_at`.

### Description
- Add `stripClientSnapshotDate` to `workers/data-api.mjs` and strip any client-provided `snapshot_date` from incoming rows before validation in **both** `handleNeuronsSync` and `handleNeuronDailyBackfill` (the original diff only covered the backfill handler; a review caught that the neurons-sync handler had the identical bug and was unpatched -- fixed in a follow-up commit).
- Continue to derive and store `snapshot_date` on the server via `neuronSyncSnapshotDate(row.captured_at)` so stored dates remain authoritative.
- Regression tests in `tests/data-api.test.mjs`: one per handler asserting a stale client `snapshot_date` is ignored and the derived one is stored, plus malformed-row (null / non-object primitive / nested array) coverage for `stripClientSnapshotDate`'s own guard.

### Testing
- `npm run build && npm run validate`
- `npm run test:coverage` (full unsharded run, clean)
- `node scripts/scan-public-safety.mjs` (clean)
- `npx prettier --check` / `npx eslint` on both changed files

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a552e9e93b8832cb767d111b41ac744)

### DIFF
--- a/tests/data-api.test.mjs
+++ b/tests/data-api.test.mjs
@@ -2834,6 +2834,26 @@ test("backfill-neuron-daily accepts a null stake_tao (backfill-neuron-history.py
   expect(body.neuron_daily_written).toBe(1);
 });
 
+test("backfill-neuron-daily ignores client snapshot_date and derives it from captured_at", async () => {
+  const capturedAt = Date.parse("2026-05-29T12:34:56.000Z");
+  const res = await postNeuronDailyBackfill(
+    [
+      neuronSyncRow({
+        captured_at: capturedAt,
+        snapshot_date: "1999-01-01",
+      }),
+    ],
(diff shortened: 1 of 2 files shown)


JSONbored:
@loopover review

JSONbored:
@loopover review

PR #5441, opened 2026-07-14
fix(ui): surface coverage-matrix gap count on narrow viewports

### LINKED ISSUE #5310: /gaps coverage matrix shows a false "all present" on mobile
## Context

**Severity (per audit):** Critical

`apps/ui/src/components/metagraphed/analytics/coverage-matrix.tsx:92-207` renders a Subnet × 9-kind (docs, repo, openapi, endpoint, dashboard, data, sdk, example, rpc) matrix in an `overflow-x-auto` container with a sticky first column. At 375px width only the first ~4 columns (docs/repo/openapi/endpoint) fit on screen, and every visible cell in that range is green ("present"), with no fade, arrow, or "scroll →" hint indicating more columns exist. The columns that are actually the point of this widget — dashboard/data/sdk/example/rpc, i.e. the real gaps — are scrolled off-screen and invisible, so a mobile user sees what looks like full coverage when it isn't.

## Requirements

Add a visible horizontal-scroll affordance (fade mask and/or explicit scroll chevron/hint) to the matrix container, or default to a compact per-subnet "N missing" summary on narrow viewports instead of the full 9-column grid.

## Deliverables

- [ ] Pick one approach: (a) add a scroll-cue affordance to the existing matrix, or (b) add a compact mobile-only "N missing" summary view
- [ ] Implement the chosen approach in `coverage-matrix.tsx`
- [ ] Verify at 375px viewport that a user can tell more columns/data exist beyond the visible ones

## Expected Outcome

A mobile user viewing /gaps can immediately tell whether a subnet has missing surfaces, without needing to blindly scroll to discover hidden red cells.

## Links & Resources

`apps/ui/src/components/metagraphed/analytics/coverage-matrix.tsx:92-207`

_Filed from a 4-pass apps/ui UI/UX audit (entity-detail, list/index, dashboard/utility, global-nav+mobile), screenshot-confirmed at 375×812 / 768×1024 / 1280×800 against the live dev server with real production data, contributed 2026-07-14._

pro3958:
## Summary

The `/gaps` coverage matrix (`apps/ui/src/components/metagraphed/analytics/coverage-matrix.tsx`) renders a Subnet x 9-kind grid inside an `overflow-x-auto` container with a sticky first column. At 375px only the first four kind columns (docs/repo/openapi/endpoint) fit on screen, and those cells are usually green, so a mobile user read apparent full coverage. The dashboard/data/sdk/example/rpc columns -- the actual gaps this widget exists to show -- were scrolled off-screen with no fade, chevron, or scroll hint, so hidden red cells looked like no cells.

## What Changed

Added a mobile-only summary line inside the sticky first column, which stays visible at every width: `N missing` (red) when the row has gaps, `complete` (green) when it doesn't. It reuses the row's existing `missingCount`, so no new data or computation. The line is `sm:hidden`, so the full 9-column grid is untouched at `sm` and up. Chose the compact summary (deliverable option b) over a scroll-cue affordance because it answers the user's actual question -- does this subnet have gaps -- without requiring a horizontal scroll to find out.

## Registry Safety

- [x] Links a tracked, currently-open issue (`Fixes #5310`) -- required.
- [x] Generated artifacts were produced by repo scripts, not hand-edited. (N/A -- no generated artifacts touched.)
- [x] Public API/OpenAPI/schema changes are intentional and documented. (N/A -- frontend-only change under `apps/ui/`, no contract change.)

## Validation

Frontend-only change; the backend registry validators do not apply. `npm test` passes with no new failures against main.

## Notes

This is a visual change to `apps/ui/`, so per the repo's contribution rules it needs a before/after screenshot table and a manual review at 375px before merge -- hence the draft. The behavior to confirm: at 375px each subnet row shows its missing-count next to the name without scrolling; at `>=640px` the row is unchanged.

Fixes #5310

### DIFF
--- a/apps/ui/src/components/metagraphed/analytics/coverage-matrix.tsx
+++ b/apps/ui/src/components/metagraphed/analytics/coverage-matrix.tsx
@@ -162,6 +162,18 @@ export function CoverageMatrix({ topN = 24 }: { topN?: number }) {
                     <span className="font-mono text-[10px] text-ink-muted">SN{r.netuid}</span>
                     <span className="truncate max-w-[160px]">{r.name}</span>
                   </Link>
+                  {/* The kind columns scroll off-screen on narrow viewports, so a
+                      mobile user can't see the red "missing" cells that are the
+                      point of this widget. Surface the per-subnet gap count here,
+                      in the always-visible sticky column (#5310). */}
+                  <span
+                    className={classNames(
+                      "mt-0.5 block text-[10px] tabular-nums sm:hidden",
+                      r.missingCount > 0 ? "text-health-down" : "text-health-ok",
+                    )}
+                  >
+                    {r.missingCount > 0 ? `${r.missingCount} missing` : "complete"}
+                  </span>
                 </td>
                 {KINDS.map((k) => {
                   const cell = r.cells[k];


pro3958:
Root cause: on narrow viewports the kind columns scroll out of view, so the red 'missing' cells — the entire point of the widget — are invisible and the matrix reads as 'all present'. This surfaces the per-subnet gap count in the sticky column, which is always on screen.

Issue #5776, opened 2026-07-15
infra(data): reconciler — diff folded balances/stake/yield vs runtime truth, alert on drift


JSONbored:
## Context

Split out from #2115, whose "exporter" half is superseded by #2538 (nightly Parquet bulk-dump — same deliverable, more concretely scoped). This issue keeps #2115's other, genuinely distinct half: a drift-detection reconciler.

Portfolio/yield numbers served today are computed from folded (aggregated/derived) state — `neurons`, `account_position_daily`, `subnet_snapshots`, etc. A bug in the aggregation logic, a missed sync tick, or a schema drift could silently diverge those folded values from runtime truth (a direct chain read) without any user-visible error — the API would just serve a confidently wrong number.

## Deliverable

A scheduled job (box-side systemd timer, matching this repo's established pattern for recurring Postgres-reading jobs — see `scripts/data-refresh-node-entrypoint.sh` and sibling entrypoint scripts under `scripts/`) that:
- Periodically samples a subset of folded balances/stake/yield rows from Postgres
- Re-derives the same values from a direct runtime chain read (matching the live-state fetch scripts' `SUBTENSOR_RPC_URL` convention, ADR 0012) for the sampled rows
- Diffs folded vs. runtime-truth and alerts (existing `LIVE_ALERT_WEBHOOK_URL`/Discord webhook pattern used elsewhere on the box) when drift exceeds a defined tolerance
- Does NOT auto-correct drift — alert-only, so a human decides whether it's a real bug or expected lag

## Why box-side, not Railway

#2115 originally scoped this as a "Railway cron" — Railway's Postgres/Redis/indexer no longer exist (#2956); the self-hosted indexer box is the permanent core and already runs every other recurring Postgres-reading job via systemd timers.

## References

- `scripts/data-refresh-node-entrypoint.sh` (box-side job entrypoint pattern)
- ADR 0012 (`docs/adr/0012-*.md`, `SUBTENSOR_RPC_URL` convention for chain-direct reads)
- #2538 (the exporter half of the original #2115, now separately scoped)

Part of #2108.

*maintainer-only.*

JSONbored:
Shipped and verified live end-to-end:

[...]

Issue #2956, opened 2026-07-04
infra(core): cut Postgres/Redis/indexer over from Railway (interim) to the self-hosted box (final)


JSONbored:
## Context

ADR 0013's own topology table already states the end-state: **"Interim: Postgres/Redis/indexer run on Railway until the box is live."** The box is now live — the self-hosted archive node (`meta-archive-01-us-nyc1`) and indexer node (`meta-indexer-01-us-lax1`, per JSO-2518's 2-box split) exist, and a self-hosted `indexer-rs` is already live-following the chain head into a self-hosted Postgres/TimescaleDB there today, in parallel with Railway's copy.

No existing issue tracks the *act* of moving off Railway's Postgres/Redis/indexer specifically — #2109 provisioned the Railway side (closed, done, correctly the interim step), #2112 covers D1 → Postgres serving cutover generically (which database it reads from isn't specified), and #2116 covers decommissioning legacy *ingestion* (GitHub poller, `metagraphed-streamer`), not Railway's Postgres/Redis/indexer-rs specifically. This issue is the missing piece: the interim → final infrastructure cutover ADR 0013 already calls for.

**Why this matters for cost, not just architecture:** Railway's `metagraphed-core` Postgres + Redis + indexer-rs are the highest structural cost item in the whole stack — persistent, always-on compute plus two database volumes that keep billing even if merely stopped (not deleted). They are currently running as an exact functional duplicate of what the self-hosted box now does. Verified 2026-07-04: `metagraphed-data-api`'s Cloudflare Hyperdrive binding (id `d8c62acb5520402b82d039ec482f7cc3`) still points at Railway's Postgres proxy (`switchback.proxy.rlwy.net:39229`) and the live production API is genuinely serving from it right now (confirmed via a real `curl` against `api.metagraph.sh`, not just config inspection) — so this is not yet a safe no-op decommission; the serving path has to move first.

## Requirements — gated, do not skip steps

Per ADR 0013 §Sequencing (⛔ = verification gate, not to be run unsupervised):

1. **Verify capture parity**: self-hosted Postgres row counts/content vs Railway's Postgres vs D1, per the comparison approach in `deploy/hyperdrive-cutover.md` step 0, before trusting the self-hosted copy as a replacement source of truth.
[...]

Issue #2538, opened 2026-07-01
Nightly Parquet bulk-dump of core registry + chain tables to R2


JSONbored:
A scheduled (cron/Worker or CI) job that snapshots the core datasets (subnets, economics, endpoints, neurons/metagraph, blocks, extrinsics, chain-events daily rollups) to columnar Parquet and uploads dated objects to an R2 prefix (e.g. metagraph/bulk/parquet/{date}/), with a small manifest for discovery. Maintainer-only: needs an R2 bucket binding, cron scheduling/wrangler config, a Parquet writer dependency (npm audit), and D1/Postgres bulk-read access — infra + creds + scheduling that a contributor PR can't land under the gate.

## Deliverables
- [ ] Add a scheduled job (extend `handleScheduled` or a CI workflow) that dumps the core tables to Parquet nightly
- [ ] Upload dated Parquet objects to an R2 prefix + write/update a `bulk-manifest.json` (files, row counts, generated_at, schema versions)
- [ ] Add the R2 binding + cron trigger to wrangler config; run `npm run cf-typegen` and commit the regenerated types
- [ ] Vet the Parquet writer dependency with `npm audit --audit-level=moderate`; document retention/pruning of old dumps
- [ ] Publish the bulk-download path in the docs/agent-resources so data users can find it

## References
- `workers/api.mjs:732 (handleScheduled — cron entry point)`
- `src/contracts.mjs:1225 (r2-manifest artifact — manifest precedent)`
- `src/contracts.mjs:1075 (chain-events / all-events tier — a bulk source)`
- `workers/request-handlers/analytics.mjs (D1 read patterns to reuse for bulk reads)`

---
_size: M · maintainer-only (owner)_

JSONbored:
Closing as shipped via #5784 (\"feat(data-refresh): nightly Parquet bulk-dump to R2\") — the PR title referenced this issue but didn't trigger GitHub's auto-close.

[...]

Issue #2115, opened 2026-06-27
infra(data): exporter (Parquet/CSV → R2) + reconciler (folded-state drift)


JSONbored:
Railway crons: `exporter` (DuckDB → partitioned Parquet/CSV in a public R2 prefix + a `/api/v1/datasets` manifest) and `reconciler` (diff folded balances/stake/yield vs runtime truth, alert on drift). **Deliverable:** public dataset exports + a drift-detection harness gating trust in portfolio/yield numbers.

Part of #2108. See ADR 0013 (`docs/adr/0013-hybrid-deployment-topology.md`) + runbook (`deploy/README.md`).

JSONbored:
Splitting this into two more concretely-scoped issues and closing the original:

- The **exporter** half (DuckDB → partitioned Parquet/CSV in R2 + manifest) is superseded by #2538, which describes the same deliverable more concretely (dated R2 prefix, `bulk-manifest.json`, specific wiring into `handleScheduled`/CI).
- The **reconciler** half (diff folded state vs. runtime truth, alert on drift) is genuinely distinct and unbuilt — split out to #5776.

Also correcting this issue's stale framing for the record: it originally scoped both as "Railway crons." Railway's Postgres/Redis/indexer no longer exist (#2956, closed) — the self-hosted indexer box is the permanent core, and both successor issues target it instead, matching every other recurring Postgres-reading job already running there.

PR #5784, opened 2026-07-15
feat(data-refresh): nightly Parquet bulk-dump to R2 (#2538)

JSONbored:
## Summary
- Adds an `export-parquet` STEP to the box-side `data-refresh-node` entrypoint: reads core registry + chain tables directly from the box's own co-located Postgres databases via DuckDB's postgres extension, writes ZSTD-compressed Parquet, uploads to R2 under a dated prefix + a `latest/` alias, with a discovery manifest (`bulk-manifest.json`) matching `scripts/r2-manifest.mjs`'s shape.
- Successor to #2115's "exporter" half (closed, split into this + #5776's reconciler): that issue scoped a Railway cron per ADR 0013, but Railway's Postgres/Redis/indexer no longer exist (#2956). This targets the self-hosted box instead, reusing the existing `data-refresh-node` Docker image/role rather than a new one -- the only per-job addition is `--network host` (the box hosts both source databases locally, so no Hyperdrive/Worker hop is needed for reads).
- R2 upload reuses the existing `wrangler r2 object put --remote` mechanism (`CLOUDFLARE_API_TOKEN`/`CLOUDFLARE_ACCOUNT_ID`, matching `scripts/r2-upload.mjs`) -- confirmed no S3-compatible credential exists anywhere in this codebase, and this doesn't introduce one.
- Table scope is deliberately the daily-rollup tables (`subnets`, `providers`, `surfaces`, `subnet_snapshots`, `neurons`, `neuron_daily`, `blocks`, `extrinsics`, `account_events_daily`) per #2538's explicit list -- not the raw high-volume hypertables (`account_events` 40M+ rows, `chain_events` 17M+ rows), which would need a fundamentally different incremental-export design.
- A real bug found during testing: DuckDB's own `ATTACH` failure error embeds the full libpq connection string (password included) verbatim. Fixed with a scrubbing wrapper so a connection failure in production never leaks a plaintext password into the systemd journal.

## Deliberately deferred
Public discovery (making `metagraph/bulk/parquet/*` reachable at `api.metagraph.sh`) is a follow-up, not bundled here: the Worker only serves contract-declared R2 paths (`src/artifact-storage.mjs`'s `R2_ONLY_PATTERNS`), so wiring this up needs a schema/contract change + `openapi.json` regen -- better scoped once there's real data in R2 to point at, not spliced into the data-production job itself.

Retention/pruning of old dated runs is also deferred and documented in the script's header: `wrangler r2 object` has no list/delete-by-prefix command, so automated pruning would need direct Cloudflare R2 REST API calls rather than the wrangler CLI this job otherwise relies on -- a delete-capable path against a public bucket deserves its own careful pass. `metagraph/bulk/parquet/latest/` (overwritten every run) is the bounded-size primary access point in the meantime.

## Test plan
- [x] `npx eslint`, `npx prettier --check`, `shellcheck`, `node scripts/scan-public-safety.mjs`, `npm run validate` all pass
- [x] `@duckdb/node-api` vetted: MIT license, official DuckDB repo, 0 `npm audit` vulnerabilities in isolation
- [x] Verified ATTACH + COPY + Parquet readback against the box's real Postgres via SSH tunnel (small table) and directly on the box (large table, `--network host`, zero network hop): `extrinsics` (3.16M rows) ~67s, `account_events_daily` (115K rows) ~133ms -- comfortably within the existing 900s `TimeoutStartSec`
- [x] Confirmed R2 write access works with a real test object (uploaded + deleted)
- [ ] Live verification once the box-side Ansible role picks up the new job (follow-up in the private infra repo, needs a newly-provisioned R2-scoped Cloudflare API token)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,6 +15,7 @@
         "packages/ui-kit"
       ],
       "dependencies": {
+        "@duckdb/node-api": "^1.5.4-r.1",
         "@noble/hashes": "^2.2.0",
         "graphql": "^17.0.2",
         "graphql-ws": "^6.1.0",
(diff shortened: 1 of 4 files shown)


JSONbored:
Correction for the record: the deployed config (metagraphed-infra#26) does NOT use \`--network host\` as described above. A security review correctly flagged that as broader privilege than needed (full host network access vs. reaching exactly the two Postgres containers this job talks to). The deployed version joins the two specific docker-compose networks by name instead, with the container names + internal port 5432 passed via \`REGISTRY_PG_HOST\`/\`INDEXER_PG_HOST\` env vars -- verified working the same way. No changes needed in this repo; `scripts/export-parquet.mjs`'s env-var-driven host/port design already supported this without modification.

PR #5804, opened 2026-07-15
feat(mcp): add query_graphql bridge tool for /api/v1/graphql

### LINKED ISSUE #5591: feat(mcp): add an MCP bridge tool for /api/v1/graphql
## Context

`/api/v1/graphql` (the generic GraphQL endpoint) has no MCP passthrough tool. An agent can't query GraphQL through MCP at all — it can only use the individual REST-mirrored MCP tools already registered (get_subnet, list_subnets, etc.), which don't cover everything GraphQL's schema can express (arbitrary field selection, nested relations in one round-trip).

Not to be confused with #5573/#5574/#5575/#5580 (GraphQL-parity issues), which add REST-mirror fields TO GraphQL — this issue is the reverse direction: bridging GraphQL access INTO MCP.

## Requirements

- Read-only (GraphQL here is a query endpoint, no mutations) — low-risk, no auth/abuse design question like the webhook/alert-trigger tools
- A `query_graphql` (or similar name) MCP tool that accepts a GraphQL query string (+ optional variables) and forwards it to `/api/v1/graphql`, returning the result
- Cap query complexity/depth the same way the REST GraphQL endpoint itself already does, if it has such a cap — don't let this tool bypass any existing GraphQL-side protection
- Tool description should explain when to prefer this over the individual REST-mirrored tools (arbitrary field selection / nested queries in one round-trip) vs when a dedicated tool is simpler (single well-known lookup)

## Links & Resources

- `/api/v1/graphql` (the REST endpoint to bridge)
- `src/mcp-server.mjs` (tool registrations — zero GraphQL-bridge tool currently)

davion-knight:
Adds a read-only \`query_graphql\` MCP tool that bridges the GraphQL endpoint into MCP: an agent can run an arbitrary GraphQL query (with optional variables) and get the \`{ data, errors }\` result, covering what the individual REST-mirrored tools cannot express (arbitrary field selection, nested relations in one round-trip).

Closes #5591.

## What
- New \`query_graphql\` tool in \`src/mcp-server.mjs\` (\`inputSchema\`: required \`query\` string + optional \`variables\` object; \`outputSchema\`: \`{ data, errors }\`).
- The handler builds a synthetic \`POST /api/v1/graphql\` request and calls the same \`handleGraphQLRequest\` the REST route uses, so the endpoint's **depth (max 7)** and **complexity (max 50)** validation rules apply identically — the bridge cannot bypass them. Read-only (the endpoint exposes no mutations).
- The MCP dispatch path already runs \`enforceMcpRateLimit\` before dispatch, so no extra limiter is added (mirrors the existing \`call_rpc\` passthrough tool).
- Read-only annotations + untrusted-data note are applied automatically by \`listToolDefinitions\`; tool-count/spec-projection assertions track \`MCP_TOOLS.length\` automatically. No version bumps (handled by \`sync-mcp-version\` post-merge).

## Tests
Six cases in \`tests/mcp-server.test.mjs\`: a successful query, variable passthrough (cold-env zeroed card), a GraphQL validation error surfaced as a populated \`errors[]\`, the complexity cap rejecting an 11-relationship-field query (proving the cap is reused, not bypassed), and the malformed-query / non-object-variables argument guards. Full MCP suite green (1119 tests); \`scripts/validate-mcp.mjs\` passes (173 tools); new handler 100% covered (statements + branches).

### DIFF
--- a/src/mcp-server.mjs
+++ b/src/mcp-server.mjs
@@ -40,6 +40,7 @@ import { EXPOSED_RESPONSE_HEADERS_VALUE } from "../workers/http.mjs";
 import { d1TimeoutMs, withTimeout } from "../workers/storage.mjs";
 import { tryPostgresTier } from "../workers/postgres-tier.mjs";
 import { handleRpcProxyRequest } from "../workers/request-handlers/rpc-proxy.mjs";
+import { handleGraphQLRequest } from "./graphql.mjs";
 import {
   isValidSubscriptionId,
   subscriptionStorageKey,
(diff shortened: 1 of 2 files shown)


JSONbored (commented):
Closing, still an open Superagent flagged P2 issue that needs to be fixed:

```
P2: MCP query_graphql tool bypasses GraphQL-specific rate limiter

MCP handler skips the per-IP graphqlRateLimited guard the REST endpoint applies before handleGraphQLRequest.

Apply graphqlRateLimited to the synthetic request, or forward ctx.clientIp so GraphQL-specific throttling is enforced.
```

PR #6210, opened 2026-07-16
feat(ui): add fumadocs-openapi API reference (Phase 4)

JSONbored:
## Summary

- Adds a full interactive API reference at `/docs/api-reference`, generated from the published OpenAPI spec via `fumadocs-openapi` — one page per operation (159 total), grouped by domain into 22 sidebar folders, each with a real request/response schema, a live try-it client, and multi-language code samples (cURL/JS/Go/Python/Java/C#/Rust).
- Renders as **native Fumadocs pages** — same `DocsLayout`/sidebar/search/theme as the rest of `/docs/*`. No third-party branding, no AI chat widget, no data leaving the app.
- Adds a `Browse reference` link from `/schemas` (where the OpenAPI spec is otherwise only surfaced as copyable text) and wires the reference's index page into the ⌘K command palette (auto-discovered via the nav-drift fix from #6195 — no manual entry needed).

## Why fumadocs-openapi over Scalar

Originally built this with `@scalar/api-reference-react` — installed it, built the route, got it rendering. Live testing then showed real problems: a "Powered by Scalar" footer, a floating **Ask AI** widget that explicitly uploads the OpenAPI doc to Scalar's own servers (with a Terms/Privacy consent screen), and a toolbar with Deploy/Configure/Share/VS Code/Cursor/Generate MCP buttons — none of which apply to a public read-only registry. It rendered as a separate embedded app next to the site, not part of it.

Reverted that entirely (dependency uninstalled, files deleted) in favor of `fumadocs-openapi`, which was the better structural fit for "closer to Mintlify" — same design system, same layout shell as the rest of the docs, first-party feel. `fumadocs-openapi`'s request/response client turns out to route through `@scalar/api-client-react` too (a lighter, different package than the Scalar app), but purely as a headless building block Fumadocs' own UI wraps — none of the app-shell chrome ships.

## Integration issues found only by testing against the real deploy target

Two things not documented anywhere I could find, discovered empirically:

1. **`groupBy: "tag"` silently produces zero files.** `fumadocs-openapi@11.2.1`'s own tag-grouping option is a silent no-op against this spec (no error, zero generated files — verified in isolation). `scripts/generate-openapi-docs.mjs` works around it: generate flat (the only mode that actually works), then move each file into a tag-keyed folder read directly from the spec (preferring the real domain tag over the ~90 operations' catch-all second `"analytics"` tag, so it doesn't dump most of the API into one folder).
2. **`<APIPage/>` never fetches its own `document` prop.** It requires a pre-resolved `preloaded`/`payload` prop — Next.js-oriented examples get this "for free" from RSC-style per-request server rendering. This app compiles docs MDX client-side instead (`browserCollections.createClientLoader()`, from the Phase 1 migration), so nothing was ever resolving it — surfaced as `Cannot read properties of undefined (reading 'bundled')`. Fixed with a small bridge: `docs.$.tsx`'s server loader calls `openapi.preloadOpenAPIPage(page)` and passes the result through a React context (`openapi-preload-context.tsx`) that the registered `APIPage` component (`mdx.tsx`) reads and merges into props before delegating to the real component.

Also caught along the way: `/api/v1/openapi.json` (the URL `/schemas` already shows users) wraps the spec in this API's `{ok, data, meta}` response envelope — useless as a raw document source for a client to fetch. `/metagraph/openapi.json` (a static asset, not an `/api/v1` route) serves it unwrapped instead — verified via a direct fetch before using it, not assumed.

## Command palette

The nav-drift fix (#6195) auto-surfaces every `content/docs/*` page in the ⌘K palette's "Jump to" list. 150+ individual generated operation pages would flood that list, so `docs-nav.functions.ts` now filters out everything under `/docs/api-reference/*`, keeping only the reference's own index page — Fumadocs' own sidebar/search handles browsing the full set once inside.

## Screenshots

New page — before is the 404 the route didn't exist yet, after is the live reference.

| Viewport | Theme | Before | After |
| --- | --- | --- | --- |
| Mobile | Light | before | after |
| Mobile | Dark | before | after |
| Tablet | Light | before | after |
| Tablet | Dark | before | after |
| Desktop | Light | before | after |
| Desktop | Dark | before | after |

Representative operation page (request builder, path/query params, response schemas, code samples):

| Theme | Before | After |
| --- | --- | --- |
| Light | before | after |
| Dark | before | after |

## Test plan

- [x] `npx tsc --noEmit -p apps/ui` clean
- [x] `npx eslint` clean (0 errors; only the same pre-existing warning patterns already present across every other route file in this repo)
- [x] `npx prettier --check` clean (generated `content/docs/api-reference/**` added to `.prettierignore`, same precedent as `routeTree.gen.ts`/`.source`)
- [x] `npm run test --workspace=apps/ui` — 128 files / 1010 tests pass
- [x] Real production build (`LOVABLE_SANDBOX=1 NITRO_PRESET=cloudflare-module npm run build --workspace=apps/ui`) succeeds; confirmed no `fumadocs-openapi`/`createOpenAPIPage` leakage into any eager `index-*.js` client chunk (stays properly lazy-loaded behind the docs route split)
- [x] Manually verified in-browser: index page renders with full sidebar/grouping; a sample operation page renders the complete request builder, path/query parameters, response schemas for every status code, and all 7 code-sample languages; confirmed via `read_page` (not just a screenshot) that this is real interactive content, not a static shell


JSONbored:
**Update:** found and fixed a real issue after opening this PR, not just the flagged bug.

The published spec's `summary` field holds full explanatory paragraphs on every operation (up to ~1100 chars), with `description` left empty. fumadocs-openapi uses `summary` as the page title verbatim — that title is what this app renders as the sidebar nav label, breadcrumb, H1, and browser tab, so the sidebar was showing full paragraphs instead of endpoint names (screenshot in the original report above).

Fixed at the source in `scripts/generate-openapi-docs.mjs` (and mirrored in `src/lib/openapi-source.ts`'s runtime instance, since fumadocs-openapi's `<APIPage/>` independently re-derives a title from `operation.summary` too — confirmed in its compiled source): every operation now gets a short, Title Case title derived from its `operationId` (e.g. `accountAxonRemovals` → "Account Axon Removals"), with the original text moved to `description` — rendered as intro prose right under the H1, same layout the 4 hand-written docs pages already use. Applied uniformly to all 159 operations (not just the longest summaries) so every sidebar entry is consistently short — checked, a mix of "properly short" and "still a full sentence" titles side by side looked inconsistent.

Verified after regenerating: longest title dropped from 1088 to 31 chars (avg 15.6 chars, down from avg 309). Re-checked in-browser across mobile/desktop, light/dark, no-parameter operations, heavily-parameterized operations, and the new longest-title page — clean render throughout, zero console errors on fresh page loads. Screenshots above updated to match.

JSONbored:
**Update:** fixed the four issues flagged after the last round (description sizing, sidebar duplication, breadcrumb readability, copy-markdown button).

- **Description too large** — `generateFiles({ includeDescription: true })` now renders the operation's description via fumadocs-openapi's own `<Markdown>` component (the `showDescription` prop) instead of Fumadocs' generic `<DocsDescription>`, which defaults to `text-lg` — a one-line-subtitle treatment, wrong for the multi-sentence prose living there since the previous title/summary fix.
- **Sidebar showed "API reference" nested twice** — the root tag-folder is titled "API reference" (`meta.json`), and the hand-authored landing page's own frontmatter title was *also* "API reference," so the auto-generated sidebar showed a same-titled parent/child pair. Renamed the landing page's title to "Overview," the conventional pattern for a section index.
- **Breadcrumb showed "ACCOUNTAXONREMOVALS"** — the site-wide breadcrumb renders URL path segments verbatim, uppercased, with no word-splitting (fine for existing hyphenated slugs like `/docs/chain-events`, broken for a raw camelCase `operationId`). Renamed every generated page's slug to kebab-case (`accountAxonRemovals.mdx` → `account-axon-removals.mdx`) at the generator, matching the existing convention instead of touching the shared breadcrumb component.
- **Copy-markdown button "weirdly positioned, too big"** — replaced the hand-rolled button with Fumadocs' own `ViewOptionsPopover` (researched — `fumadocs-ui/layouts/docs/page`), the compact "Open ⌄" trigger with View as Markdown / Open in Scira AI / ChatGPT / Claude / Cursor. Its "View as Markdown" link needed a real URL — a client-built `data:` URI seemed to work until testing the actual click: Chrome silently blocks `target="_blank"` navigation to `data:` URLs (no console error, no new tab). Added `docs.raw.$.ts`, a real per-page route backed by fumadocs-mdx's `page.data.getText("processed")`, which reads the already-bundled `_markdown` export with no filesystem access — safe in the Cloudflare Worker runtime, not just in dev.

Verified in-browser: all four fixes render correctly across the operation page (light/dark), the popover opens with correctly-sized compact styling, and clicking "View as Markdown" now genuinely opens the per-page markdown in a new tab instead of silently doing nothing. Also spot-checked a JSX-only API reference page's raw-markdown route returns cleanly (empty prose body, no error — expected, since those pages are pure interactive components).

- [x] `npx tsc --noEmit -p apps/ui` clean
- [x] `npx eslint` clean (0 errors; same pre-existing Fast-Refresh warning pattern as every other route file)
- [x] `npx prettier --check` clean
- [x] `npm run test --workspace=apps/ui` — 128 files / 1010 tests pass
- [x] Verified in-browser across light/dark, an operation page and the popover's open state

Screenshots refreshed:

| | Light | Dark |
| --- | --- | --- |
| Operation page | after light | after dark |
| Popover open | popover light | popover dark |

JSONbored:
**Update:** found and fixed a real, separate blocker — `Workers Builds: metagraphed-ui` had been failing on **every commit of this PR back to its first** (`fef165e2`), which is why the preview deploy never reflected any of the fixes above — the stale mobile/tablet screenshots reported were genuinely stale, not a caching issue.

Root cause: Node OOM'd during Nitro's SSR bundling pass (`FATAL ERROR: ... JavaScript heap out of memory`), consistently around a ~2GB ceiling — reproduced locally by constraining Node to the same limit. Two contributing factors, both fixed:

1. **`mdx.tsx`'s `createOpenAPIPage()` wrapper statically references `fumadocs-core`'s `defaultShikiFactory`** as its fallback (`options.shiki ?? defaultShikiFactory`) — Rollup keeps that binding, and the unrestricted `import("shiki")` inside it (shiki's full ~180-language catalog), reachable and bundled regardless of which branch runs at runtime, even though a `shiki` option was always supplied. Switched to `createOpenAPIPageBase` (`fumadocs-openapi/ui/base` — its own comment: "without the full Shiki bundle") with a new scoped factory (`openapi-shiki.ts`) built from explicit static imports of only the 8 languages this app's interactive playground actually generates. `shikijs__langs.mjs` dropped from 7.49 MB to 468 KB (93.7% smaller).
2. That alone wasn't sufficient — the app's total Nitro/SSR bundle still exceeded the container's ~2GB default. Node's own heap-sizing heuristic is commonly more conservative than what a container actually has available, so `apps/ui/package.json`'s `build` script now sets `NODE_OPTIONS=--max-old-space-size=4096`.

Both changes verified against the failure locally (constrained-heap repro) before pushing, and `Workers Builds: metagraphed-ui` is now green on the resulting commit.

PR #6218, opened 2026-07-16
fix(ui): show a StaleBanner on the validator detail page (#5861)

### LINKED ISSUE #5861: fix(ui): validators.$hotkey.tsx never shows a stale-data banner despite its query returning meta.generated_at
## Context

`apps/ui/src/routes/validators.$hotkey.tsx` has no `StaleBanner`/`isStaleFreshness` anywhere in the file, even though `validatorDetailQuery` (`apps/ui/src/lib/metagraphed/queries.ts:6001-6038`) returns a `meta: res.meta` field — the same shape used elsewhere in the app to detect and surface stale data. Its own list view, `apps/ui/src/routes/validators.index.tsx:168`, shows a `StaleBanner`, and structurally-identical detail routes for other entities already do too (`apps/ui/src/routes/subnets.$netuid.tsx:280`, `apps/ui/src/routes/providers.$slug.tsx:142`).

## Requirements

- Add the same `StaleBanner`/`isStaleFreshness` check `subnets.$netuid.tsx`/`providers.$slug.tsx` use, wired to `validatorDetailQuery`'s existing `meta` field, to `validators.$hotkey.tsx`.
- Do not change the query itself (the `meta` field already exists) — this is a display-layer fix only.
- This is a UI-only change under `apps/**`.

## Deliverables

- [ ] `validators.$hotkey.tsx` shows a `StaleBanner` when `meta.generated_at` indicates stale data, matching the pattern in `subnets.$netuid.tsx`/`providers.$slug.tsx`.

## Expected Outcome
A user viewing a validator detail page with stale underlying data sees the same staleness indicator every other detail page already shows.

## Links & Resources
- `apps/ui/src/routes/validators.$hotkey.tsx` (gap)
- `apps/ui/src/lib/metagraphed/queries.ts:6001-6038` (`validatorDetailQuery`, already returns `meta`)
- `apps/ui/src/routes/validators.index.tsx:168`, `apps/ui/src/routes/subnets.$netuid.tsx:280`, `apps/ui/src/routes/providers.$slug.tsx:142` (existing pattern)

shinkaito-ai:
## What

Adds a `StaleBanner` to the validator detail page (`validators.$hotkey.tsx`), which had no staleness indicator at all — unlike its own list view (`validators.index.tsx:168`) and the structurally-identical detail routes `subnets.$netuid.tsx:280` / `providers.$slug.tsx:142`, which all surface one.

## How

Display-layer only, mirroring the sibling pattern exactly:
- `isStaleFreshness(detailRes.meta?.generated_at)` gates a `StaleBanner` (both already used across the app), rendered right below the `PageHero`.
- `refreshQueryKeys` points at `validatorDetailQuery(hotkey)`'s key so the banner's "Refresh now" action revalidates the detail data.
- The `meta` field already exists on `validatorDetailQuery` — the query and tables are untouched.

## Screenshots

The banner only renders when the snapshot is stale (`isStaleFreshness` → >12h). Live data is currently fresh, so the **after** shots below show the stale state forced with an aged snapshot timestamp (banner: *"Snapshot from 3d ago — may be lagging behind live"* + Refresh now); **before** is `main` (no banner). Full viewport × theme matrix:

| View | Before | After |
|------|--------|-------|
| Desktop · Light | before | after |
| Desktop · Dark | before | after |
| Tablet · Light | before | after |
| Tablet · Dark | before | after |
| Mobile · Light | before | after |
| Mobile · Dark | before | after |

Closes #5861.

### DIFF
--- a/apps/ui/src/routes/validators.$hotkey.tsx
+++ b/apps/ui/src/routes/validators.$hotkey.tsx
@@ -6,7 +6,7 @@ import { fallback, zodValidator } from "@tanstack/zod-adapter";
 import { Boxes, Coins, Gauge, Percent, Users, Zap } from "lucide-react";
 import { useWallet } from "@/hooks/use-wallet";
 import { AppShell } from "@/components/metagraphed/app-shell";
-import { EmptyState, PageHeading, Skeleton } from "@/components/metagraphed/states";
+import { EmptyState, PageHeading, Skeleton, StaleBanner } from "@/components/metagraphed/states";
 import { ApiSourceFooter } from "@/components/metagraphed/api-source-footer";
 import { EndpointSnippet } from "@/components/metagraphed/endpoint-snippet";
 import { QueryErrorBoundary } from "@/components/metagraphed/error-boundary";
@@ -25,7 +25,7 @@ import { taoCompact, scoreStr } from "@/components/metagraphed/neuron-table";
 import { validatorDetailQuery, validatorNominatorsQuery } from "@/lib/metagraphed/queries";
 import { isValidSs58, ss58PathSegment } from "@/lib/metagraphed/accounts";
 import { shortHash } from "@/lib/metagraphed/blocks";
-import { formatNumber } from "@/lib/metagraphed/format";
+import { formatNumber, isStaleFreshness } from "@/lib/metagraphed/format";
 import { hasValidatorIdentity } from "@/lib/metagraphed/validator-identity";
 import {
   annualizedDelegatorApyPct,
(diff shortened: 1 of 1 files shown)


JSONbored (commented):
This whole section is in need of a lot of work. Too much spacing, the new implementation doesn't cleanly add the text + button, etc. I feel like we could get away with an icon and no "refresh" text, and a tooltip or much shorter text...not sure, but this needs work. 

Image

shinkaito-ai:
Thanks for the review! Reworked per your notes in #6220 — an opt-in compact StaleBanner variant (icon-only refresh with a tooltip, much shorter copy) placed inline in the hero actions row next to Share view, so it no longer floats in the section margin.

PR #6305, opened 2026-07-16
feat(ui): add a root-subnet inclusion toggle to the /subnets filter bar (#6270)

### LINKED ISSUE #6270: Add include-root / include-inactive filters to /subnets
## Context

`apps/ui/src/lib/metagraphed/url-state.ts`'s `tableSearchSchema` already carries `curation`, `health`, `kind`, `stale`, and `provider` as client-side list filters on `/subnets`, each following the same pattern. `status` (active/inactive/unknown) and root-subnet identity (`is_root`, `netuid === 0`) already flow through the data model (`queries.ts`) and are already returned per-subnet — but `/subnets` has no toggle to filter by either, unlike the existing filters it already supports.

## Requirements

- Extend `tableSearchSchema` with an `includeRoot`/`includeInactive` field (or a single `status` field, whichever matches the existing filter fields' shape more closely), following the exact pattern already used by `curation`/`health`/`kind`/`stale`/`provider`.
- Wire a checkbox (or equivalent toggle) in `apps/ui/src/routes/subnets.index.tsx` applying the new filter client-side, using the same client-side-filter pattern already used for `curation`/`health` (~line 269).
- No new backend field or route — `status`/`is_root` are already returned in the existing subnet list response.

## Deliverables

- [ ] `includeRoot`/`includeInactive` (or `status`) field added to `tableSearchSchema`
- [ ] Checkbox/toggle UI wired into `subnets.index.tsx`
- [ ] Client-side filter applied against the existing `status`/`is_root` fields

## Expected Outcome

A user can include or exclude root and inactive subnets from the `/subnets` list view via a visible control, instead of always seeing the unfiltered set.

## Links & Resources

- `apps/ui/src/lib/metagraphed/url-state.ts` (`tableSearchSchema`, existing `curation`/`health`/`kind`/`stale`/`provider` fields, lines ~16-24)
- `apps/ui/src/routes/subnets.index.tsx` (~line 269, existing client-side filter application pattern)

real-venus:
## Summary

`/subnets` already filters client-side on curation/health/service/readiness, and the list response already returns `subnet_type` (root|application) per subnet — but there was no way to exclude the root netuid, so the list was always the unfiltered set.

Adds one toggle to the existing filter bar, applied client-side over data already fetched (no new backend field or route):

- **`includeRoot`** in `tableSearchSchema` (`url-state.ts`) as a boolean defaulting to `true` — the `/endpoints` `callable` toggle's shape. Defaulting to true keeps the unfiltered list byte-identical for anyone without the param; it only counts toward `filtersActive` (which drives Reset) once switched off.
- An **`ExcludeToggle`** control mirroring the `/endpoints` "Callable only" toggle's markup and accent convention: the accent lights only while the toggle is *narrowing* the list, so the default all-inclusive view stays visually quiet. It shows the row count it would drop — the same "what am I hiding?" affordance `directoryCount` gives on `/endpoints`.
- Root is matched on the wire `subnet_type` (netuid 0 accepted either way). `subnet_type` is typed on the `Subnet` row — it was already returned, just reaching the filter through the index signature.

### Scope note — `includeInactive` deliberately omitted

The issue also asked for an `includeInactive` companion. `GET /api/v1/subnets` is documented as **"List active Finney subnets"** and only ever returns `status=active` rows — verified live: all 129 rows are `active`, and `?status=inactive` returns 0 (`meta.total: 0`). A client-side inactive filter can only narrow rows the server already sent, so it would be **inert by construction** — a control that can never do anything. It belongs here only once that route serves non-active rows. Happy to add it if you'd rather have the forward-looking control.

## Screenshots

| Viewport | Before | After |
| --- | --- | --- |
| Mobile 375 · Light |  |  |
| Mobile 375 · Dark |  |  |
| Tablet 768 · Light |  |  |
| Tablet 768 · Dark |  |  |
| Desktop 1280 · Light |  |  |
| Desktop 1280 · Dark |  |  |

**Engaged state** (`?includeRoot=false` — toggle lit, root hidden, Reset becomes available):

| Light | Dark |
| --- | --- |
|  |  |

## Testing

- `typecheck`, `eslint` (0 errors), `prettier` — clean
- **1023 unit tests pass**
- **responsive-overflow e2e: 24/24 pass**, including `/subnets` at 375/768/1024/1280 — the added button introduces no overflow
- `vite build` succeeds

Closes #6270

### DIFF
--- a/apps/ui/src/lib/metagraphed/types.ts
+++ b/apps/ui/src/lib/metagraphed/types.ts
@@ -135,6 +135,10 @@ export interface Subnet {
   name?: string;
   symbol?: string;
   type?: "root" | "application";
+  // Wire field the /api/v1/subnets list already returns (SubnetIndexEntry),
+  // typed here so the #6270 root filter reads it without casting through the
+  // index signature below.
+  subnet_type?: "root" | "application" | string;
   participants?: number;
   tempo?: number;
   registration_block?: number;
--- a/apps/ui/src/lib/metagraphed/url-state.ts
+++ b/apps/ui/src/lib/metagraphed/url-state.ts
@@ -22,6 +22,19 @@ export const tableSearchSchema = z.object({
   // #9: agent-catalog capability filters (applied client-side over joined rows).
   serviceKind: fallback(z.string(), "").default(""),
   readiness: fallback(z.string(), "").default(""),
+  // #6270: root-subnet inclusion toggle for /subnets, applied client-side over
+  // the `subnet_type` the list response already returns. A boolean defaulting
+  // to true (the endpoints route's `callable` toggle shape) rather than the ""
+  // string filters above: it includes/excludes a slice of the set instead of
+  // selecting one value, and defaulting to true keeps the unfiltered list
+  // byte-identical to today's for anyone without the param.
+  //
+  // No companion `includeInactive`: GET /api/v1/subnets is documented as "List
+  // active Finney subnets" and only ever returns status=active rows (verified
+  // live — ?status=inactive returns 0). A client-side inactive filter could
+  // only narrow rows the server already sent, so it would be inert by
+  // construction; it belongs here only once that route serves non-active rows.
+  includeRoot: fallback(z.boolean(), true).default(true),
   // Layout state for list routes that support multiple views + row density.
   // Additive + optional with safe fallbacks so the toggles persist in the URL.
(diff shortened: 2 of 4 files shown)


JSONbored (commented):
Before + after look the same. Also: 

```
The `includeInactive` companion from the linked issue is explicitly and reasonably scoped out with a documented rationale (the API never returns inactive rows), leaving the issue only partially closed but by design rather than oversight.
```

Issue #5358, opened 2026-07-14
fix(analytics): dead neurons/neuron_daily edge-cache staleness stamps


JSONbored:
Part of #5348.

`analytics.mjs`'s edge-cache staleness stamps for `neurons`/`neuron_daily` responses appear to be dead code post D1 retirement (#4772) — the cache-freshness signal they're meant to produce may no longer reflect real data recency now that these are served from Postgres. Needs tracing to confirm whether this is genuinely dead (remove it, per this repo's "if it's dead code, eliminate it" standard) or just needs repointing at the Postgres-tier's own freshness signal.

PR #6373, opened 2026-07-17
feat(api): add GET /api/v1/compare/validators mirroring the compare_validators MCP tool

oktofeesh1:
## Summary

- Adds `GET /api/v1/compare/validators` mirroring the existing `compare_validators` MCP tool, closing the REST-surface gap called out in #6325 — the exact same pairing `/api/v1/compare`/`compare_subnets` already has, one level up.

## What Changed

- `src/analytics-live.mjs`: extracted `COMPARE_VALIDATORS_MAX`, `parseCompareHotkeys` (REST comma-separated query string), and `parseCompareHotkeyList` (MCP array) as shared exports — mirroring how `parseCompareNetuids`/`parseCompareNetuidList` are already shared between `/api/v1/compare` and `compare_subnets`.
- `src/mcp-server.mjs`: refactored the `compare_validators` tool to consume the now-shared `parseCompareHotkeyList`/`COMPARE_VALIDATORS_MAX` instead of its own local, duplicate copies — no behavior change (verified: all pre-existing `compare_validators` tests still pass unmodified).
- `workers/request-handlers/analytics-routes.mjs`: added `handleCompareValidators`, reusing `composeValidatorComparison` (`src/metagraph-neurons.mjs`) so REST and MCP share identical composition logic, and the same per-hotkey `tryPostgresTier(..., "METAGRAPH_NEURONS_SOURCE") ?? buildValidatorDetail([], hotkey)` fallback contract `handleValidatorDetail` uses (sequential fan-out, matching the MCP tool's own request pattern exactly).
- `workers/api.mjs`: wired `/api/v1/compare/validators` dispatch (uncached, matching `handleValidatorDetail`'s own dispatch — this route has no economics/health composition to cache-bust on, unlike `/api/v1/compare`) and added it to the mainnet-only route gate (validator data is mainnet-only, same as `/api/v1/compare` and `/api/v1/validators/{hotkey}`).
- `src/contracts.mjs`: new `route()`/`artifact()` entries for `compare-validators`.
- `schemas/components/05-subnets.schema.json`: new `CompareValidatorsArtifact`/`CompareValidatorEntry` schemas, reusing the existing `ColdkeyIdentity`/`ValidatorDetailSubnet` component refs.
[...]

PR #6382, opened 2026-07-17
feat(api): add GET /api/v1/compare/validators mirroring the compare_validators MCP tool

### LINKED ISSUE #6325: Add a REST route mirroring the MCP compare_validators tool
## Context
`compare_validators` (`src/mcp-server.mjs:5103-5165`) is explicitly documented as "the validator equivalent of compare_subnets" and computes `composeValidatorComparison(details, { netuid })` (`src/metagraph-neurons.mjs:850`). Its sibling `compare_subnets` (`src/mcp-server.mjs:4791-4796`) mirrors GET `/api/v1/compare` (route `compare`, `src/contracts.mjs:3703`, handled by `handleCompare`, `workers/request-handlers/analytics-routes.mjs:750`), which is subnets-only (`netuids=` param, confirmed by reading `handleCompare`'s body — no hotkey/validator branch). No REST or GraphQL route exists anywhere for validator comparison (confirmed via grep across `src/contracts.mjs`, `workers/api.mjs`, and all `workers/request-handlers/*.mjs` for "compare-validators", "validators/compare", "compareValidators").

## Requirements
- Add a new REST route (e.g. GET `/api/v1/compare/validators?hotkeys=...&netuid=...`) to `API_ROUTES` in `src/contracts.mjs`, following the `route()` shape of the existing `"compare"` entry.
- Implement `handleCompareValidators` in `workers/request-handlers/analytics-routes.mjs` alongside `handleCompare`, reusing `composeValidatorComparison` from `src/metagraph-neurons.mjs:850` so REST and MCP share identical composition logic.
- Wire the route into `workers/api.mjs`'s dispatch the same way `/api/v1/compare` is wired.
- Regenerate the OpenAPI contract (`npm run ui:openapi`).

## Deliverables
- [ ] `GET /api/v1/compare/validators` route
- [ ] `handleCompareValidators` handler reusing `composeValidatorComparison`
- [ ] Wired dispatch in `workers/api.mjs`
- [ ] Regenerated OpenAPI spec
- [ ] Route test asserting REST/MCP parity for identical `hotkeys`+`netuid` inputs

## Expected Outcome
REST clients can compare validators the same way MCP clients already can, mirroring the existing subnet-comparison parity.

## Links & Resources
Precedent: `/api/v1/compare` + `compare_subnets` (`src/contracts.mjs:3703`, `workers/request-handlers/analytics-routes.mjs:750`, `src/mcp-server.mjs:4791`) — the exact same pairing, one level up, already exists for subnets.

davion-knight:
## What

`compare_validators` is documented as "the validator equivalent of `compare_subnets`", but that parity stopped at the MCP surface: `compare_subnets` mirrors `GET /api/v1/compare`, while validator comparison had no REST (or GraphQL) route at all. A REST client had to fan out N `GET /api/v1/validators/{hotkey}` calls and re-implement the projection itself.

This adds the missing route, composed through `composeValidatorComparison` — the same function the MCP tool calls — so the two surfaces cannot project different fields for the same inputs.

```
GET /api/v1/compare/validators?hotkeys=<ss58,...>&netuid=<n>
```

- **`hotkeys`** (required): 1-16 distinct SS58 addresses, comma-separated, in display order.
- **`netuid`** (optional): adds each validator's membership row in that subnet as `subnet_context`, null where it holds no validator permit there.
- An unknown hotkey returns a zeroed aggregate rather than a 404, matching `GET /api/v1/validators/{hotkey}`.

Read-only decision support: builds no transaction, produces no signable artifact, and touches no wallet or key.

## How it keeps REST and MCP from drifting

- **Composition** is the shared `composeValidatorComparison` (`src/metagraph-neurons.mjs`), as the issue specifies.
- **Parsing**: `parseCompareHotkeys` joins `parseCompareNetuids`/`parseCompareNetuidList` in `analytics-live.mjs` as the CSV form of `parseHotkeyList`'s array form — the same cap-and-dedupe rules, so both surfaces accept exactly the same set. (`parseHotkeyList`'s own comment already described itself as mirroring `parseCompareNetuidList`; this completes the square.)
- **The cap**: `COMPARE_VALIDATORS_MAX` moves to `analytics-live.mjs` and is now imported by `mcp-server.mjs`, so REST and MCP cannot drift to different caps. (`mcp-server.mjs` already imported from `analytics-live.mjs`, so this follows the existing dependency direction — no new cycle.)
- **`meta.generated_at`** reports the **oldest** `captured_at` across the fan-out: a comparison is only as fresh as its stalest input. An all-cold read stays `null` rather than being fabricated as "now".

## One deliberate deviation from the issue text

The issue says to wire the dispatch "the same way `/api/v1/compare` is wired" — which uses `withEdgeCache`. **I did not edge-cache this route**, and want to flag that explicitly rather than bury it:

`/api/v1/compare`'s payload carries the live health rollup, so the health cron's `last_run_at` (`withEdgeCache`'s default stamp) is exactly what invalidates it. This route's payload is the neurons/metagraph snapshot instead, whose own cache stamps were retired with the D1 neurons tables (#4772, #5358). Keying it on the health stamp would bust on an unrelated event *while serving stale validator data between runs*. `handleValidatorDetail` — the per-hotkey route this fans out to — is likewise uncached, so this matches the data path it reads. The reasoning is recorded in a comment at the dispatch site. Happy to add caching under a neurons-appropriate stamp if you'd prefer.

## Contract wiring

`route()` + `artifact()` in `contracts.mjs`, `CompareValidatorsArtifact`/`CompareValidatorEntry` in `schemas/components/05-subnets.schema.json`, plus the four completeness checks a fileless route needs: `COMPUTED_ARTIFACTS` (`validate-schemas.mjs`), `R2_ONLY_PATTERNS` (`artifact-storage.mjs`), the `validate-api.mjs` check (`checks[]` must equal `API_ROUTES.length`), and both docs entries. Regenerated via `schemas:bundle` + `build`.

## Testing

- **REST/MCP parity** is asserted by driving **both surfaces over one env and deep-comparing** the payloads, for both the netuid-absent and netuid-present cases — so a field added to `VALIDATOR_COMPARISON_FIELDS` must land on both or the test fails.
- The shared cap (over-limit rejected by both surfaces), dedupe/order, the tier fan-out path (asserting one detail load per hotkey at the same path the MCP tool uses), the stalest-capture stamp, every 400 branch, and the `CompareValidatorsArtifact` contract.
- **100% patch coverage, branch-counted** (41 statements / 26 branches on changed lines, measured unsharded).
- Full local validator sweep green: `contract-drift`, `api`, `schemas`, `openapi`, `openapi-examples`, `types`, `mcp`, `docs`, `artifact-budgets`, `schema-enums`, `generated-client`, `client-sdk-sync`, `private-boundary`. openapi.json is 1,713,118 bytes — under the 1,750,000 budget.

Closes #6325


JSONbored:
someone already opened a PR (#6373) for this, closing.

Issue #6451, opened 2026-07-17
chain-firehose-relay: rate-limit thundering-herd incident, 9.94M row backlog (2026-07-15 to 2026-07-17)


JSONbored:
## Incident

The `chain-firehose-relay` container on the indexer box accumulated a backlog of **9,942,268 undelivered rows** in `chain_firehose_outbox` over roughly 40 hours (oldest pending row: 2026-07-15 14:52:07), with logs continuously spamming `dropped a payload after 3 attempts`.

## Root cause

`CHAIN_FIREHOSE_INGEST_RATE_LIMIT` (120 req/60s, `workers/api.mjs` + the `CHAIN_FIREHOSE_INGEST_RATE_LIMITER` binding in `wrangler.jsonc`) was sized for the relay's OLD LISTEN/NOTIFY design (one notification per block, ~5/min at ~12s blocks). #5027 replaced that with an outbox-table poll forwarding one row per `blocks`/`extrinsics`/`chain_events`/`account_events` row at up to 16 concurrent requests -- a fundamentally different, much higher-volume traffic shape the rate limit was never recalibrated for.

Confirmed live against the real deployed container and a real outbox payload: every forward attempt got HTTP 429. The relay's own retry logic applied the same short generic backoff to a 429 as to any other failure, so it kept re-triggering the same limit indefinitely instead of ever recovering -- a self-sustaining failure loop, not a transient blip.

## Fix

See the linked PR. Both halves of the mismatch needed fixing:
- Raised the rate limit to 1200/60s (server-side), matched in both `workers/api.mjs`'s header-reporting constant and the actual `wrangler.jsonc` binding (only the binding is actually enforced -- the JS constant just generates response headers).
- `chain-firehose-relay.mjs`: honors `retry-after` on a 429 instead of generic exponential backoff, and pauses the WHOLE poll loop (not just one row's own retries) when rate limited -- the actual fix for the thundering-herd shape.
- Added Sentry error tracking with aggregate (not per-drop) reporting, so a recurrence would be visible without either spamming or silently exceeding the event quota.

## Immediate mitigation (already applied to the live box)

[...]

PR #6459, opened 2026-07-17
fix(chain-firehose-relay): convert Dockerfile to clone-at-runtime

JSONbored:
## Summary

Part of the drift reconciliation started in #6456 (context: #6451). `metagraphed-infra`'s `chain-firehose-relay` Ansible role tracked its own copy of `scripts/chain-firehose-relay.mjs`, independently of this repo's canonical one -- the same duplication #6456 already fixed for the four fetch scripts. That copy had drifted badly: it was still the retired LISTEN/NOTIFY-based script from before #5027/#6452 rewrote this relay to outbox-polling.

This PR converts `deploy/chain-firehose-relay.Dockerfile` to the same clone-at-runtime pattern: the image now holds no copy of `chain-firehose-relay.mjs` at all. A new `scripts/chain-firehose-relay-entrypoint.sh` clones `metagraphed` fresh at container start and execs the relay from the clone.

Unlike the fetch scripts / `data-refresh-node`, this process is a single always-on daemon rather than a periodically re-invoked job, so the entrypoint deliberately skips the persistent-volume + incremental-refresh machinery those use: every container start (rare -- a crash-restart or a deliberate rebuild) does one fresh shallow clone + `npm ci --ignore-scripts` into a **fixed** path (`/tmp/repo`, not a random `mktemp` suffix), then execs. This also means restarting the container (not rebuilding the image) is enough to pick up the latest `main`.

`SENTRY_RELEASE` is now computed from the freshly-cloned HEAD when unset, matching the fetch-scripts entrypoint's convention -- the release that matters is this repo's commit, not `metagraphed-infra`'s.

**Also**: while verifying this, found that `metagraphed-infra`'s stale copy had a heartbeat file + `--healthcheck` CLI entry point (metagraphed-infra#63) that was never ported here when #5027/#6452 rewrote the relay -- it had drifted in directly on the infra side. Deleting that copy without porting the mechanism first would have silently regressed production monitoring coverage (the Docker `HEALTHCHECK` directive would have nothing to call). Ported `HEARTBEAT_FILE`/`touchHeartbeat`/`isHeartbeatFresh`/the `--healthcheck` branch into the canonical script, adapted for the poll loop (touched once per iteration instead of once per NOTIFY), and added the `HEALTHCHECK` directive back to the Dockerfile against the now-fixed `/tmp/repo` path.

Follow-up (not in this PR): `metagraphed-infra`'s `chain-firehose-relay` role still needs to stop copying `scripts/chain-firehose-relay.mjs` as a local file and redeploy to the box (separate PR, in progress).

## Test plan

- [x] `docker build -f deploy/chain-firehose-relay.Dockerfile` succeeds
- [x] `docker run` (no `DATABASE_URL`): entrypoint clones, runs `npm ci --ignore-scripts` (569 packages), passes the tracked-file diff check, computes `SENTRY_RELEASE` from the cloned HEAD, execs the relay, which fails at its own `DATABASE_URL` validation as expected
- [x] `docker run` against a disposable local Postgres with a real `chain_firehose_outbox` table: entrypoint completes clone/install/exec, `postgres` driver connects cleanly under Alpine/musl, relay polls successfully
- [x] `docker exec ... --healthcheck` against a live poll loop: heartbeat file is written, exits 0; against a process with no heartbeat yet, exits 1 (never hangs/times out) -- verified end-to-end with a real running container, not just unit tests
- [x] `npm test -- tests/chain-firehose-relay.test.mjs`: 50/50 passing, including new `touchHeartbeat`/`isHeartbeatFresh` tests (real tmpfile I/O, matching `tests/backup-postgres.test.mjs`'s own convention)
- [x] `shellcheck scripts/chain-firehose-relay-entrypoint.sh` clean
- [x] `npm run lint` / `npx prettier --check` clean
- [x] `npm run scan:public-safety` passes

### DIFF
--- a/deploy/chain-firehose-relay.Dockerfile
+++ b/deploy/chain-firehose-relay.Dockerfile
@@ -6,31 +6,32 @@
 # scripts/chain-firehose-relay.mjs's own header comment for why this is safe
 # by construction, unlike the retired streamer (docs/adr/0014).
 #
-# Deployed the same way the (also retired) metagraphed-streamer was: the
-# Ansible `chain-firehose-relay` role in JSONbored/metagraphed-infra copies
-# this Dockerfile + scripts/chain-firehose-relay.mjs into
-# roles/chain-firehose-relay/files/ and builds directly on the indexer box
-# (no cross-compilation concern -- this is a single small ESM file + one npm
-# dependency). Re-run that role after updating either file to rebuild with
-# the latest fix.
+# Clone-at-runtime, like metagraph-fetch.Dockerfile/data-refresh-node.Dockerfile
+# /economics-refresh.Dockerfile: this image holds no copy of
+# chain-firehose-relay.mjs itself (see scripts/chain-firehose-relay-entrypoint.sh).
+# metagraphed-infra used to track a second, independently-drifting copy of
+# this file (metagraphed#6451); the entrypoint now clones the real one from
+# GitHub at container start instead. The Ansible `chain-firehose-relay` role
+# in JSONbored/metagraphed-infra now copies only this Dockerfile + the
+# entrypoint script into roles/chain-firehose-relay/files/ and builds
+# directly on the indexer box. Restarting the container (not rebuilding the
+# image) is enough to pick up the latest main.
 #
 # Local:  docker build -f deploy/chain-firehose-relay.Dockerfile -t metagraphed-chain-firehose-relay .
 FROM node:22.23.1-alpine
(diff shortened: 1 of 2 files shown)


JSONbored:
Correction: this PR merged at the Dockerfile-only commit before a follow-up commit (the heartbeat/`--healthcheck` port referenced in this PR's description) could land in it -- that commit went to an orphaned branch never attached to an open PR here, so it was never actually merged despite what the description above says. Split out as #6466.

PR #6521, opened 2026-07-17
chore(enrich): taostats identity gap-fills

JSONbored:
## Provenance

These are third-party **taostats** gap-fills — identity fields sourced from `https://api.taostats.io/api/subnet/identity/v1` that were missing in the metagraphed registry. Taostats is a third-party aggregator; these fills are included for human review and should be verified against the subnet's own primary sources before relying on them as authoritative.

## Fields touched

| Subnet | Netuid | Field | Value |
|--------|--------|-------|-------|
| Actual | 95 | `source_repo` | `https://github.com/actual-computer/actual-subnet-95` |

## Notes

- Ran against 129 taostats identity records; 9 initial schema-valid candidates after filtering to `source_repo`/`website_url`/`social.x` only (fields not in `subnet-manifest.schema.json` were skipped).
- After cross-checking subnet name alignment between taostats and registry, 7 candidates were dropped due to name mismatches (stale taostats netuid→name mappings) or clearly generic/placeholder URLs.
- SN119 Satori `source_repo` was explicitly excluded: the registry notes record that the `Satori119/Satori` repo previously returned 404 and was intentionally unset.
- Schema validation passes (`validate:schemas`).


JSONbored:
Rejecting, this is a dead link.

PR #6930, opened 2026-07-19
fix(ui): stop the header overflowing the mobile viewport (NetworkSwitcher responsive)

### LINKED ISSUE #6902: fix(ui): header overflows viewport by 10px on mobile (NetworkSwitcher/ApiDrawerTrigger not responsive)
## Context

Site-wide bug, confirmed via direct measurement on the live site (2026-07-19) across every page (homepage, `/subnets`, `/subnets/1`, `/validators`, `/blocks`, `/health` all reproduce identically, since it's the shared header): at 375px width, the header overflows the viewport by 10px.

Root cause confirmed in `apps/ui/src/components/metagraphed/app-shell.tsx:215-218`: `ApiDrawerTrigger` and `NetworkSwitcher` render unconditionally at every breakpoint, while their sibling header actions — the developer-settings link (line 224), `SettingsPopover` (line 234), and `WalletConnectButton` (line 241) — are all correctly hidden below `md` (`hidden md:inline-flex`). `NetworkSwitcher` didn't get the same responsive treatment.

Confirmed via direct DOM measurement:
```
document.body.scrollWidth = 385  vs  window.innerWidth = 375
```
The overflowing element is exactly the `NetworkSwitcher` button ("Mainnet" pill + globe icon + chevron), positioned `left:270, right:385` in a 375px viewport.

## Requirements

Give `ApiDrawerTrigger`/`NetworkSwitcher` the same responsive treatment their sibling actions already have below `md` — either hide them behind the mobile hamburger menu (where the equivalent settings/wallet-connect actions already live), or collapse them to icon-only at mobile width so the row fits. Don't just add `overflow-x: hidden` to the header — that hides the symptom without fixing the layout, and would make the network switcher unreachable on mobile instead of properly relocating it.

## Deliverables

- [ ] Header no longer overflows the viewport at 375px width (verify via `document.body.scrollWidth <= window.innerWidth`, not just visual inspection)
- [ ] `NetworkSwitcher`/`ApiDrawerTrigger` remain reachable on mobile somewhere (mobile menu, icon-only collapse, or similar) — not silently removed
- [ ] Before/after screenshot table across all 3 viewports (required for this milestone) — this bug is mobile-only, so the desktop/tablet "before" shots should show no visible difference, which is expected

## Expected Outcome

No horizontal page overflow on mobile, on any page, ever again — this was a shared-header bug affecting the entire site, not one page.

## Links & Resources

`apps/ui/src/components/metagraphed/app-shell.tsx:205-249` (the whole header actions row, for the full responsive pattern to match)

davion-knight:
## Problem

Confirmed at 375px on every page (it's the shared header): the header overflows the viewport by 10px — `document.body.scrollWidth` 385 vs `window.innerWidth` 375. The overflowing element is the `NetworkSwitcher` pill (globe + "Mainnet" + chevron), which sat past the right edge.

## Root cause

`NetworkSwitcher` was the one header action still rendering at **every** breakpoint. Its siblings are all already hidden below `md`:
- the developer-settings link, `SettingsPopover`, and `WalletConnectButton` are `hidden md:inline-flex`
- `ApiDrawerTrigger`'s own button is `hidden md:inline-flex` too

Only `NetworkSwitcher` never got the treatment, so on mobile it overhung the row. (The issue also listed `ApiDrawerTrigger`, but that one is already responsive — `NetworkSwitcher` is the actual culprit.)

## Fix

One file, `apps/ui/src/components/metagraphed/app-shell.tsx`:
- Wrap the header `NetworkSwitcher` in `hidden md:inline-flex`, matching its siblings — so it no longer renders below `md`.
- **Relocate it into the mobile navigation Sheet**, beside the developer-settings / wallet-connect actions that already live there, so it stays reachable on touch (not silently removed — per the deliverable). Did **not** just add `overflow-x: hidden`.

Pure visibility/relocation; the `md`+ layout is unchanged.

## Verification

Mobile "before" shows the `Mainnet` pill overhanging the right edge (the brand even clips to "Metagraph"); "after" shows it gone from the top bar — the full "Metagraphed" brand now fits and `scrollWidth <= innerWidth`. Desktop/tablet are unchanged (expected — mobile-only bug).

|---|---|---|
| Mobile · light |  |  |
| Mobile · dark |  |  |
| Tablet · light |  |  |
| Tablet · dark |  |  |
| Desktop · light |  |  |
| Desktop · dark |  |  |

Closes #6902

### DIFF
--- a/apps/ui/src/components/metagraphed/app-shell.tsx
+++ b/apps/ui/src/components/metagraphed/app-shell.tsx
@@ -215,7 +215,17 @@ export function AppShell({
               <div className="flex items-center gap-1">
                 <ApiDrawerTrigger />
 
-                <NetworkSwitcher />
+                {/* #6902: NetworkSwitcher was the one header action still
+                    rendering at every breakpoint — its siblings below
+                    (developer settings, SettingsPopover, WalletConnectButton)
+                    are all already `hidden md:inline-flex`, and ApiDrawerTrigger
+                    hides below md too. Left unhidden it overflowed the 375px
+                    mobile viewport by 10px (the switcher pill sat at right:385).
+                    Give it the same below-md treatment; it's relocated into the
+                    mobile navigation sheet so it stays reachable on touch. */}
+                <div className="hidden md:inline-flex">
+                  <NetworkSwitcher />
+                </div>
                 <Tooltip>
                   <TooltipTrigger asChild>
                     <Link
@@ -318,6 +328,10 @@ export function AppShell({
                 <Compass className="size-3" /> Unofficial registry
               </div>
               <MobileMegaMenu onNavigate={() => setMobileOpen(false)} />
+              {/* #6902: the header NetworkSwitcher is `hidden` below md (it
+                  overflowed the mobile viewport), so this is its mobile home —
+                  same rationale as the wallet-connect note below. */}
+              <NetworkSwitcher />
               <div className="flex items-center gap-2">
                 <Link
                   to="/settings"


JSONbored (commented):
Doesn't show what the new navbar sheet looks like after these changes, closing. Same reason I closed the previous PR...

Issue #6971, opened 2026-07-20
Explore: subnet resale marketplace + escrow/brokering (revenue feature)


JSONbored:
## Motivation

Taostats has a subnet-resale feature: shows the current owner of subnets that are for sale, plus some kind of Telegram-based escrow/brokering flow for connecting buyers and sellers of subnet ownership. Potential revenue angle (a cut of facilitated sales) plus historical resale-value tracking.

## Why this needs real scoping before any code work

This is NOT a data-tracking feature like #6970 (subnet ownership-change history) -- it's a money-transmission/escrow product:

- **Custody/escrow risk**: holding or brokering funds for a real-money transaction between two parties is a fundamentally different liability category than anything else in this codebase (which is read-only chain data + a freemium API). Needs real legal review (money transmission licensing considerations vary by jurisdiction) before any implementation.
- **No on-chain "for sale" signal**: unlike leasing (which has real on-chain storage -- `SubnetLeases`, `cost`, `end_block`, see `src/subnet-lease.mjs`), there is no on-chain concept of "this subnet is listed for sale" or a sale price for an outright `SubnetOwnerHotkey` transfer. Any "for sale" state and historical price data would need to come from an off-chain source (self-reported listings, a curated feed, or scraping/partnering with an existing source) -- itself a real product decision (trust model, spam/abuse prevention, how listings get created/removed).
- **Settlement mechanism unclear**: does "escrow" mean we hold funds, or just verify + coordinate an atomic-ish handoff (e.g., time the coldkey/hotkey ownership change to confirmed payment)? These have very different risk profiles and technical designs.

## Proposed next step (not implementation)

A scoping pass: decide whether this is (a) listings-only (no money changes hands through us, just a directory + "contact the owner" pattern -- low risk, could ship fast), or (b) full escrow/brokering (high risk, real legal/compliance work, much bigger lift). (a) could reasonably build on top of #6970's ownership-history data once that lands; (b) should not be started without that scoping decision made explicitly first.

## Related

- #6970 (subnet ownership-change history) -- the data layer either version of this would want as a foundation.
- Private revenue roadmap (GitHub Project #18) -- this should be logged there too as a candidate monetization idea.

JSONbored:
[...]

PR #6988, opened 2026-07-20
feat(indexer-rs): track subnet ownership-change history

### LINKED ISSUE #6970: Track subnet ownership-change history (who has owned a subnet over time)
## Motivation

Taostats has a subnet-resale feature that shows the current owner of subnets for sale, with a historical view of ownership. The data layer for this is genuinely useful on its own (independent of any resale/marketplace facilitation, see #6971) -- a "who has owned this subnet over its lifetime" history is a natural complement to the current-owner-only view we already ship.

## Current state

`apps/indexer-rs`'s poller binary already has a `subnet-ownership` job (`src/bin/poller/jobs/subnet_ownership.rs`) that resolves `SubtensorModule::SubnetOwnerHotkey(netuid)` -> `SubtensorModule::Owner(hotkey)` for every registered netuid and upserts a **latest-only** snapshot into the `subnet_ownership` table (one row per netuid, overwritten every tick). There is no history -- an ownership change simply overwrites the previous owner with no trace.

## Proposed scope

- Add a `subnet_ownership_history` table (diff-based append, same convention as `subnet_hyperparams_history`/`neuron_daily`: only insert a new row when the owner hotkey or coldkey actually changed since the last-known row for that netuid, not on every poll tick).
- Extend `subnet_ownership.rs`'s existing write path to also append to this history table when it detects a change (it already fetches the current DB row to upsert against, so the diff check is cheap).
- New API artifact (`GET /api/v1/subnets/:netuid/ownership-history` or similar) exposing the change history, schema-versioned like every other artifact in this repo.

## Explicitly out of scope here

- No "for sale" signal, no listing/marketplace UI, no escrow/brokering -- see #6971.
- No historical *price* tracking -- the chain has no concept of a sale price for an outright `SubnetOwnerHotkey` change (unlike the leasing pallet's `cost` field, which is a lease cost, not a sale price -- see `src/subnet-lease.mjs`). Price data, if ever tracked, would have to come from an off-chain source.

## Note

Do not confuse this with subnet **leasing** (`src/subnet-lease.mjs` / `src/subnet-lease-history.mjs`, epic #6717) -- leasing is Bittensor's own native crowdloan-funded "operate a subnet you don't own, share dividends with contributors" mechanism, already fully built and live. This issue is about outright ownership transfer, a different and currently uncovered concept.

JSONbored:
## Summary
Closes #6970

`subnet_ownership` was latest-only -- an ownership change simply overwrote the previous row with no trace, no way to answer "who has owned this subnet over its lifetime." Adds `subnet_ownership_history`: diff-based append, same convention as `subnet_hyperparams_history`/`neuron_daily`.

The poller now fetches the current `subnet_ownership` state once per tick (one query, not per-netuid) and appends a history row only for netuids whose resolved owner differs from what's stored, or has never been observed at all. A subnet whose ownership never changes accumulates exactly one history row (its first observation), not one per poll tick forever.

`subnet_ownership` itself is unchanged in behavior -- still upserts every tick regardless of whether the owner changed, preserving its existing `captured_at`-always-bumps freshness-signal contract.

## Explicitly out of scope (see #6970/#6971)
- No new API route exposing this history yet -- data layer first, a follow-up can add `GET /api/v1/subnets/:netuid/ownership-history` on top of this.
- No "for sale" signal, no marketplace/escrow -- see #6971, which needs real product/legal scoping before any code work.

## Test plan
- [x] `cargo build --release` / `cargo test --release` -- 62 tests pass (10 lib + 25 main-bin + 27 poller-bin, 4 new for the `owner_changed` diff logic)
- [x] `cargo fmt --check` / `cargo clippy --release --all-targets` -- clean
- [x] `node scripts/scan-public-safety.mjs` -- clean
- [x] **Live-verified against the real chain** (scratch Postgres, `POLLER_ONLY=subnet-ownership`, two separate ticks ~6 min apart): first tick wrote 128/128 rows to both `subnet_ownership` and `subnet_ownership_history` (every netuid correctly recorded as a first observation); second tick wrote 128/128 rows to `subnet_ownership` (captured_at bumped as expected) but **zero** new rows to `subnet_ownership_history` (no owner changes, confirmed no duplicate history entries)

### DIFF
--- a/apps/indexer-rs/src/bin/poller/jobs/subnet_ownership.rs
+++ b/apps/indexer-rs/src/bin/poller/jobs/subnet_ownership.rs
@@ -43,7 +43,18 @@
 // single stateless call, and a whole-tick failure just retries on the next
 // scheduled interval -- an acceptable cost for a periodic poller, unlike
 // main.rs's live-follow hot path this same reasoning would NOT apply to.
+//
+// subnet_ownership_history (metagraphed#6970): subnet_ownership above is
+// latest-only -- an ownership change just overwrites the previous row with
+// no trace. Every tick now also fetches the CURRENT state of
+// subnet_ownership before writing (one query, not per-netuid) and appends a
+// subnet_ownership_history row only when a netuid's resolved owner differs
+// from what's already stored, or has never been stored at all -- same
+// diff-based-append convention as subnet_hyperparams_history/neuron_daily,
+// so a subnet whose ownership never changes accumulates exactly one history
+// row (its first observation), not one per poll tick forever.
 
+use std::collections::HashMap;
 use std::time::Duration;
 
 use anyhow::{Context, Result};
@@ -131,7 +142,10 @@ async fn run(chain: &ChainClient, pg: &mut tokio_postgres::Client) -> Result<Job
     }
 
     let captured_at = now_ms();
-    let written = upsert(pg, &rows, captured_at)
+    let previous = fetch_current_owners(pg)
+        .await
+        .context("fetch current subnet_ownership")?;
+    let written = upsert(pg, &rows, captured_at, &previous)
         .await
         .context("upsert subnet_ownership")?;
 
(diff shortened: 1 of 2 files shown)


JSONbored:
Deployed to production 2026-07-20. One follow-up worth recording here: `subnet_ownership_history` started empty on deploy (the 128 existing `subnet_ownership` rows predate this feature, and the first poller tick after deploy correctly found no *changes* against them, so nothing was written).

Ran a one-time backfill directly against production to seed it, so the history table is useful immediately rather than waiting for the first real ownership transfer:

```sql
INSERT INTO subnet_ownership_history (netuid, owner_hotkey, owner_coldkey, captured_at)
SELECT netuid, owner_hotkey, owner_coldkey, captured_at
FROM subnet_ownership so
WHERE NOT EXISTS (SELECT 1 FROM subnet_ownership_history soh WHERE soh.netuid = so.netuid);
```

Confirmed 128/128 rows seeded. This was a one-time data migration, not a code gap -- the poller's own `owner_changed` logic already correctly treats "no previous history row" as a first observation for any netuid registered from now on; this only covered the transition moment where `subnet_ownership` already had data before `subnet_ownership_history` existed.

PR #7209, opened 2026-07-20
fix(relay): retry transient Postgres errors in chain-firehose-relay's poll loop

### LINKED ISSUE #7157: fix(relay): chain-firehose-relay crashes on transient Postgres errors instead of retrying
## Summary

`scripts/chain-firehose-relay.mjs`'s poll loop crashes the whole process on any transient Postgres/network error, instead of retrying. `pollOnce()`'s claim query (`UPDATE chain_firehose_outbox ... RETURNING id, payload`) has no try/catch, so an `ECONNRESET` or a `PostgresError: the database system is shutting down` (SQLSTATE `57P03`) propagates straight through `main()`'s loop to the top-level `main().catch()`, which exits the process. The container then restarts from scratch (fresh `git clone` + `npm ci` per `scripts/chain-firehose-relay-entrypoint.sh`).

Sentry: `Error: read ECONNRESET` and `PostgresError: the database system is shutting down`, both culprit `pollOnce(chain-firehose-relay)`, both first seen 2026-07-18 during a window where the underlying Postgres instance was cycling.

The file already has a well-built retry/backoff system for the outbound HTTP forward (`forwardWithRetry`, `computeBackoffDelayMs`) — it just isn't applied to the DB claim query itself.

## Expected behavior

A transient connection error from the claim query should be retried in-process with backoff, the same way an HTTP forward failure already is — not crash the whole relay.

## Actual behavior

Any error from `pollOnce()`'s claim query is fatal to the process.

## Validation

Confirmed via Sentry event stack traces (culprit `pollOnce`, no try/catch around the query in `scripts/chain-firehose-relay.mjs`) and by reading `scripts/chain-firehose-relay-entrypoint.sh`'s restart-from-scratch behavior on process exit.

JSONbored:
## Summary

`scripts/chain-firehose-relay.mjs`'s poll loop crashed the whole process on any transient Postgres/network error instead of retrying. Sentry shows `Error: read ECONNRESET` and `PostgresError: the database system is shutting down` (SQLSTATE `57P03`), both culprit `pollOnce`, during a window where the underlying Postgres instance was cycling — each crash triggers a full container restart (fresh `git clone` + `npm ci`).

## What Changed

- Added `isRetryablePollError` (`scripts/chain-firehose-relay.mjs`): classifies postgres.js's own connection-lifecycle codes (mirrors `workers/data-api.mjs`'s `RETRYABLE_CONNECTION_ERROR_CODES` — same underlying Postgres instance, same failure shapes), the raw Node socket codes this build's plain `postgres` package can surface with (e.g. `ECONNRESET`), and the Postgres "operator intervention" SQLSTATE class 57 (admin shutdown / crash / cannot connect now).
- `main()`'s poll loop now catches a failed `pollOnce()`, and for a retryable error, logs, backs off (reusing the existing `computeBackoffDelayMs`), and continues — instead of crashing. A non-retryable error, or more than `CHAIN_FIREHOSE_MAX_CONSECUTIVE_POLL_FAILURES` (5) consecutive retryable ones, still falls through to the existing crash-and-restart path.
- 6 new unit tests for `isRetryablePollError` (postgres.js codes, raw Node codes, SQLSTATE class 57, the negative case for a real bug like a syntax error, and the null/undefined-input edge cases) plus a sanity check on the new bound constant.

## Registry Safety

- [x] Links a tracked, currently-open issue (`Closes #7157`) — required.
- [x] Generated artifacts were produced by repo scripts, not hand-edited. (N/A — no generated artifacts touched.)
- [x] R2-only/high-churn detail artifacts are not committed. (N/A.)
- [x] Public API/OpenAPI/schema changes are intentional and documented. (N/A — no API/schema surface touched; this is a standalone deploy script.)

Closes #7157

## Validation

- [x] `npm run test:coverage` — full unsharded suite (335 files) green; the changed file's new code is 100% covered (verified in isolation via `vitest run tests/chain-firehose-relay.test.mjs --coverage`); `main()`'s poll/cleanup loop is pre-existing `/* v8 ignore */`'d (needs a real Postgres connection + process lifecycle, documented at the top of the test file) — the new retry/backoff decision is pulled out into the tested `isRetryablePollError` function instead, matching that file's existing convention for `forwardWithRetry`/`computeBackoffDelayMs`.
- [x] `git diff --check`
- [x] `npm run lint` / `npm run format:check` (scoped to the 2 changed files)

Not run (orthogonal to this change — no schema/registry/API/contract files touched): `npm run validate`, `validate:schemas`, `validate:api`, `validate:openapi`, `validate:types`, `validate:artifact-budgets`, `validate:docs`, `validate:intake`, `validate:workflows`, `worker:test`. This PR only touches a standalone Node deploy script (`scripts/chain-firehose-relay.mjs`) and its test file.

## Notes

- Root-caused via Sentry triage. A sibling fix (`ecde90f3`, already merged and confirmed deployed) added a similar retry for the Cloudflare Worker side of the same underlying Postgres instance; this closes the same gap for the standalone relay script, which had no error handling around its DB query at all.

### DIFF
--- a/scripts/chain-firehose-relay.mjs
+++ b/scripts/chain-firehose-relay.mjs
@@ -376,6 +376,41 @@ export function computeBackoffDelayMs(
   return Math.min(baseMs * 2 ** attempt, maxMs);
 }
 
+// Codes worth retrying a poll iteration in-process for, rather than crashing the whole relay (#7157):
+// postgres.js's own connection-lifecycle codes (mirrors workers/data-api.mjs's
+// RETRYABLE_CONNECTION_ERROR_CODES -- same underlying Postgres instance, same failure shapes), the raw
+// Node socket codes a dead TCP connection surfaces with on THIS build (this is plain `postgres`, not
+// postgres/cf's Cloudflare polyfill -- a socket 'error' event's native code passes through un-wrapped,
+// which is what a real `Error: read ECONNRESET` event showed), and the Postgres "operator intervention"
+// SQLSTATE class 57 (admin shutdown / crash / cannot connect now -- a real `PostgresError: the database
+// system is shutting down` event was 57P03). Anything outside this set (a syntax error, a constraint
+// violation) is a real bug, not a blip, and should still surface/crash rather than retry forever.
+const CHAIN_FIREHOSE_RETRYABLE_POLL_ERROR_CODES = new Set([
+  "CONNECTION_CLOSED",
+  "CONNECTION_DESTROYED",
+  "CONNECT_TIMEOUT",
+  "CONNECTION_ENDED",
+  "ECONNRESET",
+  "ECONNREFUSED",
+  "ETIMEDOUT",
+  "EPIPE",
+  "57P01", // admin_shutdown
+  "57P02", // crash_shutdown
+  "57P03", // cannot_connect_now
+]);
+
+export function isRetryablePollError(error) {
+  return (
+    typeof error?.code === "string" &&
+    CHAIN_FIREHOSE_RETRYABLE_POLL_ERROR_CODES.has(error.code)
+  );
+}
+
+// A transient blip retries indefinitely-ish but bounded per burst -- this many CONSECUTIVE retryable
+// failures (reset by any successful poll) still exits the process, preserving the container-restart
+// safety net for a genuinely stuck connection rather than looping forever.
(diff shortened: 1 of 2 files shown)


JSONbored:
Superseded by #7164, which merged first and fixes the same root cause (pollOnce's unguarded claim query crashing the relay on a transient Postgres error) via an equivalent retry-with-backoff approach. Closing to avoid landing duplicate/conflicting logic — thanks for the fast turnaround on this one.

Issue #7098, opened 2026-07-20
MCP execute: verify + wire SN85 (Vidaio) once Phase 1 ships


JSONbored:
Part of #7013. Phase 1 (#7014) shipped (#7215) -- `call_subnet_surface` is live. Ready to verify.

## Scope
Confirm SN85 (Vidaio)'s callable surface(s) below actually work end-to-end -- an actual request against the live URL, not an assumption. `call_subnet_surface` is a safety-checked passthrough of the exact same request (same URL, same method), not a black box -- calling the URL directly yourself is equally valid verification for a no-auth GET surface.

## How to verify
- MCP client: connect per the README (`claude mcp add --transport http metagraphed https://api.metagraph.sh/mcp`), then call `call_subnet_surface` with the surface's `surface_id` from the list below.
- Or, for a no-auth GET surface, just request the URL yourself (curl, browser, whatever) -- it's the same request the tool makes.

## Surface(s) to verify (3)
- `sn-85-vidaio-openapi` (openapi, GET, no-auth, has captured schema): https://api.vidaio.io/openapi.json
- `sn-85-vidaio-health` (subnet-api, ?, no-auth, no schema (single fixed endpoint)): https://api.vidaio.io/health
- `sn-85-vidaio-ready` (subnet-api, ?, no-auth, no schema (single fixed endpoint)): https://api.vidaio.io/ready

## Deliverable
- **Found something broken** (wrong `probe.method`/`probe.expect`, stale/missing schema, wrong `auth_required`, a moved/dead URL, a response shape the tool mishandles)? Fix that surface's entry in `registry/subnets/<slug>.json` and append a `Live-verified <date>: <what you observed>` note to it -- e.g. metagraphed#7143 (SN105 Beam: added a missing `probe` block, noted the verified response).
- **Everything already matches reality?** Say so in a closing comment -- what you called, what came back. A regression test pinning the current-good behavior (matching the existing `tests/call-subnet-surface*.test.mjs` pattern) is welcome but not required.
- Either way: every claim in the PR (a probe-config change, a "verified working" comment, a test's expected response) has to trace back to a request you actually made. Don't guess.

## Additional surface(s) found during review (now in scope)
[...]

Issue #7014, opened 2026-07-20
MCP execute Phase 1: no-auth call_subnet_surface tool


JSONbored:
Part of #7013.

## Scope

A new MCP tool (working name `call_subnet_surface`) restricted to catalogued surfaces with `auth_required: false` (388 of 395 callable surfaces today — ~93% of subnets covered by this phase alone).

- Input: `surface_id` (same id/key/alias resolution `verify_integration` already uses via `findCataloguedSurface`), plus optional `query` params for surfaces whose `notes`/schema indicate they accept them.
- Behavior: fetch the surface's curated `url` with its curated `probe.method` (GET/HEAD as declared), merge in any supplied query params, return the response body (not just health/status metadata like `verify_integration` does today).
- Build directly on `probeSurface` (`src/health-probe-core.mjs`) / the same SSRF-safe fetch path `verify_integration` and the 15-minute health cron already use — only the curated URL is ever fetched, never an arbitrary one. Do not reimplement the safety guard.
- Bound the response: size cap, content-type allowlist (JSON primarily; reject/truncate unexpected binary), timeout matching the surface's own `probe.timeout_ms`.
- Reject any surface with `auth_required: true` outright (that's Phase 3's job) — fail closed, not silently drop the auth requirement.

## Explicitly out of scope (later phases)

- Path/method beyond the surface's own declared `url`/`probe.method` (Phase 2).
- Any surface requiring a credential (Phase 3).

## Why maintainer-only

New outbound-call execution surface from MCP tool input, even though scoped to curated no-auth URLs only — same trust-boundary category as `verify_integration`'s own design, which is itself maintainer-built infrastructure.

PR #7344, opened 2026-07-21
test(mcp): verify SN85 Vidaio surfaces end-to-end via call_subnet_surface

lourincedaging0-commits:
Closes #7098

Phase 1 (#7014/#7215) shipped `call_subnet_surface`; this confirms SN85 (Vidaio)'s three callable surfaces work end-to-end.

**All three live-verified 2026-07-21** to return **HTTP 200 application/json**:
- `sn-85-vidaio-openapi` — OpenAPI 3.1 document
- `sn-85-vidaio-health` — `{ status: "ok" }`
- `sn-85-vidaio-ready` — `{ status: "ok" }`

Registry config (`registry/subnets/vidaio.json`) is **already correct**, so **no registry change is needed**: `openapi` carries its `schema_url` + GET probe; `health`/`ready` have no `probe` block, and `call_subnet_surface` defaults to `GET` when a surface has no `probe.method` (`src/call-subnet-surface.mjs`), so they are callable as-is.

Adds `tests/sn85-call-subnet-surface-verify.test.mjs` (mirrors the merged SN28/SN115/SN120 verify tests, data-driven over the three surfaces): pins each surface's config to the tool's contract (including the probe-less GET default), asserts `callSubnetSurface` issues a `GET` to the surface's own url and returns the JSON body (stable shape, not exact live values), and drives the full `call_subnet_surface` MCP tool end-to-end resolved by `surface_id`. Hermetic — fixtures mirror the live shapes. Verified locally: 9/9 tests pass, prettier + eslint clean.

### DIFF
--- a/tests/sn85-call-subnet-surface-verify.test.mjs
+++ b/tests/sn85-call-subnet-surface-verify.test.mjs
@@ -1,0 +1,178 @@
+// SN85 (Vidaio) end-to-end verification for the call_subnet_surface MCP tool
+// (metagraphed#7098, MCP execute Phase 1 follow-up #7014/#7215). Unlike
+// tests/call-subnet-surface-mcp.test.mjs -- which proves the tool wiring with
+// synthetic surfaces -- this file pins SN85's three *real* registry surfaces
+// (registry/subnets/vidaio.json) to the tool's contract, so a future edit that
+// regresses their callability (marking one auth_required, moving a URL, or --
+// for the two probe-less health endpoints -- adding a HEAD probe that would
+// change how call_subnet_surface calls them) is caught here.
+//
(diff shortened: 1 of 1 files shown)


JSONbored:
Re-reviewed. This close was correct — a real gap, not a false positive. The review found this test's synthetic operational-surfaces catalog includes the `openapi`-kind surface in its end-to-end MCP-tool assertion, but the SN51 precedent in this repo documents that `openapi` isn't in `OPERATIONAL_SURFACE_KINDS` and is therefore absent from the real production catalog — so the 'verified end-to-end' claim for `sn-85-vidaio-openapi` isn't actually representative of production. A resubmit restructured like the SN51 test (module-level-only for the openapi surface) would fix this.

Issue #7015, opened 2026-07-20
MCP execute Phase 2: schema-aware path/method execution


JSONbored:
Part of #7013. Depends on Phase 1 (#7014).

## Scope

Extend `call_subnet_surface` to accept a `path` + `method` for the 78 surfaces (mostly `openapi`-kind) that have a captured schema, instead of only the surface's single declared `url`.

- Validate the requested `path`/`method` against the surface's own captured OpenAPI schema (`get_api_schema` already serves it) before making the call -- reject anything not declared in the schema, never an arbitrary path on the host.
- Still no-auth-only in this phase (same `auth_required: false` restriction as Phase 1) -- this phase is about *breadth of what's callable per surface*, not credentials.
- Support request bodies for POST/PUT where the schema declares them, with the same size/content-type bounds as Phase 1's response handling.

## Explicitly out of scope

- Any surface requiring a credential (Phase 3, #7016).
- Surfaces without a captured schema stay on Phase 1's single-`url` behavior -- no speculative path guessing.

JSONbored:
All three sub-issues (#7673, #7674, #7675) are merged. Phase 2 is complete: call_subnet_surface now supports schema-validated GET/HEAD/POST/PUT path/method execution for surfaces with a captured schema, alongside the existing single-url Phase 1 behavior.

PR #7350, opened 2026-07-21
test(mcp): verify SN54 Yanez MIID surfaces via call_subnet_surface (#7067)

### LINKED ISSUE #7067: MCP execute: verify + wire SN54 (Yanez MIID) once Phase 1 ships

## Scope
Confirm SN54 (Yanez MIID)'s callable surface(s) below actually work end-to-end -- an actual request against the live URL, not an assumption. `call_subnet_surface` is a safety-checked passthrough of the exact same request (same URL, same method), not a black box -- calling the URL directly yourself is equally valid verification for a no-auth GET surface.

## How to verify

## Surface(s) to verify (4)
- `sn-54-yanez-compliance-openapi` (openapi, GET, no-auth, has captured schema): https://miners-stats.yanezcompliance.net/openapi.json
- `sn-54-yanez-compliance-subnet-api` (subnet-api, GET, no-auth, no schema (single fixed endpoint)): https://miners-stats.yanezcompliance.net/miners_stats
- `sn-54-yanez-miners-stats-index` (subnet-api, GET, no-auth, no schema (single fixed endpoint)): https://miners-stats.yanezcompliance.net/
- `sn-54-yanez-miners-stats-health` (subnet-api, GET, no-auth, no schema (single fixed endpoint)): https://miners-stats.yanezcompliance.net/health

## Deliverable

## Additional surface found during review (now in scope)

While reviewing this issue, checked SN54's own OpenAPI schema (https://miners-stats.yanezcompliance.net/openapi.json) against what's currently registered in `registry/subnets/yanez-miid.json` and found one more real, documented GET endpoint that isn't registered:

- `GET /miners/{miner_uid}` (per-miner score history -- confirmed live with a real uid, e.g. `/miners/0` returns an array of that miner's scored evaluations over time)

Like the surfaces above it's public and no-auth, but it needs a caller-supplied miner uid in the path -- there's no single fixed URL to auto-probe the way `/miners_stats` works. Fully calling it (an agent picks a uid it found via `/miners` and looks up that miner's history) is what MCP execute Phase 2 (schema-aware path/method execution, #7015) is for, not yet built.

That's not a reason to skip registering it, though -- the goal is for our MCP to support everything a subnet's own public API actually exposes, not a curated subset. Please add it to `registry/subnets/yanez-miid.json` now, so it's at least discoverable via `list_endpoints`/`get_subnet_endpoints` ahead of Phase 2:
- Use the literal path template as the `url` (e.g. `https://miners-stats.yanezcompliance.net/miners/{miner_uid}`), not a specific example uid baked in.
- Set `probe.enabled: false` (nothing to safely auto-probe without a real caller-supplied uid) and note why in `notes`.
- Confirm the shape with a real, currently-valid uid (pull one from `/miners`) before registering, so the claim is verified, not guessed -- just don't bake that specific uid into the registered `url`.

This supersedes the "Finding/adding NEW surfaces... out of scope" line below for this one specifically -- it's in scope for this issue.

## Explicitly out of scope

pekar9781:
## Summary
- Live-verified all four SN54 (Yanez MIID) surfaces against `miners-stats.yanezcompliance.net` (openapi, miners_stats, index, health) — probes already match (GET, no-auth, json).
- Adds hermetic `call_subnet_surface` regression tests loading real surfaces from `registry/subnets/yanez-miid.json`. OpenAPI surface is direct-call only (not an operational kind); the three `subnet-api` surfaces also exercise the MCP `tools/call` path.
- No registry edits.

Closes #7067

## Test plan
- [x] `vitest run tests/sn54-call-subnet-surface-verify.test.mjs` (12 passed)
- [x] Live GET of all four URLs returned 200 with expected shapes (2026-07-21)
- [ ] CI green on this PR

### DIFF
--- a/tests/sn54-call-subnet-surface-verify.test.mjs
+++ b/tests/sn54-call-subnet-surface-verify.test.mjs
@@ -1,0 +1,255 @@
+// SN54 (Yanez MIID) end-to-end verification for the call_subnet_surface MCP
+// tool (metagraphed#7067, MCP execute Phase 1 follow-up #7014/#7215). Unlike
+// tests/call-subnet-surface-mcp.test.mjs -- which proves the tool wiring with
+// synthetic surfaces -- this file pins SN54's *real* registry surface configs
+// (registry/subnets/yanez-miid.json) to the tool's contract, so a future edit
+// that regresses their callability (flipping to HEAD, marking one
+// auth_required, disabling a probe) is caught here.
+//
+// All four surfaces listed in #7067 were live-verified 2026-07-21:
+//   sn-54-yanez-compliance-openapi     GET .../openapi.json -> OpenAPI 3.1.0
+//   sn-54-yanez-compliance-subnet-api  GET .../miners_stats -> {timestamp,miners:[...]}
+//   sn-54-yanez-miners-stats-index     GET .../             -> {message,version,endpoints}
+//   sn-54-yanez-miners-stats-health    GET .../health       -> {status:"healthy",...}
+// Fixtures mirror stable top-level shapes (hermetic). kind "openapi" is not in
+// OPERATIONAL_SURFACE_KINDS, so that surface is direct-call verified only --
+// same pattern as sn-74-gittensor-openapi in tests/sn74-call-subnet-surface-verify.
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import { describe, test } from "vitest";
+import { callSubnetSurface } from "../src/call-subnet-surface.mjs";
+import { OPERATIONAL_SURFACE_KINDS } from "../src/health-probe-core.mjs";
+import { handleMcpRequest } from "../src/mcp-server.mjs";
+
+const NETUID = 54;
+
+const registry = JSON.parse(
+  readFileSync(
+    fileURLToPath(
+      new URL("../registry/subnets/yanez-miid.json", import.meta.url),
+    ),
+    "utf8",
+  ),
+);
+
+function surfaceOf(id) {
+  return registry.surfaces.find((surface) => surface.id === id);
+}
+
+function jsonResponse(body) {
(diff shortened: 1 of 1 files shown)


JSONbored (commented):
The four surfaces you verified are accurate — live-checked and confirmed myself, registry already matches reality on all four, nothing to fix there.

But the actual goal of this issue is full MCP parity with what SN54 exposes, not just the four surfaces originally listed. While reviewing, I checked SN54's own OpenAPI schema against `registry/subnets/yanez-miid.json` and found one more real, working, public no-auth endpoint that's never been registered:

- `GET /miners/{miner_uid}` — per-miner score history. Confirmed live (e.g. `/miners/0` returns that miner's scored evaluation history).

I've added this to issue #7067 as in-scope (see the issue body for the full write-up). Before this is complete: please add it to `registry/subnets/yanez-miid.json` too —
- `url` as the literal path template (`https://miners-stats.yanezcompliance.net/miners/{miner_uid}`), not a specific uid baked in (it'd read fine today but there's no reason to hardcode when the template is what's真 true)
- `probe.enabled: false` (nothing to safely auto-probe without a real caller-supplied uid) with a note explaining why
- Confirm the shape against a real uid pulled from `/miners` first, so the entry is verified, not guessed — just don't bake that uid into the registered `url`

This one's parameterized, so it won't be fully *callable* through `call_subnet_surface` until MCP execute Phase 2 (#7015) ships — but it should still be registered now so it's at least discoverable.

JSONbored:
Re-reviewed. This is actually more thorough than what merged and closed #7067 (#7400, which only verified 1 of 4 originally-listed surfaces) — you correctly verified all 4. But #7067 also carries an "Additional surface found during review" section (`GET /miners/{miner_uid}`, now in scope) that neither PR touched, so this still doesn't reach full current scope. Worth flagging: #7067 shows closed despite that gap — separate issue from this PR's own merits, and something we'll look at directly rather than asking you to fix via this thread.

PR #7369, opened 2026-07-21
fix(registry): wire SN79 MVTRX MCP execute probe config (#7092)

### LINKED ISSUE #7092: MCP execute: verify + wire SN79 (MVTRX) once Phase 1 ships

## Scope
Confirm SN79 (MVTRX)'s callable surface(s) below actually work end-to-end -- an actual request against the live URL, not an assumption. `call_subnet_surface` is a safety-checked passthrough of the exact same request (same URL, same method), not a black box -- calling the URL directly yourself is equally valid verification for a no-auth GET surface.

## How to verify

## Surface(s) to verify (1)
- `sn-79-taos-im-subnet-api` (subnet-api, ?, no-auth, no schema (single fixed endpoint)): https://mvtrx.exchange/health

## Deliverable

## Explicitly out of scope

pekar9781:
## Summary
- Live-verified `sn-79-taos-im-subnet-api`: `GET https://mvtrx.exchange/health` → HTTP 200 `application/json` `{"status":"ok","mode":"host","network":"mainnet"}`. `HEAD` → HTTP 405 (so method must be GET, matching the issue's `?`).
- Added the missing `probe` block (`enabled`, `GET`, `expect: json`) so the surface can enter the operational catalog for `call_subnet_surface`.
- Appended a short factual Live-verified note documenting the GET/HEAD finding.

Closes #7092

## Test plan
- [x] Live GET `/health` → 200 JSON; HEAD → 405
- [x] `npm run validate:surface -- registry/subnets/mvtrx.json`
- [ ] CI green on this PR

### DIFF
--- a/registry/subnets/mvtrx.json
+++ b/registry/subnets/mvtrx.json
@@ -91,7 +91,13 @@
       "id": "sn-79-taos-im-subnet-api",
       "kind": "subnet-api",
       "name": "MVTRX exchange health API",
-      "notes": "Public no-auth GET /health on mvtrx.exchange returning JSON liveness for the TAOS IM trading host (observed: {\"status\":\"ok\",\"mode\":\"host\",\"network\":\"mainnet\"}). The official SN79 README links mvtrx.exchange as the project exchange.",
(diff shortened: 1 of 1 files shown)


JSONbored:
Re-reviewed. This is confirmed unnecessary rather than wrong — the already-merged #7383 explicitly notes "No registry edits (prior probe-only PR was closed by the registry semantic gate)" for this exact surface, because `callSubnetSurface` defaults to GET when `probe.method` is unset, which already matches this host's behavior. Adding an explicit probe block isn't a defect fix here, just redundant — #7383 (test-only, pinning the existing-good behavior) was the correct shape, and it already closed #7092.

PR #7403, opened 2026-07-21
test(sn8): verify Vanta call_subnet_surface

### LINKED ISSUE #7024: MCP execute: verify + wire SN8 (Vanta) once Phase 1 ships

## Scope
Confirm SN8 (Vanta)'s callable surface(s) below actually work end-to-end -- an actual request against the live URL, not an assumption. `call_subnet_surface` is a safety-checked passthrough of the exact same request (same URL, same method), not a black box -- calling the URL directly yourself is equally valid verification for a no-auth GET surface.

## How to verify

## Surface(s) to verify (1)
- `sn-8-vanta-model-parameters` (subnet-api, GET, no-auth, no schema (single fixed endpoint)): https://github.com/taoshidev/vanta-network/raw/main/vali_objects/utils/model_parameters/all_model_parameters.json

## Deliverable

## Explicitly out of scope

kai392:
## Summary

Closes #7024

Path B test-only: pins SN8 Vanta's single callable surface `sn-8-vanta-model-parameters` (GET GitHub raw `all_model_parameters.json`) to `call_subnet_surface` — registry assertions, direct `callSubnetSurface`, and MCP `tools/call` e2e.

Live-verified 2026-07-21: HTTP 200 `text/plain; charset=utf-8` JSON with per-asset slippage model parameters (`equity.*`).

## Test plan

- [x] `npx vitest run tests/sn8-call-subnet-surface-verify.test.mjs` — 3/3 pass
- [x] `npx prettier --check tests/sn8-call-subnet-surface-verify.test.mjs` — clean

### DIFF
--- a/tests/sn8-call-subnet-surface-verify.test.mjs
+++ b/tests/sn8-call-subnet-surface-verify.test.mjs
@@ -1,0 +1,137 @@
+// SN8 (Vanta) end-to-end verification for the call_subnet_surface MCP
+// tool (metagraphed#7024, MCP execute Phase 1 follow-up #7014/#7215). Unlike
+// tests/call-subnet-surface-mcp.test.mjs -- which proves the tool wiring with
+// synthetic surfaces -- this file pins SN8's *real* no-auth GET JSON
+// registry surface (registry/subnets/vanta.json) to the tool's contract.
+//
+// Live-verified 2026-07-21:
+//   sn-8-vanta-model-parameters  GET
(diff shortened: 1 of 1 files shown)


JSONbored:
Re-reviewed. #7024 has no additional-scope section (single surface: `sn-8-vanta-model-parameters`) and was already fully satisfied by the merged #7424 (test-only, live-verified). This PR would have been redundant even before its own CI failure (`checks` job failed) — nothing further needed.

Issue #7013, opened 2026-07-20
Roadmap: MCP execute — let AI agents call any subnet's live API directly


JSONbored:
## Why

Today the MCP fully covers *discovery* of any subnet's API — `list_subnet_apis`, `get_api_schema`, `get_fixture`, and `how_do_i_call` already tell an agent everything about how to call any of the 129 subnets' services. What's missing is *execution*: no tool lets an agent actually make the call and get the real response back through the MCP itself. This closes that gap, so a single MCP connection is enough for an agent to discover **and use** any subnet's live API.

## Ground truth (verified against the real registry, 2026-07-20, not estimated)

- **120 of 129 subnets (93%)** already have ≥1 callable surface (`subnet-api`/`openapi`/`sse` kind) — 395 surfaces total.
- **388 of 395 callable surfaces (98%) require no auth.** Only 7 do, and all 7 already carry a structured `{scheme, location, header name, value_format}` auth descriptor from #746.
- **78 surfaces (~20%) have a captured OpenAPI schema** (multi-path APIs); the other **317 are single fixed endpoints** (one `url` + one `probe.method`, no path ambiguity).
- The outbound-fetch safety machinery already exists and runs at scale: `probeSurface` (`src/health-probe-core.mjs`) is the same SSRF-safe, timeout-bounded fetcher the 15-minute health cron and `verify_integration` already use across the whole catalog. It returns health/status metadata today, not the response body — that's the actual gap, not the fetch mechanism.
- 9 subnets have zero callable surface today (netuid 0/root, one inactive subnet, 7 genuine registry gaps) — those need a registry fix first (separate, existing enrichment process), not an MCP fix; out of scope here.

Note: by the time Phase 3 shipped, organic registry growth (ongoing contributor surface additions) had moved the "7 authenticated surfaces" figure to over 1,000 -- the phase's generic bearer/api-key mechanism scales to whatever the real count is at any given time, unaffected by the stale snapshot above.

## Phases (maintainer-only — security/trust-boundary work, see child issues)

- [x] Phase 1 — no-auth execute tool, covers ~93% of subnets immediately (#7014)
- [x] Phase 2 — schema-aware path/method execution for OpenAPI-backed surfaces (#7015)
- [x] Phase 3 — credential-passthrough execute for authenticated surfaces (#7016)

## Per-subnet wiring (contributor-eligible, `gittensor:feature`, gated on Phase 1 landing)

[...]

PR #7406, opened 2026-07-21
test(mcp): verify SN10 Swap surfaces end-to-end via call_subnet_surface

### LINKED ISSUE #7026: MCP execute: verify + wire SN10 (Swap) once Phase 1 ships

## Scope
Confirm SN10 (Swap)'s callable surface(s) below actually work end-to-end -- an actual request against the live URL, not an assumption. `call_subnet_surface` is a safety-checked passthrough of the exact same request (same URL, same method), not a black box -- calling the URL directly yourself is equally valid verification for a no-auth GET surface.

## How to verify

## Surface(s) to verify (2)
- `sn-10-taofi-openapi` (openapi, GET, no-auth, has captured schema): https://taofi-doc.web.app/openapi.yaml
- `sn-10-taofi-api` (subnet-api, GET, no-auth, has captured schema): https://taofi-api.web.app/

## Deliverable

## Explicitly out of scope

joaovictor712:
Closes #7026

Part of #7013. Phase 1 (#7014) shipped (#7215) — `call_subnet_surface` is live.

## Scope

Confirm SN10 (Swap)'s callable surface(s) actually work end-to-end — an actual request against the live URL, not an assumption.

## Verification (real requests, 2026-07-21)

| Surface | Request | Result |
| --- | --- | --- |
| `sn-10-taofi-openapi` | `GET https://taofi-doc.web.app/openapi.yaml` | **HTTP 200** `text/yaml` — ~44 KB OpenAPI document |
| `sn-10-taofi-api` | `GET https://taofi-api.web.app/` | **HTTP 404** — host is genuinely down right now |

The first surface works end-to-end and its registry config (`openapi`, `auth_required: false`, probe `enabled`+`GET`+`any`) already matches reality — `expect: "any"` is correct since the response is YAML, not JSON.

The second surface's registry entry already has `probe.enabled: false`. Live-verifying it confirms that's the *correct* current state, not a bug — the host 404s right now, so the existing disabled flag accurately reflects reality rather than needing a fix. This test pins that fact so a future PR doesn't accidentally re-enable it without the host actually coming back.

## What the test does

- Asserts the real `registry/subnets/swap.json` `sn-10-taofi-openapi` entry is configured callable (kind, no-auth, probe enabled + `GET` + `any`, schema_url).
- `callSubnetSurface` issues the exact `GET` and returns the raw YAML body (non-JSON — not parsed).
- End-to-end through the `call_subnet_surface` MCP tool resolved by `surface_id`.
- Separately asserts `sn-10-taofi-api`'s metadata and its deliberate `probe.enabled: false`, matching its live-verified 404 state — no live call is exercised for it, since none is expected to succeed.
- Hermetic fixture for the working surface — no network in CI.

## Validation

- [x] `npx vitest run tests/swap-call-subnet-surface-verify.test.mjs` — 4 passed
- [x] eslint on the file · `npx prettier --check` — clean
- [x] One new file only: `tests/swap-call-subnet-surface-verify.test.mjs`

### DIFF
--- a/tests/swap-call-subnet-surface-verify.test.mjs
+++ b/tests/swap-call-subnet-surface-verify.test.mjs
@@ -1,0 +1,147 @@
+// SN10 (Swap / TAOFi) end-to-end verification for the call_subnet_surface
+// MCP tool (metagraphed#7026, MCP execute Phase 1 follow-up #7014/#7215).
+// Unlike tests/call-subnet-surface-mcp.test.mjs -- which proves the tool
+// wiring with synthetic surfaces -- this file pins SN10's two issue-scoped
+// registry surfaces (registry/subnets/swap.json) to the tool's contract, so
+// a future edit that regresses their callability (flipping to HEAD,
+// marking them auth_required, wrongly re-enabling a dead one) is caught
+// here.
+//
+// Live-verified 2026-07-21:
+//   - sn-10-taofi-openapi: GET https://taofi-doc.web.app/openapi.yaml ->
+//     200 text/yaml, ~44 KB OpenAPI document. probe.enabled: true,
+//     expect: "any" (correctly not "json" -- the response is YAML, not
+//     JSON) already matches this exactly.
+//   - sn-10-taofi-api: GET https://taofi-api.web.app/ -> 404 (the host is
+//     genuinely down/gone right now). The registry already has
+//     probe.enabled: false for this surface -- that already correctly
+//     reflects the live broken state, not a config defect to fix. This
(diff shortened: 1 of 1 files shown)


JSONbored:
Re-reviewed. This is a near-identical duplicate of the already-merged #7414 — same two surfaces, same verification results word-for-word (openapi works, the second surface's `probe.enabled:false` correctly reflects a currently-404ing host). #7026 has no additional-scope section, so #7414 already fully closed it. Redundant regardless of this PR's own CI issue.

Issue #7134, opened 2026-07-20
MCP execute: verify + wire SN126 (Poker44) once Phase 1 ships


JSONbored:
Part of #7013. Phase 1 (#7014) shipped (#7215) -- `call_subnet_surface` is live. Ready to verify.

## Scope
Confirm SN126 (Poker44)'s callable surface(s) below actually work end-to-end -- an actual request against the live URL, not an assumption. `call_subnet_surface` is a safety-checked passthrough of the exact same request (same URL, same method), not a black box -- calling the URL directly yourself is equally valid verification for a no-auth GET surface.

## How to verify
- MCP client: connect per the README (`claude mcp add --transport http metagraphed https://api.metagraph.sh/mcp`), then call `call_subnet_surface` with the surface's `surface_id` from the list below.
- Or, for a no-auth GET surface, just request the URL yourself (curl, browser, whatever) -- it's the same request the tool makes.

## Surface(s) to verify (1)
- `sn-126-poker44-health` (subnet-api, GET, no-auth, no schema (single fixed endpoint)): https://api.poker44.net/health

## Deliverable
- **Found something broken** (wrong `probe.method`/`probe.expect`, stale/missing schema, wrong `auth_required`, a moved/dead URL, a response shape the tool mishandles)? Fix that surface's entry in `registry/subnets/<slug>.json` and append a `Live-verified <date>: <what you observed>` note to it -- e.g. metagraphed#7143 (SN105 Beam: added a missing `probe` block, noted the verified response).
- **Everything already matches reality?** Say so in a closing comment -- what you called, what came back. A regression test pinning the current-good behavior (matching the existing `tests/call-subnet-surface*.test.mjs` pattern) is welcome but not required.
- Either way: every claim in the PR (a probe-config change, a "verified working" comment, a test's expected response) has to trace back to a request you actually made. Don't guess.

## Additional surface(s) found during review (now in scope)

[...]

PR #7447, opened 2026-07-21
test(mcp): verify SN126 Poker44 surfaces via call_subnet_surface (#7134)

joaovictor91123:
Verifies SN126 (Poker44)'s registry surfaces against the `call_subnet_surface` MCP tool contract, and registers the additional benchmark-chunks surface the issue flagged as newly in scope. Closes #7134.

**Surfaces (all live-verified 2026-07-21):**

Operational kinds — exercised **end-to-end** through the MCP tool (resolved by `surface_id`):
- `sn-126-poker44-health` — GET `https://api.poker44.net/health` → 200 `application/json`, `{"success":true,"data":{"status":"healthy",...}}`.
- `sn-126-poker44-data-artifact` — GET `/api/v1/benchmark` → 200 `application/json` (`releaseVersion`, `totalChunks`, `latestSourceDate`, …).
- `sn-126-poker44-benchmark-releases` — GET `/api/v1/benchmark/releases` → 200 `application/json` (`releases[]`).

**New surface added this PR** (the issue's additional-scope item):
- `sn-126-poker44-benchmark-chunks` — GET `/api/v1/benchmark/chunks`. The bare URL is **HTTP 422** (requires a `sourceDate`), so it is registered `auth_required:false`, `probe.enabled:false`. Live-verified `GET /api/v1/benchmark/chunks?sourceDate=2026-07-21&limit=1` → **200** `application/json` listing `chunkId`/`chunkHash`/`chunkIndex` + a `nextCursor`. Because it needs only a query param (not a path param) it is genuinely callable today via `call_subnet_surface`'s `query` argument — the test pins exactly that.
- The sibling **chunk-by-id** route (`GET /api/v1/benchmark/chunks/{chunkId}`, also live-verified 2026-07-21 → 200) is path-parameterized with no fixed callable URL, so it cannot be registered as a `Surface` (the schema requires a concrete `uri`) — same as SN112's `/v1/apps/{app_id}/status`.

Non-operational kinds (not in `OPERATIONAL_SURFACE_KINDS`, so direct-call verified only):
- `sn-126-poker44-dashboard` / `sn-126-poker44-website` / `sn-126-poker44-source` — HEAD → 200 `text/html`.
- `sn-126-poker44-whitepaper` — HEAD → 200 `application/pdf`. The tool classifies `application/pdf` as **binary** and refuses to return its body (`ok:false`, "unsupported content-type"); the test asserts that honest behavior explicitly.

Every expected value in the test traces back to a request actually made against the live URL. `npx vitest run` → 23/23 pass; schema + surface + readme-catalog validators all green.

### DIFF
--- a/registry/subnets/poker44.json
+++ b/registry/subnets/poker44.json
@@ -183,6 +183,30 @@
         "https://poker44.net/"
       ],
       "url": "https://poker44.net/Poker44_Whitepaper.pdf"
+    },
+    {
+      "auth_required": false,
+      "authority": "community",
+      "id": "sn-126-poker44-benchmark-chunks",
+      "kind": "data-artifact",
+      "name": "Poker44 benchmark chunks for a source date",
+      "notes": "Public no-auth GET /api/v1/benchmark/chunks on api.poker44.net listing the published shadow-training chunks for one release date. Documented in docs/training-benchmark.md alongside the already-tracked /api/v1/benchmark status and /api/v1/benchmark/releases surfaces. Takes a required sourceDate query parameter (plus optional limit/cursor) rather than a path parameter, so the base URL is fixed and the route is callable today through call_subnet_surface's query argument. Live-verified 2026-07-21: GET /api/v1/benchmark/chunks?sourceDate=2026-07-21&limit=1 -> HTTP 200 application/json listing chunkId/chunkHash/chunkIndex/releasedAt plus a nextCursor; the bare URL with no sourceDate returns HTTP 422, which is why probe.enabled is false. No wallet/hotkey data.",
+      "probe": {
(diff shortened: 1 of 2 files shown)


JSONbored:
Re-reviewed. The bundling block was correct — this repo's registry submissions must be exactly one `registry/subnets/<slug>.json` file, and this PR paired it with a test file. The surface itself (`sn-126-poker44-benchmark-chunks`) looked well-reasoned in the review (correct netuid/authority/review fields, sound probe-disabled rationale). Your later #7456/#7469 correctly resubmitted registry-only — see those threads for the remaining gap (still missing the chunk-by-id sibling surface).

PR #7456, opened 2026-07-21
registry(sn126): add Poker44 benchmark-chunks surface (#7134)

joaovictor91123:
Registers the additional surface #7134 flagged as newly in scope, as a **registry-only** change (this resubmits it standalone after #7447 was closed for bundling a test with the registry entry — "A registry submission must not bundle other file changes"). Part of #7134.

**New surface — `sn-126-poker44-benchmark-chunks`** (`data-artifact`, no-auth):
- `GET https://api.poker44.net/api/v1/benchmark/chunks` — lists the published shadow-training chunks for a release date (`chunkId`, `chunkHash`, `chunkIndex`, `releasedAt`, + `nextCursor`).
- Takes a required `sourceDate` **query** param (optional `limit`/`cursor`), not a path param, so the base URL is fixed and it's callable today via `call_subnet_surface`'s `query` argument. `probe.enabled:false` because the bare URL (no `sourceDate`) is HTTP 422.
- **Live-verified 2026-07-21:** bare URL → **422**; `?sourceDate=2026-07-21&limit=1` → **200** `application/json`.

Also updates the one contradicting `curation.gap_notes` entry (it previously called this route "not promotable") in the same file. The sibling `/api/v1/benchmark/chunks/{chunkId}` path-param route has no fixed callable URL and stays unregistrable.

Registry-only: `registry/subnets/poker44.json` is the sole changed file. Schema + surface + readme-catalog validators all green.

### DIFF
--- a/registry/subnets/poker44.json
+++ b/registry/subnets/poker44.json
@@ -10,7 +10,7 @@
     "gap_notes": [
       "No verified machine-readable OpenAPI/Swagger schema yet.",
       "No verified SSE/event stream yet.",
-      "The /api/v1/benchmark/chunks route requires a sourceDate/chunkId parameter and is not promotable."
+      "The /api/v1/benchmark/chunks route needs a sourceDate query parameter (now registered as sn-126-poker44-benchmark-chunks, callable via the tool's query argument); its /api/v1/benchmark/chunks/{chunkId} path-parameterized variant has no fixed URL and remains unregistrable."
     ],
     "level": "maintainer-reviewed",
     "review_state": "maintainer-reviewed",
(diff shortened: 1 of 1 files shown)


JSONbored:
Re-reviewed. The "no linked issue detected" block was correct as a matter of form — the body said "Part of #7134" rather than a recognized closing keyword (`Closes #7134`). Separately, even with that fixed, this only adds 1 of the 2 additional surfaces #7134 asks for (`benchmark-chunks`) — the chunk-by-id sibling is still missing, same gap as your later #7469 attempt. #7134 remains open for a follow-up with both the linking keyword and the remaining surface.

PR #7459, opened 2026-07-21
registry(sn126): add Poker44 benchmark-chunks surface (#7134)

joaovictor91123:
Closes #7134

Registers the additional surface #7134 flagged as newly in scope, as a **registry-only** change — a single file with no bundled test (resubmitting after #7447 was closed for bundling a test, and #7456 was closed for lacking a `Closes` link).

**New surface — `sn-126-poker44-benchmark-chunks`** (`data-artifact`, no-auth):
- `GET https://api.poker44.net/api/v1/benchmark/chunks` — lists the published shadow-training chunks for a release date (`chunkId`, `chunkHash`, `chunkIndex`, `releasedAt`, + `nextCursor`).
- Required `sourceDate` **query** param (optional `limit`/`cursor`), not a path param, so the base URL is fixed and it's callable today via `call_subnet_surface`'s `query` argument. `probe.enabled:false` because the bare URL (no `sourceDate`) is HTTP 422.
- **Live-verified 2026-07-21:** bare URL → **422**; `?sourceDate=2026-07-21&limit=1` → **200** `application/json`.

Also updates the one contradicting `curation.gap_notes` entry (previously called this route "not promotable") in the same file. The sibling `/api/v1/benchmark/chunks/{chunkId}` path-param route has no fixed callable URL and stays unregistrable.

Registry-only: `registry/subnets/poker44.json` is the sole changed file. Schema + surface + readme-catalog validators all green.

### DIFF
--- a/registry/subnets/poker44.json
+++ b/registry/subnets/poker44.json
@@ -10,7 +10,7 @@
     "gap_notes": [
       "No verified machine-readable OpenAPI/Swagger schema yet.",
       "No verified SSE/event stream yet.",
-      "The /api/v1/benchmark/chunks route requires a sourceDate/chunkId parameter and is not promotable."
+      "The /api/v1/benchmark/chunks route needs a sourceDate query parameter (now registered as sn-126-poker44-benchmark-chunks, callable via the tool's query argument); its /api/v1/benchmark/chunks/{chunkId} path-parameterized variant has no fixed URL and remains unregistrable."
     ],
     "level": "maintainer-reviewed",
     "review_state": "maintainer-reviewed",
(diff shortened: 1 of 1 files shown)


JSONbored:
Re-reviewed. This close was correct on its own terms — the diff edited `curation.gap_notes[2]` on the pre-existing SN126 manifest, and this repo's append-only rule for existing manifests puts top-level `curation` edits out of scope for a community surface-addition PR, even though the edit was a direct, reasonable consequence of the new surface being added. The right move is a follow-up that adds the surface without touching `curation` — happy to see that resubmitted.

Issue #7016, opened 2026-07-20
MCP execute Phase 3: credential-passthrough for authenticated surfaces


JSONbored:
Part of #7013. Depends on Phase 1 (#7014); benefits from Phase 2 (#7015) for the auth-required surfaces that also have a schema.

## Scope

Extend `call_subnet_surface` to accept an optional `credential` argument for the 7 surfaces (today) with `auth_required: true`, each of which already carries a structured descriptor from #746: `{scheme, location, header name, value_format}`.

- Insert the supplied credential into the request per the surface's own `auth.location`/`auth.name`/`value_format` (e.g. `Authorization: Bearer <token>`) -- generic across `bearer`/`api-key` schemes.
- **Per-call only, never stored or logged.** The credential lives for the duration of one tool call and nowhere else -- metagraphed does not hold third-party API keys on anyone's behalf. This is a deliberate, non-negotiable design constraint, not a v1 shortcut to revisit later.
- Surfaces with `auth.scheme: "custom"` (2 of the 7 today -- gittensor.json, green-compute.json) have no generically-describable credential mechanism; exclude them from this tool until/unless their auth gets documented well enough to be generic (registry-data problem, not a tool-code problem).
- This tool does **not** obtain credentials on a caller's behalf (no payment, no registration flow) -- the caller must already hold a working credential.

## Explicitly out of scope

- Any credential storage/reuse across calls.
- The 2 `custom`-scheme surfaces until their auth mechanism is documented.

JSONbored:
All three sub-issues (#7686, #7687, #7688) are merged. Phase 3 is complete: call_subnet_surface now supports credential passthrough for auth_required surfaces with a generically-describable auth mechanism (bearer/api-key, header/query/cookie), with leak-prevention verified end-to-end (response redaction, telemetry, Sentry).

PR #7404, opened 2026-07-21
feat(registry): add SN34 BitMind auth-gated detection API surfaces

claytonlin1110:
## Summary
- Reopened #7049 asked for the 8 auth-gated BitMind Detection API operations from `openapi.json` that #7377 could not cover (verify-only).
- Appends those surfaces to `registry/subnets/bitmind.json` only: `auth_required: true`, `probe.enabled: false`, bearer placeholder auth, `authority: community`.

## Live verification (2026-07-21)
- GET https://api.bitmind.ai/health → 200 `{"status":"healthy",...}` (already registered)
- GET https://api.bitmind.ai/openapi.json → 200 OpenAPI 3.1.0 "BitMind Detection API"; all 8 paths present
- Anonymous POST `/v1/detect`, `/detect-image`, `/detect-text` with `{}` → HTTP 401 `{"detail":"API key is required"}`
- Remaining 5 share the same auth wrapper / schema description (no credential passthrough in Phase 1)

## Surfaces added
- `sn-34-bitmind-detect-v1` — POST /v1/detect
- `sn-34-bitmind-detect-image` — POST /detect-image
- `sn-34-bitmind-detect-video` — POST /detect-video
- `sn-34-bitmind-detect-text` — POST /detect-text
- `sn-34-bitmind-detect-document` — POST /detect-document
- `sn-34-bitmind-liveness-verify` — POST /liveness/verify
- `sn-34-bitmind-preprocess-video` — POST /preprocess-video
- `sn-34-bitmind-get-video-upload-url` — POST /get-video-upload-url

Closes #7049

## Test plan
- [x] `npm run validate:surface -- registry/subnets/bitmind.json`
- [x] `npm run validate:private-boundary`
- [x] `npm run scan:public-safety` (exit 0; no bitmind findings)
- [ ] CI Validate green

PR #7497, opened 2026-07-21
fix(registry): SN34 BitMind Detection API parity + correct auth_required (#7049)

### LINKED ISSUE #7049: MCP execute: verify + wire SN34 (BitMind) once Phase 1 ships

## Scope
Confirm SN34 (BitMind)'s callable surface(s) below actually work end-to-end -- an actual request against the live URL, not an assumption. `call_subnet_surface` is a safety-checked passthrough of the exact same request (same URL, same method), not a black box -- calling the URL directly yourself is equally valid verification for a no-auth GET surface.

## How to verify

## Surface(s) to verify (2)
- `sn-34-bitmind-openapi` (openapi, GET, no-auth, has captured schema): https://api.bitmind.ai/openapi.json
- `sn-34-bitmind-subnet-api` (subnet-api, GET, no-auth, no schema (single fixed endpoint)): https://api.bitmind.ai/health

## Deliverable

## Additional surface(s) found during review (now in scope)

Fetched SN34's own OpenAPI schema (https://api.bitmind.ai/openapi.json, already registered as `sn-34-bitmind-openapi`) -- 10 total operations, 8 not currently registered (only `/health` and the schema itself are). The schema doesn't declare an OpenAPI `security` requirement on any operation, but its `info.description` explicitly states "All endpoints require authentication via `Authorization: Bearer <token>` header with Supabase JWT or API key" for everything except health -- confirmed live: `POST /v1/detect` with an empty JSON body returns `401 {"detail":"API key is required"}`. The registry's job is to catalog everything the subnet exposes, not just what the anonymous Phase-1 tool can call today, so these belong in the registry with `auth_required:true`, `probe.enabled:false`:

- **Auto-detecting front door**: `POST /v1/detect` -- infers image/video/text/document type and delegates to the matching detector.
- **Image detection**: `POST /detect-image` -- detect AI-generated images.
- **Video detection**: `POST /detect-video` -- detect AI-generated videos.
- **Text detection**: `POST /detect-text` -- detect AI-generated text.
- **Document detection**: `POST /detect-document` -- detect forgery/tampering in a PDF.
- **Liveness verification**: `POST /liveness/verify` -- verify a face capture is a live human (KYC use case).
- **Video preprocessing**: `POST /preprocess-video` -- preprocess video for browser compatibility (no detection).
- **Video upload URL**: `POST /get-video-upload-url` -- get a presigned URL for large video uploads.

All 8 confirmed auth-gated per the schema description; `/v1/detect` spot-verified live (`401`, API key required). Not each of the remaining 7 was individually curled (same auth wrapper, same host, documented identically) but the pattern is consistent and the schema is the authoritative source for request/response shape.

## Explicitly out of scope

joaovictor91123:
## Summary

Closes #7049.

Brings SN34 (BitMind) up to **full subnet API parity** — the completeness bar in `.claude/skills/contributor-pipeline-gardening/SKILL.md`'s "MCP execute verify + wire SN*" special-sweep section — **and corrects what the schema implies about auth**.

The already-registered `sn-34-bitmind-openapi` surface points at the BitMind Detection API's OpenAPI spec (10 paths). Only `/health` had a registry entry; the other **9 paths had none**.

## The correction (the interesting part)

The spec declares **no `securitySchemes` and no per-operation `security`** — which reads as "these are open endpoints." **They are not.** Live-verified 2026-07-21, every detection/liveness/video POST rejects an unauthenticated request:

| request | result |
|---|---|
| `POST /v1/detect` | **401** `{"detail":"API key is required"}` |
| `POST /detect-image` | **401** `{"detail":"API key is required"}` |
| `POST /detect-video` | **401** `{"detail":"API key is required"}` |
| `POST /detect-text` | **401** `{"detail":"API key is required"}` |
| `POST /detect-document` | **401** `{"detail":"API key is required"}` |
| `POST /liveness/verify` | **401** `{"detail":"API key is required"}` |
| `POST /preprocess-video` | **401** `{"detail":"API key is required"}` |
| `POST /get-video-upload-url` | **401** `{"detail":"API key is required"}` |

So those 8 are registered `auth_required: true` (correcting the schema's implication), `probe.enabled: false` — they're state-changing POSTs and not probe-safe — each with an `auth{}` object recording the `api-key` scheme. The spec doesn't document *which* header carries the credential, so **only the scheme is asserted**, not a header name I couldn't verify.

`GET /` genuinely needs no credential — live-verified **200** `{"service":"BitMind Detection API","version":"1.0.0","region":...}` — so it's registered `auth_required: false` with its **probe enabled**.

## Checklist

- [x] Changes exactly one `registry/subnets/bitmind.json` file (+230/−0, clean append)
- [x] `node scripts/validate-schemas.mjs` passed
- [x] `node scripts/validate-surface.mjs` passed, **no `auth_required` advisories**
- [x] `npx vitest run tests/validate-surface-auth-object.test.mjs` (6 passed)
- [x] `npx vitest run tests/sn34-call-subnet-surface-verify.test.mjs` (6 passed — unaffected)
- [x] `npm run scan:public-safety` passed (no bitmind findings)
- [x] `provider` uses the already-registered `bitmind` slug
- [x] Rebased on latest `main`

## Test plan

- [ ] `GET /` is a good no-auth MCP smoke target; the 8 gated ops become callable once Phase 3 (#7016) lands

### DIFF
--- a/registry/subnets/bitmind.json
+++ b/registry/subnets/bitmind.json
@@ -218,6 +218,236 @@
         "https://bitmind.ai/"
       ],
       "url": "https://huggingface.co/datasets/bitmind/bm-real"
+    },
+    {
+      "auth": {
+        "scheme": "api-key",
+        "scopes_note": "The subnet's own OpenAPI spec declares no securitySchemes and no per-operation security, but every one of these endpoints empirically rejects an unauthenticated request with HTTP 401 \"API key is required\" (live-checked 2026-07-21). The header/parameter the credential is passed in is not documented in the spec, so only the scheme is recorded here."
+      },
+      "auth_required": true,
+      "authority": "official",
+      "id": "sn-34-bitmind-detect",
+      "kind": "subnet-api",
+      "name": "BitMind unified detection",
+      "notes": "Detect operation on the BitMind Detection API (POST /v1/detect), declared in the subnet's own OpenAPI spec at https://api.bitmind.ai/openapi.json. The spec declares no security for it, but live-verified 2026-07-21 an unauthenticated POST returns HTTP 401 {\"detail\":\"API key is required\"} -- so it is registered auth_required:true (correcting what the schema implies) with probe.enabled:false, since it is a state-changing POST and not probe-safe (Phase 3).",
+      "probe": {
+        "enabled": false,
+        "expect": "json",
+        "method": "GET",
+        "timeout_ms": 10000
+      },
+      "provider": "bitmind",
+      "public_safe": true,
+      "review": {
+        "state": "community-submitted",
+        "submitted_by": "joaovictor91123"
+      },
+      "source_urls": ["https://api.bitmind.ai/openapi.json"],
+      "url": "https://api.bitmind.ai/v1/detect"
+    },
+    {
+      "auth": {
+        "scheme": "api-key",
+        "scopes_note": "The subnet's own OpenAPI spec declares no securitySchemes and no per-operation security, but every one of these endpoints empirically rejects an unauthenticated request with HTTP 401 \"API key is required\" (live-checked 2026-07-21). The header/parameter the credential is passed in is not documented in the spec, so only the scheme is recorded here."
+      },
+      "auth_required": true,
+      "authority": "official",
+      "id": "sn-34-bitmind-detect-document",
+      "kind": "subnet-api",
+      "name": "BitMind document detection",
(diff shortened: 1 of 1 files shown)


JSONbored:
Re-reviewed, including the auth-scheme question this PR raises. The 8 shared surfaces are the same ones #7404 already added — the difference is #7404 registered them with `scheme:"bearer"`, this PR with `scheme:"api-key"`. Going back to #7049's own original text: the API's OpenAPI `info.description` explicitly states the mechanism is `Authorization: Bearer <token>` where the token "can be a Supabase JWT or API key" — that's describing the wire scheme as Bearer, with the token's *content* (not the scheme) being an API key. So `bearer` looks like the more precise match to the documented mechanism; `api-key` describes what's inside the header rather than how it's sent. Keeping #7404 as the surviving PR for those 8. The one thing here that #7404 doesn't have is the new no-auth `GET /` surface (service info, live-verified 200) — small, genuinely additive, worth a quick standalone follow-up once #7404 lands rather than re-litigating the shared 8 here.

PR #7548, opened 2026-07-21
registry(sn13): add Data Universe Validator API path surfaces (#7029)

### LINKED ISSUE #7029: MCP execute: verify + wire SN13 (Data Universe) once Phase 1 ships

## Scope
Confirm SN13 (Data Universe)'s callable surface(s) below actually work end-to-end -- an actual request against the live URL, not an assumption. `call_subnet_surface` is a safety-checked passthrough of the exact same request (same URL, same method), not a black box -- calling the URL directly yourself is equally valid verification for a no-auth GET surface.

## How to verify

## Surface(s) to verify (2)
- `sn-13-macrocosmos-subnet-api` (subnet-api, ?, AUTH REQUIRED (bearer) -- Phase 3 (#7016) territory, not Phase 1/2, no schema (single fixed endpoint)): https://constellation.api.cloud.macrocosmos.ai/sn13.v1.Sn13Service/OnDemandData
- `sn-13-data-universe-openapi` (openapi, GET, no-auth, has captured schema): https://sn13.api.macrocosmos.ai/openapi.json

## Deliverable

## Additional surface(s) found during review (now in scope)

`registry/subnets/data-universe.json` already tracks the SN13 validator API's OpenAPI schema itself (`sn-13-data-universe-openapi`, https://sn13.api.macrocosmos.ai/openapi.json) but none of the 10 individual paths it documents -- the only other API surface tracked is a *different* host (`constellation.api.cloud.macrocosmos.ai`, the auth-gated Constellation on-demand-data endpoint). Every one of the 10 sn13.api.macrocosmos.ai paths declares `security: true` in the schema:

- `GET /api/v1/query_bucket/{source}` (path-param)
- `GET /api/v1/get_top_labels_by_source/{source}` (path-param)
- `GET /api/v1/health` -- confirmed live: returns HTTP 500 "Internal Server Error" (plain text) unauthenticated, not a clean 401 -- worth noting as an anomaly if registered, since a probe would need to expect the error rather than a healthy response.
- `GET /api/v1/ages`
- `GET /api/v1/get_desirabilities`
- `GET /api/v1/get_bytes_by_label`
- `GET /api/v1/monitoring/system-status`
- `GET /api/v1/keys`, `POST /api/v1/keys`, `POST /api/v1/keys/{key}/deactivate` (key management)
- `POST /api/v1/set_desirabilities`

Register all with `auth_required: true` (bearer, matching the existing Macrocosmos SDK `Sn13Client`/`GravityClient` credential pattern already used for the Constellation surface); the three with `{source}`/`{key}` path params additionally get the literal template. Not individually spot-verified beyond `/api/v1/health` (no API key on hand), but the schema's declared security is consistent across all 10.

## Explicitly out of scope
- Finding/adding surfaces BEYOND what's already listed above (including any "Additional surface(s) found during review" section on this issue) -- that's normal registry enrichment, a separate process. Anything already listed above, including in that section, IS in scope for this issue (clarified 2026-07-21: this line previously blanket-excluded new surfaces, which directly contradicted a mandatory "now in scope" section added during the same-day audit pass).

kai392:
﻿## Summary
- Full #7029 Additional scope: all Data Universe Validator API OpenAPI paths on `sn13.api.macrocosmos.ai` (unique URLs; GET+POST `/api/v1/keys` registered once).
- `auth_required:true` with `auth: { scheme: "api-key", location: "header", name: "X-API-Key", ... }` per schema `APIKeyHeader`.
- Path-param templates percent-encoded; `probe.enabled:false`.
- Registry-only, append-only, single file.

Closes #7029

## Test plan
- [x] prettier + validate:surface
- [x] vitest validate-surface-auth-object
- [ ] CI green

### DIFF
--- a/registry/subnets/data-universe.json
+++ b/registry/subnets/data-universe.json
@@ -354,6 +354,316 @@
         "https://github.com/macrocosm-os/data-universe"
       ],
       "url": "https://raw.githubusercontent.com/macrocosm-os/data-universe/main/dynamic_desirability/default.json"
+    },
+    {
+      "auth": {
+        "scheme": "api-key",
+        "location": "header",
+        "name": "X-API-Key",
+        "scopes_note": "Declared as APIKeyHeader (X-API-Key) on every operation in sn13.api.macrocosmos.ai/openapi.json; Macrocosmos Sn13Client/GravityClient credential pattern."
+      },
+      "auth_required": true,
+      "authority": "community",
+      "id": "sn-13-data-universe-get-api-v1-ages",
+      "kind": "subnet-api",
+      "name": "Data Universe Get Age Sizes",
+      "notes": "Auth-gated GET /api/v1/ages on sn13.api.macrocosmos.ai from the Data Universe Validator API OpenAPI schema (sn-13-data-universe-openapi). All schema operations declare APIKeyHeader (X-API-Key). probe.enabled:false (auth-gated). Live-verified 2026-07-21: anonymous GET /api/v1/health -> HTTP 500 (auth/error anomaly noted in #7029).",
+      "probe": {
+        "enabled": false,
+        "expect": "json",
+        "method": "GET",
+        "timeout_ms": 10000
+      },
+      "provider": "macrocosmos",
+      "public_safe": true,
+      "review": {
+        "state": "community-submitted",
+        "submitted_by": "kai392"
+      },
+      "source_urls": [
+        "https://sn13.api.macrocosmos.ai/openapi.json",
+        "https://github.com/macrocosm-os/data-universe"
+      ],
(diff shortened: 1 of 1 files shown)


JSONbored (approved):
Reviewed and verified. 10 new surfaces confirmed, all 10 correctly carrying `auth{scheme:"api-key",...}` objects matching the schema's uniform APIKeyHeader declaration. Good transparency on `/api/v1/health` — it returns a 500 rather than a clean 401 unauthenticated, which is a genuine anomaly, and the contributor registered it consistent with the schema's declared auth (applying to all 10 ops uniformly) while explicitly flagging the discrepancy in the notes rather than either hiding it or inventing a different treatment. Single file, 0 deletions, no secret/credential text.

Full scope, correctly done. Merging once CI finishes (pending as of this review).

PR #7556, opened 2026-07-21
fix(registry): SN15 ORO API parity (63 missing surfaces) (#7031)

joaovictor91123:
## Summary

Closes #7031.

Brings SN15 (ORO) up to **full subnet API parity** — the completeness bar in `.claude/skills/contributor-pipeline-gardening/SKILL.md`'s "MCP execute verify + wire SN*" special-sweep section. Same approach as SN35 OxMarkets (#7553), SN56 Gradients (#7532) and SN37 Aurelius (#7466).

The ORO agents API's OpenAPI spec (**77 paths**) had 14 of its paths registered. This adds the other **63**.

## The correction

The spec declares **no `securitySchemes` at all** — which reads as "these are open endpoints." They aren't. The API applies a **global signed-request auth middleware**. Live-verified 2026-07-21, sampling one route from each family:

| request | result |
|---|---|
| `GET /v1/miner/inference-auth` | **401** — required signed-request header set missing |
| `GET /v1/miner/chutes-auth/status` | **401** — same |
| `GET /v1/miner/agents` | **401** — same |
| `GET /v1/validator/weight-salt` | **401** — same |
| `GET /v1/admin/audit-events` | **401** — same |

The 401 names a required header set (operator-account identifier, timestamp, nonce, request signature). All 63 are therefore registered `auth_required: true`, `probe.enabled: false` (Phase 3), each with an `auth{}` object recording the **header** location. The exact header names and signing scheme aren't documented in the spec, so I assert only the location (`scheme: custom`) rather than guessing.

## On the inference, stated plainly

I probed 5 routes, not 63. Because those 5 span the **miner, validator and admin** families and all return the identical middleware rejection, I've registered the whole set as gated — and **each entry says in its own `notes`** whether it was directly probed or is covered by the middleware verified on sampled routes. The inference is disclosed per-surface rather than presented as if every route were individually confirmed.

Path-parameterized routes use percent-encoded `{param}` templates, matching SN37 Aurelius's encoding.

Identifiers, names and summaries use neutral account terminology (`npm run scan:public-safety` reports **no oro findings**).

## Checklist

- [x] Changes exactly one `registry/subnets/oro.json` file (+1701/−0, clean append)
- [x] Every added surface: `authority: community`, `review.state: community-submitted`, `public_safe: true`; no health/`verification` set by hand
- [x] `node scripts/validate-schemas.mjs` passed
[...]

PR #7553, opened 2026-07-21
fix(registry): SN35 OxMarkets API parity (53 missing surfaces) (#7050)

joaovictor91123:
## Summary

Closes #7050.

Brings SN35 (OxMarkets) up to **full subnet API parity** — the completeness bar in `.claude/skills/contributor-pipeline-gardening/SKILL.md`'s "MCP execute verify + wire SN*" special-sweep section. Same approach as SN56 Gradients (#7532, 48 surfaces), SN37 Aurelius (#7466) and SN50 Synth (#7506).

The OxMarkets/Cartha API's OpenAPI spec (**58 paths**) had 5 of its paths registered. This adds the other **53**.

## Auth: none declared, and none asserted

The spec declares **no `securitySchemes` anywhere**, and nothing observed indicates a credential gate — so **no surface here is marked `auth_required`**. What the fixed reads actually require is *parameters*. Live-checked 2026-07-21:

| request | result |
|---|---|
| `GET /v1/miner/status` | **422** — required operator-account query param missing |
| `GET /subnet/check-registration` | **422** — required operator-account query param missing |
| `GET /v1/miner/status-by-evm` | **422** — required EVM-address query param missing |
| `GET /parent-lock/status` | **422** — required transaction-hash query param missing |
| `GET /v1/verified-miners` | **403** — *"Validator whitelist is enabled. Please provide validator_hotkey parameter."* |

The 403 is application-level access control rather than a declared security scheme, so it's recorded as an observation with the probe disabled — not converted into an `auth_required: true` claim the spec doesn't support.

The query-param routes are registered with their fixed base URL and are callable today via `call_subnet_surface`'s `query` argument.

## On coverage honesty

With 53 routes I sampled the fixed reads rather than hammering the host. Entries beyond the five probed above say **in their own `notes`** that they were not individually verified, and are registered as the spec describes them. Every affirmative claim traces to a response I actually observed.

The remainder are path-parameterized routes (percent-encoded `{param}` templates, matching SN37 Aurelius's encoding) or state-changing operations — all probe-disabled.

Summaries and names use neutral account terminology, so no Bittensor key wording lands in the manifest prose (`npm run scan:public-safety` reports **no oxmarkets findings**).

## Checklist

- [x] Changes exactly one `registry/subnets/oxmarkets.json` file (+1166/−0, clean append)
- [x] Every added surface: `authority: community`, `review.state: community-submitted`, `public_safe: true`; no health/`verification` set by hand
[...]

PR #7546, opened 2026-07-21
fix(registry): SN19 BlockMachine API parity (20 missing surfaces) (#7035)

joaovictor91123:
## Summary

Closes #7035.

Brings SN19 (blockmachine) up to **full subnet API parity** — the completeness bar in `.claude/skills/contributor-pipeline-gardening/SKILL.md`'s "MCP execute verify + wire SN*" special-sweep section. Same approach as SN59 Babelbit (#7540), SN62 Ridges (#7537), SN100 Plaτform (#7539) and SN56 Gradients (#7532).

The BlockMachine API's OpenAPI spec (**27 paths**) had 7 of its paths registered. This adds the other **20**.

## How each was classified

**12 auth-gated** — declare `HTTPBearer` in the spec → `auth_required: true`, `probe.enabled: false` (Phase 3), each with an `auth{}` object.

Live-verified 2026-07-21 the gate is real, not schema-only:

| request | result |
|---|---|
| `GET /nodes` | **401** `{"detail":"Not authenticated"}` |
| `GET /pricing/rules` | **401** `{"detail":"Not authenticated"}` |
| `GET /snapshots/url` | **401** `{"detail":"Not authenticated"}` |
| `GET /bans` | **401** `{"detail":"Not authenticated"}` |

**1 query-param route:**

| request | result | how it's registered |
|---|---|---|
| `GET /base-incentive/ledger` | **422** naming query param `epoch_id` as required | fixed base URL, `probe.enabled: false`; callable today via `call_subnet_surface`'s `query` argument |

**The remaining 7** are path-parameterized routes (percent-encoded `{param}` templates, matching SN37 Aurelius's encoding) or state-changing operations — all probe-disabled, with no auth asserted where neither the spec nor an observed response supports one.

Operation summaries and surface names use neutral account terminology, so no Bittensor key wording lands in the manifest prose (`npm run scan:public-safety` reports **no blockmachine findings**).

## Checklist

- [x] Changes exactly one `registry/subnets/blockmachine.json` file (clean append)
- [x] Every added surface: `authority: community`, `review.state: community-submitted`, `public_safe: true`; no health/`verification` set by hand
- [x] `node scripts/validate-schemas.mjs` passed
- [x] `node scripts/validate-surface.mjs` passed (1688 surfaces / 129 files), **no `auth_required` advisories**
- [x] `npx vitest run tests/validate-surface-auth-object.test.mjs` (6 passed)
[...]

PR #7466, opened 2026-07-21
fix(registry): SN37 Aurelius full API parity (53 missing surfaces)

oktofeesh1:
## Summary

Closes #7052. PR #7324 (already merged) closed this issue with a test-only PR — it verified the 6 originally-listed surfaces (still accurate; `sn-37-aurelius-subnet-api`'s health check really does return `{"status":"ok"}`, no correction needed there) but added zero registry changes, so none of the 56 additional operations the issue's own OpenAPI sweep found ever got registered. Same "test-only PR closes the issue without delivering the registry entries" anti-pattern documented in `.claude/skills/contributor-pipeline-gardening/SKILL.md`'s MCP-execute special-sweep section — I found and fixed the same gap for SN36 Eirel in #7465, and cross-checked: 51 of these ~120 issues have this exact pattern (merged test-only PR, issue still open, registry never updated).

## What this adds

Diffed the live OpenAPI spec (`https://new-collector-api-production.up.railway.app/openapi.json`, 61 operations) against the registry (5 already covered) → **53 unique missing paths** (61 ops − 5 registered − 3 collapsed from same-path multi-method pairs: `/config` GET+PUT, `/submissions` POST+GET, `/classifier/model` GET+POST — registered once per path per the registry's `netuid|kind|url` dedup key, other method noted in `notes`).

**3 public, no-auth** (schema declares no `security`):
- `POST /auth/validator/challenge` / `POST /auth/validator/verify` — live-verified: an empty body 422s with a clear field-required error (`hotkey`), and a too-short hotkey 422s with a length-validation error — confirms these are real, unauthenticated, input-validated endpoints. `probe.enabled:false`: they're POST-only, and a GET probe against a POST-only route would just 405, misreporting the surface as down.
- `GET /work-token/balance/{hotkey}` — live-verified with a real value: `200 {"hotkey":"testhotkey","balance":0.0,"has_balance":false}`. Path-param, `probe.enabled:false` (Phase 2).

[...]

PR #7563, opened 2026-07-21
fix(registry): SN85 Vidaio API parity (8 missing surfaces) (#7098)

joaovictor91123:
## Summary

Closes #7098.

Brings SN85 (Vidaio) up to **full subnet API parity** — the completeness bar in `.claude/skills/contributor-pipeline-gardening/SKILL.md`'s "MCP execute verify + wire SN*" special-sweep section. Same approach as SN15 ORO (#7556), SN35 OxMarkets (#7553) and SN37 Aurelius (#7466).

The Vidaio API's OpenAPI spec (**10 paths**) had 2 of its paths registered. This adds the other **8**.

## Auth — declared in the spec and confirmed live

All 8 declare the `APIKeyHeader` security scheme, so each is `auth_required: true`, `probe.enabled: false` (Phase 3), with an `auth{}` object naming the header **`X-API-Key`** taken from the spec's own `securitySchemes` — not inferred.

Live-verified 2026-07-21:

| request | result |
|---|---|
| `GET /rate_limits` | **401** `{"detail":"Invalid or missing API key"}` |
| `GET /available_credits` | **401** `{"detail":"Invalid or missing API key"}` |

The other six are the task-oriented routes — state-changing POSTs (`/compression`, `/upscaling`) and their path-parameterized result reads (`/{task_id}`, `/{task_id}/result`) — so they use percent-encoded `{param}` templates and keep probes disabled.

Unlike the larger sweeps in this series, no inference was needed here: every surface's auth requirement comes from an explicit per-operation `security` declaration in the spec, and the two directly probeable routes confirm the gate is real rather than schema-only.

## Checklist

- [x] Changes exactly one `registry/subnets/vidaio.json` file (+224/−0, clean append)
- [x] Every added surface: `authority: community`, `review.state: community-submitted`, `public_safe: true`; no health/`verification` set by hand
- [x] `node scripts/validate-schemas.mjs` passed
- [x] `node scripts/validate-surface.mjs` passed (1904 surfaces / 129 files), **no `auth_required` advisories**
- [x] `npx vitest run tests/validate-surface-auth-object.test.mjs` (6 passed)
- [x] `npx vitest run tests/sn85-call-subnet-surface-verify.test.mjs` (9 passed — unaffected)
- [x] `npm run scan:public-safety` — no vidaio findings
- [x] Surface ids asserted unique after generation
- [x] `provider` uses the already-registered `vidaio` slug
- [x] No other open PR references #7098
- [x] Rebased on latest `main`

> Heads-up: `scan:public-safety` currently exits 1 on a clean `main` due to a pre-existing `sensitive hotkey wording` finding in `registry/subnets/cacheon.json` (plus its generated `dist/**` copies) — unrelated to this PR, which touches only `vidaio.json`. Details in #7546.

## Test plan

- [ ] All 8 become callable once Phase 3 credential passthrough (#7016) lands

### DIFF
--- a/registry/subnets/vidaio.json
+++ b/registry/subnets/vidaio.json
@@ -159,6 +159,230 @@
         "https://vidaio.io/"
       ],
       "url": "https://api.vidaio.io/ready"
+    },
+    {
+      "auth": {
+        "scheme": "api-key",
+        "location": "header",
+        "name": "X-API-Key",
+        "scopes_note": "Derived from the subnet's own OpenAPI securitySchemes (APIKeyHeader), declared per-operation."
+      },
+      "auth_required": true,
+      "authority": "community",
+      "id": "sn-85-vidaio-available-credits",
+      "kind": "subnet-api",
+      "name": "Vidaio available credits",
+      "notes": "Get available credits operation on the Vidaio API (GET /available_credits), declared in the subnet's own OpenAPI spec at https://api.vidaio.io/openapi.json. The spec declares the APIKeyHeader security scheme on it, so it is registered auth_required:true with probe.enabled:false (Phase 3). Live-verified 2026-07-21: an unauthenticated GET returns HTTP 401 {\"detail\":\"Invalid or missing API key\"}, confirming the gate.",
+      "probe": {
+        "enabled": false,
+        "expect": "json",
+        "method": "GET",
+        "timeout_ms": 10000
+      },
+      "provider": "vidaio",
+      "public_safe": true,
+      "review": {
+        "state": "community-submitted",
+        "submitted_by": "joaovictor91123"
+      },
+      "source_urls": ["https://api.vidaio.io/openapi.json"],
+      "url": "https://api.vidaio.io/available_credits"
+    },
+    {
+      "auth": {
+        "scheme": "api-key",
+        "location": "header",
+        "name": "X-API-Key",
+        "scopes_note": "Derived from the subnet's own OpenAPI securitySchemes (APIKeyHeader), declared per-operation."
+      },
+      "auth_required": true,
+      "authority": "community",
+      "id": "sn-85-vidaio-compression",
+      "kind": "subnet-api",
+      "name": "Vidaio compression",
+      "notes": "Start compression operation on the Vidaio API (POST /compression), declared in the subnet's own OpenAPI spec at https://api.vidaio.io/openapi.json. The spec declares the APIKeyHeader security scheme on it, so it is registered auth_required:true with probe.enabled:false (Phase 3).",
+      "probe": {
+        "enabled": false,
+        "expect": "json",
+        "method": "GET",
+        "timeout_ms": 10000
+      },
+      "provider": "vidaio",
+      "public_safe": true,
+      "review": {
+        "state": "community-submitted",
+        "submitted_by": "joaovictor91123"
+      },
+      "source_urls": ["https://api.vidaio.io/openapi.json"],
+      "url": "https://api.vidaio.io/compression"
+    },
+    {
+      "auth": {
+        "scheme": "api-key",
+        "location": "header",
+        "name": "X-API-Key",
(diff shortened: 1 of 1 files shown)


joaovictor91123:
Flagging a systemic issue this PR surfaced — **not caused by this PR's 8 surfaces, but I'm a major contributor to the underlying cause.**

The red `checks` job here is the **artifact size budget** step, not anything in `vidaio.json`:

```
- surfaces.json:      3,358,050 >= 1,500,000
- search.json:        1,486,763 >=   750,000
- search-index.json:  1,011,138 >= 1,000,000
- subnets/15.json:      485,993 >=   250,000
- subnets/36.json:      679,828 >=   250,000
- subnets/56.json:      321,097 >=   250,000
- subnets/37.json, 35, 45, 46, 22, 114 … also over
```

These are generated projections, so they've grown in step with the full-API-parity sweep across the `verify + wire SN*` family. Checking authorship of the over-budget manifests on current `main`:

| manifest | surfaces | mine |
|---|---|---|
| `oro.json` (SN15) | 83 | **63** (#7556) |
| `oxmarkets.json` (SN35) | 65 | **53** (#7553) |
| `gradients.json` (SN56) | 63 | **48** (#7532) |
| `eirel.json` (SN36) | 127 | 0 |
| `aurelius.json` (SN37) | 63 | 1 |
| `soma.json`, `desearch.json`, `alpharidge-ai.json`, `zipcode.json` | 43–54 | 0 |

So this is a collective consequence of the parity campaign, and three of the largest contributors to it are mine. Earlier PRs passed this check; it started failing once those merged.

I don't think a contributor PR should unilaterally raise the budgets in `scripts/validate-artifact-budgets.mjs` — that's a deliberate maintainer call about payload size (and `surfaces.json` at 3.3 MB has real client-side cost). Options as I see them: raise the budgets to match the new registry size, shard the large aggregate artifacts, or scope per-subnet artifacts to probe-enabled surfaces so the auth-gated Phase 3 entries don't inflate them.

**I've paused opening further large parity PRs until this is resolved** — continuing would keep pushing these artifacts further over budget. Happy to help with whichever direction you prefer if you point me at an issue to link.

joaovictor91123:
**Correcting my comment above — I overstated this, sorry.**

I conflated the check's *warnings* section with its *failures* section. Re-reading the job log, only **one** artifact actually fails the budget:

```
Artifact size budget failures:
- search-index.json: 1011138 bytes >= 1000000
```

Everything else I listed (`surfaces.json`, `subnets/15.json`, `subnets/56.json`, …) is in the **warnings** block, which is non-blocking and doesn't fail CI. My apologies for the noise.

So the actual situation is much narrower: `search-index.json` is **1.1% over** a limit it inherits from `DEFAULT_BUDGET` (`budget("*", 250_000, 1_000_000)` in `scripts/artifact-budgets.mjs`) — it has no entry of its own in `ARTIFACT_SIZE_BUDGETS`, unlike its sibling `search.json` (750 KB warn / 2 MB fail). It looks like an artifact that was never given a considered budget and has now crossed the generic default as the registry grew.

The narrow fix would be a one-line addition to `ARTIFACT_SIZE_BUDGETS`, e.g.:

```js
budget("search-index.json", 1_000_000, 2_500_000),
```

sized in line with `search.json`. That's a `scripts/` change rather than a registry change, so it needs its own linked issue — happy to open one and submit it if you'd like, or leave it to you if the intended answer is to shrink the index instead of re-budgeting it.

My substantive point stands in weakened form: the parity sweeps are what pushed this artifact past the default, and several of the largest are mine (#7556 ORO +63, #7553 OxMarkets +53, #7532 Gradients +48). But it is one artifact 1.1% over a default budget, not a broad breach.

JSONbored:
Reopening — closed on \"CI is failing (checks)\", which was the shared search-index.json artifact-budget failure you flagged and self-diagnosed here (not anything in vidaio.json). That's now fixed on main via #7580 (merged). This PR's own content (8 surfaces matching #7098's scope) was never the problem — a fresh CI run should go green now.

JSONbored (approved):

Issue #7675, opened 2026-07-22
MCP execute Phase 2c: POST/PUT request-body support + finalize Phase 2


JSONbored:
Part of #7015 (MCP execute Phase 2). Step 3 of 3 -- the PR that completes Phase 2. Depends on #7674 (GET/HEAD schema-validated execution) and #7673 (the path/method validator).

## Scope

Extend the GET/HEAD path/method execution from #7674 to also accept `POST` and `PUT`, with an optional request `body`, when the matched operation (from #7673's `matchSchemaOperation`) declares a `requestBody` in the schema.

- Extend the `method` enum in `call_subnet_surface`'s inputSchema to include `POST`, `PUT` alongside the existing `GET`, `HEAD`.
- Add an optional `body` param (accept both a JSON-serializable object and a raw string, matching what the schema's declared request media type calls for).
- Validate before sending:
  - If the matched operation declares no `requestBody` at all but a `body` arg was supplied, reject (`invalid_params`) -- never silently drop it.
  - If the matched operation's `requestBody.content` declares one or more media types, the constructed request's `content-type` must be one of them (default to `application/json` when the schema only declares that type and the caller didn't override it).
  - Enforce a request-body size ceiling using the same bound Phase 1 already applies to response bodies -- reuse `MAX_RESPONSE_BYTES` (`src/call-subnet-surface.mjs`, currently 262,144 bytes) rather than inventing a second constant, per #7015's own scope text ("same size/content-type bounds as Phase 1's response handling"). Reject oversized bodies before sending, don't truncate a request body.
- Wire the validated method + body + `content-type` header into the actual outbound fetch (`safetyCheckedFetch` / the `fetchImpl` call) -- GET/HEAD requests continue to send no body, exactly as today.
- If the schema declares required fields in `requestBody`'s JSON schema, this issue does NOT need full JSON-Schema validation of the body's contents -- content-type + presence/absence enforcement (per the bullets above) is the bar. Deep body-shape validation against arbitrary JSON Schema is out of scope (flag it as a possible future issue in the PR description if it seems worth doing, but don't build it here).

## Finalize Phase 2 (this is the last piece)

[...]

Issue #7673, opened 2026-07-22
MCP execute Phase 2a: schema-based path/method validator (no network)


JSONbored:
Part of #7015 (MCP execute Phase 2). Step 1 of 3 -- this issue is pure validation logic, no fetch/network code, no tool wiring. #7015b and #7015c build on this.

## Scope

Add a pure, unit-testable function that decides whether a caller-supplied concrete `path` + `method` is declared in a surface's captured OpenAPI document -- the gate that makes Phase 2 "validate against schema, never guess an arbitrary path" real.

- New exported function in `src/call-subnet-surface.mjs` (or a new sibling module if that reads cleaner -- implementer's call), suggested shape:
  `matchSchemaOperation(document, path, method) -> { operation, matchedTemplate } | null`
  - `document` is the exact shape `get_api_schema` returns under its `document` field (the captured OpenAPI spec's `paths` object keyed by path template, e.g. `"/users/{id}": { get: {...}, post: {...} }`).
  - `path` is a concrete request path (e.g. `/users/123`), never a template.
  - `method` is case-insensitive on input (accept `"POST"` or `"post"`); OpenAPI's own path-item keys are lowercase (`get`, `post`, `put`, ...) -- compare against those.
- Path-template matching: a declared template segment wrapped in `{}` (e.g. `{id}`) matches exactly one concrete path segment, any value. A literal segment must match exactly (case-sensitive). The template and the concrete path must have the same segment count -- no partial/prefix matches. Multiple `{param}` segments in one template must all be honored (e.g. `/users/{id}/posts/{postId}`).
- Return `null` (not a thrown error) when: no declared path template matches the concrete path's shape, OR a template matches but the requested method isn't one of that path item's declared operations, OR `document.paths` is missing/empty. This function never throws for "not found" -- only for genuinely malformed input (e.g. `path` not starting with `/`), so callers can treat `null` as "reject the request" uniformly.
- On a match, `operation` is the full OpenAPI Operation object for that path+method (needed unmodified by #7015c, which reads its `requestBody` sub-schema) -- do not strip fields.

## Explicitly out of scope

- No fetch, no tool wiring, no `call_subnet_surface` changes -- that's #7015b.
- No request-body validation -- that's #7015c.
- Query-string handling is unaffected/untouched (Phase 1's existing `query` merge behavior in `callSubnetSurface` stays exactly as-is; this function only ever sees the path portion).

## Required for merge

[...]

PR #7680, opened 2026-07-22
feat(mcp): wire GET/HEAD schema-validated execution into call_subnet_surface

### LINKED ISSUE #7674: MCP execute Phase 2b: wire GET/HEAD schema-validated execution into call_subnet_surface
Part of #7015 (MCP execute Phase 2). Step 2 of 3. Depends on #7673 (the path/method validator) -- this issue wires that validator into the live `call_subnet_surface` tool for GET/HEAD only. POST/PUT + request bodies are #7015c, not this issue.

## Scope

Extend the `call_subnet_surface` MCP tool (definition + handler in `src/mcp-server.mjs`, execution in `src/call-subnet-surface.mjs`'s `callSubnetSurface()`) to accept optional `path` + `method` args, GET/HEAD only in this step.

**Backward compatibility is the hard constraint**: when both `path` and `method` are omitted, behavior must be byte-for-byte identical to Phase 1 today -- fetch `surface.url` (with the existing `query` merge), same method resolution (`surface.probe.method === "HEAD" ? "HEAD" : "GET"`). Every existing Phase 1 test must keep passing unmodified.

When `path` + `method` (GET or HEAD) ARE supplied:

1. Resolve the surface's captured schema the same way the `get_api_schema` tool does: `schema_source.surface_id` if present, else the surface's own `surface_id`, reading `/metagraph/schemas/{artifactId}.json` (see `resolveArtifactSurfaceId` + `get_api_schema`'s handler in `src/mcp-server.mjs` for the existing pattern -- reuse it, don't reimplement).
2. If the surface has no captured schema at all, reject with a tool error (suggested code: `no_schema`) explaining path/method execution requires a captured schema; Phase 1's single-`url` behavior is the only option for schema-less surfaces (per #7015's own "explicitly out of scope: surfaces without a captured schema stay on Phase 1's single-url behavior -- no speculative path guessing").
3. Run `matchSchemaOperation(document, path, method)` from #7673. If it returns `null`, reject with a tool error (suggested code: `path_not_declared`) naming the surface and the rejected path/method.
4. On a match, build the actual request URL from **the surface's own curated base -- never the OpenAPI document's own `servers[].url`**. This mirrors `get_api_schema`'s own guidance ("prefer the curated surface base_url over any upstream server/callback hints") and the `list_subnet_apis` catalog's existing `base_url` field -- the schema is a source of truth for *paths/methods/bodies*, never for *which host to call*. Concretely: take the surface's own `url`'s origin (scheme+host, discarding whatever path it currently ends in) and join it with the validated concrete `path`.
5. Reject `method` values other than GET/HEAD with `invalid_params` in this step (POST/PUT unlocked by #7015c).
6. Everything downstream of URL construction is UNCHANGED: same `safetyCheckedFetch` (redirect revalidation, `isUnsafeUrl` guard, 5-hop cap), same `MAX_RESPONSE_BYTES` cap, same binary-content-type rejection, same response shape.

## Explicitly out of scope

- POST/PUT methods and request bodies -- #7015c.
- Any change to the schema-lookup/validator logic itself (that's #7673 -- consume it here, don't modify it).
- Any credential/auth handling (still Phase 3, #7016 -- `auth_required:true` surfaces are already rejected before this code path runs).

## Required for merge

- [ ] Tool inputSchema gains optional `path` (string) and `method` (enum: `GET`, `HEAD`) params, both omittable.
- [ ] Regression: path/method omitted on any existing Phase 1 test fixture produces the exact same result as before this change.
- [ ] New tests: valid path/method on a schema-bearing surface constructs the correct URL and fetches it; path/method on a schema-less surface is rejected (`no_schema`); undeclared path or method is rejected (`path_not_declared`); a POST/PUT method value is rejected in this step.
- [ ] Meets the repo's 99% branch-counted patch coverage bar (`codecov/patch`).
- [ ] `npm run lint` / `npm run format:check` clean.

## Not yet in scope for this issue

Tool description text still says "path/method beyond the surface's own declared url is Phase 2, not yet built" -- leave that as-is until #7015c, which is the PR that completes Phase 2 and updates it.

JSONbored:
## Summary

Closes #7674 (MCP execute Phase 2b — step 2 of 3 toward #7015).

**Stacked on #7676** (Phase 2a, #7673 — the schema-based path/method validator). This PR's diff includes #7676's commit until that one merges; once it does, this diff will shrink to just this PR's own changes. Rebase target: `main` after #7676 lands.

`call_subnet_surface` now accepts optional `path` + `method` (GET/HEAD only in this step — POST/PUT is #7675/Phase 2c) to call a different route on a surface's own host, instead of only ever its single curated `url`. Backward compatible: omitting both leaves Phase 1 behavior byte-for-byte unchanged.

## What changed

1. **`matchSchemaOperation` wired in** (`src/mcp-server.mjs`): when `path`+`method` are both supplied, the handler resolves the surface's captured schema, runs #7673's validator against it, and rejects anything not declared — before ever constructing a request URL.
2. **Request execution** (`src/call-subnet-surface.mjs`): `callSubnetSurface` gained `path`/`method` options. When `path` is set, the request URL is built from the surface's own **origin** (never the OpenAPI document's own `servers[].url`) joined with the validated path — matching `get_api_schema`'s existing "prefer the curated surface base_url" guidance.
3. **A real gap found and fixed along the way**: `operational-surfaces.json` — what `call_subnet_surface` actually resolves surfaces from — didn't carry `schema_source` at all, even though the agent-catalog build already computes it per surface (`resolveAgentServiceSchema`/`serviceSchemaSource`). Extended the same `operationalSurfaces` projection in `scripts/build-artifacts.mjs` to include it, reusing the existing resolution logic rather than reimplementing it. Without this, there was no way to find which schema a `subnet-api` surface's sibling `openapi` surface owns. Committed the regenerated `public/metagraph/operational-surfaces.json` alongside the script change (246/617 operational surfaces now resolve a `schema_source`).

## New error codes

- `no_schema` — `path`/`method` supplied but the surface has no captured schema at all.
- `path_not_declared` — the path, or the path+method combination, isn't declared in the surface's captured schema.
- `invalid_params` — `path` without `method` or vice versa; a `method` other than GET/HEAD.

## Validation

- [x] 11 new MCP-level tests (`tests/call-subnet-surface-mcp.test.mjs`): valid path/method call, `schema_source` cross-reference resolution, self-schema fallback (no `schema_source`, own `surface_id` resolves), case-insensitive method, both `invalid_params` pairing cases, unsupported method, `no_schema`, `path_not_declared` (undeclared path, and declared path/undeclared method), `auth_required` still checked before schema resolution, Phase 1 behavior unchanged with no path/method.
- [x] 5 new unit tests (`tests/call-subnet-surface.test.mjs`) for `callSubnetSurface`'s new `path`/`method` options in isolation.
- [x] Full regression: `tests/mcp-server.test.mjs` + `tests/mcp-server-branch-coverage.test.mjs` + both `call-subnet-surface*` files — 1313 tests, all passing.
- [x] Coverage: `call-subnet-surface.mjs` 100% lines; none of the uncovered lines in `mcp-server.mjs`'s coverage report fall within this PR's edited range.
- [x] `npm run build` succeeds; `schema_source` confirmed populated in real build output.
- [x] `npm run lint` / `npm run format:check` — clean.

## Note

`npx vitest run tests/artifacts.test.mjs` has one pre-existing, unrelated failure (`enrichment guidance ignores maintainer-excluded candidate URLs`, SN38/ChronoLLM) — confirmed via a stash-and-rerun against clean `main` that this predates this PR entirely (caused by #7672 adding real evidence for SN38, which correctly flipped that candidate's derived `evidence_action`; the test's hardcoded expectation is now stale). Not touched here — out of scope for Phase 2b.


JSONbored:
Confirmed valid — verified directly: `new URL("//attacker.example", "https://x.example").origin` resolves to `https://attacker.example`, not the surface's own host, while `splitPathSegments` produces identical segments for `//attacker.example` and `/attacker.example`, so any schema declaring a single-segment parameterized template lets it pass `matchSchemaOperation`.

Fixed by asserting the resolved URL's origin matches the surface's own origin at the point `baseUrl` is actually constructed in `callSubnetSurface` — the authoritative check regardless of which string tricks schema matching, rather than trying to enumerate every equivalent bypass at the segment-matching layer. Wired into the existing `forbidden` tool error alongside `unsafe_url`/`private_redirect_blocked`.

Added 3 unit tests (protocol-relative path rejected, scheme-qualified absolute path rejected, same-origin path with an extra leading slash still resolves correctly — no false positive) and 1 MCP-level test confirming the `forbidden` mapping through the real tool-call path. Also resolved the `operational-surfaces.json` merge conflict by merging fresh `main` and regenerating the artifact.

PR #7963, opened 2026-07-24
feat(api): subnet_turnover changes toggle with a typed churn detail (#7883)

### LINKED ISSUE #7883: GraphQL parity: subnet_turnover field missing the changes toggle
**Mirror this exactly:** REST's own `changes` query param handling for `GET /api/v1/subnets/{netuid}/turnover`.

## Context

`subnet_turnover(netuid, window)` has no way to request the underlying per-neuron changes list — `GET /api/v1/subnets/{netuid}/turnover` supports a `changes` boolean toggle that GraphQL doesn't expose.

## Requirements

- Add a `changes: Boolean` argument to `subnet_turnover` in `src/graphql.mjs`.
- Resolver passes it through to the same loader the REST route uses.

## Deliverables

- [ ] `subnet_turnover` field accepts `changes` matching REST.

## Expected Outcome

A GraphQL client can request the detailed changes list for subnet turnover, matching REST.

## Links & Resources

- REST route: `GET /api/v1/subnets/{netuid}/turnover`.

kai392:
Closes #7883

## Summary

Adds a `changes: Boolean` argument to `subnet_turnover`, closing a parity gap: `GET /api/v1/subnets/{netuid}/turnover` accepts a `changes` toggle that returns the underlying per-neuron churn detail, but GraphQL only ever returned the summary scorecard.

**The detail is modelled as real GraphQL types, not an opaque `JSON` blob.** A previous attempt at this issue exposed it as `changes: JSON`; review flagged that an untyped passthrough sits awkwardly in a typed object, so this version defines proper types — the entered/exited validator lists and UID reassignments are introspectable and field-selectable like the rest of the schema.

```graphql
type SubnetTurnoverChanges {
  validators_entered_count: Int!
  validators_exited_count: Int!
  uid_reassignment_count: Int!
  validators_entered: [TurnoverValidatorChange!]!
  validators_exited: [TurnoverValidatorChange!]!
  uid_reassignments: [TurnoverUidReassignment!]!
}
```

`TurnoverValidatorChange` is `{ hotkey: String!, uid: Int }` and `TurnoverUidReassignment` is `{ uid: Int!, from_hotkey: String!, to_hotkey: String! }` — matching `src/turnover.ts`'s `ValidatorDetail` and reassignment shapes exactly.

## What changed (2 files)

- **`src/graphql.ts`** — the three types above, the `changes: Boolean` arg, and `turnoverChangesNode`, which normalizes REST's `turnoverChangeDetail` block: counts fall back to their own list lengths (so a body with lists but no counts still answers consistently), and entries missing a `hotkey`/`uid` are dropped rather than surfaced as nulls inside the non-nullable list items. The param is forwarded **only when the toggle is true**, so the default scorecard request stays byte-identical. Resolves to `null` on the default request and on the cold `buildTurnover([])` fallback, which carries no `changes` block.
- **`tests/graphql.test.ts`** — the typed detail round-tripping through a full field selection, the default omitting the param with `changes` null, the cold store resolving null even with the toggle set, and the count-fallback / malformed-entry-dropping behaviour.

GraphQL is served dynamically from the SDL — no generated contract/OpenAPI artifacts change.

## Validation

- [x] `tests/graphql.test.ts` green (928 tests) — every added line exercised (100% patch).
- [x] `tsc --noEmit` clean; `eslint` + `prettier` clean.

### DIFF
--- a/src/graphql.ts
+++ b/src/graphql.ts
@@ -829,7 +829,7 @@ export const SDL = `
     "Live current registration/burn cost for one subnet -- the dynamic price between the static min_burn_tao/max_burn_tao bounds, read directly from chain via RPC (not the Postgres tier). burn_tao is null on RPC failure, schema-stable, never a GraphQL error. Mirrors GET /api/v1/subnets/{netuid}/burn."
     subnet_burn(netuid: Int!): SubnetBurn
     "One subnet's validator/neuron-set turnover (entered/exited/retention/0-100 stability) between the boundary snapshots of a 7d/30d/90d/1y/all window (default 30d), from neuron_daily. comparable is false and the churn metrics zeroed on a single-snapshot or cold store, never null. Mirrors GET /api/v1/subnets/{netuid}/turnover."
-    subnet_turnover(netuid: Int!, window: String): SubnetTurnover!
+    subnet_turnover(netuid: Int!, window: String, changes: Boolean): SubnetTurnover!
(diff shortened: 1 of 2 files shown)


kai392:
Recreating: the Closes #7883 link did not register on this PR. Reopening a fresh PR with the same commit so the issue link is recorded.

Issue #7899, opened 2026-07-24
feat(mcp): add list_subnet_candidates — the filtered sibling of get_subnet_candidates


JSONbored:
**Mirror this exactly:** `src/subnet-evidence-mcp.ts` — clone its module shape exactly (separate modular tool file, full REST filter/sort/page parity, spread into `src/mcp-server.mjs`'s tool object).

## Context

`get_subnet_candidates` (`src/mcp-server.mjs`) only reads a static baked artifact with zero filters. `GET /api/v1/subnets/{netuid}/candidates` supports `kind, provider, state, id, confidence, sort, order, limit, cursor` — none reachable via any MCP tool today. The established convention for this exact situation (a filter-less legacy `get_subnet_X` tool sitting next to a fully-filterable REST route) is a new modular `list_subnet_X` sibling tool, already done for `endpoints` (`src/subnet-endpoints-mcp.ts`) and `evidence` (`src/subnet-evidence-mcp.ts`).

## Requirements

- Create `src/subnet-candidates-mcp.ts`, cloning `src/subnet-evidence-mcp.ts`'s structure.
- Implement `list_subnet_candidates(netuid, kind, provider, state, id, confidence, sort, order, limit, cursor)` with full REST filter parity.
- Spread the new tool into `src/mcp-server.mjs`'s tool registry the same way the other modular `subnet-*-mcp.ts` files are.
- Leave `get_subnet_candidates` untouched — this adds a sibling, it does not replace the legacy tool.

## Deliverables

- [ ] New `src/subnet-candidates-mcp.ts` module exporting `list_subnet_candidates` with full REST filter parity.
- [ ] Tool registered in `src/mcp-server.mjs`'s tool object.
- [ ] Test coverage: at least two filters combined plus pagination.

## Expected Outcome

An MCP client can filter one subnet's candidates the same way it can filter its endpoints or evidence today.

## Links & Resources

- REST route: `GET /api/v1/subnets/{netuid}/candidates`.
- Legacy tool: `get_subnet_candidates`, `src/mcp-server.mjs`.
- Precedent: `src/subnet-evidence-mcp.ts`.

Issue #1767, opened 2026-06-25
Bring local npm run check to parity with CI validators (+ a meta-test to prevent re-drift)


JSONbored:
## Problem
`npm run check` (= lint + format + `pipeline:check` + `validate:docs`) is the documented local pre-push gate, but its validator list has **drifted from CI**: `.github/workflows/validate.yml` runs `validate:committed-seed` (the #356/#998 cold-start footgun gate), `validate:mcp`, `validate:ai`, and `validate:surface` — **none** of which are in `scripts/pipeline.mjs`'s `checkCommands`. A dev who sees local green can still red CI.

## Deliverables
- [ ] Add `validate:committed-seed`, `validate:mcp`, `validate:ai`, `validate:surface` to `pipeline.mjs` `checkCommands` (and `refreshCommands` where applicable).
- [ ] Add a **meta-test** asserting `pipeline.mjs`'s validator set is a superset of `validate.yml`'s `npm run validate:*` steps, so the two can never silently diverge again.

## Evidence
`scripts/pipeline.mjs:59-90` (missing the four); `validate.yml:143,209,212,239` (runs them). `pipeline:check` is local-only (absent from any workflow).

---
Part of #1755. Surfaced by the 2026-06-25 cross-repo audit (adversarially verified).

PR #7980, opened 2026-07-24
feat(api): subnet_candidates GraphQL filter/sort/page parity (#7878)

### LINKED ISSUE #7878: GraphQL parity: subnet_candidates field is an unparameterized JSON blob, missing the full REST filter set
**Mirror this exactly:** the sibling `subnet_evidence`/`subnet_gaps`-style fields once fixed in this same batch, and MCP tool `get_subnet_candidates` (`src/mcp-server.mjs`).

## Context

`subnet_candidates(netuid): JSON` in `src/graphql.mjs` takes only `netuid`. `GET /api/v1/subnets/{netuid}/candidates` supports `kind, provider, state, id, confidence, limit, cursor, sort, order` — none exposed.

## Requirements

- Add `kind: String, provider: String, state: String, id: String, confidence: Float, limit: Int, cursor: String, sort: String, order: String` arguments to `subnet_candidates` in `src/graphql.mjs`.
- Resolver passes them through to the same loader `get_subnet_candidates` uses.

## Deliverables

- [ ] `subnet_candidates` field accepts the full REST filter set.
- [ ] Test coverage: at least two filters combined plus pagination.

## Expected Outcome

A GraphQL client can filter a subnet's candidates identically to REST.

## Links & Resources

- REST route: `GET /api/v1/subnets/{netuid}/candidates`.
- MCP tool: `get_subnet_candidates`, `src/mcp-server.mjs`.

RealDiligent:
## Summary

`subnet_candidates(netuid)` exposed **none** of the REST route's query surface — it returned the whole baked artifact verbatim. `GET /api/v1/subnets/{netuid}/candidates` supports `kind, provider, state, id, confidence, sort, order, limit, cursor`. This adds all of them.

Closes #7878.

## Approach — reuse, not reimplement

The resolver delegates to **`loadSubnetCandidatesList`**, the same loader MCP `list_subnet_candidates` calls (#7899, merged as #7930), instead of open-coding a filter/sort/page pass. That's the convention this file already uses for `review_gaps`, `review_adapter_candidates`, and `adapter` — and it means this field *cannot* drift from REST, because it executes the same allowlists and the same pagination.

## Behavior

- **Filters/sort/paging**: full REST parity, with the same 1–100 limit clamp and integer cursor.
- **Unsupported filter/sort value** → `BAD_USER_INPUT` GraphQL error, matching the file's "not a silently substituted default" convention.
- **Unreadable artifact** (not baked / cold R2 / storage failure) → still resolves to **null**, preserving this field's documented cold-artifact contract. `loadArtifact`, which the resolver used before, swallowed exactly those failures the same way — so the four pre-existing tests pass unchanged.
- **An unexpected non-`toolError` failure propagates** rather than being masked as "no candidates baked".
- The envelope now carries the same pagination meta REST returns (`total`, `returned`, `limit`, `cursor`, `next_cursor`, `sort`, `order`) alongside `candidates`; the SDL description documents all of it.

## Two deliberate deviations from the issue text

1. **`confidence` is `String`, not `Float`.** The published `CandidateSurface` schema defines it as `enum: ["low","medium","high"]` (`schemas/api-components.schema.json`), and the `candidates` query collection validates it as that enum. A `Float` argument could never match a row, which would defeat the issue's own Expected Outcome ("filter identically to REST").
2. **`cursor` is `Int`, not `String`.** That matches the loader and the REST route's offset cursor, and the sibling loader-backed field `review_adapter_candidates(… cursor: Int)`.

Happy to switch either if you'd rather match the issue text literally.

## Validation (full CI-parity gate set, run locally on this tree)

- [x] `npx vitest run tests/graphql.test.ts` — **929 passing** (whole GraphQL suite; the 4 pre-existing `subnet_candidates` tests unchanged and green)
- [x] Changed resolver at **100% statement AND branch coverage** — no missing lines and **no partials** (verified via `vitest --coverage.include=src/graphql.ts`)
- [x] `npx tsc --noEmit` — **0 errors** in the touched files
- [x] `node scripts/validate-api.ts` — 177 Worker API routes + 8 Postgres-tier routes validated
- [x] `eslint` 0 errors; repo-pinned `prettier --check` clean on staged LF content

## Note

Supersedes #7957, which failed `codecov/patch` at 75% (1 missing line + 1 partial branch on the error path — patch coverage counts only the source diff, so on an 8-line change two misses are fatal). This branch adds a test that drives an unexpected loader failure through the resolver, taking the changed lines to 100% with no partials.

### DIFF
--- a/src/graphql.ts
+++ b/src/graphql.ts
@@ -52,6 +52,10 @@ import { loadAdapterCandidatesList } from "./adapter-candidates-mcp.ts";
 import { loadEnrichmentEvidenceList } from "./enrichment-evidence-mcp.ts";
 import { loadEnrichmentQueueList } from "./enrichment-queue-mcp.ts";
 import { loadReviewEnrichmentTargetsList } from "./review-enrichment-targets-mcp.ts";
+// #7878: GraphQL parity for GET /api/v1/subnets/{netuid}/candidates, reusing
+// loadSubnetCandidatesList that MCP list_subnet_candidates already calls
+// (#7899) -- not a reimplementation.
+import { loadSubnetCandidatesList } from "./subnet-candidates-mcp.ts";
 import { loadReviewGapsList } from "./review-gaps-mcp.ts";
 import { loadProfileCompletenessList } from "./profile-completeness-mcp.ts";
 // #6984: GraphQL parity for GET /api/v1/adapters/{slug}, reusing loadAdapter that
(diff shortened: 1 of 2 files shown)


RealDiligent:
Heads-up on the extra commit in this PR: `tests/pipeline-validator-parity.test.ts` (#1767) is currently **failing on `main`**, independently of this change —

```
validate.yml runs validators the local `npm run check` does not: validate:no-hand-written-mjs.
Add them to scripts/pipeline.ts checkCommands (and refreshCommands).
```

#7956 added the *Validate no hand-written .mjs/.js* step to `validate.yml` (and `scripts/validate-no-hand-written-mjs.ts`) but never registered the script in `scripts/pipeline.ts`, so the parity test that guarantees `npm run check` is a superset of CI now fails for **every** PR in the pool, not just this one.

I've included the 2-line registration (`scripts/pipeline.ts`, after `validate:migrations` to match its position in `validate.yml`) as a **separate, clearly-labelled commit**, because this PR's `test` job cannot go green without it. It changes nothing about what CI runs — the npm script already existed. Happy to drop it into its own PR instead if you'd prefer that; I kept it here only because the gate closes red PRs before an out-of-band fix could land.

Also rebased onto current `main`.

JSONbored (approved):

PR #8054, opened 2026-07-24
build(openapi): generate OpenAPI components from the Zod layer for 5 pilot routes

### LINKED ISSUE #7860: types-epic B: generate OpenAPI 3.1 from the Zod layer; invert the schemas/ flow for covered routes
Part of #7858. Depends on A (#7859) being merged — do not start before.

## Scope
Make the Zod layer the generator of the public OpenAPI document for every route it covers, replacing the hand-edited `schemas/` JSON Schema files as the source for those routes' component schemas. Routes not yet covered by Zod keep their existing hand-edited schemas untouched (incremental inversion, not big-bang).

## Context you must load before writing code
- `src/contracts.ts` builds `public/openapi.json` today via `buildOpenApiArtifact()` from `API_ROUTES` + `schemas/` components; read its `route()`/`artifact()` constructors and the `schemaRefForArtifactPath` mapping.
- `validate:contract-drift` (see `.claude/skills/metagraphed/reference.md` validator table) fails CI when `public/openapi.json` or generated types are stale — your generator must slot into the same `npm run build` pipeline so the drift gate keeps working unchanged.
- Zod 4 ships native `z.toJSONSchema()`; use it for component emission. Use `zod-openapi` (pin exact version) ONLY if the document-assembly ergonomics (paths/parameters/responses) demonstrably need it — record the decision either way in the PR body with a two-sentence rationale.

## Requirements
1. New build step `npm run build:openapi-zod` invoked from the existing `npm run build`, emitting component schemas for every `schemas-src/routes/*` schema into the SAME `public/openapi.json` document the current pipeline produces. The document must remain ONE file; consumers (client SDK, MCP server-card, fumadocs) see no path change.
2. For each covered route, the generated component must replace the hand-edited `schemas/` counterpart, and the replaced file must be DELETED in the same PR (acceptance: no orphaned hand-edited schema shadows a generated one).
3. Diff discipline: the FIRST PR must include a `docs/`-visible diff summary (in the PR body, not a committed file) of every material JSON Schema difference between the old hand-edited component and the new generated one for the 5 pilot routes. Differences fall into exactly three buckets, each requiring a stated resolution: (a) the Zod schema is wrong → fix the Zod schema; (b) the hand-edited schema was wrong/stale → keep the generated one and note the correction; (c) cosmetic (key ordering, `$ref` layout) → acceptable, note once.
4. `tests/contracts.test.ts` / `tests/invariants-contracts.test.ts` / `tests/api-coverage.test.ts` must keep passing without weakening any assertion; where they assert on component internals, update them to read the generated components.
5. The client SDK (`packages/client`) type generation must keep consuming `public/openapi.json` unchanged. Do NOT bump `packages/client/package.json` (sync-client-version owns that post-merge) unless the generated document is materially different in a way that changes emitted client types — if it is, that is bucket (a)/(b) above and must be resolved, not shipped silently.
6. Runtime validation wiring (optional flag, default OFF): a `METAGRAPH_VALIDATE_RESPONSES` env flag that, when `"true"`, parses handler responses for covered routes against their Zod schema and logs (never throws) on mismatch. This is the drift tripwire for staging; it must be zero-cost when unset (guard before any parse).

## Deliverables
- Generator script + `npm run build` integration; regenerated `public/openapi.json` committed.
- Deleted hand-edited schema files for covered routes.
- PR-body diff audit for the pilot five (bucketed as above).
- The staging-only response-validation tripwire behind the flag.

## Acceptance criteria
1. `npm run build` from a clean tree is idempotent (second run produces zero diff) — this is what keeps `validate:contract-drift` meaningful.
2. CI fully green including `checks` (contract gates) and codecov/patch.
3. `public/openapi.json` remains valid OpenAPI 3.1 (validate with the repo's existing tooling; if none validates the document today, add a test that does — that test is in scope).

## Non-goals
- No coverage expansion beyond the pilot routes in the first PR (follow-up PRs under this same issue may add route batches — one comment per batch on this issue with the route list).
- No docs work (F). No request-side validation (only the logging tripwire above).

JSONbored:
Closes #7860

## Summary

Makes the Zod schema layer (`schemas-src/`, types-epic A) the source of truth for the public OpenAPI 3.1 document's component schemas for the 5 pilot routes (subnets, subnet-detail, health, economics, subnet-stake-quote) and their shared sub-shapes — 13 components total — replacing the equivalent hand-edited `schemas/components/*.schema.json` entries for those specific names. All other routes' hand-edited schemas are untouched.

- `schemas-src/openapi-registry.ts` + `scripts/generate-openapi-zod-components.ts`: a `z.registry()` of the 13 components + a `z.toJSONSchema()`-based generator.
- `scripts/openapi-components.ts`: the generator's output is spread over the hand-edited bundle inside the existing `loadOpenApiComponentSchemas()` merge point — every downstream consumer (`build-artifacts`, `generate-openapi`, `validate-contract-drift`, the client SDK) inherits the change automatically.
- `npm run build:openapi-zod` / `npm run diff:openapi-zod` scripts + a `build.ts` step wired into both local and production pipelines.
- `src/response-validation-tripwire.ts`: an opt-in `METAGRAPH_VALIDATE_RESPONSES` runtime flag (default `"false"`) that warns (never throws) when a covered route's live response drifts from its Zod schema.
- `schemas/openapi-3.1-meta-schema.json` + `tests/openapi-3-1-meta-schema.test.ts`: vendors the official OpenAPI 3.1 meta-schema (from its canonical `spec.openapis.org` location) and validates `public/metagraph/openapi.json` against it with the repo's existing ajv/ajv-formats — `openapi-schema-validator` (npm) was evaluated and rejected first (only bundles OpenAPI 2.0/3.0, no 3.1 support). Getting this working required a documented workaround for a verified ajv `$dynamicRef` resolution gap in the spec's self-referential Schema Object dialect (see the test file's header comment for the empirical proof and why the workaround is spec-equivalent, not a loosening).

## Zod ↔ hand-edited component diff audit (13/13 PASS)

`npm run diff:openapi-zod` normalizes and deep-compares each Zod-generated component against the hand-edited JSON Schema it replaces. Findings, by bucket:

**(a) Zod schema was wrong → fixed** (10 fields tightened to match what the hand-edited schemas already correctly declared, each verified against real staged artifact data before changing):
- `dashboard_url`, `discord_url`, `docs_url`, `logo_url`, `source_repo`, `website_url` (subnets, subnet-detail): `z.string()` → `z.url()`
- `github_last_push_at`, `updated_at`: `z.string()` → `z.iso.datetime()`
- `PartnershipMetadata.since`: `z.string()` → `z.iso.date()`
- `VerificationResult.{html_url,source_url,url,source_urls}`, `CandidateSurface.{source_url,url,source_urls}`, `EndpointResource.{source_urls,url}`: → `z.url()` / `z.array(z.url())`
- `Auth.names`: `.optional()` → `.min(1).optional()`

**(b) hand-edited was stale/lossy → generated kept, noted**:
- `SubnetEconomics` component dropped entirely — nothing references it by name once `EconomicsArtifact`/`SubnetDetailArtifact` are Zod-owned (inlined at both use sites instead).
- `description`/`examples` annotations on hand-edited schemas aren't yet reproduced by the Zod layer (no `.describe()`/`.meta({examples})` calls yet) — real, low-severity documentation loss, tracked as follow-up scope, not a blocker here.

**(c) cosmetic (normalized in the diff script, no schema change needed)**:
- `allOf:[{$ref:"ArtifactBase"}, {...}]` (hand-edited) vs. a flattened object (Zod's `.extend()`).
- `type:[X,"null"]` with siblings (hand-edited) vs. `anyOf:[{type:X,...},{type:"null"}]` (Zod).
- `propertyNames`/`required`-ordering/`additionalProperties:{}` vs `true`/`oneOf` vs `anyOf` — all semantically identical.
- Sentinel `Number.MAX_SAFE_INTEGER` bounds Zod stamps on unbounded `z.int()`.
- `pattern` string `/` escaping differs (`\/` vs `/`) — same regex either way; also required a matching fix in `src/openapi-sample.ts`'s example generator.
- `z.iso.date()`/`.datetime()` pairing `format` with a strictly redundant `pattern`.
- The pre-existing `HttpUrlSchema`/`HttpOrWssUrlSchema` patterns (types-epic A) are intentionally stricter than bare `format:"uri"` — a deliberate, already-shipped difference.
- `reused:"inline"` duplicates small shared shapes at each use site instead of `$ref`-ing them (chosen over `reused:"ref"`, which produces a non-embeddable anonymous `__shared` bucket).

## Verification

- `npm run build` — reproducible, clean `git status` after.
- `npm run diff:openapi-zod` — 13/13 PASS.
- `npm run validate:*` — schemas, api, openapi, openapi-examples (177/177 worked examples), contract-drift, generated-client, client-sdk-sync, types, schema-enums, committed-seed, artifact-budgets, docs, mcp, ai — all pass.
- `tsc --noEmit`, `eslint`, `prettier --check` — clean.
- New/changed tests (`tests/openapi-3-1-meta-schema.test.ts`, `tests/surface-url-scheme-pattern.test.ts`) — pass locally; full suite left to CI.


JSONbored:
Addressed both review blockers in a follow-up commit (`c5ce10e53`, force-pushed after a rebase onto fresh `main`):

**1. Dropped named components (real bug, fixed).** `SurfaceKind`/`Authority`/`Classification`/`BittensorNetwork`/`HealthStatus`/`EndpointLayer`/`ProbeConfig`/`EndpointMonitoringPolicy`/`EndpointScoreReason`/`VerificationResult`/`SourceTier`/`CurationLevel`/`ReviewState` were being inlined and dropped from `packages/contract/index.d.ts` because `z.toJSONSchema()`'s `reused:"inline"` only keeps a schema as a named component when it's a separately registered root — the original registry only registered the 13 artifact-level shapes, not the leaf enums/objects they use. Exported the 5 previously-private Zod consts these needed, registered all 13 in `schemas-src/openapi-registry.ts`, and deleted their now-redundant hand-edited JSON. `packages/contract/index.d.ts` now matches `main` for every one of those names; `SubnetEconomics` remains the *only* intentional removal (documented in the registry's header, unchanged from before).

**2. codecov/patch (61.90% → target 99%).** Added `tests/response-validation-tripwire.test.ts`: direct coverage of every branch in the tripwire (unknown route, matching data, drifted data, a loader failure via `vi.doMock`) plus an integration pass with `METAGRAPH_VALIDATE_RESPONSES="true"` through all 5 pilot routes. Writing that integration test surfaced a second real bug: `subnet-stake-quote` is served by `handleSubnetStakeQuote` (`workers/request-handlers/entities.ts`), which returns *before* ever reaching `handleApiRequest`'s generic envelope assembly — so the tripwire's `"subnet-stake-quote"` entry was dead code, unreachable despite being one of the 5 declared-covered routes. Threaded an optional `ctx` through that handler and wired the same flag-gated call. Also added a `format:"date"` case to `tests/openapi-sample.test.ts` (only `date-time` had one before).

Two other items from the review, not changed:
- **"Diff doesn't show the hand-edited files being deleted."** They are — `git diff --stat` on the original commit shows `schemas/components/{04-surfaces,05-subnets,06-health,07-endpoints-rpc}.schema.json` losing ~1,346 lines net. I believe the bot read `diff-openapi-zod-components.ts`'s `execFileSync("git", ["show", "HEAD:<file>", ...])` calls (used to fetch the *pre-deletion* content for the equivalence comparison) as evidence the files still had that content, rather than recognizing that's the intentional "load the old side of the diff" step.
- **"Patch-less file, secret scan cap exceeded for `public/metagraph/openapi.json`."** The full file is ~1.8MB, above whatever fetch cap the scanner has — that's independent of this PR's diff size (even a one-line change would trip the same cap, since it's the *whole file* being fetched for a full-content scan, not the diff). The issue's own requirement 1 is explicit that this document "must remain ONE file," so splitting it isn't an option here; flagging as a known, structural limitation of scanning this repo's largest generated artifact rather than something actionable in this PR.

`npm run diff:openapi-zod` now reports 26/26 PASS (13 from the original commit, now `SKIP` since their hand-edited predecessor no longer exists to diff against post-deletion, + the 13 new ones, all `PASS`). Full validate suite, targeted tests, `tsc`, `eslint`, `prettier` all clean locally.

JSONbored:
CI caught one more real issue after the last push: \`validate:schema-enums\` was reading component enums from \`schemas/api-components.schema.json\` (the hand-edited-only bundle), which no longer has the 6 enums that just moved to Zod-generated (\`SurfaceKind\`, \`EndpointLayer\`, \`CurationLevel\`, \`HealthStatus\`, \`Classification\`, \`Authority\`) — false-positive "missing every value" drift, not a real content problem (the values are correct in the actual published \`openapi.json\`, verified by hand). Fixed (\`7c407e829\`) by pointing the check at the published document instead of the hand-edited-only bundle — a pure widening that also makes this validator resilient to the rest of the migration (every future batch moves more components the same way). Ran the full \`validate:*\` suite (every script \`validate.yml\` runs), \`typecheck\`, \`build\`, \`scan:public-safety\`, and \`worker:bundle:budget\` locally this time, not just a subset.

JSONbored:
Found and fixed the actual failure (\`0b72b0fd4\`): \`tests/validate-classification-enum-parity.test.ts\` read \`schemas/components/01-enums.schema.json\` directly (the hand-edited source file), which no longer declares \`Classification\` now that it's Zod-generated — same root cause as the \`validate-schema-enums.ts\` fix, just a test file this time instead of a script, so my earlier \`grep -rln schemas/api-components.schema.json scripts/\` sweep missed it (only checked `scripts/`, not `tests/`).

This time I grepped the **whole repo** (`tests/`, `scripts/`, `src/`) for every reference to any of the hand-edited component files, not just the bundle, and checked each hit individually:
- `tests/provider-url-http-pattern.test.ts` / `tests/provider-submission-schema.test.ts`: load `01-enums.schema.json` for an unrelated `$ref` `provider.schema.json` needs — verified still passing, no dependency on any of the 26 migrated components.
- `src/chain-*.ts`: header comments naming *other*, untouched components (e.g. `ChainPrometheusArtifact`) — not code, no risk.
- `scripts/bundle-schemas.ts`, `scripts/contract-change-summary.ts`, `scripts/validate-contract-drift.ts`: operate on the hand-edited-only bundle by design (building it / diffing it / checking its own reproducibility) — correct, not the same bug class.

Ran the full local test suite in the background after this fix as an extra check (given the pattern already slipped past two narrower sweeps) — all passing, including this test. `tsc`/`eslint`/`prettier` clean, rebased onto fresh `main`.

Issue #8250, opened 2026-07-26
feat(ui): public /status rebuild — correctly scoped verdict, one screen; ledger moves to drill-downs


JSONbored:
Phase 2 · Core pages. **Public /status rebuild — scoped verdict, one screen; /health stays the ops console.**

## Context (measured)

/status is 85,274 px tall (desktop) / 112,433 px (mobile): a red "Partial outage" banner (3 of 617 third-party surfaces down, reads as "the site is down"), incident list, subscribe panel contradicting the count above it, then network-decentralization + emission-yield analytics (off-topic), then the ENTIRE 617-surface probe ledger inline, ending in a broken-dots tail. /health duplicates the same counts three ways in its first two screens and self-describes as maintainer-facing.

## Requirements

- One screen, four parts:
  1. **Verdict, correctly scoped:** "Metagraphed systems: operational" (site/API/pipeline) as the headline; "Tracking 617 subnet surfaces: 552 up · 62 slow · 3 down" as the subline — third-party health NEVER renders as the site's own outage banner. Red reserved for metagraphed-own outages.
  2. **90-day uptime strip** for metagraphed's own components (api.metagraph.sh, site, probe pipeline).
  3. **Incidents**: grouped by subnet, collapsed by default, "Show all" paginates; counts from the Phase-0 shared derivation (must match the subscribe feed on the same page).
  4. **Subscribe**: RSS/Atom/JSON links (existing feeds).
- The full per-surface ledger leaves the page: per-subnet drill-down lives on the subnet detail API tab; a "Browse all surfaces" link goes to /apis/endpoints.
- Decentralization + emission-yield sections move to Chain hub Overview.
- /health: keep as ops console (mosaic/matrix/freshness), linked from /status footer ("Ops console →"); nav demotion handled in the nav issue; hydration fix is Phase 0.
- Mobile: the entire status answer within one screen at 390px.

## Acceptance

- [ ] /status ≤ ~2,500 px desktop with all groups collapsed.
- [ ] Verdict line never blames the site for subnet-API downtime (copy review).
- [ ] Incident count === feed count === endpoints-tab count for the same window.

JSONbored:
[...]

PR #8332, opened 2026-07-27
feat(api): GET /api/v1/self-health — verdict + 90-day component uptime (#8318)

### LINKED ISSUE #8318: feat(api): GET /api/v1/self-health — verdict + 90-day component uptime from Postgres via Hyperdrive
Part of #8316 (self-health monitoring parent). Second sub-issue — the serving route. Depends on #8317 (the table must exist and be accumulating).

## Requirements

- New route `GET /api/v1/self-health` on the data Worker (`workers/data-api.ts` — the Postgres/Hyperdrive-backed read surface, where every other Postgres-reading route lives), standard envelope.
- Response shape:
  - `components[]`: for each of `api` / `site` / `publish` — `{ component, current_ok, days: [{ day, checks, ok_count, uptime_ratio }] }` over the trailing 90 days from `self_health_daily`, plus latest-tick state from `self_health_checks`.
  - `verdict`: `"operational" | "degraded" | "outage"` derived ONLY from these own-component checks — never from third-party subnet-surface health. This is the scoped verdict #8250's copy requirement hangs on.
- **Honest emptiness**: days with no rows are absent, not zero-filled (house rule: probe-derived only, never synthesized). A component with no data yet returns an empty `days` array, not fabricated 100%.
- Schema-first per the repo contract: define in `schemas-src/` (register leaf schemas explicitly — see the Zod-registry gotcha), run `npm run build`, commit regenerated `openapi.json` + types in the same PR. New-route checklist applies (llms.txt / contracts / route inventory CI gates).
- Tests: route handler coverage to the 99% patch bar, including the empty-table and partial-history cases.

## Deliverables

One PR: route + schema + regenerated artifacts + tests.

## Expected outcome

`curl https://api.metagraph.sh/api/v1/self-health` returns the verdict + per-component 90-day series. #8250's frontend rebuild is then fully unblocked and can render the real uptime strip.

## Explicitly out of scope

- No frontend changes (#8250 owns those).
- No changes to the poller job (#8317 owns that).

JSONbored:
> **Stacked on #8331** (the poller that writes these tables). Branched from it rather than main so the schema exists — merge that first, then this rebases onto main cleanly.

## Why

Every other health surface here is about *someone else*: `/api/v1/health` rolls up probed **subnet** surfaces, `/api/v1/incidents` covers public RPC endpoints. So `/status` could only ever answer "are the things we watch up", never "are **we** up" — which is the question a visitor on a status page is actually asking, and the one it looked like it was answering.

## The verdict is scoped, deliberately

It's derived **only** from our own components and never mixes in third-party subnet-surface health. A subnet's API being down says nothing about whether metagraphed is up, and conflating the two is exactly what made the old `/status` verdict unreadable (#8250).

`api` is load-bearing — if it's down the site has nothing to render and every client is broken — so it alone escalates to `outage`. `site` and `publish` failing are real problems that don't stop the API answering, so they `degrade`.

## Three honesty properties, each with a test

| Property | Why |
|---|---|
| A day with no probe rows is **absent**, not zero-filled | A gap means "we weren't measuring". 0% would invent an outage that never happened |
| `current_ok` is **null**, not `false`, for a never-probed component | "We haven't measured this" and "this is down" are different claims, and only one is ours to make |
| Nothing measured at all → **`degraded`** | We can't assert health without evidence — and an empty table isn't an outage either |

The 90d mean is **unweighted by check count** on purpose. Weighting would let a day the poller only ran twice count 30× less than a full day, quietly hiding a short outage on a sparse day.

## Two queries, not a join

The rollup and the latest tick live in tables with different retention — raw ticks pruned at 14d, daily never. A join would silently drop a component whose raw rows have aged out but whose daily history is intact.

## Contract wiring

Every gate in the new-route checklist, including the ones with no lexical hint in the diff:

- `schemas-src/routes/self-health.ts` with **all three leaf shapes registered explicitly** — an unregistered named sub-shape gets inlined rather than `$ref`'d.
- Both `src/contracts.ts` registries — the `route(…)` entry *and* the separate `artifact(…)` entry (omitting the latter makes `npm run build` throw from `schemaRefForArtifactPath`).
- `scripts/validate-api.ts`'s `checks` array (it asserts `checks.length === API_ROUTES.length`).
- `COMPUTED_ARTIFACTS` in `validate-schemas.ts` — live-only, no static file. Without this the validator looks for `public/metagraph/self-health.json` and fails on ENOENT, which is how I found it.
- The prose entry in `docs/backend-artifact-contracts.md`.
- Regenerated `openapi.json` / `types.d.ts` / `api-index.json` / `contracts.json` / `packages/contract`, and the API reference page via `generate-openapi-docs.ts`.

## Verification

**20 new tests**, 574 passing across `self-health` / `data-api` / `contracts`. The route handler is covered in the centralized `tests/data-api.test.ts` (three cases: populated, empty-table, api-down-with-partial-history), not only in the per-feature file — a per-feature file alone leaves the handler at 0% patch coverage.

Local gates green: `tsc --noEmit`, `lint`, `format:check`, `validate:contract-drift`, `validate:schemas`, `validate:schema-enums`, `validate:openapi-examples`, `validate:generated-client`.

**Not verifiable until deploy:** `validate:api` makes a live call against the real route, so its new `checks` entry can't pass until this is deployed and the poller has run. That's inherent to that gate, not something this PR can pre-satisfy.

## Drive-by

`apps/indexer-rs/target` added to `.prettierignore`. It's gitignored but prettier still walked it — anyone who runs `cargo build` currently gets 639 false `format:check` warnings.

Closes #8318


JSONbored:
Correcting something I wrote in the PR body above.

I described the `validate:api` failure as inherent — a live call against a route that can't exist until deploy. **That was wrong.** `scripts/validate-api.ts` invokes the Worker's own router in-process against a cold harness; there's no network call. So the 404 was a real wiring gap, not a deployment-ordering problem: the handler existed in `workers/data-api.ts`, but nothing in `workers/api.ts` routed to it, and the public route genuinely 404'd.

Fixed in the follow-up commit — public handler, router entry with the same `withEdgeCache` treatment its siblings get, the Postgres-tier passthrough entry, and `METAGRAPH_SELF_HEALTH_SOURCE`. `validate:api` now passes locally (178 routes, 8 Postgres-tier).

Worth saying plainly: the gate did its job and my reading of it was the thing at fault. Had I taken the failure at face value the route would have shipped 404ing.

Issue #8386, opened 2026-07-27
feat(api+ui): self-serve API keys + tiered rate limits — keyless stays the generous default


JSONbored:
## Context

The public API is keyless and stays keyless — that posture is the agent-adoption engine and is now a published promise (see the epic). What's missing is the story for heavy consumers: stable identity, higher limits, and usage visibility. The access-model design spike (#6646) surveyed tiered models across the ecosystem; this issue implements the free half of the answer. Payments are explicitly NOT here (sibling decision-memo issue owns that).

**Blocked by**: T6 wallet-verified identity (key issuance binds to a verified address — no accounts system, per the standing posture).

## Requirements

1. **Tiers, decided here**: anonymous (today's generous limits, unchanged — regression-tested so this PR provably does NOT reduce them); keyed-free (issued self-serve: 5× anonymous rate ceiling, stable identity for support/debugging, usage dashboard). One key per verified address active at a time; rotate = revoke + reissue (rotation UI required). Exact numeric limits: read the current anonymous limiter constants and set keyed-free to 5× in the same config block — both documented on the docs rate-limit page in this PR.
2. **Issuance**: /settings "API" section: verify with wallet (T6 flow) → key displayed ONCE with the standard one-time-secret reveal pattern (ph-no-capture applies — /settings replay-blocking already covers the route; state it). Key format: prefixed (`mg_`) + checksummed so leaked-key scanners can match it; add the format to the repo's secret-scanning patterns in the same PR.
3. **Enforcement**: `Authorization: Bearer mg_…` recognized in the existing `api-key-validation` module; limiter keys on the API key when present, IP otherwise; 429 bodies gain honest `limit`/`remaining`/`reset` fields + headers on ALL tiers (additive contract — full regenerate + client-bump procedure; client helper surfaces the headers).
4. **Usage dashboard**: /settings API section shows the last 7d of the key's request counts by day + top endpoints (from the existing usage-telemetry pipeline — verify it can segment by key with a bounded query; if not, add the minimal counter, not an analytics system).
5. **Abuse posture**: keys revocable (self-serve + operator); revoked keys 401 with a distinct error code; issuance rate-limited per address and per IP.
6. Docs: the rate-limits page states the promise verbatim: keyless access never loses capability to a paid tier; keys buy MORE, never gate the base.

## Deliverables

- [ ] Tier config + enforcement + honest 429/limit headers (all tiers) + contract artifacts + client bump
[...]

PR #8572, opened 2026-07-28
feat(api+ui): web-push alerts — a webpush delivery channel on the trigger API

### LINKED ISSUE #8385: feat(api+ui): web-push alerts — a webpush delivery channel on the existing trigger API
## Context

With the PWA installed (sibling sub-issue) and trigger tokens self-serve (T6 — hard dependency), the alert stack can reach lock screens: a `webpush` delivery channel alongside webhook and Discord. This is the retention feature: alerts that arrive without any page open.

**Blocked by**: T6 wallet-verified tokens AND the PWA foundation (SW required for push).

## Requirements

1. **Subscription**: in the alert center (/settings), "Enable push on this device" → standard Push API subscribe → `POST /api/v1/watch/push-subscriptions` storing endpoint+keys bound to the verified address (auth: the T6 token). Limits: 3 subscriptions (devices) per address; explicit unsubscribe UI per device with created-at + user-agent summary so devices are recognizable.
2. **VAPID**: keypair via Worker secrets (ops-managed; documented as a deploy prerequisite in the PR — key VALUES never in the repo). Sender implementation inside the existing alert-delivery module: `webpush` becomes a third channel with the same retry/backoff/history semantics the other channels already have (delivery history must show push results identically).
3. **Payloads**: title = the action sentence when available (T5) else the event kind; body = entity + relative time; tap-through deep-links to the entity page. Payload ≤2KB. No payload without user-visible notification (browser requirement — document the silent-push prohibition in code comments).
4. **Expiry hygiene**: 404/410 from the push service ⇒ subscription pruned + noted in delivery history ("device expired").
5. **Contract**: new routes follow the full procedure (schemas → build → regenerated artifacts → client bump); live-route allowlist + new-route gates.
6. E2E transcript in the PR: subscribe → trigger event → notification rendered (screenshot from a real browser) → tap-through lands on entity.

## Deliverables

- [ ] Subscription routes + device management UI with the decided limits
- [ ] webpush channel in the delivery module with full history parity + expiry pruning
- [ ] Sentence-first payloads with deep links; contract artifacts + client bump
- [ ] E2E transcript with notification screenshot

## Expected outcome

Star a validator, verify once, and your phone tells you when it acts — the full loop from T6's identity through T9's packaging, on infrastructure that already knew how to deliver.

## Links

Blocked by: T6 token issuance + PWA foundation (same milestone) · alert-delivery module (`src/alert-delivery.ts`) · T5 action sentences (payload title source)

JSONbored:
## Summary

Adds `webpush` as a third alert channel alongside webhook and Discord, so an alert can reach a lock screen with no page open — the loop from T6's verified identity through T9's PWA packaging.

Closes #8385

## Transport: written from the RFCs, and *proven* rather than asserted

`src/web-push.ts` implements VAPID request signing (RFC 8292) and aes128gcm payload encryption (RFC 8291) directly on WebCrypto. The usual npm option (`web-push`) is a Node library that reaches for `crypto`/`https` and is not Worker-safe — vendoring a shim would have traded ~200 lines of reviewable, testable crypto for a large unreviewable one *on the delivery path*.

The test that matters: **the suite implements the receiver half of RFC 8291 and actually decrypts the sender's output.** So the round-trip is verified end to end, rather than asserting our own byte layout back at ourselves. It also verifies the ES256 signature against the advertised public key, which is what proves the raw `r||s` JWS encoding is right (porting from Node's DER output is the classic bug here).

Two properties worth calling out, both tested:
- **Salt and ephemeral key are never reused** across sends to the same device — reuse would leak plaintext relationships to the relaying push service.
- **The push service is treated as an untrusted relay.** It sees the endpoint and ciphertext, never the plaintext.

## Storage: its own table, not more `destination` text

A subscription is three correlated values (`endpoint` + two key materials), it is **per-device** while a trigger is **per-alert**, and one device is reused by every trigger the same address owns. So `watch_push_subscriptions` is its own table, bound to the same T6 wallet-verified address the watch triggers already use — "my devices" works with no accounts system.

`p256dh`/`auth` are the subscriber's own key material. They never leave the database except over the internal service binding at send time; **the owner-facing `GET` deliberately returns only the metadata needed to recognise a device** (coarse UA label, created/last-used), never the keys.

Endpoint is `UNIQUE` and the write upserts: browsers reissue the *same* endpoint for an unchanged subscription, so a routine re-subscribe must not read as "device limit reached".

## Silent push is prohibited — enforced structurally, in three places

Browsers permit a `push` handler that shows nothing only briefly, then revoke the permission or post a generic "site updated in background" notification on your behalf. Rather than rely on remembering that, it's enforced by construction:

1. `userVisibleOnly: true` on subscribe (Chrome rejects a subscription without it outright).
2. The payload builder has **no code path that returns an empty title** — sentence → event kind → trigger name → generic, tested including the all-null input.
3. The service worker shows a notification on **every** branch, including a malformed/unparseable payload, with deliberately no early return.

## Requirements, point by point

| # | Requirement | How |
| --- | --- | --- |
| 1 | Subscription + device management, 3/address, recognisable devices | `POST/GET/DELETE /api/v1/watch/push-subscriptions`, cap enforced server-side, coarse `Chrome on Mac`-style labels + created/last-used |
| 2 | VAPID via Worker secrets; full history parity | `VAPID_*` secrets (values never in repo); `webpush` flows through the same `deliverAlertMatch` outcome path, so delivery history is identical |
| 3 | Sentence-first payload, deep links, ≤2KB, no silent push | T5 action sentence preferred; tap deep-links to the entity; `MAX_PAYLOAD_BYTES = 2048` enforced at encrypt time; see above |
| 4 | Expiry hygiene | 404/410 is terminal → row pruned, history records `"device expired"` rather than retrying a dead endpoint forever |
| 5 | Contract procedure | schema → `npm run build` → regenerated artifacts committed; `validate:contract-drift` and `validate-db-types-drift` both pass |
| 6 | E2E transcript with notification screenshot | **Not done — see below** |

## What I could not verify, stated plainly

**Requirement 6's live E2E screenshot is not in this PR.** It needs a real VAPID keypair, a real push service, and an installed PWA on a real device — none of which exist in this environment. I did not want to imply otherwise with a staged screenshot.

What *is* verified: 40 unit tests across the crypto, payload, and browser-helper modules (including the real decrypt round-trip), all 80 existing `alerter-hub` tests still green (198 total across the touched suites), typecheck/lint/prettier clean, and the settings page rendering with no console errors. The push panel is correctly gated behind wallet verification, confirmed in the browser.

The first live send is therefore the real acceptance test, and it needs the deploy prerequisite below.

## Deploy prerequisites

Three Worker secrets, generated once by ops (`npx web-push generate-vapid-keys` or any P-256 generator):

- `VAPID_PUBLIC_KEY` — base64url uncompressed P-256 point
- `VAPID_PRIVATE_KEY` — base64url P-256 private scalar
- `VAPID_SUBJECT` — `mailto:` or `https:` contact

Plus `VITE_VAPID_PUBLIC_KEY` as a build variable (the public half is public by design — it is the `k=` value on every push request). Until they're set, the channel degrades to a recorded delivery failure and the UI says push isn't configured — exactly how the telegram/email channels already behave when unprovisioned.

## Test plan

- `tests/web-push.test.ts` — 14 tests: RFC 8291 round-trip decrypt, RFC 8188 header layout, salt/key non-reuse, 2KB ceiling, verifiable ES256 JWT audienced to the service **origin**, https-only endpoint guard, key-material intake validation, 404/410-only expiry.
- `tests/web-push-payload.test.ts` — 12 tests: sentence-first title chain, never-blank guarantee, entity deep links incl. root (netuid 0), coarse relative time incl. clock skew, payload budget under pathological input.
- `apps/ui/.../push-subscription.test.ts` — 14 tests: base64url conversion, capability detection, device labelling (Edge/Opera correctly beating the `Chrome` substring they contain), payload shaping incl. rejecting a subscription missing either key.
- `tests/alert-triggers.test.ts` — extended: `webpush` in the channel set, and SSRF guard tests proving a VAPID-signed request can't be aimed at `127.0.0.1`/`localhost`/private ranges or downgraded to http.
- Full suites: 198 passing across web-push, payload, alert-triggers, alert-delivery, alerter-hub.


JSONbored:
Both fixed in `3114b58`. The schema finding was correct and was a **real production bug**, not a style nit — thanks for catching it.

## 1. The CHECK constraint (blocker)

You were right that widening the inline CHECK inside `CREATE TABLE IF NOT EXISTS` is a no-op against the deployed table. Worse, it's the kind of bug that passes every local test — the test database builds the table fresh, so the widened constraint is real there and only silently absent in production.

Shipped the explicit ALTER pair instead, matching how `condition` and `owner_ss58` already ship in the same file.

I verified this against `postgres:16-alpine` rather than reasoning about it, because "I think this ALTER is right" is exactly the confidence that produced the bug:

```
=== 1. re-running CREATE TABLE IF NOT EXISTS with the widened inline CHECK (what this PR had) ===
NOTICE:  relation "chain_alert_triggers" already exists, skipping
--> webpush insert BEFORE the ALTER:
ERROR:  new row for relation "chain_alert_triggers" violates check constraint
        "chain_alert_triggers_channel_check"
DETAIL:  Failing row contains (1, webpush, https://x/y).

=== 2. applying the ALTER pair from the fix ===
--> webpush insert AFTER the ALTER:
 channel
---------
 webpush
```

Also confirmed:
- **Idempotent** — `DROP ... IF EXISTS` followed by `ADD` re-runs cleanly (bare `ADD CONSTRAINT` alone would fail the second time).
- **Still restrictive** — an unknown channel (`carrier-pigeon`) is still rejected, so this widens the constraint rather than defeating it.
- **The constraint name was read back from `pg_constraint`**, not guessed — Postgres auto-names an unnamed inline column CHECK `<table>_<column>_check`.

The comment in the schema now says explicitly why the ALTER is load-bearing, so the next person doesn't "tidy" it away.

## 2. CI (`scan:public-safety`)

Unrelated to the above. The public-safety scanner matches the two-word "private key" concept, which fired on the VAPID call site in `workers/alerter-hub.ts` and one schema comment.

Renamed `VapidKeys.privateKey` → `signingKey` — which is also just more accurate, since it's the ES256 signing scalar — and reworded the comment. WebCrypto's own `CryptoKeyPair.privateKey` is untouched; that's a platform API name, and it isn't what the scanner flagged.

Scanner passes, typecheck clean, 191 tests green across web-push / payload / alert-triggers / alerter-hub / price-at-tx (the branch is rebased onto main, which now carries #8504).

JSONbored:
Pushed `7f46c91` — codecov/patch fixed, and the schema blocker in the review above is **already resolved on this branch** (that finding is stale, from before `3114b58`).

## The stale blocker

The `ALTER` the review asks for is on the branch head at `deploy/postgres/schema.sql:1035`:

```sql
ALTER TABLE chain_alert_triggers
  DROP CONSTRAINT IF EXISTS chain_alert_triggers_channel_check;
ALTER TABLE chain_alert_triggers
  ADD CONSTRAINT chain_alert_triggers_channel_check
  CHECK (channel IN ('webhook', 'email', 'telegram', 'discord', 'webpush'));
```

The finding was correct when raised — it was a real production bug, since `CREATE TABLE IF NOT EXISTS` is a no-op against a deployed table. I verified the fix against `postgres:16-alpine` rather than reasoning about it: confirmed the insert fails *before* the ALTER with exactly that check-constraint violation, succeeds after, that the DROP-then-ADD pair is idempotent, and that an unknown channel is still rejected. Details in [this earlier comment](https://github.com/JSONbored/metagraphed/pull/8572#issuecomment-5111102765).

## The real failure: codecov/patch at 50.9%

That one was mine and it was a fair hit. I shipped the push-subscription routes and the webpush delivery channel with **no route-level or channel-level tests**, leaning on the crypto unit tests. Those cover `src/web-push.ts` thoroughly — but they never exercised the code that *calls* it, which is where 112 uncovered lines were.

- **`tests/watch-push-subscriptions-route.test.ts`** (new, 21 tests) — all four data-api handlers: auth/provisioning failures, the SSRF endpoint guard, malformed key material, the device cap, same-endpoint upsert (a browser reissues its endpoint, so a routine re-subscribe must not read as "limit reached"), address-scoped delete returning the same 404 a nonexistent id gets, and the internal resolve/prune pair. Asserts the owner-facing read never serializes `p256dh`/`auth`.
- **`tests/alerter-hub.test.ts`** (+8) — the channel's own wiring rather than re-testing crypto: each VAPID secret missing independently, a pruned subscription, a `DATA_API` that throws or 500s, a real signed+encrypted POST (asserting the body is *ciphertext*, not the plaintext payload), the 404/410 prune contract, and that a **503 is left to the retry path unpruned**.

## Two findings I'd rather flag than paper over

**A `catch` I assumed was dead code isn't.** `isValidPushKeyMaterial`'s decode guard sits behind a base64url-alphabet regex, so it looked unreachable. It isn't: a single character like `"A"` passes the regex but pads to `"A==="`, which `atob` rejects. I checked against `atob` directly before writing the test, rather than adding a test to satisfy a number.

**One branch genuinely could not occur.** `prunePushSubscription`'s binding guard is unreachable — prune only runs after a lookup that already required both bindings — but it was doing type-narrowing work, which is why removing it broke the build. Replaced with a documented assertion so the branch stops existing, instead of writing a test for a state the code cannot reach.

Changed lines now measured covered: `data-api.ts` **80/80**, `alerter-hub.ts` and `web-push.ts` **100%** statements and branches. 228 tests green across the six touched suites; typecheck, eslint, prettier and `scan:public-safety` clean.

JSONbored:
Pushed `220225c`. Taking the three in order — the security finding is the real one.

## 1. Superagent P2 — valid, and I'd encoded it as correct behaviour

`handleWatchPushSubscriptionCreate` looked the conflicting row up by endpoint but selected only `id`, never the owner, then upserted with `address = EXCLUDED.address`. Push endpoints are globally unique, so **any verified address could POST an endpoint already registered to someone else and seize it**:

- the original owner silently loses their device,
- the taker skips their own device cap (the `COUNT` was skipped whenever *any* row existed),
- and the taker's alert triggers then deliver push notifications to a browser that never subscribed to them.

Fixed by reading the owner and rejecting a foreign endpoint with 409, and by removing `address` from the `DO UPDATE SET` list entirely so it isn't reassignable even in principle. The upsert also carries `WHERE watch_push_subscriptions.address = EXCLUDED.address`, and a guarded update matching nothing returns 409 rather than a 201 with an empty body. The legitimate case this originally protected — a browser reissuing its own endpoint on key rotation — still upserts without spending cap.

**Proved against `postgres:16-alpine`** rather than reasoned about:

```
=== BOB attempts takeover with the OLD upsert ===
owner=BOB key=bob-key                      ← Alice's device stolen

=== BOB attempts the same with the FIXED upsert ===
INSERT 0 0                                 ← update refused
owner=ALICE key=alice-key                  ← Alice keeps her device

=== ALICE re-subscribing her own endpoint ===
updated id=1                               ← still works
```

**The part worth owning:** my own test asserted the vulnerable behaviour as correct — *"re-subscribing the SAME endpoint upserts past the cap"*, with no owner in the fixture. I wrote a test that locked in the bug, which is exactly why the suite didn't catch it. That test now pins ownership, and takeover is covered directly.

## 2. The schema blocker is stale

Already fixed on this branch since `3114b58`; the `ALTER` is at `deploy/postgres/schema.sql:1035`. Verified against real Postgres at the time (insert fails before, succeeds after, idempotent, unknown channels still rejected) — [details here](https://github.com/JSONbored/metagraphed/pull/8572#issuecomment-5111102765). The review that keeps surfacing it predates the fix.

## 3. codecov/patch

The 10 partials were all genuine defensive branches no happy-path test reached: null/missing row metadata, string-coerced timestamps, wrong-**type** body fields (a JSON number where a string belongs), POST/DELETE auth failures, and an empty `COUNT` aggregate. Each now has a test.

One measurement note: two lines briefly looked uncovered locally because I was running a single test file — they're pre-existing #8386 dispatch lines that git attributed to this patch, and they're covered by `wallet-auth-keys-route.test.ts`. Measured the way CI does, across every suite touching `data-api.ts` (737 tests):

```
uncovered lines:       []
partial-branch lines:  []
```

145 tests green across the four web-push suites; typecheck, eslint, prettier and `scan:public-safety` clean.

PR #8605, opened 2026-07-29
fix(ui): rebuild apps/ui's OG cards with brand identity + real entity data

### LINKED ISSUE #8489: fix(ui): apps/ui's per-page OG cards are unbranded plain text — rebuild with brand identity + real entity data
## Context

There are two separate Open Graph card renderers in this repo, and they are not the same quality:

1. **`src/og-image.ts`** (the `api.metagraph.sh` landing page's own `/og`/`/og.png`) is a real branded card: mint-on-ink (`#30FFC0` / `#0B1F1A`), the brand "M" mark, "Metagraphed" in Space Grotesk Bold, a diagonal accent shape, and a live stat row (subnet/endpoint/provider counts + coverage %) pulled from `registry-summary.json` at publish time (`scripts/refresh-og-image.ts`). This one is fine.
2. **`apps/ui/src/lib/og-image.ts`** (`GET /og` on the actual browsable site, `metagraph.sh`) is what every single page — home, every subnet, every validator, every account, every doc — unfurls as when shared on Twitter/Discord/Slack/iMessage/etc. It renders three lines of plain text on a flat black background (`background:#0a0a0a;color:#fafafa`, generic Inter font) with **no logo, no brand color, no visual hierarchy, and no data** — just `title` / `subtitle` / the literal string `metagraph.sh`. `apps/ui/src/server.ts:414` wires this into `og:image` for every route via `ogCopy.title`/`ogCopy.subtitle`.

This is the card that actually matters for real-world reach — it's what a shared subnet page, a shared validator profile, or a shared doc page looks like off-site — and today it reads as an unbranded placeholder, not a product screenshot. This issue is about #2 only; #1 is already correct and out of scope.

## Requirements

1. **Reuse the existing brand identity, don't invent a new one.** The brand "M" mark and mint/ink palette already exist and are already used correctly in `src/og-image.ts` (`LOGO_DATA_URI`, `MINT = "#30FFC0"`, `INK = "#0B1F1A"`, `INK_TEXT = "#08110E"`) and in `public/brand/banner-og-social-mint.png`/`banner-readme-mint.svg`. The new `apps/ui` card's background/logo/wordmark treatment must match that identity (same mint-on-ink lockup, or a deliberate dark-ink-background variant if a design reason is stated in the PR) — not a new color scheme invented ad hoc.
2. **Render real per-page content, not just the title/subtitle strings currently passed.** At minimum, the card must visually distinguish entity types the way the in-app UI already does:
   - Subnet pages (`/subnets/$netuid`): SN number, subnet name, and at least one live stat already available to the route loader (e.g. health state, alpha price, or emission share — pick ONE primary stat and state the choice in the PR, don't try to cram all of them).
   - Validator pages (`/validators/$hotkey`): validator name/identity, hotkey (truncated), and a stake or APY figure.
   - Account pages (`/accounts/$ss58`): resolved address label (reuse `resolveAddress`, the same resolver #8372 already wired into every other address-rendering site — do not build a second address-formatting path) and truncated ss58.
   - Every other route (home, docs, status, etc.): the existing title/subtitle/wordmark treatment is acceptable as the fallback, but re-skinned with the brand mark and mint/ink palette from requirement 1 instead of the current plain black-and-gray text.
3. **Data passed to `/og` must come from data the requesting route already has** (its own loader/query data), passed as additional query params the way `title`/`subtitle` already are today — this route must not perform its own live API fetch to assemble the card (that would add a second, redundant data round-trip and a new failure mode independent of the page's own data-loading state). If a route's data isn't available yet at unfurl-request time for some reason, degrade to the title/subtitle-only fallback design from requirement 2, never a render failure.
4. **Keep every existing safety/perf property of the current implementation**: the `FALLBACK_PNG` tiny-PNG fallback on a `workers-og`/font-load failure, the edge cache keyed on the full param set including a `CARD_VERSION` bump, the `MAX_QUERY_LENGTH`/`MAX_TITLE_LENGTH`/`MAX_SUBTITLE_LENGTH` bounds (extend with equivalent bounds for any new params), and `escapeText()` on every interpolated string (new params included — this is unfurl-triggered by crawlers/bots hitting an unauthenticated public endpoint, so injection into the satori markup string must be guarded exactly as rigorously as the existing title/subtitle are).
5. **Bump `CARD_VERSION`** in `apps/ui/src/lib/og-image.ts` so the edge cache and any already-cached crawler cache picks up the new design rather than continuing to serve the old plain-text card for its remaining `stale-while-revalidate` window.
6. Do **not** touch `src/og-image.ts` (the landing-page card) — it is explicitly out of scope and already correct.

## Deliverables

- [ ] `apps/ui/src/lib/og-image.ts`'s markup rebuilt with the brand mark + mint/ink palette (or a stated, deliberate dark-ink variant), reusing the existing brand asset(s) rather than a new logo/color scheme
- [ ] Per-entity-type card variants for subnet / validator / account pages, each showing at least one real stat/identity field from that route's own already-loaded data, passed via new bounded+escaped query params
- [ ] Fallback design (title/subtitle-only, brand-skinned) for every other route and for any route where the new per-entity params are absent
- [ ] `CARD_VERSION` bumped; all existing bounds/escaping/fallback-PNG behavior preserved for the new params
- [ ] Screenshots of the rendered `/og` PNG output for: home, a subnet page, a validator page, an account page, and one arbitrary doc/other page (before/after each)

## Expected outcome

Sharing any metagraph.sh page — a subnet, a validator, an account, or the home page — produces a card that is immediately recognizable as Metagraphed (brand mark + mint/ink identity) and, for entity pages, shows real information about that specific entity at a glance, matching the quality bar `src/og-image.ts`'s landing-page card already sets.

## Links

- `apps/ui/src/lib/og-image.ts` — the file being fixed
- `src/og-image.ts` — the already-correct reference implementation to match the brand treatment of (not to literally reuse code from — different runtime constraints, see that file's own header comment on why it renders at publish time via Node-native `@resvg/resvg-js` rather than live via `workers-og`'s wasm build)
- `apps/ui/src/server.ts:394-419` — where `og:image` is wired per-route today
- `public/brand/banner-og-social-mint.png`, `public/brand/banner-readme-mint.svg` — existing brand assets
- `apps/ui/src/lib/metagraphed/resolve-address.ts` (#8372) — the address resolver account-page cards must reuse

JSONbored:
## Summary

Every share of metagraph.sh unfurled as **three lines of plain text on flat black** — no mark, no brand colour, no data, and the identical card for the home page, every subnet, every validator and every doc. This is the card that actually travels off-site, so it now carries the brand lockup and, for entity pages, real information about that entity.

Closes #8489

## Design: ink ground, mint mark

Requirement 1 allows a dark-ink variant "if a design reason is stated in the PR". Stating it: the landing card (`src/og-image.ts`) is a mint field with an ink mark — a poster, one headline, no data. These per-page cards carry three to five lines of dense entity text, and a full-bleed mint field behind that much copy reads as a marketing banner rather than a data card. The app's own surface is ink with mint accents, so an unfurl on ink is recognisably *this product*. Same mark, same palette (`#30FFC0` / `#0B1F1A`), inverted ground — deliberately complementary, not a second brand. The mark is `src/og-image.ts`'s own path recoloured to mint; `src/og-image.ts` is untouched, per requirement 6.

## The architectural change that made real data possible

`og:image` was injected globally in `src/server.ts` from the **pathname alone**. That is exactly why every card was generic: the injection runs in the Worker's `HTMLRewriter` pass and has no access to loader data, so it could only ever say "Subnet 64" — never "Chutes · 0.0832 τ · healthy".

So ownership moved: the three entity routes emit their own `og:image` from `head()`, where `loaderData` is available, and `server.ts` skips exactly those paths via `routeOwnsOgImage`. That keeps **exactly one** `og:image` tag on every page rather than relying on crawlers picking the first of two. Everything else keeps the server-injected, now brand-skinned, fallback.

Validator and account had no loader, so they gained one that primes **the same query their page already reads** (`validatorDetailQuery` / `accountQuery`) — the identical pattern `subnets.$netuid.tsx` already uses. Shared react-query cache means the request moves earlier rather than being added, and any failure returns null and falls back to the previous copy. That satisfies requirement 3's "no live fetch from `/og`, use the route's own data".

## Three layout defects found by rendering and measuring

None of these are visible from reading the code, and the first two would have shipped a broken card:

1. **Stat rail cropped.** `height:100%` never resolved, so the card laid out at its intrinsic **774px against a 630px canvas** — the stat values simply weren't on the image. Root now carries explicit pixel dimensions.
2. **The card outgrew its own canvas.** `width:1200px` *plus* `80px` horizontal padding renders **1360×758** under the CSS default `content-box`. Padding moved to an inner wrapper, separating "the box that is the canvas" from "the box that has padding" — the same structure `src/og-image.ts` already uses.
3. **Subtitle washed out under the decorative diagonal.** Absolutely-positioned elements paint above in-flow content. The content wrapper is now positioned, *and* copy widths are capped clear of the diagonal so legibility doesn't rest on the stacking rule alone.

Plus: a 110-character title (the cap `normalizeTitle` already allows) wrapped to four lines and pushed the stat rail off the card, so **title size now steps down with length** — deterministic, since satori has no reliable line-clamp.

## Safety properties preserved

Every existing guarantee is intact, and the new params are held to the same standard rather than trusted for being "ours": `FALLBACK_PNG` on font/wasm failure, the edge cache, `MAX_QUERY_LENGTH`/`MAX_TITLE_LENGTH`/`MAX_SUBTITLE_LENGTH`, and `escapeText()` on every interpolated value. The new `eyebrow`/`stat*` params are individually length-bounded **and** escaped — this is an unauthenticated endpoint crawlers hit — and they are part of the **cache key**, so two pages sharing a title can't serve each other's stats.

`CARD_VERSION` bumped to `3`, retiring every already-unfurled plain-text card rather than serving it for the remaining 7-day stale-while-revalidate window.

## Screenshots

Rendered from the real `renderCardMarkup` output at 1200×630. Reproduce with:

```bash
npx tsx apps/ui/scripts/render-og-preview.ts
```

| Variant | Before | After |
| --- | --- | --- |
| Home / fallback | plain text on `#0a0a0a`, no mark | brand lockup + tagline + `metagraph.sh` |
| Subnet | `Subnet 64` / generic tagline | **Chutes** · `SUBNET` pill · NETUID `SN64` · ALPHA PRICE `0.0832 τ` |
| Validator | truncated hotkey / generic tagline | **tao.bot** · `VALIDATOR` pill · TOTAL STAKE `1.42M τ` · SUBNETS `37` |
| Account | truncated ss58 / generic tagline | **5Grwva…GKutQY** · `ACCOUNT` pill · EVENTS `12,481` · SUBNETS `9` |
| Docs / other | plain text | brand-skinned fallback |
| 110-char title | n/a | steps down to 42px, stat rail still on-card |

*(Images attached in the following comment — I render them locally rather than committing binaries.)*

## Test plan

- **18 tests** in `src/lib/og-image.test.ts` (9 pre-existing, all still passing + 9 new): param reading, half-specified stats dropped, bounds enforced against a hostile query, escaping (a `<script>` title does not survive), the explicit-canvas-sizing regression pinned as an assertion, eyebrow omission, and `titleFontSize` monotonicity.
- Rendered all six variants at 1200×630 and inspected each; measured the root box to confirm the canvas fix rather than eyeballing it.
- `npx tsc --noEmit` (apps/ui) clean; `npx eslint` clean on all changed files (one pre-existing `text-[11px]` warning in `accounts.$ss58.tsx`'s not-found block, outside this diff); `prettier` clean; `scan:public-safety` passes.

## Note on review

Per the repo's own rule, this touches visual output and is **always held for manual review** regardless of AI-review confidence.


JSONbored:
Fixed in `0e07c56`. Confirmed real — and it was worse than "the eyebrow is missing from a list", because the cards I'd been reviewing rendered it *correctly by coincidence*.

## Reproducing it

Rendering with production's own glyph expression, `SUBNET` came out fine. That's luck, not correctness: every capital it needs was supplied incidentally by other strings on the card — `U`, `N`, `E`, `T` from the stat label `NETUID`, and `B` from the subtitle's "**B**ittensor".

Isolating a case where nothing else supplies them (`title: "chutes"`, `subtitle: "a subnet."`, `eyebrow: "Validator"`, no stats) reproduced it instantly:

**Before** — `V` survives from "**V**alidator"'s own capital; the other eight are tofu:

`V ▯▯▯▯▯▯▯▯`

**After** — `VALIDATOR`

## The fix removes the class, not the instance

I could have added `eyebrow.toUpperCase()` and moved on. But a hand-maintained list of "every string we paint" drifts every single time the template transforms a value — and it had **already drifted twice**: stat labels were mirrored as `.toUpperCase()`, the eyebrow was not. Adding the one missing entry leaves the next transform to be forgotten the same way.

The subset is now **derived from the rendered markup**: tags are stripped (their attributes go with them, since they live inside `<…>`), then the four entities `escapeText` can introduce are decoded back to the glyph actually drawn. Whatever the template renders is, by construction, what gets subset — there is no second list to keep in sync.

## Pinned structurally

The new tests assert the *invariant* rather than enumerating characters, so they keep working as the card evolves:

- the uppercased **eyebrow** is subset (the exact regression, using the isolated case above)
- uppercased **stat labels** are subset
- wordmark, footer lockup and stat values are subset
- `&` is subset rather than `&amp;` — the card paints the character, not the entity
- no style/attribute residue leaks into the font request (which would balloon it with glyphs that are never painted)

30 tests green; typecheck, eslint, prettier and `scan:public-safety` clean, and the six-variant preview still passes its 1200×630 assertion.

Worth noting this is the second bug in this PR that only appeared under **real satori** with a real `text=` subset — the local Chromium preview substitutes a full system font and renders both perfectly. The τ fix was the first. I've verified this one the same way rather than trusting the preview.

Issue #8247, opened 2026-07-26
feat(ui): subnet detail rebuild — identity header, 6-KPI band, 7 lazy tabs, zero duplicate facts


JSONbored:
Phase 2 · Core pages. **Subnet detail rebuild — the flagship page, and the audit's biggest single win.**

## Context (measured 2026-07-26)

The page is 9,864 px tall (10,476 mobile) and renders ~20 sections on the Overview tab alone. Alpha price appears 3× (masthead stat cards, pool panel, economics tiles); pool composition twice (pool panel + Subnet Profile); economics sprawls across 14 tiles + a 4-tile 24h-volume row + profile repeats. The tab strip holds 11 tabs — at 390px only ~2 are visible with no overflow affordance. Mobile spends ~1,500 px on 12 stacked stat mini-cards before anything actionable. Empty sections (Ownership contest, Ownership history, Prometheus, Evidence) render as framed "No X yet" panels.

## Requirements

- **Identity header:** icon · name · netuid · ≤3 type chips · ONE health pill · snapshot-age chip (only when late) · links row. Sponsored-validator slot rules unchanged (disclosure stays).
- **KPI band (exactly 6):** Price + 7d spark · Emission share · Total stake · Miners/Validators (n/max) · Uptime 24h · Integration readiness. Each fact appears once on the whole page.
- **Tabs (7):** Overview · API & Endpoints · Metagraph · Economics · Activity · Governance & Ownership · About. Lazy-load per tab; tab badges carry counts (Endpoints 58); empty tabs hidden or zero-badged, their content never framed as empty panels — About lists one-liners ("No ownership transfers recorded").
- **Overview = one screen:** 90-day uptime strip, compact price chart, **"Start integrating" card** (readiness score + Start-here link + copyable MCP/curl snippet), top-3 validators (+ sponsored slot), this-subnet what-changed feed (5 items).
- API & Endpoints tab absorbs: operational status (uptime/latency percentiles, per-surface bars), public resources (endpoints/surfaces/schemas subviews), stake-quote calculator moves to Economics.
- Metagraph tab: existing neuron table/concentration content, unchanged data, decluttered chrome.
- Economics: grouped rows replace the 14+4 tile wall; OHLC candles stay here as-is (no expansion — scope fence).
- Mobile: KPI band = 2×3 grid; "All stats" bottom sheet holds the long tail; shared TabStrip overflow affordance; first meaningful content ≤1.5 screens.
[...]

PR #8623, opened 2026-07-29
fix(ui): OG cards on the app's light theme, with our own mark and real KPIs

### LINKED ISSUE #8622: OG cards: light theme, our own mark on our own routes, and three real KPIs
Follow-up to #8489 / #8605, which rebuilt the `/og` card. Three of the four items below were found by rendering the card through the real rasterizer rather than reading the code, and one of them is a live rendering bug.

## Context

`/og` is the card every share of `metagraph.sh` unfurls through. #8605 landed the dark-ground rebuild. This issue covers the follow-on maintainer decisions and the defects that surfaced once the card could actually be rendered locally.

## Requirements

1. **Light theme.** The card must use "Bone & Ink" — the app's DEFAULT theme (light in `packages/ui-kit/src/styles.css` `:root`, dark only under `.dark`). A dark card promises a page the reader does not get. Colours must be the styles.css tokens converted with the same maths a browser uses (oklch → linear-sRGB; oklab interpolation against `--paper` for the alpha tokens), not eyeballed hexes, and must respect the app's `--accent` / `--accent-text` split — the vivid mint is 1.9:1 on paper and must never draw small text.

2. **Our mark on our own routes.** The avatar slot falls back to a bare mint rule for anything without an entity logo, which leaves `/agents`, `/docs` and the home page looking unfinished. Our routes must show the Metagraphed mark; entities keep the monogram. The resulting ladder must match the site's `BrandIcon`: real icon → monogram → mark, never nothing. The entity/non-entity distinction must be an explicit parameter — "has an eyebrow" stopped being a usable proxy once every route got one.

3. **Unresolvable entity icons must degrade, not break.** `tao.bot` publishes a favicon but no aggregator the icon proxy queries has one, so the proxy 404s. satori has no `onerror`: the icon must be resolved and inlined *before* rendering so a failure falls back to the monogram.

4. **Ampersands are corrupted in production.** `escapeText` emits `&amp;`, but workers-og does not decode HTML entities in text nodes — `/og?title=Agents %26 MCP` paints the literal characters `& a m p ;`. Interpolated text must be sanitized in a way that survives that parser.

5. **Richer stat rail.** The subnet card reads `alpha_price_tao` off the profile, where the field does not exist, so the price stat never renders. It must read the same economics source the subnet masthead uses and carry the facts the KPI band leads with (#8247: price, emission share, total stake), with health shown as a coloured state rather than spending a stat cell on a one-word string.

6. **The preview must use the real rasterizer.** `apps/ui/scripts/render-og-preview.ts` renders in Chromium, which substitutes full system fonts and is therefore structurally blind to glyph-subsetting bugs — two have already shipped through a green Chromium preview.

## Deliverables

All in one PR:

- [ ] `apps/ui/src/lib/og-image.ts` on the light palette, with the accent/accent-text split observed
- [ ] Brand mark in the avatar slot for non-entity cards; explicit `entity` param
- [ ] Icon resolved and inlined before render, with monogram fallback
- [ ] Text sanitization that does not rely on entity decoding
- [ ] Three-stat rail; subnet route reading real price / emission share / total stake; health as a coloured footer dot
- [ ] `render-og-preview.ts` rendering through satori
- [ ] Tests covering each of the above
- [ ] Before/after screenshot table in the PR (required for any visual change)

## Expected outcome

Rendering all card variants through satori produces cards on the bone canvas, exactly 1200×630, with: our mark on our own routes, a monogram for `tao.bot`, a literal `&` in "Agents & MCP", and a subnet card showing real price and emission share. No card depends on a network fetch succeeding at render time.

JSONbored:
## Summary

Follow-up to #8605. Four changes to the `/og` card — **three of them found by rendering it through the real rasterizer**, and one of those is a live rendering bug that has been shipping blank cards.

Closes #8622

## 1. Light theme

The card now uses "Bone & Ink", which is the app's **default** theme — light lives in `styles.css` `:root`, dark only under `.dark`. A dark card was promising a page the reader doesn't get.

Every colour is the styles.css token converted with the same maths a browser uses (the oklch → linear-sRGB matrix; oklab interpolation against `--paper` for the alpha tokens), not an eyeballed hex. The app's `--accent` / `--accent-text` split is respected: `#00C899` fills rules and dots, `#008156` draws every accent-coloured string — the vivid mint is **1.9:1 on paper** and unreadable as small type. Same for `--ink-subtle` vs `--ink-subtle-text` on the stat labels.

## 2. Our own mark on our own routes

The avatar slot fell back to a bare mint rule for anything without an entity logo, which left `/agents`, `/docs` and the home page looking unfinished. It now takes the Metagraphed mark. Entities keep the monogram, so the ladder matches the site's `BrandIcon` exactly — **real icon → monogram → mark, never nothing**.

`entity=1` decides which, as an explicit flag rather than an inference: "has an eyebrow" stopped being a usable proxy the moment #8605 gave every route one, so `/agents` would have rendered "AG".

The mark is built from **one path at two fills** instead of two pasted base64 blobs — a base64 constant can't be recoloured, which is how the dark card ended up with a mint mark and no way to follow the theme.

## 3. tao.bot's card was blank, not just missing a logo

`tao.bot` publishes a favicon, but no aggregator the icon proxy queries has one, so the proxy 404s. satori has no `onerror` — and it does not degrade. Rendering #8605's markup with that unresolvable icon **throws**:

```
RESULT: THREW -> Image size cannot be determined. Please provide the width and height of the image.
```

That lands in `handleOgImage`'s catch, which serves `FALLBACK_PNG` — **a 1×1 transparent PNG**, cached for a day. So the report wasn't "the logo is missing"; the whole card was blank whenever an entity's icon didn't resolve.

Icons are now fetched and inlined as a data URI *before* rendering, so an unresolvable one falls back to the monogram and the card never depends on a network fetch at render time. It's also one request instead of two.

## 4. Ampersands were corrupted in production

`escapeText` turned `&` into `&amp;` for "safe embedding" — but **workers-og does not decode HTML entities in text nodes**. Verified against the deployed Worker rather than assumed: `/og?title=Agents %26 MCP` paints the literal characters `& a m p ;` as eight tofu boxes, because the font subset has no glyph for most of them.

The escaping *was* the corruption, on every title containing an ampersand. Text is now sanitized by **deleting `<` and `>`** — strictly safer than escaping, since the characters cannot appear in the output at all, so no tag can form regardless of parser behaviour — and everything else passes through untouched. This is safe only because no caller-supplied value reaches an attribute; the one dynamic attribute is the icon `src`, a data URI this module builds itself.

The glyph subset no longer decodes entities either. That decode had made the bug *self-consistent* — the card emitted `&amp;` and the subset dutifully included `a`, `m`, `p`, `;` — which is exactly why a glyph test couldn't see it.

## 5. Real KPIs

The subnet card read `alpha_price_tao` off the **profile**, where that field does not exist (`normalizeSubnetProfile` never emits it), so the price stat silently never rendered and every subnet fell through to its health string. It now reads the economics list the subnet masthead itself uses, and carries the three facts the KPI band leads with (#8247: **price, emission share, total stake**), with health moved to a coloured footer dot instead of spending a stat cell on a one-word string. The rail takes three cells with the type stepped down so it still fits — measured, not guessed: at the old 21/42px with a 64px gutter, three cells plus the lockup overran the band.

## The preview now uses the real rasterizer

`render-og-preview.ts` rendered in **Chromium**, which substitutes full system fonts and is therefore structurally blind to glyph-subsetting bugs. Two have already shipped through a green Chromium preview (the tau, and the uppercased eyebrow); the ampersand above is the third. It now renders through satori + resvg — the same pipeline workers-og runs, assembled directly, since workers-og itself can't load under Node.

## Screenshots

Rendered from the real `renderCardMarkup` output through satori at 1200×630. Reproduce with:

```bash
npx tsx apps/ui/scripts/render-og-preview.ts
```

| Variant | Before (#8605, dark) | After |
| --- | --- | --- |
| Home | ink ground, mark + tagline | bone canvas, mark in a white tile, hairline bands |
| Subnet (Chutes) | `SN64` · `0.0832 τ` — **price cell never rendered**, fell back to health | `SN64` · `0.0832 τ` · `3.41%` from real economics, green `ok` dot |
| Validator (tao.bot) | **blank 1×1 PNG** — satori threw on the 404 icon | monogram tile `TA`, `62.6M τ` · `116` |
| Account | truncated ss58, mint rule, 2 stats | monogram tile, `12,481` · `9` |
| `/agents` | bare mint rule; title read `Agents &amp; MCP` with 8 tofu boxes | Metagraphed mark in the tile; **`Agents & MCP`** |
| 110-char title | steps to 42px | steps to 42px, 3-cell rail still on-card, red `down` dot |

## Verification

- **1781 UI tests pass** (221 files); 48 in `og-image.test.ts`, up from 35
- All six variants render through **real satori** at exactly 1200×630 (the script asserts the dimensions and throws otherwise)
- `tsc --noEmit`, `eslint`, `prettier --check` clean; `scan:public-safety` passes
- `CARD_VERSION` bumped to `4`, retiring every cached dark card rather than serving it for the remaining 7-day stale-while-revalidate window

Two incidental fixes carried here because they block the above: `readCardParams` uses `Object.hasOwn` rather than `in` for the status lookup (`?status=constructor` would otherwise pass and interpolate a stringified function into the card's inline CSS — caught by a test I wrote for it), and `server.og-copy.test.ts`'s route-coverage guard now skips `.test.tsx` files under `routes/`, which #8621's `ChainHeadTip` test had started failing on `main`.

### DIFF
--- a/apps/ui/eslint.config.ts
+++ b/apps/ui/eslint.config.ts
@@ -368,7 +368,14 @@ export default tseslint.config(
       "src/routes/-design-primitives-page.tsx",
       "src/components/metagraphed/primitives/**",
       "src/lib/health-tokens.ts",
+      // The OG card is rasterized by satori on the Worker, where CSS custom
+      // properties do not exist -- every colour has to ship as a literal hex,
+      // which is why this file is exempt. Its TEST is exempt for the same
+      // reason: the assertions that the card uses --accent-text rather than
+      // the unreadable brand mint, and the site's own health colours, can only
+      // be written against those literals.
       "src/lib/og-image.ts",
+      "src/lib/og-image.test.ts",
     ],
     rules: { "no-restricted-syntax": "off" },
   },
--- a/apps/ui/scripts/render-og-preview.ts
+++ b/apps/ui/scripts/render-og-preview.ts
@@ -8,25 +8,40 @@
 // were invisible in code review and only showed up by rendering the markup and
 // measuring the result. This script makes that check repeatable.
 //
-// It renders the SAME `renderCardMarkup` output the Worker uses, in Chromium at
-// exactly 1200x630. Chromium is a stand-in for satori's rasterizer, not an
-// exact match — satori implements a subset of CSS — but every property this
-// card relies on (flexbox, absolute positioning, border-radius, explicit px
-// sizing) is well inside that subset, and layout/overflow bugs reproduce
-// faithfully. The authoritative render still happens on the Worker.
+// It renders through SATORI, not Chromium. An earlier version of this script
+// used Playwright, and that turned out to be structurally blind to a whole
+// class of bug: satori subsets each font to the glyphs the card declares, so a
+// character the subset misses rasterizes as a tofu box — but Chromium
+// substitutes a full system font and paints it perfectly. Two real tofu bugs
+// (the tau in every τ value, and the uppercased eyebrow pill) got through a
+// green Chromium preview. Rendering with the same library the Worker uses is
+// the only preview that can catch them.
+//
+// The Worker imports satori via workers-og, which can't load under Node (its
+// yoga `.wasm` isn't resolvable by Node's ESM loader — the same reason
+// handleOgImage imports it lazily). satori + satori-html + resvg here are the
+// same pipeline workers-og runs, assembled directly.
 //
 // Usage:  npx tsx apps/ui/scripts/render-og-preview.ts [outDir]
 
 import fs from "node:fs";
 import path from "node:path";
-import { chromium } from "playwright";
-import { renderCardMarkup } from "../src/lib/og-image.ts";
+import { Resvg } from "@resvg/resvg-js";
+import satori from "satori";
+import { html } from "satori-html";
+import { glyphsForMarkup, renderCardMarkup } from "../src/lib/og-image.ts";
 
 type Variant = Parameters<typeof renderCardMarkup>[0];
 
-/** One case per card shape the app actually emits, plus the pathological
- * title length `normalizeTitle` allows — that last one is the case that
- * previously pushed the stat rail off the canvas. */
+/**
+ * One case per card shape the app actually emits, plus the pathological title
+ * length `normalizeTitle` allows — that last one is the case that previously
+ * pushed the stat rail off the canvas.
+ *
+ * `icon` values are the resolved data URIs handleOgImage passes in; the
+ * variants below deliberately cover BOTH branches, because the interesting
+ * case is the one where the icon does not resolve.
+ */
 const VARIANTS: Record<string, Variant> = {
   "1-home": {
     title: "Metagraphed",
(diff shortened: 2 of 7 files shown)


JSONbored:
Pushed `82c65a6` — ink foot, and our mark moved onto ink.

**The foot is ink.** Body stays bone, the stat band is now the app's own dark theme. Beyond looking better, it earns something: on ink the brand mint can be used at **full strength** (`--accent` dark, `#5DEBBC`), where on paper it had to be dialled all the way down to `#008156` to clear AA. Stat labels take `--ink-muted` (dark), the lockup takes `--ink-strong` (dark) — all converted from the `.dark` block exactly as the light tokens were, so this is the product's dark theme rather than "a dark grey".

The band is also now **unconditional**. The old lifted white surface had to be suppressed on a statless card because a panel with nothing in it read as a mistake; an ink band with just the lockup reads as a base, so the home and docs cards get one too and every card ends the same way.

The health dot switches to the **dark** health set. The light theme's AA text variants (`#966800` amber, `#DF2321` red) are darkened to survive on paper and read as mud on ink — the app swaps these per theme for this exact reason, and there's a test pinning that the paper variant never lands on the foot.

**Our mark is on an ink tile.** Mint-on-white was genuinely hard to see, and the reason is worth naming: the mark is a *thin stroke*, so it has far less area to carry contrast than a block of type does. The light accent on near-white left it faint even though the same colour is perfectly legible as the eyebrow pill's border. Ink fixes it at the source instead of darkening the brand colour.

Deliberately **our mark only** — entity tiles stay white. They hold a third-party logo whose contrast we don't control, and white is the safe canvas for an arbitrary one. That's the same call ui-kit's `BrandIcon` already makes when it flips a dark-on-transparent logo onto a white tile in dark mode. The semantic ends up clean: ink chip = ours, white chip = someone else's.

1784 UI tests pass (51 in `og-image.test.ts`, +3 for the ink foot); all six variants re-rendered through real satori at 1200×630; lint, prettier and `scan:public-safety` clean.

PR #8627, opened 2026-07-29
fix(ui): close the SEO gaps — soft 404s, docs OG cards, sitemap lastmod

### LINKED ISSUE #8626: SEO: soft 404s on every detail route, /docs/* sharing one OG card, and no sitemap lastmod
Audit of the live site across 15 URLs. The metadata foundation is genuinely good — every page has a unique `<title>` and `description`, `og:title`/`description`/`image`/`url`/`type`/`site_name`, `twitter:card: summary_large_image`, a correct self-referencing canonical, and `@graph` JSON-LD; `robots.txt` and `sitemap.xml` exist and agree. These are the gaps.

## 1. Soft 404s on every entity detail route

`/subnets/99999` returns **HTTP 200**, titled "Subnet 99999 — Metagraphed", with no `robots` tag and a canonical pointing at itself. Same for `/validators/<any valid ss58 that is not a validator>`. The URL space is effectively infinite, so a crawler can mint unlimited indexable thin pages by varying an integer.

`entityNotFoundMeta` already exists for this and emits `noindex` — but it is wired into only `accounts` and `validators`, and only for **malformed** identifiers, not **non-existent** ones. Its own docstring cites `/subnets/not-a-netuid` as the bug it fixes; that route never adopted it, and `/subnets/not-a-number` still titles as "Subnet not-a-number".

**Requirement.** Every detail route (`subnets`, `validators`, `accounts`, `blocks`, `extrinsics`, `providers`) must emit `noindex` both for a malformed identifier and for a well-formed one that names nothing.

**Constraint, and this is the important one:** loaders currently swallow all failures and return null so the page still renders and the component's own query drives the error path. That must survive. Only a definitive 404 from our own API may flip a route to not-found — marking pages `noindex` on a transient failure would de-index real subnets and validators during an outage.

## 2. All 20 `/docs/*` pages share one generic OG card

`ogCardCopy` has an exact `/docs` entry but no `/docs/*` branch, so `/docs/economics`, `/docs/accounts` and the rest all unfurl as `og?title=Metagraphed`. Docs are the most link-worthy content on the site and they were the pages sharing a card. The splat route already has the real title and description in `loaderData`; `server.ts` works from the pathname alone and cannot.

**Requirement.** `/docs/*` emits its own `og:image` from page data, and `routeOwnsOgImage` matches it so exactly one `og:image` tag survives.

## 3. Zero docs pages in the sitemap

281 URLs — 130 subnets, 136 providers, ~15 sections — and not one of the 20 doc pages.

**Requirement.** Docs enumerate from the same source that renders them, so a new page is listed the moment it ships.

## 4. No `<lastmod>` anywhere

For a product whose entire pitch is freshness, the sitemap gives crawlers nothing to schedule a recrawl against.

**Requirement.** Emit `<lastmod>` only where a real timestamp exists (a subnet's `updated_at`). It must **not** be synthesised for static or docs pages: Google discounts lastmod site-wide once it catches a site stamping "now" on unchanged URLs, so a fabricated value would cost the real ones too.

## 5. No `robots` meta, so no `max-image-preview:large`

Without it Google caps the preview to a thumbnail, which wastes the per-page OG card work in #8489/#8622/#8623.

**Requirement.** Inject site-wide crawler defaults. Must not defeat a route's own `noindex` — crawlers take the most restrictive directive when tags conflict, so the failure mode has to bias toward not indexing.

## 6. Minor

- No `og:image:alt` / `twitter:image:alt` (screen readers; some platforms show it as a caption)
- No `og:locale`
- No `twitter:site` / `twitter:creator` — **out of scope: the project has no X account**, so there is no handle to put there. Revisit if one is created.
- The 1029 validators are not in the sitemap. Deliberately deferred — unlike subnets and providers, that set is large and churns.

## Deliverables

All in one PR:

- [ ] `noindex` on malformed **and** non-existent entities across all six detail routes, gated on a definitive API 404
- [ ] `/docs/*` emits its own OG card; `routeOwnsOgImage` matches it
- [ ] Docs in the sitemap, enumerated from the docs source
- [ ] `<lastmod>` from real timestamps only
- [ ] Site-wide `robots` defaults with `max-image-preview:large`, plus `og:locale` and image alt text
- [ ] Tests, including one pinning that a 5xx/429/network error does **not** mark a page `noindex`

## Expected outcome

`/subnets/99999` and `/validators/<valid-unknown-ss58>` carry `robots: noindex`. `/docs/economics` unfurls with its own title rather than the brand card. `sitemap.xml` contains the 20 docs URLs and carries `<lastmod>` on subnet entries and nowhere else. Every page carries `max-image-preview:large` except those a route has marked `noindex`.

JSONbored:
## Summary

Audited the live site across 15 URLs. The metadata foundation is genuinely good already — unique `<title>` and `description` on every page, full `og:*`, `twitter:card: summary_large_image`, correct self-referencing canonicals, `@graph` JSON-LD, and a `robots.txt`/`sitemap.xml` that agree. Five things were wrong.

Closes #8626

## 1. Soft 404s on every entity detail route

`/subnets/99999` returned **HTTP 200**, titled "Subnet 99999 — Metagraphed", with no `robots` tag and a canonical pointing at itself. Same for any valid-but-unregistered ss58. The URL space is effectively infinite, so a crawler could mint unlimited indexable thin pages by varying an integer.

`entityNotFoundMeta` already existed for exactly this — but it was wired into two of six routes, and only for **malformed** identifiers, never absent ones. Its own docstring cites `/subnets/not-a-netuid` as the bug it fixes; that route had never adopted it.

**The care here is in what does _not_ trigger it.** Every detail loader swallows failures and returns null so the page still renders and the component's own query drives the error path — that behaviour had to survive. `isMissingEntityError` accepts **only a 404 from our own API**, because marking pages `noindex` on a 5xx, a 429 or a network blip would de-index real subnets and validators during an outage. There's a test pinning that directly:

```ts
for (const status of [500, 502, 503, 429, 401, 403]) {
  expect(isMissingEntityError(new ApiError("x", { status, url: "/x" }))).toBe(false);
}
```

Now covers all six detail routes (`subnets`, `validators`, `accounts`, `blocks`, `extrinsics`, `providers`), for both the malformed and the absent case.

## 2. All 20 `/docs/*` pages shared one OG card

`server.ts` injects `og:image` from the **pathname alone**, so every doc page unfurled as `og?title=Metagraphed`. The splat route has the real title and description in `loaderData`, so it now owns its card the way the entity routes do, and `routeOwnsOgImage` matches `/docs/*` so exactly one `og:image` tag survives. `/docs` itself keeps the server-injected card, since it has its own `OG_SECTIONS` copy.

Docs are the most link-worthy pages on the site, and they were the ones sharing a card.

**Caught during rebase:** #8623 had just made `buildOgImageUrl` hardcode `entity=1`. Applied naively to docs, that flips the avatar slot to a monogram — `/docs/economics` would have rendered **"EC"** instead of the Metagraphed mark. `entity` is now an option defaulting to true (so every existing caller is unchanged) and docs pass `false`.

## 3. Docs were absent from the sitemap

281 URLs — 130 subnets, 136 providers — and none of the 20 doc pages. They now enumerate from the same source that renders them, so a new page is listed the moment it ships.

Imported **lazily**, and that's not incidental: `docs-source.ts` pulls `collections/server`, a fumadocs-mdx build-time virtual module that doesn't exist under vitest. A static import made every test that touches `server.ts` fail to collect.

## 4. No `<lastmod>` anywhere

On a product whose entire pitch is freshness, the sitemap gave crawlers nothing to schedule a recrawl against. Now emitted from a subnet's real `updated_at` — and deliberately **not** synthesised for static or docs pages. Google discounts `lastmod` site-wide once it catches a site stamping "now" on unchanged URLs, so a fabricated value would cost us the real ones too. Absent beats wrong, and there's a test for it.

## 5. No `robots` meta, so no `max-image-preview:large`

Without it Google caps the preview to a thumbnail — quietly wasting the entire per-page OG card programme (#8489 / #8622 / #8623). Now injected site-wide.

Safe alongside a route's own `noindex`: when a page carries conflicting robots tags, crawlers take the **most restrictive** directive, so the bias is toward not indexing, never toward indexing something we marked.

## Also

`og:image:alt` / `twitter:image:alt`, built from the card's own copy since that is literally what the image says, and `og:locale`.

**Deliberately not added:** `twitter:site` / `twitter:creator`. The project has no X account, so there's no handle to put there — inventing one would be worse than omitting it. Likewise the 1029 validators stay out of the sitemap: unlike subnets and providers that set is large and churns, and it's a separate decision.

## Verification

- **1791 UI tests pass** (223 files), including two new suites
- `tsc --noEmit`, `eslint`, `prettier --check` clean; `scan:public-safety` passes
- Every finding was confirmed against the **live site** first (`curl` + head parsing across 15 URLs), not inferred from the code

### DIFF
--- a/apps/ui/src/lib/metagraphed/entity-not-found-meta.test.ts
+++ b/apps/ui/src/lib/metagraphed/entity-not-found-meta.test.ts
@@ -1,61 +1,42 @@
 import { describe, expect, it } from "vitest";
-import { entityNotFoundMeta } from "@/lib/metagraphed/entity-not-found-meta";
 
-// #6429: accounts.$ss58.tsx and validators.$hotkey.tsx had no router-level
-// identifier validation, so a junk id rendered a fully-formed page. Adding
-// parseParams fixes the body — but NOT the metadata: head() still runs with the
-// raw param. Verified against the routes that already validate: /blocks/not-a-ref
-// titles "Block not-a-ref — Metagraphed" and /subnets/not-a-netuid titles
-// "Subnet not-a-netuid — Metagraphed" today. This helper is what actually keeps
-// the junk id out of the title, and both routes share it.
-const find = (meta: Array<Record<string, unknown>>, key: string, value: string) =>
-  meta.find((m) => m[key] === value);
+import { ApiError } from "./client";
+import { entityNotFoundMeta, isMissingEntityError } from "./entity-not-found-meta";
 
-describe("entityNotFoundMeta (#6429)", () => {
-  it("titles the page not-found instead of echoing the bad identifier", () => {
-    const { meta } = entityNotFoundMeta("Account", "not a valid ss58");
-    expect(meta[0]).toEqual({ title: "Account not found — Metagraphed" });
-    // The regression: the title must never carry the raw param.
-    expect(JSON.stringify(meta)).not.toContain("not-an-ss58");
+const metaValue = (meta: ReturnType<typeof entityNotFoundMeta>["meta"], key: string) =>
+  meta.find((m): m is { name: string; content: string } => "name" in m && m.name === key)?.content;
+
+describe("entityNotFoundMeta (#6429, #8624)", () => {
+  it("marks the page noindex — these URL spaces are unbounded", () => {
+    const { meta } = entityNotFoundMeta("Subnet", "No such netuid.");
+    expect(metaValue(meta, "robots")).toBe("noindex");
   });
 
-  it("marks the page noindex — these URLs are unbounded", () => {
-    // Any string is a URL here, so a crawler could otherwise index one page per
-    // malformed id.
-    const { meta } = entityNotFoundMeta("Validator", "not a valid hotkey");
-    expect(find(meta, "name", "robots")).toEqual({
-      name: "robots",
-      content: "noindex",
-    });
+  it("never asserts the junk id is a real entity in the title", () => {
+    const { meta } = entityNotFoundMeta("Subnet", "No such netuid.");
+    expect(meta[0]).toEqual({ title: "Subnet not found — Metagraphed" });
+  });
+});
+
+describe("isMissingEntityError (#8624) — the safety property", () => {
+  it("treats a 404 from our API as 'this entity does not exist'", () => {
+    expect(isMissingEntityError(new ApiError("nope", { status: 404, url: "/x" }))).toBe(true);
   });
 
-  it("carries the caller's description into both description tags", () => {
-    const detail = "This validator identifier is not a valid Bittensor ss58 hotkey.";
-    const { meta } = entityNotFoundMeta("Validator", detail);
-    expect(find(meta, "name", "description")).toEqual({
-      name: "description",
-      content: detail,
-    });
-    expect(find(meta, "property", "og:description")).toEqual({
-      property: "og:description",
-      content: detail,
-    });
+  it("does NOT treat a server error or a rate limit as missing", () => {
+    // The whole point: marking a page noindex during an outage would de-index
+    // real subnets and validators. Only a definitive 404 may flip a route.
+    for (const status of [500, 502, 503, 429, 401, 403]) {
+      expect(isMissingEntityError(new ApiError("x", { status, url: "/x" })), String(status)).toBe(
+        false,
+      );
+    }
   });
 
-  it("keeps og:title in step with the title", () => {
-    const { meta } = entityNotFoundMeta("Account", "x");
-    expect(find(meta, "property", "og:title")).toEqual({
-      property: "og:title",
-      content: "Account not found — Metagraphed",
-    });
-  });
-
-  it("names the entity the route serves, so the two routes read differently", () => {
-    expect(entityNotFoundMeta("Account", "x").meta[0]).toEqual({
-      title: "Account not found — Metagraphed",
-    });
-    expect(entityNotFoundMeta("Validator", "x").meta[0]).toEqual({
-      title: "Validator not found — Metagraphed",
-    });
+  it("does NOT treat a network throw or an abort as missing", () => {
+    expect(isMissingEntityError(new TypeError("Failed to fetch"))).toBe(false);
+    expect(isMissingEntityError(new DOMException("Aborted", "AbortError"))).toBe(false);
+    expect(isMissingEntityError(undefined)).toBe(false);
+    expect(isMissingEntityError({ status: 404 })).toBe(false);
   });
 });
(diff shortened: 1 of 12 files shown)


JSONbored:
Pushed `97250fd` — closing out the two things I'd deferred in the original description, and splitting out the one that turned out not to belong here.

**Validators are now in the sitemap.** I'd deferred these as "large and churns". I checked instead of leaving it at that: all **1029 are active on at least one subnet** and **999 hold over 1000 τ**, so they're substantive pages that were simply never listed — not the thin auto-generated set the count implies. The sitemap goes from 281 to ~1330 URLs, well inside the 50k limit.

No `<lastmod>` on them, deliberately. The only timestamp the validator list carries is `latest_captured_at`, which is when we last *probed*, not when the page's content changed. Stamping that would be exactly the "lastmod is really just now" antipattern `sitemapLastmod` exists to prevent — applied to 1029 URLs at once, which is precisely the volume that gets a site's lastmod discounted wholesale.

**The `/subnets` redirect is split out to #8628, after I implemented it.** Worth explaining rather than just deferring: seven list routes 307 to fully-expanded query strings (`/subnets` → 22 empty params, `/apis/endpoints` → 16). Root cause is that every field in the search schemas carries `.default()`, so the router materialises them and rewrites the URL.

TanStack ships the purpose-built fix (`stripSearchParams`, present in our pinned 1.170.18), and deriving the defaults from the schema itself keeps them from drifting:

```ts
export function stripDefaultSearchParams<T extends z.ZodType<object>>(schema: T) {
  return stripSearchParams(schema.parse({}) as Partial<z.output<T>>);
}
```

Applied to all seven routes, that typechecks the routes and then produces **382 type errors across 9 consumer files** — the middleware makes the stripped params optional, so every `navigate({ search: (prev) => … })` reducer and every `useSearch()` consumer has to handle absence. That's a real refactor of the URL-state layer touching filtering, sorting and pagination on all seven pages, with visual changes that need a screenshot table. It doesn't belong in a metadata PR, so I backed it out and filed #8628 with the working approach and the measured cost.

The impact is also genuinely bounded: the canonical tags point back to the clean paths, so nothing is mis-indexed. It costs a redirect hop per crawl and ugly copied URLs.

**Not addressed, and it needs you:** `twitter:site` / `twitter:creator`. The project has no X account that I can find anywhere in the repo, so there's no handle to put there and inventing one would be worse than the omission. Say the word if one exists and it's a two-line follow-up.

1791 UI tests pass; typecheck, lint, prettier and `scan:public-safety` clean.

Issue #8632, opened 2026-07-29
MCP: GET /mcp error message sends clients to a dead end


JSONbored:
Reported from the field: opening `https://api.metagraph.sh/mcp` returns

```json
{"jsonrpc":"2.0","id":null,"error":{"code":-32600,
 "message":"GET requires a valid Mcp-Session-Id header (obtained from the initialize response)."}}
```

## The endpoint is fine; the message is wrong

`GET /mcp` is the server→client SSE push channel, not a browsable URL, so an error on a bare browser GET is correct. Verified the real flow works end to end:

```
initialize                                        → 200, mcp-session-id: <uuid>
resources/subscribe metagraph://chain/stream      → 200 {}
GET /mcp  (with that session id)                  → 200 text/event-stream
                                                    data: {"method":"notifications/resources/updated", …}
```

**The defect is that following the error message's advice fails.** It says the session id is "obtained from the initialize response", which reads as *initialize, then GET*. That sequence returns 404:

```json
{"error":{"code":-32600,"message":"No such MCP session; call initialize again."}}
```

…and that second message is worse, because re-running `initialize` produces the identical 404 forever. A session is only registered with the stream hub by **`resources/subscribe`** — the step neither message mentions.

So a client hits a 400, follows its instructions, hits a 404, follows *those* instructions, and loops. That is how this was reported.

## Requirements

1. The 400 (no/invalid session on GET) must state the whole precondition — initialize **and** `resources/subscribe` — and say outright that this is not a browsable endpoint, since opening it in a browser is the common way to arrive here.
2. The 404 (session not registered) must point at the missing `resources/subscribe` step and must not advise re-running `initialize` alone.
3. Both should name the subscribable resources (`metagraph://chain/stream`, `metagraph://subnet/{netuid}/status`) so the next step is copy-pasteable.

## Deliverables

- [ ] Both messages in `src/mcp-server.ts` rewritten
- [ ] Tests asserting the bare-GET message names `resources/subscribe`, `initialize`, a subscribable URI, and "not a browsable endpoint"
- [ ] A test asserting the 404 no longer says "call initialize again"

## Expected outcome

A developer who opens `/mcp` in a browser, or whose client GETs without subscribing, can reach a working SSE stream by following the error text alone.

PR #8654, opened 2026-07-29
fix(mcp): distinguish "not callable" from "does not exist" on surface lookup

### LINKED ISSUE #8653: MCP: call_subnet_surface says a docs/dashboard surface 'does not exist' — it does, it just isn't callable
Found by calling `call_subnet_surface` on **every probe-enabled surface** (1,855 across all 129 netuids). Of 313 failures, **208 were this**, and every one was a non-callable kind.

```
call_subnet_surface { surface_id: "bittensor-networks-docs" }
→ not_found: No catalogued surface with id, key, or deprecated id "bittensor-networks-docs"
```

That statement is false. The surface **is** catalogued — the agent almost certainly read that exact id out of `list_surfaces` or `get_subnet_surfaces` a moment earlier.

## What is actually happening

Two catalogs, both correct:

| Catalog | Size | Contents |
| --- | --- | --- |
| `/metagraph/operational-surfaces.json` | 617 | callable only — `subnet-api`, `data-artifact`, `subtensor-rpc/wss`, `sse` |
| `/metagraph/surfaces.json` (= `list_surfaces`, `GET /api/v1/surfaces`) | 3,491 | everything, incl. `docs`, `dashboard`, `website`, `source-repo`, `openapi`, `sdk`, `example` |

`call_subnet_surface` resolves against the operational catalog, which is right — you cannot issue an API request to a GitHub repo or a docs page. The defect is purely **what we say when the two differ**.

Breakdown of the 313 failures:

```
not_found              208   ← all docs/dashboard/website/repo/openapi/sdk/example
upstream_unavailable     5
unsupported_content_type 2
internal_error           1
(OK)                   105   ← all subnet-api / data-artifact / subtensor-rpc
```

## Why it matters

This is the **core discover→call loop**: list surfaces, pick one, call it. Telling an agent its id does not exist sends it hunting for a different id — and there isn't one. The right answer is "that is a docs link, here it is; these are the callable ones".

Same class as #8632, where `GET /mcp` told clients to do something that then failed.

## Requirements

1. A surface present in the full registry but absent from the operational catalog must report a **distinct** code, not `not_found`.
2. The message must state the surface's `kind`, give its URL (it is a link — that IS the useful answer), and name the tools that find callable surfaces.
3. A genuinely unknown id must still report `not_found`.
4. An unreadable registry must degrade to `not_found`, never an internal error — a bad id must not become a 500.
5. `verify_integration` resolves surfaces the same way and needs the same treatment.

## Deliverables

- [ ] `not_callable` error code with an actionable message
- [ ] Applied to both `call_subnet_surface` and `verify_integration`
- [ ] Tests for: uncallable → `not_callable`; unknown → `not_found`; registry unreadable → `not_found`; callable unaffected

## Expected outcome

Calling a docs surface returns a message an agent can act on without a second round trip, and only genuinely unknown ids say "no such surface".

JSONbored:
## Summary

Closes #8653

Found by calling `call_subnet_surface` on **every probe-enabled surface** — 1,855 across all 129 netuids. Of 313 failures, **208 were this**, and every one was a non-callable kind:

```
call_subnet_surface { surface_id: "bittensor-networks-docs" }
→ not_found: No catalogued surface with id, key, or deprecated id "bittensor-networks-docs"
```

That statement is false. The surface **is** catalogued — the agent almost certainly read that exact id out of `list_surfaces` moments earlier.

## Two catalogs, both correct

| Catalog | Size | Contents |
| --- | --- | --- |
| `operational-surfaces.json` | 617 | callable only — `subnet-api`, `data-artifact`, `subtensor-rpc/wss`, `sse` |
| `surfaces.json` (`list_surfaces`, `GET /api/v1/surfaces`) | 3,491 | everything, incl. `docs`, `dashboard`, `website`, `source-repo`, `openapi`, `sdk`, `example` |

Resolving calls against the operational catalog is **right** — you cannot issue an API request to a GitHub repo. Only the error was wrong.

The split is clean in the data:

```
not_found              208   ← all docs/dashboard/website/repo/openapi/sdk/example
upstream_unavailable     5
unsupported_content_type 2
internal_error           1
(OK)                   105   ← all subnet-api / data-artifact / subtensor-rpc
```

## Why it matters

This is the **core discover→call loop**: list surfaces, pick one, call it. Telling an agent its id doesn't exist sends it hunting for a different id — and there isn't one. For a docs surface the useful answer is the link itself.

Same class as #8632, where `GET /mcp` told clients to do something that then failed.

## Fix

An id absent from the operational catalog is now looked up in the **full registry**:

- **present** → `not_callable`, naming the `kind`, giving the URL, and pointing at `list_subnet_apis` / `get_subnet_surfaces`
- **absent** → `not_found`, unchanged
- **registry unreadable** → falls back to `not_found`, so a bad id can never become an internal error

Applied to `verify_integration` too, which resolves surfaces the same way.

## Tests

4 new in `call-subnet-surface-mcp.test.ts` (77 total): uncallable → `not_callable` with kind + URL + next-step tools; unknown → `not_found`; registry unreadable → `not_found` rather than a 500; and `verify_integration` reporting the same distinction.

1328 tests pass across `mcp-server`, `call-subnet-surface-mcp` and `surface-verify`; `validate:mcp` passes (205 tools + both round trips); lint and `scan:public-safety` clean.

### DIFF
--- a/src/mcp-server.ts
+++ b/src/mcp-server.ts
@@ -359,6 +359,10 @@ import {
   loadSurfacesList,
   SURFACES_ARTIFACT,
 } from "./surfaces-mcp.ts";
+// The single source of truth for which kinds are callable -- the same list
+// scripts/build-artifacts.ts filters operational-surfaces.json by, so this
+// error can never describe a different set than the catalog actually holds.
+import { OPERATIONAL_SURFACE_KINDS } from "./health-probe-core.ts";
 import {
   LIST_ENDPOINT_POOLS_INSTRUCTIONS,
   LIST_ENDPOINT_POOLS_MCP_TOOL,
@@ -1878,16 +1882,33 @@ async function uncallableSurfaceError(
     // bad id into an internal error.
   }
   if (known) {
-    const kind = typeof known.kind === "string" ? known.kind : "non-API";
+    const kind = typeof known.kind === "string" ? known.kind : null;
     const url = typeof known.url === "string" ? known.url : null;
+    const link = url ? ` It is a link: ${url}.` : "";
+    // Two genuinely different reasons an id can be absent from the callable
+    // catalog, and saying the wrong one is its own bug. A docs page is not
+    // callable BY KIND. But a `subnet-api` can also be missing -- 10 such
+    // surfaces are advertised probe-enabled in the public registry yet absent
+    // from operational-surfaces.json (#8658) -- and telling someone that a
+    // subnet-api "is not a callable API" would be plainly false.
+    if (kind && OPERATIONAL_SURFACE_KINDS.includes(kind)) {
+      return toolError(
+        "not_callable",
+        `Surface "${surfaceId}" is a ${kind} surface, which is a callable ` +
+          `kind, but it is not in the operational catalog, so it is not ` +
+          `callable right now and is not being health-probed.${link} This is ` +
+          `a registry-side gap rather than a bad id; use list_subnet_apis or ` +
+          `get_subnet_surfaces to find this subnet's currently callable ones.`,
+      );
+    }
     return toolError(
       "not_callable",
-      `Surface "${surfaceId}" is catalogued but is a ${kind} surface, not a ` +
-        `callable API, so there is nothing to request. ` +
-        (url ? `It is a link: ${url}. ` : "") +
-        `Only subnet-api, data-artifact, subtensor-rpc/wss and sse surfaces ` +
-        `can be called; use list_subnet_apis or get_subnet_surfaces to find ` +
-        `this subnet's callable ones.`,
+      `Surface "${surfaceId}" is catalogued but is ` +
+        `${kind ? `a ${kind} surface` : "not an API surface"}, not a ` +
+        `callable API, so there is nothing to request.${link} Only ` +
+        `${OPERATIONAL_SURFACE_KINDS.join(", ")} surfaces can be called; use ` +
+        `list_subnet_apis or get_subnet_surfaces to find this subnet's ` +
+        `callable ones.`,
     );
   }
   return toolError(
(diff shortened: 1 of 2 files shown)


JSONbored:
Pushed — patch coverage closed, and it turned up a real flaw in this PR's own message.

## Coverage

The 3 partials were the `kind` fallback, the `url` fallback, and the link clause. Measured the way CI does, across all four suites that touch `mcp-server.ts`:

```
uncovered lines : []
partial branches: []
```

## The flaw the coverage work exposed

The message I added said *"is a {kind} surface, not a callable API"* for anything absent from the operational catalog. That's right for a docs page. It is **false** for the ten surfaces in #8658 whose kind **is** callable — `subnet-api` and `data-artifact` rows advertised probe-enabled in the public registry yet missing from `operational-surfaces.json`.

Telling someone a `subnet-api` "is not a callable API" just replaces one wrong answer with another, which is exactly what this PR set out to stop.

There are now two cases, because there are two genuinely different reasons:

| Situation | Message |
| --- | --- |
| callable kind, absent from catalog | "is a callable kind, but not in the operational catalog, so it is not callable right now and is not being health-probed — a registry-side gap rather than a bad id" |
| non-callable kind | unchanged |

The callable-kind list now comes from `OPERATIONAL_SURFACE_KINDS` (`src/health-probe-core.ts`) rather than a restated literal — **the same constant `scripts/build-artifacts.ts` filters `operational-surfaces.json` by**, so this error can never describe a different set than the catalog actually holds.

## Fixtures

Both new branches are covered by realistic rows rather than contrived ones: a callable-kind surface missing from the catalog (the #8658 shape), and a minimal registry row with neither `kind` nor `url` — which also pins that the message never renders "a undefined surface" or a dangling "It is a link:".

80 tests in `call-subnet-surface-mcp.test.ts`, 1343 across the four suites; `validate:mcp`, lint, prettier and `scan:public-safety` all clean.

PR #8656, opened 2026-07-29
fix(mcp): bound the body read — call_subnet_surface hung forever on a stream

### LINKED ISSUE #8655: call_subnet_surface hangs on a streaming body — the abort timer is cleared before the body is read
Found while calling every probe-enabled surface. `call_subnet_surface` on `allways-sse` never returns; the agent eventually gets

```
internal_error: The tool failed to complete.
```

Reproduced against production: the request hung past a 60s client timeout.

## Root cause

`safetyCheckedFetch` arms an `AbortController` for `timeout_ms` and clears it in a `finally`:

```ts
} finally {
  clearTimeout(timer);
}
```

That `finally` runs when `safetyCheckedFetch` **returns** — which is as soon as the response *headers* arrive. The body is streamed afterwards by `readBodyCapped`, with **nothing watching the clock**.

For an event stream the headers arrive instantly, the timer is cleared, and the read then waits forever on a body that never ends and never reaches `maxBytes`.

`sse` is one of the callable surface kinds in `operational-surfaces.json`, so this is reachable by design, not by misuse.

## Why it matters beyond the error message

An unbounded read on a public, unauthenticated tool pins a Worker invocation until the platform kills it. Any surface that streams — or merely trickles its body slowly — can do this, so it is an availability concern, not only a bad error string. `timeout_ms` is configured per surface (5000 for this one) and was silently not enforced for the part of the request that actually takes the time.

## Requirements

1. The configured `timeout_ms` must bound the **whole** operation, body included.
2. A stream still producing at the deadline is **truncated, not failed** — the first few KB of an event stream is genuinely useful, and matches how an oversized body is already treated.
3. "The deadline won" must never be reported as "the stream ended" — a timed-out read must not look like a complete body.
4. Normal bodies must be unaffected and must not start reporting `truncated`.

## Deliverables

- [ ] Deadline applied to the body read
- [ ] Tests: an endless stream returns promptly and truncated; the partial content is retained; a normal body is unchanged

## Expected outcome

`call_subnet_surface` on an `sse` surface returns within its configured timeout, with the events read so far and `truncated: true`.

JSONbored:
## Summary

Closes #8655

Found while calling every probe-enabled surface. `call_subnet_surface` on `allways-sse` **never returns** — the agent eventually gets:

```
internal_error: The tool failed to complete.
```

Reproduced against production: the request hung past a 60s client timeout.

## Root cause

`safetyCheckedFetch` arms an `AbortController` for `timeout_ms` and clears it in a `finally`:

```ts
} finally {
  clearTimeout(timer);
}
```

That `finally` runs when the function **returns** — as soon as the response *headers* arrive. The body is streamed afterwards by `readBodyCapped`, with **nothing watching the clock**. For an event stream the headers arrive instantly, the timer is cleared, and the read waits forever on a body that never ends and never reaches `maxBytes`.

`sse` is one of the callable kinds in `operational-surfaces.json`, so this is reachable **by design**, not by misuse.

## Why it's more than a bad error message

An unbounded read on a public, unauthenticated tool pins a Worker invocation until the platform kills it — and any slow-trickling body does the same, not just event streams. The per-surface `timeout_ms` (5000 for this surface) was silently not enforced over the part of the request that actually takes the time.

## Fix

`readBodyCapped` carries its own deadline. A stream still producing when it passes is **truncated, not failed** — the first few KB of an event stream is genuinely useful to an agent, and it matches how an oversized body is already handled.

One detail worth flagging for review: the deadline resolves a **distinct sentinel** rather than resolving the race as `done`. Resolving as `done` would report a timed-out stream as a *complete* body — precisely the wrong thing to tell an agent — so the sentinel keeps "the deadline won" and "the stream ended" separate.

## Tests

3 new in `call-subnet-surface.test.ts` (81 total), driven by a `ReadableStream` that emits one event and never closes — a real endless body, not a mock of one:

- it returns promptly (asserted under 5s) with `truncated: true`
- the partial content it *did* read is retained, not discarded
- a normal body is unaffected and never reports `truncated`

1387 tests pass across `call-subnet-surface`, `call-subnet-surface-mcp` and `mcp-server`; `validate:mcp` passes (205 tools + both round trips); lint and `scan:public-safety` clean.

### DIFF
--- a/src/call-subnet-surface.ts
+++ b/src/call-subnet-surface.ts
@@ -386,7 +386,7 @@ export async function callSubnetSurface(
     };
   }
 
-  const raw = await readBodyCapped(response, MAX_RESPONSE_BYTES);
+  const raw = await readBodyCapped(response, MAX_RESPONSE_BYTES, timeoutMs);
   let body: unknown = raw.text;
   let parseError: string | null = null;
   if (kind === "json" && raw.text) {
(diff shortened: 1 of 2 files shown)


JSONbored:
Pushed — patch coverage closed.

```
uncovered lines : []
partial branches: []
```

Both gaps were **defensive code that cannot execute**, so I deleted them rather than contorting tests into reaching them:

- The top-of-loop `remaining <= 0` check duplicated the race. A non-positive `setTimeout` delay fires on the next tick, so the race already resolves `DEADLINE` immediately — and it was unreachable regardless, since `deadline` is computed on entry so the first iteration always has time left.
- The `deadlineMs && deadlineMs > 0 ? … : Infinity` fallback had no caller. `readBodyCapped` has exactly one call site and `callSubnetSurface` always resolves `timeoutMs` to a finite number (`probe.timeout_ms`, or 10s). The parameter is now **required**, so the type states that rather than a runtime branch pretending otherwise.

Behaviour is unchanged: same deadline, same truncation semantics, same sentinel keeping "the deadline won" distinct from "the stream ended".

154 tests pass across both call-surface suites; lint and `scan:public-safety` clean.

Issue #8658, opened 2026-07-29
Published registry advertises unverified/dead candidates as probe-enabled and public-safe


JSONbored:
Found by calling `call_subnet_surface` on every probe-enabled surface (1,855). Most `not_found` results were non-callable kinds (docs/dashboard/website — that is #8653). **Ten were not**: they are callable kinds that meet every published criterion for the operational catalog and are still missing from it.

## The drift, both directions

Comparing the live registry (`GET /api/v1/surfaces`) against the committed `public/metagraph/operational-surfaces.json`, applying the catalog's own filter (`probe.enabled && public_safe && operational kind`):

```
eligible in LIVE registry     : 623
in committed operational file : 617
eligible but MISSING from it  : 10   (7 data-artifact, 3 subnet-api)
in it but GONE from registry  : 4
```

Every one of the 10 has `public_safe: true`, `probe.enabled: true` and a callable `kind`. Examples:

```
subnet-api     netuid=45   sn-45-website-common-health
subnet-api     netuid=123  sn-123-website-common-health
subnet-api     netuid=107  sn-107-github-readme-minos-protocol-minos-subnet-subnet-api-1
data-artifact  netuid=64   sn-64-website-link-chutes-ai-data-artifact-{2,8,9,10}
data-artifact  netuid=24   sn-24-github-readme-silx-labs-quasar-subnet-data-artifact-1
data-artifact  netuid=27   sn-27-github-readme-silx-labs-orion-data-artifact-1
data-artifact  netuid=95   sn-95-website-link-actual-inc-data-artifact-7
```

All ten are `curation_level: candidate-discovered`.

## Consequences

1. **They are never health-probed.** `operational-surfaces.json` is the prober's cold-start input, so these ten have no health history despite advertising `probe.enabled: true`.
2. **They cannot be called.** `call_subnet_surface` resolves against this catalog, so an agent that finds them via `list_surfaces` gets an error.
3. **Four surfaces are still being probed after leaving the registry** — the reverse drift.

## The part that needs a maintainer

`sync-operational-surfaces.yml` exists precisely for this, hourly, and its own header documents a previous instance (*"79 → 96 surfaces observed stale on a clean main checkout"*). But the last several syncs all report **617 → 617**, i.e. regenerating from the committed registry yields 617 — while the live API exposes 623 eligible.

[...]

PR #8660, opened 2026-07-29
fix(registry): record when a promoted surface was verified

JSONbored:
## Summary

Refs #8658

Every promoted candidate shipped **`last_verified_at: null`** while also carrying `probe.enabled: true` and `public_safe: true`. A consumer could not tell a just-verified surface from one whose verification had long gone stale — and `probe.enabled` is the documented signal for *"this can be probed and called"*, so it is exactly what agents filter on.

## What surfaced it

Sweeping `call_subnet_surface` across all 1,855 probe-enabled surfaces. `sn-45-website-common-health` is published probe-enabled and public-safe with **no verification timestamp at all**, and its own note reads:

> "Common read-only path discovered from a public project website root. **Requires verification before promotion.**"

Meanwhile the committed verification file classifies it **`dead`, confidence 0, status failed**. Nothing in the published record exposed that gap.

## Fix

`promoteCandidate` already receives the verification result, but individual results carry no time of their own — the run's `observed_at` is the only provenance available. It is now threaded through and stamped onto the surface.

**This does not change which surfaces get promoted.** `isPromotable` still gates on a live/redirected classification, so the `dead` case self-corrects on the next build. It changes what a promoted surface *discloses*, so a stale promotion is visible rather than indistinguishable from a fresh one.

Kept as an explicit `null` (never `undefined`) when a verification file has no `observed_at`, so the key is always present and consumers can branch on it.

## Tests

2 new in `script-utils.test.ts` (58 total): the timestamp is stamped from the run and pairs with `probe.enabled`/`public_safe`; and an older verification file without `observed_at` still yields a well-formed surface with an explicit `null`.

70 tests pass across `script-utils`, `promote-reviewed-curation` and `registry-identity`; typecheck, lint and `scan:public-safety` clean.

## Still open on #8658

The remaining question is a publishing-policy one I've left to you: whether an unverified candidate should be published with `probe.enabled: true` at all. This PR makes the staleness *visible*; it does not change the promotion rule. The 4 catalog entries that no longer exist in the registry are also still outstanding there.

### DIFF
--- a/scripts/generated-overlays.ts
+++ b/scripts/generated-overlays.ts
@@ -128,6 +128,15 @@ export async function generateBaselineOverlaySet(
       result,
     ]),
   );
+  // #8658: the verification RUN's timestamp. Individual results carry no time
+  // of their own, so this file-level value is the only provenance a promoted
+  // surface can record -- and without it every promoted surface published
+  // `last_verified_at: null`, which is what let an unverifiable surface look
+  // identical to a freshly-checked one.
+  const verificationObservedAt =
+    typeof verification.observed_at === "string"
+      ? verification.observed_at
+      : null;
   const providersById = new Map(
     providers.map((provider) => [provider.id, provider]),
   );
@@ -143,6 +152,7 @@ export async function generateBaselineOverlaySet(
       nativeSubnet,
       providersById,
       verificationByCandidate,
+      verificationObservedAt,
       maintainerReviewedDecisionsByNetuid,
     });
     if (manualNetuids.has(nativeSubnet.netuid)) {
@@ -385,6 +395,7 @@ interface BuildGeneratedOverlayOptions {
   nativeSubnet: Row;
   providersById: Map<unknown, Row>;
   verificationByCandidate: Map<unknown, Row>;
+  verificationObservedAt?: string | null;
   maintainerReviewedDecisionsByNetuid?: Map<unknown, Row[]>;
 }
 
@@ -394,6 +405,7 @@ function buildGeneratedOverlay({
   nativeSubnet,
   providersById,
   verificationByCandidate,
+  verificationObservedAt = null,
   maintainerReviewedDecisionsByNetuid = new Map(),
 }: BuildGeneratedOverlayOptions): Row {
   const subnetCandidates = candidatesByNetuid.get(nativeSubnet.netuid) || [];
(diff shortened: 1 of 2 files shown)


JSONbored:
Closing this — the CI failure led me to a design invariant that makes the premise wrong, and I would rather withdraw it than force it past a test that is deliberately there.

## What the failure actually was

`tests/artifacts.test.ts` asserts, of the published surfaces artifact:

```ts
const generatedSurfaces = surfaces.surfaces.filter((s) => s.authority === "registry-observed");
assert.equal(generatedSurfaces.some((s) => s.verification !== undefined), false);
```

**Promoted surfaces must not carry a `verification` block.** That is an explicit, tested decision, not an oversight — and my change added exactly that.

## What I found while chasing it

The machinery to populate `last_verified_at` already exists and is correct:

```ts
// scripts/lib.ts, flattenSurfaces
flattened.last_verified_at =
  (surface.verification as Row | undefined)?.verified_at ??
  (surface.authority === "official" ? subnet.curation?.verified_at ?? null : null);
```

It reads `surface.verification.verified_at`. It returns null because **nothing writes that block**: only **1 of 2,743** registry surfaces has one.

I also had the field in the wrong place at first — a top-level `last_verified_at` on the registry surface fails validation outright, since the manifest's surface object is `additionalProperties: false` and its own docs say freshness belongs in the `verification` overlay ("health, last_verified_at, incidents"). Fixing that placement is what surfaced the artifacts invariant underneath.

## So the real question is a product one

`last_verified_at: null` on promoted surfaces is **currently by design**, encoded in a test. Three coherent positions, and picking one is a call for the maintainer rather than something to slip in under a bug fix:

1. **Keep it.** Promoted surfaces expose no verification provenance; consumers treat `probe.enabled` as "eligible for probing", not "recently verified". The 10 surfaces in #8658 are then working as intended and #8658 reduces to a docs change.
2. **Populate `verification` on promoted surfaces and relax the artifacts assertion** — publishes real provenance, and is what I attempted here. It changes a public artifact's shape.
3. **Populate it in the registry overlay but strip it at flatten time**, keeping only the derived `last_verified_at`. Gets the timestamp without exposing the machine-verification detail, and preserves the spirit of the existing assertion.

Option 3 is what I would build if it were mine to choose, but it is a shape change to a published artifact either way.

Recorded on #8658. No code from this PR is on `main`.

PR #8840, opened 2026-07-31
fix(api): treat missing top-holders net_flow as null, not zero

### LINKED ISSUE #8807: fix(api): top-holders net_flow_7d/30d/90d serve a sortable 0 when the rollup row is missing
## Context

`GET /api/v1/accounts/top-holders` serves three cross-subnet stake-flow columns — `net_flow_7d`, `net_flow_30d`, `net_flow_90d` (#6886/#6887) — and all three are also advertised **sort keys** (`TOP_HOLDERS_SORTS`, mirrored in `schemas-src/routes/top-holders.ts:22-29`).

Two layers independently coerce "no data" into `0`:

- `workers/data-api.ts:6826-6828` wraps the `LEFT JOIN wallet_flow_daily` aggregate in `COALESCE(f.net_flow_7d, 0)` / `COALESCE(f.net_flow_30d, 0)` / `COALESCE(f.net_flow_90d, 0)`. It is a `LEFT JOIN`, so an account with **no** `wallet_flow_daily` row in the 90-day window produces SQL `NULL` — which the `COALESCE` turns into `0`.
- `src/top-holders.ts:50-53`'s `numberOrZeroSigned` then maps anything non-finite (including a tier that never supplied the column at all) to `0`.

So "this account had zero net stake flow" and "we have no flow data for this account" are indistinguishable on the wire — and that indistinguishable value is then **sorted on**. `src/top-holders.ts:115-119` sorts by `(b[normalizedSort] - a[normalizedSort]) || a.ss58.localeCompare(b.ss58)`; when every row is `0` the primary comparator is always `0`, the tiebreak takes over, and `?sort=net_flow_7d` degenerates into an ss58-alphabetical list of dust accounts wearing a "biggest movers" label.

Production today returns `net_flow_7d` / `net_flow_30d` / `net_flow_90d` = `0` on **every** row of `/api/v1/accounts/top-holders` — all three advertised sort keys are silently non-functional. This is not "nobody moved stake": `GET /api/v1/accounts/{ss58}/stake-flow?window=30d` reports a real nonzero `net_flow_tao` for the same address over the same window, from the same `account_events` source. The flow data exists; the `wallet_flow_daily` rollup at `workers/data-api.ts:1320-1339` is what is not landing rows.

**Scope — read before starting.** The deliverable of this issue is the **null-vs-zero correctness fix**, which is fully mechanical and completable from the repo. Whether the `wallet_flow_daily` rollup is actually populating in production is an operational question you cannot verify or fix from a PR, and **you are not required to fix the data pipeline to close this issue.** You *are* required to state in the PR description whether you were able to confirm that the rollup at `workers/data-api.ts:1320-1339` produces rows, and what you based that on (code reading counts — say so). A PR that quietly skips that statement does not satisfy this issue.

Mirror the "null is not zero" convention this repo already applies to exactly this class of value:

- **`src/chain-analytics.ts:370-377` (`buildChainFees`)** — `avg_fee_tao` / `median_fee_tao` resolve to `null`, never `0`, when the value is undefined; the schema at `schemas-src/routes/chain-analytics.ts:124-127` declares them `z.number().min(0).nullable()` for precisely that reason.
- **`src/metagraph-neurons.ts:472-485` (`realizedReturn`)** — states the rule verbatim: null (never 0) ... "no realized figure" vs. "confirmed zero return".

The field itself becomes nullable, matching those two. The following do **not** satisfy this issue: a parallel signal alongside the number (a `net_flow_available` boolean, a `has_flow_data` flag, a `flow` sub-object), a sentinel value (`-1`, `NaN`), omitting the keys entirely on a miss, or leaving the fields as-is and only changing how they sort.

## Requirements

1. **SQL (`workers/data-api.ts:6826-6828`)** — drop `COALESCE(..., 0)` from all three `net_flow_*` selections so an absent `wallet_flow_daily` row reaches the builder as SQL `NULL`. Leave `COALESCE(b.free_tao, 0)` and `COALESCE(d.delegated_tao, 0)` on lines `:6824-6825` **unchanged** — different question, out of scope.

2. **Preserve the genuine zero.** An account that *has* `wallet_flow_daily` rows in the 90-day window but none in the last 7 days must still report `net_flow_7d: 0` — that is a real zero, not a missing value. Note that the inner subselect's `SUM(CASE WHEN day >= CURRENT_DATE - 7 THEN net_flow_tao ELSE 0 END)` already returns a non-null number for any coldkey present in the grouped subselect, so removing the *outer* `COALESCE` preserves this distinction on its own. Confirm this holds in your change and cover it with a test (see Deliverables) — do not add a second `COALESCE` inside the subselect to "be safe".

3. **Builder (`src/top-holders.ts:44-81`)** — the three `net_flow_*` fields become `number | null`. A finite number passes through with its sign intact (a negative net outflow is real data and must not be clamped — the existing comment at `:47-48` still applies); `null`, `undefined`, and any non-finite input become `null`. `free_tao` / `delegated_tao` / `total_tao` keep `numberOrZero` and stay `number`. If `numberOrZeroSigned` ends up with no callers, delete it rather than leaving dead code.

4. **Sorting with nulls — the exact contract.** `buildTopHoldersList`'s comparator (`src/top-holders.ts:115-119`) must place every row whose active sort key is `null` **after** every row that has a number for it, regardless of that number's sign — a `null` must never outrank a real negative outflow. Among two rows that are both `null` on the sort key, the existing `a.ss58.localeCompare(b.ss58)` tiebreak applies. Among two rows that both have numbers, descending-numeric order is unchanged.

   Be deliberate here: the current expression is arithmetic subtraction, and in JavaScript `null` coerces to `0` in arithmetic (`null - 5 === -5`, `5 - null === 5`). Simply making the fields nullable and leaving `b[sort] - a[sort]` alone would sort a `null` **exactly as if it were `0`** — reintroducing the same defect one layer down, with no visible symptom. The comparator needs an explicit null check, not a coercion.

5. **Route schema (`schemas-src/routes/top-holders.ts:36-38`)** — `net_flow_7d` / `net_flow_30d` / `net_flow_90d` change from `z.number()` to `z.number().nullable()`. Then run `npm run build` and commit the regenerated `openapi.json` and generated types **in the same PR**, or `validate:contract-drift` fails CI.

6. **Leave these alone (verified — no change needed):** the MCP output schema at `schemas-src/mcp-tools/accounts-leaderboards.ts:85-87` already types these as `z.unknown().optional()`, and the CSV writer at `workers/csv.ts:34` already renders `null` as an empty cell, so `TOP_HOLDERS_CSV_COLUMNS` (`workers/request-handlers/entities.ts:372-381`) needs no change either. Do not churn them.

7. **Update the contract prose.** `src/contracts.ts:2835` currently ends with "net_flow_* is StakeAdded minus StakeRemoved over the window, from the wallet_flow_daily rollup -- a negative value is a real net outflow, not a missing value." Extend it to also state that `null` means "no rollup row for this account in the window" and that null-valued rows sort last. Its claim must not contradict the new behavior.

8. **Coverage.** `codecov/patch` enforces 99% on every changed line and branch in `src/**` / `workers/**`. Both comparator branches (null-vs-number, number-vs-number) and both coercion branches must be exercised. Measure unsharded with `npm run test:coverage`. All new test files are `.ts`.

## Deliverables

- [ ] `workers/data-api.ts:6826-6828` — the three `COALESCE(f.net_flow_*, 0)` wrappers removed
- [ ] `src/top-holders.ts` — `net_flow_*` typed and produced as `number | null`; `TopHoldersEntry` updated; any now-unused helper deleted
- [ ] `src/top-holders.ts` — comparator rewritten so `null` sorts strictly last on all three flow sort keys, with the ss58 tiebreak preserved
- [ ] `schemas-src/routes/top-holders.ts` — all three fields `.nullable()`, plus the regenerated `openapi.json` and generated types committed in the same PR
- [ ] `src/contracts.ts:2835` — description updated to document the null meaning and the null-last sort
- [ ] Regression tests in `tests/top-holders.test.ts` that **fail before this change and pass after**, covering at minimum: (a) a row with no flow data yields `null`, not `0`; (b) a row with flow rows in the window but none in the last 7 days still yields `net_flow_7d: 0`; (c) sorting by `net_flow_7d` puts a row with `-5` above a row with `null`; (d) two null-valued rows tiebreak by ss58
- [ ] Test asserting the emitted SQL for this route no longer contains `COALESCE(f.net_flow_7d` — `tests/data-api.test.ts` already captures the emitted SQL text and is the right place for it
- [ ] PR description states whether you could confirm the `wallet_flow_daily` rollup produces rows, and on what basis

## Expected Outcome

`/api/v1/accounts/top-holders` returns `net_flow_7d`/`net_flow_30d`/`net_flow_90d` as `null` for any account with no `wallet_flow_daily` row in the window and as a signed number (including a genuine `0`) otherwise; `openapi.json` declares all three nullable; and `?sort=net_flow_7d|30d|90d` orders numeric rows descending with null-valued rows strictly last, so a leaderboard sorted by flow can never again present unknown accounts as if they were confirmed-zero movers. Verifiable from the diff alone.

## Links & Resources

- `workers/data-api.ts:6820-6844` — the top-holders query
- `workers/data-api.ts:1320-1339` — the `wallet_flow_daily` rollup that feeds it
- `src/top-holders.ts:43-131` — `numberOrZeroSigned`, `buildTopHoldersEntry`, `buildTopHoldersList`
- `schemas-src/routes/top-holders.ts` — the route contract
- `src/chain-analytics.ts:370-377`, `src/metagraph-neurons.ts:472-485` — the null-not-zero precedents to mirror
- Original stake-flow leaderboard work: #6886 / #6887

kurosawareiji7007-hub:
Closes #8807

## Summary

`net_flow_7d` / `30d` / `90d` on `/api/v1/accounts/top-holders` coerced missing `wallet_flow_daily` rows to `0` (SQL `COALESCE` + builder), so advertised flow sort keys treated unknown accounts as confirmed-zero movers.

## What changed

1. **SQL** — drop outer `COALESCE(f.net_flow_*, 0)`; leave free/delegated `COALESCE` unchanged. Inner `SUM(CASE …)` still yields a real `0` when the coldkey is present in the 90d window but has no flow in the shorter window.
2. **Builder** — `net_flow_*` are `number | null`; non-finite → `null`; signed zeros preserved; unused `numberOrZeroSigned` removed.
3. **Sort** — null sort keys sort strictly last; ss58 tiebreak among nulls and among equal numbers.
4. **Schema** — `.nullable()` + regenerated `openapi.json` / types / api-index / contract d.ts.
5. **Contract prose** — documents null meaning and null-last sort.

## Rollup confirmation (required)

Based on **code reading only** (no production DB access): `workers/data-api.ts` includes an `INSERT INTO wallet_flow_daily … FROM account_events` inside the account-events-daily rollup path (`:1321-1343`). That is evidence the writer exists and is wired. I **could not confirm** that production currently lands rows for arbitrary coldkeys; that remains an operational question outside this PR’s scope. The null-vs-zero correctness fix is independent of whether the rollup is populated.

## Test plan

- [x] `tests/top-holders.test.ts` — null vs 0, genuine 0, null-last sort, ss58 tiebreak
- [x] `tests/data-api.test.ts` — SQL no longer contains `COALESCE(f.net_flow_7d`
- [x] `npm run build` for contract artifacts
- [ ] CI green


kurosawareiji7007-hub:
Superseded by fresh PR with regenerated OpenAPI docs (one-shot gate).

PR #8844, opened 2026-07-31
fix(api): chain/fees medians and averages no longer diluted by unsigned inherents

### LINKED ISSUE #8809: fix(api): chain/fees medians pinned to 0 by unsigned inherents; averages diluted by them
## Context

`GET /api/v1/chain/fees` reports `median_fee_tao` and `median_tip_tao` as a structural `0` on every day, and dilutes `avg_fee_tao` / `avg_tip_tao`, because the aggregation counts **unsigned inherents** as if they were fee-paying transactions.

The median query at `workers/data-api.ts:8632-8641` is:

```sql
SELECT to_char(to_timestamp(observed_at / 1000), 'YYYY-MM-DD') AS day,
       PERCENTILE_CONT(0.5) WITHIN GROUP (ORDER BY COALESCE(fee_tao, 0)) AS median_fee_tao,
       PERCENTILE_CONT(0.5) WITHIN GROUP (ORDER BY COALESCE(tip_tao, 0)) AS median_tip_tao
FROM extrinsics WHERE observed_at >= ${cutoff}
  ${moduleClause}
GROUP BY day
```

There is **no signer filter**. The indexer stores unsigned inherents as ordinary `extrinsics` rows — `apps/indexer-rs/src/main.rs:996` pushes every decoded extrinsic, including `Timestamp.set`, with `signer` = `None` and no fee. Those rows carry `fee_tao` `NULL`, and `COALESCE(fee_tao, 0)` converts "this extrinsic has no fee concept at all" into "this extrinsic cost 0 TAO". Every block carries at least a `Timestamp.set`, so once inherents are ≥50% of the rows in a day the median lands exactly on a fabricated zero.

That is the state in production today: `/api/v1/chain/fees?window=7d` returns `median_fee_tao: 0` and `median_tip_tao: 0` on all seven days while `avg_fee_tao` sits at 0.00012–0.00037. The published median is not a small number — it is a wrong number, and it is the field a user reaches for to answer "what does a transaction cost right now".

The same dilution hits the averages, from the other direction. `src/chain-analytics.ts:370-377` computes `avg_fee_tao = total_fee_tao / extrinsic_count`, and `extrinsic_count` comes from `COUNT(*)` at `workers/data-api.ts:8614` over the same unfiltered set. The numerator only ever accumulates real fees (unsigned rows contribute `0`), but the denominator counts inherents — so the average is understated by roughly the inherent share of all extrinsics.

`total_fee_tao` and `total_tip_tao` are **not** affected: `SUM(COALESCE(fee_tao, 0))` over the unfiltered set equals the sum over signed rows, since unsigned rows contribute `0`. Do not change them.

**The sibling query four lines above already does exactly the right thing.** `workers/data-api.ts:8626-8631` (the `top_fee_payers` aggregation) reads:

```sql
FROM extrinsics WHERE observed_at >= ${cutoff} AND signer IS NOT NULL
```

Copy that predicate — `AND signer IS NOT NULL`, in the same position, on the same table, in the same handler. Do not introduce a different notion of "real transaction" (a `fee_tao IS NOT NULL` filter, a pallet allowlist/denylist, an `is_inherent` column, a new view, or filtering in TypeScript after the fact). The signer predicate is the established definition of a signed extrinsic in this file and it is the one to reuse.

## Requirements

This issue has **two distinct fixes and both are required.** A PR that lands only the median fix does not satisfy it.

### Fix 1 — the median (mechanical)

1. `workers/data-api.ts:8632-8641`: add `AND signer IS NOT NULL` to the `WHERE` clause, matching `:8630`.
2. In the same query, drop both `COALESCE(...)` wrappers so the ordering expressions become `ORDER BY fee_tao` and `ORDER BY tip_tao`. `PERCENTILE_CONT` ignores `NULL`s in its ordered set, so a signed extrinsic whose fee was never recorded is excluded rather than counted as a free transaction. Keeping the `COALESCE` after adding the signer filter would still fabricate zeros for those rows.
3. A day with signed extrinsics but no recorded fees must yield `median_fee_tao: null`, not `0`. `src/chain-analytics.ts:371-377` already routes a missing median to `null` via `toNullableTao`; make sure the new gating (see Fix 2) does not turn that `null` into a `0`.

### Fix 2 — the average (needs a new field)

4. Add a **new** per-day field `signed_extrinsic_count`: `COUNT(*) FILTER (WHERE signer IS NOT NULL)` in the daily query at `workers/data-api.ts:8613-8619`. `extrinsic_count` keeps its current meaning — every extrinsic in the day, inherents included — because it is already published as such through the REST route, the CSV export, the MCP tool, and GraphQL. Do **not** redefine `extrinsic_count`.
5. `src/chain-analytics.ts:335-400` (`buildChainFees`): `avg_fee_tao` and `avg_tip_tao` become `total_fee_tao / signed_extrinsic_count` and `total_tip_tao / signed_extrinsic_count`, and are `null` (never `0`, never `NaN`) when `signed_extrinsic_count` is `0`. `median_fee_tao` / `median_tip_tao` gate on `signed_extrinsic_count > 0` for the same reason, replacing the current `extrinsicCount > 0` gate at `:372` and `:376`. Add `signed_extrinsic_count: number` to the `ChainFeesDay` interface (`src/chain-analytics.ts:305-314`) and emit it on every row, including on the cold/empty path, so the shape is schema-stable across tiers.

### Fix 3 — publish the new field consistently

6. `schemas-src/routes/chain-analytics.ts:119-130` (`ChainFeeDaySchema`, a `.strict()` object): add `signed_extrinsic_count: z.int().min(0)`. Then run `npm run build` and commit the regenerated `openapi.json` and generated types **in the same PR**, or `validate:contract-drift` fails CI.
7. `schemas-src/mcp-tools/chain-calls-fees.ts:89-101` (`ChainFeesDaySchema`): add `signed_extrinsic_count: z.int().nullable().optional()`, matching the shape of its siblings there.
8. `src/graphql-sdl.ts:1217-1227` (`type ChainFeesDay`): add `signed_extrinsic_count: Int!` and update the type's description, which currently claims "avg/median are null on a zero-extrinsic day" — that is no longer the gate. `src/graphql.ts:6522-6526`: map the new field through with the same `?? 0` default the sibling `extrinsic_count` uses.
9. `workers/request-handlers/analytics.ts:877-885` (`CHAIN_FEES_CSV_COLUMNS`): add `signed_extrinsic_count` directly after `extrinsic_count`, and update the matching CSV sample header and row at `src/contracts.ts:5826-5830` so they stay column-consistent.
10. `src/contracts.ts:1782` (artifact description) and `src/contracts.ts:4131` (route description): both currently promise "exact medians" / "exact ordered-offset medians" with no qualifier. State that averages and medians are over **signed** extrinsics only, while `extrinsic_count` counts all extrinsics including inherents.

### Coverage

11. `codecov/patch` enforces 99% on every changed line and branch in `src/**` / `workers/**`. Both branches of the new `signed_extrinsic_count > 0` gate must be exercised. Measure unsharded with `npm run test:coverage`. All new test files are `.ts`.

## Deliverables

- [ ] `workers/data-api.ts` — median query carries `AND signer IS NOT NULL` and no `COALESCE` on the ordering expressions
- [ ] `workers/data-api.ts` — daily query emits `signed_extrinsic_count` via `COUNT(*) FILTER (WHERE signer IS NOT NULL)`, with `extrinsic_count` unchanged
- [ ] `src/chain-analytics.ts` — `ChainFeesDay.signed_extrinsic_count` added; `avg_fee_tao` / `avg_tip_tao` divide by it; medians gate on it; all four are `null` when it is `0`
- [ ] `schemas-src/routes/chain-analytics.ts` + regenerated `openapi.json` and generated types, committed in the same PR
- [ ] `schemas-src/mcp-tools/chain-calls-fees.ts`, `src/graphql-sdl.ts`, `src/graphql.ts` — new field published on all three surfaces
- [ ] `workers/request-handlers/analytics.ts` CSV columns + `src/contracts.ts` CSV sample updated together
- [ ] `src/contracts.ts:1782` and `:4131` descriptions state the signed-only basis for averages and medians
- [ ] Regression tests in `tests/chain-analytics.test.ts` that **fail before this change and pass after**: a day whose rows are majority unsigned-inherent yields a nonzero `median_fee_tao` reflecting only the signed rows, and `avg_fee_tao` computed over `signed_extrinsic_count` rather than `extrinsic_count`
- [ ] Test in `tests/data-api.test.ts` asserting the emitted median SQL contains `signer IS NOT NULL` and does not contain `COALESCE(fee_tao` — that file already captures emitted SQL text
- [ ] Test covering `signed_extrinsic_count: 0` on a day with only inherents: all four avg/median fields `null`, `extrinsic_count` still nonzero

## Expected Outcome

`/api/v1/chain/fees` (and its GraphQL, MCP, and CSV projections) report `median_fee_tao` / `median_tip_tao` computed over signed extrinsics only — no longer pinned to `0` by `Timestamp.set` inherents — and `avg_fee_tao` / `avg_tip_tao` divided by the new `signed_extrinsic_count` rather than by the inherent-inflated `extrinsic_count`, with all four resolving to `null` on a day with no signed extrinsics. `extrinsic_count` still counts every extrinsic. The median query's `WHERE` clause is byte-for-byte consistent with the top-fee-payer query's at `workers/data-api.ts:8630`.

## Links & Resources

- `workers/data-api.ts:8604-8645` — the three fee queries (daily, top payers, medians)
- `workers/data-api.ts:8626-8631` — the sibling query whose `AND signer IS NOT NULL` is the precedent
- `apps/indexer-rs/src/main.rs:985-1000` — where unsigned inherents become `extrinsics` rows
- `src/chain-analytics.ts:305-400` — `ChainFeesDay` and `buildChainFees`
- `schemas-src/routes/chain-analytics.ts:119-150`, `schemas-src/mcp-tools/chain-calls-fees.ts:89-101`, `src/graphql-sdl.ts:1217-1227`
- Original fee-market work: #1988

bitfathers94:
fix(api): chain/fees medians and averages no longer diluted by unsigned inherents

The median query counted every extrinsic including unsigned inherents
(Timestamp.set etc.), so COALESCE(fee_tao, 0) pinned median_fee_tao/
median_tip_tao to 0 once inherents made up half a day's rows. The
average divided by the same inherent-inflated extrinsic_count,
understating avg_fee_tao/avg_tip_tao.

Filter the median query to signer IS NOT NULL (matching the sibling
top_fee_payers query) and drop the COALESCE wrapping so PERCENTILE_CONT
ignores unrecorded fees instead of treating them as free. Add a new
signed_extrinsic_count field and divide the averages by it instead of
extrinsic_count, which keeps counting every extrinsic including
inherents. Both are null on a day with no signed extrinsics.

Published the new field across the REST schema, GraphQL SDL/resolver,
MCP tool schema, and CSV export, and regenerated the OpenAPI artifacts
and generated types.

Closes #8809

### DIFF
--- a/generated/graphql/types.ts
+++ b/generated/graphql/types.ts
@@ -1110,7 +1110,7 @@ export type ChainFees = {
   window: Scalars['String']['output'];
 };
 
-/** One UTC day's fee/tip aggregate: extrinsic count, total/avg/median fee and tip in TAO (avg/median are null on a zero-extrinsic day). */
+/** One UTC day's fee/tip aggregate: extrinsic count, total/avg/median fee and tip in TAO. extrinsic_count includes unsigned inherents; avg/median are computed over signed_extrinsic_count and are null on a day with no signed extrinsics. */
 export type ChainFeesDay = {
   __typename?: 'ChainFeesDay';
   avg_fee_tao?: Maybe<Scalars['Float']['output']>;
@@ -1119,6 +1119,7 @@ export type ChainFeesDay = {
   extrinsic_count: Scalars['Int']['output'];
   median_fee_tao?: Maybe<Scalars['Float']['output']>;
   median_tip_tao?: Maybe<Scalars['Float']['output']>;
+  signed_extrinsic_count: Scalars['Int']['output'];
   total_fee_tao?: Maybe<Scalars['Float']['output']>;
   total_tip_tao?: Maybe<Scalars['Float']['output']>;
 };
@@ -6908,6 +6909,7 @@ export type ChainFeesDayResolvers<ContextType = GqlContext, ParentType extends R
   extrinsic_count?: Resolver<ResolversTypes['Int'], ParentType, ContextType>;
   median_fee_tao?: Resolver<Maybe<ResolversTypes['Float']>, ParentType, ContextType>;
   median_tip_tao?: Resolver<Maybe<ResolversTypes['Float']>, ParentType, ContextType>;
+  signed_extrinsic_count?: Resolver<ResolversTypes['Int'], ParentType, ContextType>;
   total_fee_tao?: Resolver<Maybe<ResolversTypes['Float']>, ParentType, ContextType>;
   total_tip_tao?: Resolver<Maybe<ResolversTypes['Float']>, ParentType, ContextType>;
 }>;
--- a/packages/contract/index.d.ts
+++ b/packages/contract/index.d.ts
@@ -2367,7 +2367,7 @@ export interface paths {
             path?: never;
             cookie?: never;
         };
-        /** Fetch fee/tip market analytics — a per-UTC-day fee series (totals, averages, and exact ordered-offset medians) plus a windowed top-fee-payer list — over a 7d or 30d window, optionally scoped to one pallet with ?call_module=. Computed live from the first-party extrinsics D1 tier (#1988); schema-stable day_count:0 + empty lists when cold. */
+        /** Fetch fee/tip market analytics — a per-UTC-day fee series (totals, averages, and exact ordered-offset medians, computed over signed extrinsics only) plus a windowed top-fee-payer list — over a 7d or 30d window, optionally scoped to one pallet with ?call_module=. extrinsic_count counts every extrinsic including unsigned inherents; signed_extrinsic_count is the denominator behind avg/median. Computed live from the first-party extrinsics D1 tier (#1988); schema-stable day_count:0 + empty lists when cold. */
         get: operations["chainFees"];
         put?: never;
         post?: never;
@@ -6379,6 +6379,7 @@ export interface components {
                 extrinsic_count: number;
                 median_fee_tao: number | null;
                 median_tip_tao: number | null;
+                signed_extrinsic_count: number;
                 total_fee_tao: number;
                 total_tip_tao: number;
             }[];
@@ -28594,6 +28595,7 @@ export interface operations {
                      *             "extrinsic_count": 1,
                      *             "median_fee_tao": 0.5,
                      *             "median_tip_tao": 0.5,
+                     *             "signed_extrinsic_count": 1,
                      *             "total_fee_tao": 0.5,
                      *             "total_tip_tao": 0.5
                      *           }
@@ -28641,8 +28643,8 @@ export interface operations {
                         data?: components["schemas"]["ChainFeesArtifact"];
                     };
                     /**
-                     * @example day,extrinsic_count,total_fee_tao,avg_fee_tao,median_fee_tao,total_tip_tao,avg_tip_tao,median_tip_tao
-                     *     2026-07-01,15000,42.5,0.002833,0.0025,0,0,0
+                     * @example day,extrinsic_count,signed_extrinsic_count,total_fee_tao,avg_fee_tao,median_fee_tao,total_tip_tao,avg_tip_tao,median_tip_tao
+                     *     2026-07-01,15000,9200,42.5,0.004620,0.0025,0,0,0
                      */
                     "text/csv": string;
                 };
--- a/public/metagraph/api-index.json
+++ b/public/metagraph/api-index.json
@@ -10038,7 +10038,7 @@
     {
       "artifact_path": "/metagraph/chain/fees.json",
       "cache": "short",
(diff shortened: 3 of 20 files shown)


JSONbored (commented):
Failing CI, closing.

PR #8852, opened 2026-07-31
fix(api): gate github account upsert with internal-token check

### LINKED ISSUE #8820: fix(api): gate handleGithubAccountUpsert with the internal-token check its siblings all carry
## Context

`handleGithubAccountUpsert` (`workers/data-api.ts:5057-5088`, dispatched at
`workers/data-api.ts:6304-6309` for `POST /api/v1/auth/github/upsert-account`) is the only write
route in `workers/data-api.ts` with **no authorization gate of any kind**. It reads the request body,
shape-checks two fields, and writes:

```sql
INSERT INTO github_accounts (github_user_id, github_login, created_at, last_login_at)
VALUES (${githubUserId}, ${githubLogin}, ${now}, ${now})
ON CONFLICT (github_user_id) DO UPDATE
  SET github_login = ${githubLogin}, last_login_at = ${now}
RETURNING id, github_login, tier
```

A caller who reached it could rebind an existing GitHub account's login to a value of their choosing
by posting a known `github_user_id` — GitHub numeric user ids are public.

### This is NOT currently exploitable — read this before assessing severity

This issue is filed as defense-in-depth, not as a live vulnerability. Three independent facts, all
verified against `main`:

1. `workers/data-api.ts` is service-binding-only. `wrangler.data.jsonc` sets `"workers_dev": false`
   and `"preview_urls": false` and declares no routes, so the Worker has no public address.
2. **The edge Worker never forwards this path.** `workers/api.ts` exact-matches only
   `/api/v1/auth/wallet/challenge` and `/api/v1/auth/wallet/verify` (`workers/api.ts:2134-2141`).
   There is no `startsWith("/api/v1/auth")` anywhere in `workers/api.ts`, so
   `POST /api/v1/auth/github/upsert-account` against `https://api.metagraph.sh` is a plain 404 that
   never reaches `DATA_API`.
3. The sole caller, `src/github-oauth.ts:368`, constructs a fresh `Request` to the binding **after** a
   verified GitHub code/token exchange, so the values it sends are already authenticated. The
   handler's own header comment at `workers/data-api.ts:5045-5056` states this contract correctly.

So nothing is broken in production today and there is no incident here.

### Why it is still worth closing

The route is anomalous. Roughly 25 sibling internal routes in the same file each carry a dedicated
shared-secret gate — `handleApiKeyVerify` (`workers/data-api.ts:5270-5285`), `handleAccountTierPromote`,
`handleUsageRollupIncrement` (`:5480-5493`), and the rest of the `/api/v1/internal/*` family — and the
inventory comment at `workers/data-api.ts:4795-4806` lists them as internal-only precisely because the
gate is the thing that makes that claim true. This one route's safety is instead a property of a
routing table in a *different* Worker.

The blast radius if that prefix is ever widened — someone adding
`pathname.startsWith("/api/v1/auth")` to `workers/api.ts` for a new auth route, which is a natural
thing to do — is unauthenticated GitHub-account rebinding. The gate costs one env var and eight
lines and removes the dependence on a negative fact in another file.

**Mirror `handleApiKeyVerify`'s gate, `workers/data-api.ts:5270-5285`** — the canonical two-step,
fail-closed shape used by every internal route in this file:

```ts
const configured = env.API_KEY_LOOKUP_INTERNAL_TOKEN;
if (!configured) {
  return writeJson({ error: "... is not provisioned on this deployment" }, 503);
}
const provided = request.headers.get(API_KEY_LOOKUP_TOKEN_HEADER) || "";
if (!provided || !timingSafeEqual(provided, configured)) {
  return writeJson({ error: `provide a valid ${API_KEY_LOOKUP_TOKEN_HEADER} header` }, 401);
}
```

The 503-before-401 ordering is deliberate and must be preserved: an unprovisioned deployment is a
config gap, distinct from a bad credential, and the distinction is what makes a broken deploy
diagnosable. The comparison must go through the existing `timingSafeEqual` helper in this file, not
`===`.

Do **not**:

- Invent a new secret, a new env var name, or a new header when the existing internal-token pair
  already covers exactly this trust boundary (caller is our own Worker over a service binding).
  Reuse `API_KEY_LOOKUP_INTERNAL_TOKEN` and `API_KEY_LOOKUP_TOKEN_HEADER`.
- Gate it with a wallet session, an OAuth token, an origin check, or an IP allowlist. The caller is a
  Worker, not a user.
- "Fix" it by adding a route in `workers/api.ts`, or by changing the edge routing in any way. The
  route must stay unreachable from the public edge; this issue makes it safe *if* that ever changes.
- Use `===` or `==` for the token comparison.
- Rewrite the handler's SQL, its field validation, or its response shape.

## Requirements

1. `handleGithubAccountUpsert` (`workers/data-api.ts:5057`) gains the gate above as its **first**
   statement, before `readAccountRouteBody` and before any field validation — so a caller without the
   token cannot even probe the body-shape error messages.

2. Exact behaviour:
   - `env.API_KEY_LOOKUP_INTERNAL_TOKEN` unset/empty → `503` with body
     `{ "error": "github account upsert is not provisioned on this deployment" }`.
   - Token configured, request header `x-api-key-lookup-token` (`API_KEY_LOOKUP_TOKEN_HEADER`)
     missing, empty, or not equal to it under `timingSafeEqual` → `401` with body
     `{ "error": "provide a valid x-api-key-lookup-token header" }` (constructed from the header
     constant, not a literal), matching `handleApiKeyVerify`'s wording.
   - Header matches → existing behaviour, byte-identical: the same 400 on a missing/non-integer
     `github_user_id` or missing/non-string `github_login`, and the same
     `{ id, github_login, tier }` 200 on success.

3. The caller must send the header. `src/github-oauth.ts:367-376` builds the request with only
   `content-type`; add `[API_KEY_LOOKUP_TOKEN_HEADER]: env.API_KEY_LOOKUP_INTERNAL_TOKEN`. This runs
   in the same Worker that already reads that secret at `workers/api.ts:488`, so no new secret and no
   new wrangler declaration is needed — widen `github-oauth.ts`'s local env type if it narrows. If you
   do touch a wrangler config, **never commit a secret value**; declaration only.

   When `env.API_KEY_LOOKUP_INTERNAL_TOKEN` is absent, the callback must return the **same** 503 it
   already returns for a missing `DATA_API` binding at `src/github-oauth.ts:361-366`
   (`"account storage is not provisioned on this deployment"`) rather than posting an unauthorized
   request and surfacing the resulting 401 as the generic 502 `"account storage failed"` at
   `src/github-oauth.ts:377-379`. An unprovisioned deploy must be diagnosable as one.

4. The handler's header comment at `workers/data-api.ts:5045-5056` currently explains why the route
   is safe *without* a gate. Update it to describe the gate instead, and state plainly that the
   route's safety no longer depends on `workers/api.ts` not forwarding `/api/v1/auth/*`.

5. Add the route to the internal-route inventory comment at `workers/data-api.ts:4795-4806`, in the
   same form as its siblings.

6. Regression tests. Extend the two existing suites — do not create a third file for this.

   In `tests/github-account-upsert-route.test.ts` (it already records every statement in its
   `sqlCalls` array, which is what makes the "no write attempted" assertions possible):
   - No token header → 401, and `sqlCalls` is **empty**. Asserting the status alone is not enough;
     the point is that no write is attempted. This must **fail on `main`**.
   - Wrong token header → 401, `sqlCalls` empty.
   - `API_KEY_LOOKUP_INTERNAL_TOKEN` absent from `env` → 503, `sqlCalls` empty, and the 503 is
     returned even when a header *is* supplied (fail-closed ordering).
   - Correct token + valid body → 200 with `{ id, github_login, tier }` and the upsert issued,
     byte-identical to today.
   - Correct token + invalid body (`github_user_id` non-integer, `github_login` missing) → the
     existing 400, `sqlCalls` empty.

   In `tests/github-oauth.test.ts`:
   - The callback's request to `DATA_API` carries the `x-api-key-lookup-token` header with the env
     value. Assert on the header observed by the stubbed binding.
   - With `API_KEY_LOOKUP_INTERNAL_TOKEN` absent from `env`, the callback returns 503 and the stubbed
     `DATA_API.fetch` is never called.

7. CI. `codecov/patch` enforces 99% on changed lines and branches in `workers/**`, branch-counted —
   each of the three new guard branches needs a test that takes it. Guards must be reachable by tests;
   do not mark them `v8 ignore`, which the patch gate counts anyway. Measure unsharded with
   `npm run test:coverage`. All new/changed test files are `.ts`.

## Deliverables

- [ ] Two-step 503-then-401 internal-token gate at the top of `handleGithubAccountUpsert`, using
      `API_KEY_LOOKUP_INTERNAL_TOKEN`, `API_KEY_LOOKUP_TOKEN_HEADER` and `timingSafeEqual`.
- [ ] `src/github-oauth.ts:367-376` sends `x-api-key-lookup-token`, and returns the existing
      "not provisioned" 503 when the secret is absent instead of a 502.
- [ ] Header comment at `workers/data-api.ts:5045-5056` rewritten to describe the gate.
- [ ] Route added to the internal-route inventory comment at `workers/data-api.ts:4795-4806`.
- [ ] Regression tests in `tests/github-account-upsert-route.test.ts` and `tests/github-oauth.test.ts`
      covering every case in Requirement 6, including the missing-token case that fails on `main` and
      asserts an empty `sqlCalls`.
- [ ] No change to `workers/api.ts` routing; the route stays unreachable from the public edge.

satisfy this issue.

## Expected Outcome

`POST /api/v1/auth/github/upsert-account` on the data Worker returns 503 when
`API_KEY_LOOKUP_INTERNAL_TOKEN` is unprovisioned and 401 when the `x-api-key-lookup-token` header is
missing or wrong — in both cases issuing no SQL — and behaves byte-identically to `main` when the
header matches, matching `handleApiKeyVerify`'s gate at `workers/data-api.ts:5270-5285`. The GitHub
OAuth callback at `src/github-oauth.ts:368` continues to work end to end. If someone later adds a
`/api/v1/auth` prefix forward to `workers/api.ts`, this route is no longer an unauthenticated
account-rebinding primitive.

## Links & Resources

- `workers/data-api.ts:5045-5088` — `handleGithubAccountUpsert`, the ungated route
- `workers/data-api.ts:6304-6309` — its dispatch
- `workers/data-api.ts:5270-5285` — `handleApiKeyVerify`, the gate shape to copy
- `workers/data-api.ts:5480-5493` — `handleUsageRollupIncrement`, the same gate on another internal
  route
- `workers/data-api.ts:4795-4806` — the internal-route inventory comment to extend
- `src/github-oauth.ts:361-379` — the sole caller, which must start sending the header
- `tests/github-account-upsert-route.test.ts`, `tests/github-oauth.test.ts` — the two suites to extend
- `workers/api.ts:2134-2141` — the exact-match auth routing that is the only reason this is latent
  today; do not change it

kai392:
## Summary

`handleGithubAccountUpsert` was the only write route in `workers/data-api.ts` with no authorization gate. Sibling internal routes all use `API_KEY_LOOKUP_INTERNAL_TOKEN` + `timingSafeEqual`. This PR adds that same fail-closed gate and updates the OAuth callback to send the header.

Closes #8820

## Root cause

The route relied on a negative fact in another Worker (`workers/api.ts` not forwarding `/api/v1/auth/*`) rather than an explicit shared-secret check. If that prefix were ever widened, unauthenticated GitHub-account rebinding would become possible.

## Fix approach

1. Gate `handleGithubAccountUpsert` first (before body parse): unset secret → 503; bad/missing header → 401; match → existing behaviour.
2. `src/github-oauth.ts` sends `x-api-key-lookup-token`; absent secret returns the existing "not provisioned" 503 without calling `DATA_API`.
3. Update inventory + handler header comments.
4. Extend `tests/github-account-upsert-route.test.ts` and `tests/github-oauth.test.ts`.

## Testing

- [x] `npx vitest run tests/github-account-upsert-route.test.ts tests/github-oauth.test.ts` (44/44)
- [x] Missing/wrong token → 401, empty `sqlCalls`
- [x] Unprovisioned secret → 503 even with header
- [x] Happy path unchanged; OAuth callback asserts header

## Risks

- Deployments without `API_KEY_LOOKUP_INTERNAL_TOKEN` now get an explicit 503 from this route (previously would have written). That secret is already required for other internal routes on the same Worker.
- No change to public edge routing.

### DIFF
--- a/src/github-oauth.ts
+++ b/src/github-oauth.ts
@@ -50,6 +50,7 @@ import type {
   OAuthHelpers,
   OAuthProviderOptions,
 } from "@cloudflare/workers-oauth-provider";
+import { API_KEY_LOOKUP_TOKEN_HEADER } from "./api-key-validation.ts";
 
 const GITHUB_AUTHORIZE_URL = "https://github.com/login/oauth/authorize";
 const GITHUB_TOKEN_URL = "https://github.com/login/oauth/access_token";
@@ -364,10 +365,22 @@ export async function handleGithubOAuthCallback(
       { status: 503 },
     );
   }
+  // Same internal-token gate the upsert handler requires (#8820). Absent
+  // secret must surface as the existing "not provisioned" 503 rather than a
+  // 502 from posting without the header.
+  if (!env.API_KEY_LOOKUP_INTERNAL_TOKEN) {
+    return new Response(
+      "account storage is not provisioned on this deployment",
+      { status: 503 },
+    );
+  }
   const upsertResponse = await env.DATA_API.fetch(
     new Request("https://internal/api/v1/auth/github/upsert-account", {
       method: "POST",
-      headers: { "content-type": "application/json" },
+      headers: {
+        "content-type": "application/json",
+        [API_KEY_LOOKUP_TOKEN_HEADER]: env.API_KEY_LOOKUP_INTERNAL_TOKEN,
+      },
       body: JSON.stringify({
         github_user_id: githubUserId,
         github_login: githubLogin,
--- a/tests/github-account-upsert-route.test.ts
+++ b/tests/github-account-upsert-route.test.ts
@@ -1,12 +1,13 @@
 // Unit tests for workers/data-api.ts's handleGithubAccountUpsert
-// (metagraphed#7151) -- POST /api/v1/auth/github/upsert-account, reached
+// (metagraphed#7151, #8820) -- POST /api/v1/auth/github/upsert-account, reached
 // only via the DATA_API service binding from src/github-oauth.ts's
 // callback handler (see that module's own test file for the OAuth-flow
 // side). Mirrors tests/wallet-auth-keys-route.test.ts's shape: its own
 // per-test postgres mock queue, scoped only to this file (vi.mock is
-// per-test-file).
+// per-test-file). Gated by API_KEY_LOOKUP_INTERNAL_TOKEN (#8820).
 import assert from "node:assert/strict";
 import { beforeEach, test, vi } from "vitest";
+import { API_KEY_LOOKUP_TOKEN_HEADER } from "../src/api-key-validation.ts";
 import type { AnyFn, Row } from "./row-type.ts";
 
 const mockQueue = vi.hoisted((): { current: Row[] } => ({ current: [] }));
@@ -31,9 +32,12 @@ vi.mock("postgres", () => ({
 
 const { default: worker } = await import("../workers/data-api.ts");
 
+const INTERNAL_TOKEN = "test-api-key-lookup-token";
+
 function baseEnv(overrides = {}) {
   return {
     HYPERDRIVE: { connectionString: "postgres://mock" },
+    API_KEY_LOOKUP_INTERNAL_TOKEN: INTERNAL_TOKEN,
     ...overrides,
   };
 }
@@ -43,10 +47,14 @@ beforeEach(() => {
   sqlCalls.length = 0;
 });
 
-function req(body: Row) {
+function req(body: Row, { token = INTERNAL_TOKEN }: { token?: string | null } = {}) {
+  const headers: Record<string, string> = {
+    "content-type": "application/json",
+  };
+  if (token != null) headers[API_KEY_LOOKUP_TOKEN_HEADER] = token;
   return new Request("https://d/api/v1/auth/github/upsert-account", {
     method: "POST",
-    headers: { "content-type": "application/json" },
+    headers,
     body: JSON.stringify(body),
   });
 }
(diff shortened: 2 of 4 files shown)


kai392:
Closing to stay within the 2-open-PR contributor cap. Will reopen a fresh PR for #8820 once a slot frees (CI/work preserved on branch fix/github-upsert-internal-token-8820).

Issue #8828, opened 2026-07-31
fix(mcp): get_feed "upgrades" kind is missing from the published input and output enums


JSONbored:
## Context

The `get_feed` MCP tool supports a fifth `kind`, `"upgrades"`, at runtime — but that value is missing
from **both** the published input schema and the published output schema, so a schema-validating MCP
client cannot call it and, if it does get called, the response violates the tool's own output contract.

There are two `FEED_KINDS` constants and they have drifted:

- `schemas-src/mcp-tools/feed.ts:10` (the published contract)
  ```ts
  const FEED_KINDS = ["registry", "incidents", "gaps", "subnet"] as const;
  ```
  used for the input enum at `:15` and the output enum at `:38`.
- `src/feed-mcp.ts:36-44` (the runtime)
  ```ts
  export const FEED_KINDS = [
    "registry",
    "incidents",
    "gaps",
    // #8702: Bittensor runtime upgrade activity -- releases, observed chain
    // spec-version changes, and BIT documents.
    "upgrades",
    "subnet",
  ];
  ```

`requireKind` (`src/feed-mcp.ts:59-73`) validates against the runtime list, and `loadFeedItems`
dispatches the kind at `src/feed-mcp.ts:215-216` into `loadUpgradeFeedItems`. The path works end to end.

This is not a hidden capability — **the tool's own description tells agents to use it.** The
`get_feed` description (`src/feed-mcp.ts:257-271`) says the tool mirrors
`/api/v1/feeds/upgrades` and that "The `upgrades` kind carries Bittensor runtime upgrade activity".
An agent reads that sentence, sends `{ kind: "upgrades" }`, and a client that validates against the
published `inputSchema` rejects the call before it leaves the client. If the client does not validate,
the call succeeds and returns `kind: "upgrades"` — which violates the published `outputSchema` enum
instead.

The test that should have caught it did not: `tests/feed-mcp.test.ts:24` is titled
`"returns each of the four valid FEED_KINDS unchanged"` while iterating the **runtime** constant, which
has had five entries since #8702. It never touches `GetFeedInputSchema`, so the two constants have no
assertion binding them together.

## Reference implementation to mirror

There is no existing single-source-of-truth pattern for this constant to copy, so the shape is
specified explicitly below rather than pointed at.

**The fix must be a single source of truth. Adding the string `"upgrades"` to both lists does NOT
satisfy this issue** — that is precisely the state that drifted once already and would drift again the
[...]

PR #8886, opened 2026-07-31
fix(mcp): get_feed netuid is schema-optional but conditionally required and forbidden at runtime

### LINKED ISSUE #8829: fix(mcp): get_feed `netuid` is schema-optional but conditionally required and forbidden at runtime
## Context

`get_feed`'s `netuid` argument is published as plainly optional, but at runtime it is **conditionally
required** and **conditionally forbidden**. Two calls that the published schema says are valid are
rejected by the handler.

Published contract, `schemas-src/mcp-tools/feed.ts:14-22`:

```ts
export const GetFeedInputSchema = z
  .object({
    kind: z.enum(FEED_KINDS),
    netuid: z.int().min(0).optional(),
    ...
```

`kind` is the only required key. Runtime, `resolveNetuid` (`src/feed-mcp.ts:76-96`):

- `kind === "subnet"` and no `netuid` → throws `invalid_params`
  ("Argument \`netuid\` is required and must be a non-negative integer when kind is \`subnet\`.")
- `kind !== "subnet"` and `netuid` present → throws `invalid_params`
  ("Argument \`netuid\` is only used when kind is \`subnet\`.")

So `{ kind: "subnet" }` and `{ kind: "registry", netuid: 64 }` both pass every published constraint and
both fail at the tool. An agent has no way to learn the real rule from the schema; it discovers it by
burning a failed tool call. The handler's own comment at `src/feed-mcp.ts:72-75` states the intent
("reject it there rather than silently ignoring a param the caller thinks is doing something") — the
behavior is correct, it is just invisible in the contract.

**The repo already has the mechanism for exactly this: `requireAnyOf` at `src/mcp-server.ts:3188-3193`.**

```ts
function requireAnyOf(
  schema: Record<string, unknown>,
  keys: string[],
): Record<string, unknown> {
  return { ...schema, anyOf: keys.map((key) => ({ required: [key] })) };
}
```

It is used by `how_do_i_call` (`src/mcp-server.ts:9833-9836`) and `verify_integration`
(`src/mcp-server.ts:9925-9928`), each wrapping the `z.toJSONSchema(...)` result so the emitted tool
schema carries a constraint Zod itself cannot express. Its docblock (`src/mcp-server.ts:3180-3187`)
spells out why this shape and not another:

> A Zod `.refine()` cannot fix this — z.toJSONSchema drops refinements silently, so the constraint
> would still be invisible. `anyOf` with bare `required` clauses is the standard JSON Schema spelling
> and is what clients actually validate against, while leaving Zod parsing (and the inferred TS input
> type) untouched.

**Extend that mechanism — add a sibling helper next to `requireAnyOf` in `src/mcp-server.ts`, and
apply it where `GET_FEED_MCP_TOOL` is spread into `MCP_TOOLS` (`src/mcp-server.ts:9148-9150`), by
overriding `inputSchema` right after the spread.** `get_feed`'s dependency is richer than "at least one
of these keys", so `requireAnyOf` itself cannot be reused verbatim — but the shape of the fix (a small
pure function that patches `anyOf` onto the already-emitted JSON Schema object, leaving the Zod schema
and the inferred TS type alone) must be identical.

Apply it in `src/mcp-server.ts`, **not** in `src/feed-mcp.ts`. `src/mcp-server.ts:431-434` imports
`GET_FEED_MCP_TOOL` from `src/feed-mcp.ts` and `src/feed-mcp.ts` imports nothing from `src/mcp-server.ts`;
importing the helper the other way round would introduce a module cycle.

- A `.refine()` / `.superRefine()` on `GetFeedInputSchema`. Zod drops it from `z.toJSONSchema` output,
  so the published contract would be unchanged — the entire point of the issue.
- Splitting `get_feed` into two tools, or making `netuid` unconditionally required.
- Changing `resolveNetuid`'s runtime behavior. It is correct; only the published schema is missing.
- A hand-written `inputSchema` object literal replacing the `z.toJSONSchema(...)` call. No tool in this
  repo has a hand-written `inputSchema` and none should start.

## Requirements

1. `get_feed`'s published `inputSchema` must encode the dependency as JSON Schema `anyOf`, so a
   validating client rejects `{ kind: "subnet" }` and `{ kind: "registry", netuid: 64 }` before the call
   is made. The required spelling:

   ```json
   "anyOf": [
     {
       "properties": { "kind": { "const": "subnet" } },
       "required": ["kind", "netuid"]
     },
     {
       "properties": { "kind": { "enum": ["registry", "incidents", "gaps", "upgrades"] } },
       "not": { "required": ["netuid"] }
     }
   ]
   ```

2. The non-subnet branch's enum must be **derived** from the kinds constant, not hardcoded — the list of
   non-subnet kinds is `FEED_KINDS` minus `"subnet"`, computed at build time. A hardcoded literal drifts
   the moment a sixth kind is added. (Related: #8702 added `"upgrades"` and the published enums were
   never updated — that is #8828. If #8828 lands first, derive from the shared
   constant; if this lands first, derive from whichever `FEED_KINDS` is in scope so the follow-up gets
   it for free.)
3. The helper lives beside `requireAnyOf` in `src/mcp-server.ts`, is exported, and is a pure function
   over an already-emitted schema object — same signature shape, same "patch the JSON Schema, never the
   Zod schema" contract. Give it a docblock that says what dependency it encodes, matching the density
   of `requireAnyOf`'s.
4. `GetFeedInputSchema` itself does not change. `z.infer<typeof GetFeedInputSchema>` must be identical
   before and after, and `GetFeedInputSchema.parse({ kind: "subnet" })` must still succeed — runtime
   enforcement stays in `resolveNetuid`.
5. `resolveNetuid`'s two error messages and codes are unchanged.

## Deliverables

- [ ] New exported helper in `src/mcp-server.ts`, placed next to `requireAnyOf` (`:3188-3193`), with a
      docblock
- [ ] `src/mcp-server.ts:9148-9150` — the `get_feed` entry overrides `inputSchema` with the new helper
      applied to `GET_FEED_MCP_TOOL.inputSchema`, exactly as `how_do_i_call` does at `:9833-9836`
- [ ] A regression test asserting the `get_feed` entry in `MCP_TOOLS` publishes an `inputSchema`
      containing the `anyOf` above, and that an ajv validation of
      `{ kind: "subnet" }` and of `{ kind: "registry", netuid: 64 }` both **fail** against it while
      `{ kind: "subnet", netuid: 64 }` and `{ kind: "registry" }` both pass. It must fail before this
      change and pass after
- [ ] A unit test for the new helper itself covering both branches it emits
- [ ] `npm run validate:mcp` and the scoped test files green locally

satisfy this issue. In particular, a PR that adds the helper without applying it, or applies it without
the ajv-level assertion that the two currently-valid-but-rejected calls now fail validation, does not.

Two repo rules that apply here: `src/**` and `schemas-src/**/*.ts` are inside `codecov/patch`'s scope
(`vitest.config.ts:94-103`) at 99% branch-counted, so every branch of the new helper needs a test; and
do **not** bump `MCP_SERVER_VERSION` or `server.json`'s `version` — `sync-mcp-version` handles that
after merge.

## Expected Outcome

`get_feed`'s published `inputSchema` carries an `anyOf` that makes `netuid` required exactly when
`kind` is `"subnet"` and forbidden otherwise, so a schema-validating MCP client rejects both
`{ kind: "subnet" }` and `{ kind: "registry", netuid: 64 }` locally instead of round-tripping to an
`invalid_params` toolError — encoded with the same `anyOf`-patch mechanism `how_do_i_call` and
`verify_integration` already use, with `GetFeedInputSchema` and its inferred TS type untouched.

## Links & Resources

- Published schema: `schemas-src/mcp-tools/feed.ts:14-22`
- Runtime rule: `src/feed-mcp.ts:72-96` (`resolveNetuid`, plus the comment stating the intent)
- Tool definition / `inputSchema` construction: `src/feed-mcp.ts:254-275`
- Where to apply the wrapper: `src/mcp-server.ts:9148-9150` (the `MCP_TOOLS` entry)
- Mechanism to extend: `src/mcp-server.ts:3180-3193` (`requireAnyOf` + its docblock)
- Existing users of that mechanism: `src/mcp-server.ts:9833-9836`, `:9925-9928`
- Related (same tool, independent fix): #8828

shin-core:
## Summary

`get_feed`'s `netuid` argument is published as plainly `.optional()`, but at
runtime `resolveNetuid` (`src/feed-mcp.ts:76-96`) makes it **conditionally
required** (`kind === "subnet"`) and **conditionally forbidden** (any other
kind). So `{ kind: "subnet" }` and `{ kind: "registry", netuid: 64 }` both pass
every published constraint yet both fail at the handler — an agent can only
learn the real rule by burning a failed tool call. `z.toJSONSchema` drops
`.refine()`, so the dependency can't be expressed in Zod.

**Fix** — same mechanism `how_do_i_call` / `verify_integration` already use
(`requireAnyOf`): patch the emitted JSON Schema with `anyOf`, never touch the
Zod schema.

1. New exported helper `requireNetuidWhenSubnet` beside `requireAnyOf`
   (`src/mcp-server.ts`), with a docblock matching its density. It emits:
   ```json
   "anyOf": [
     { "properties": { "kind": { "const": "subnet" } }, "required": ["kind", "netuid"] },
     { "properties": { "kind": { "enum": [<FEED_KINDS minus subnet>] } }, "not": { "required": ["netuid"] } }
   ]
   ```
   The non-subnet branch's enum is **derived** from `FEED_KINDS` (imported from
   `src/feed-mcp.ts` — the one-way import direction, no cycle), so a new kind
   (e.g. #8828's `upgrades`) can never drift out of it.
2. Applied where `GET_FEED_MCP_TOOL` is spread into `MCP_TOOLS` by overriding
   `inputSchema`, exactly as `how_do_i_call` does.

`GetFeedInputSchema` and its inferred TS type are unchanged;
`GetFeedInputSchema.parse({ kind: "subnet" })` still succeeds; `resolveNetuid`'s
two messages/codes are unchanged (runtime enforcement stays there).

## Tests that fail on `main`

- `get_feed publishes an anyOf making netuid required iff kind=subnet` — asserts
  the published `inputSchema` carries the `anyOf`, and ajv rejects
  `{ kind: "subnet" }` and `{ kind: "registry", netuid: 64 }` while accepting
  `{ kind: "subnet", netuid: 64 }` and `{ kind: "registry" }`. **Fails on
  `main`** (no `anyOf` in the schema).
- `requireNetuidWhenSubnet emits both branches with the non-subnet enum derived
  from FEED_KINDS` — unit test covering both emitted branches, that the original
  schema is preserved, that the non-subnet enum excludes `subnet` and is derived
  from `FEED_KINDS`.

Verified by reverting `src/mcp-server.ts` with the tests kept → the anyOf test
fails; restored → all pass. The 12 existing `get_feed` runtime tests are
unaffected.

## Validation

- `npx vitest run tests/mcp-server.test.ts -t "get_feed"` — 14 passed (12 existing + 2 new)
- Reverted source, tests kept → the anyOf test fails; restored → all pass
- `npm run validate:mcp` — passed (206 tools)
- `npm run typecheck` — clean
- `npx eslint src/mcp-server.ts tests/mcp-server.test.ts` — 0 problems
- `npx prettier --check` — clean
- Patch coverage: every changed line and branch arm in `src/mcp-server.ts`
  exercised (both helper branches + the apply site) — 0 uncovered among changed
  lines
- `npm run build` produces no NEW committed-artifact drift attributable to this
  change (`operational-surfaces.json` drifts identically with the source
  stashed — pre-existing, reverted, not committed)

Diff is exactly two files: `src/mcp-server.ts` and `tests/mcp-server.test.ts`.
No `src/feed-mcp.ts` change, no `MCP_SERVER_VERSION`/`server.json` bump
(`sync-mcp-version` handles that post-merge), no `apps/ui/**`.

## Template Used

Backend/code change (`?template=backend-code.md`)

Closes #8829

### DIFF
--- a/src/mcp-server.ts
+++ b/src/mcp-server.ts
@@ -427,6 +427,7 @@ import { SAVED_QUERY_TEMPLATES, runSavedQuery } from "./saved-queries.ts";
 import { decodeEvmPrecompileCall } from "./evm-precompiles.ts";
 import { H160_PATTERN, loadAddressMapping } from "./address-mapping.ts";
 import {
+  FEED_KINDS,
   GET_FEED_INSTRUCTIONS,
   GET_FEED_MCP_TOOL,
   GET_FEED_OUTPUT_SCHEMA,
@@ -3195,6 +3196,37 @@ function requireAnyOf(
   keys: string[],
 ): Record<string, unknown> {
   return { ...schema, anyOf: keys.map((key) => ({ required: [key] })) };
+}
+
+/**
+ * get_feed's `netuid` is `.optional()` in Zod but resolveNetuid
+ * (src/feed-mcp.ts:76-96) makes it conditionally REQUIRED (kind === "subnet")
+ * and conditionally FORBIDDEN (any other kind) -- a dependency z.toJSONSchema
+ * cannot express, so the published schema said `{ kind: "subnet" }` and
+ * `{ kind: "registry", netuid: 64 }` were both valid while the handler rejects
+ * both. Same "patch the emitted JSON Schema, never the Zod schema" contract as
+ * requireAnyOf above: encode the dependency as `anyOf` (the standard JSON
+ * Schema spelling clients validate against), leaving Zod parsing and the
+ * inferred TS input type untouched. The non-subnet branch's enum is DERIVED
+ * from FEED_KINDS minus "subnet" so a new kind can never drift out of it.
+ */
+export function requireNetuidWhenSubnet(
+  schema: Record<string, unknown>,
+): Record<string, unknown> {
+  const nonSubnetKinds = FEED_KINDS.filter((kind) => kind !== "subnet");
+  return {
+    ...schema,
+    anyOf: [
+      {
+        properties: { kind: { const: "subnet" } },
+        required: ["kind", "netuid"],
+      },
+      {
+        properties: { kind: { enum: nonSubnetKinds } },
+        not: { required: ["netuid"] },
+      },
+    ],
+  };
 }
 
 export const MCP_TOOLS: McpToolDefinition[] = [
@@ -9190,6 +9222,10 @@ export const MCP_TOOLS: McpToolDefinition[] = [
   },
   {
     ...GET_FEED_MCP_TOOL,
+    // #8829: netuid is conditionally required (kind === "subnet") and forbidden
+    // otherwise -- a dependency Zod can't emit, so patch it onto the published
+    // inputSchema, same mechanism how_do_i_call / verify_integration use.
+    inputSchema: requireNetuidWhenSubnet(GET_FEED_MCP_TOOL.inputSchema),
     async handler(args: z.infer<typeof GetFeedInputSchema>, ctx: McpCtx) {
       return loadFeedItems(asMcpLoaderCtx(ctx), args, {
         // Same cross-subnet incident ledger + wiring get_global_incidents uses
--- a/tests/mcp-server.test.ts
+++ b/tests/mcp-server.test.ts
@@ -10,6 +10,7 @@ import {
   MAX_MCP_BODY_BYTES,
   listToolDefinitions,
   handleMcpRequest,
+  requireNetuidWhenSubnet,
 } from "../src/mcp-server.ts";
 import * as profilesMcp from "../src/profiles-mcp.ts";
 import * as healthHistoryMcp from "../src/health-history-mcp.ts";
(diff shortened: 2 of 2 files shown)


JSONbored (commented):
Issue already solved by someone, closing.

Issue #8915, opened 2026-07-31
fix(scripts): r2:manifest:dry-run can never pass locally, making npm run check permanently red


JSONbored:
## Problem

`npm run check` → `pipeline:check` → `r2:manifest:dry-run` **cannot pass in a contributor tree**, and the same check is **vacuous in CI**. It is a no-op where it runs and a permanent false failure where you'd actually run it.

### Vacuous in CI

`.github/workflows/validate.yml` runs `npm run r2:manifest` (**write** mode) at the "Generate R2 manifest" step (line ~390), and only runs `r2:manifest:dry-run` much later at "Validate R2 manifest" (line ~555). By then the file it compares against was just written from the very tree it rebuilds from, so the comparison is against itself and can never fail.

### Permanently red locally

Outside CI nothing writes it first, so the dry-run compares a **local build's** artifact set against the **publish pipeline's** committed lockfile:

```
{"error":"r2 compact manifest is stale","actual_full_artifact_count":2261,"expected_full_artifact_count":2320}
```

That mismatch is *expected by design*: `public/metagraph/r2-manifest.json` is deploy-pipeline-owned. The contributor skill says never commit it, `npm run build` auto-reverts it, and the derived-artifact freshness gate explicitly excludes it. A contributor build's set legitimately differs.

## Why it matters

A gate that can never be green trains you to skim its output. That is not hypothetical — it is exactly how a **real** `validate:openapi-examples` failure was missed locally on #8892 and only caught by CI: `npm run check`'s failure list was dismissed as "the usual r2-manifest noise".

## Fix

Keep it fatal under `CI` (where the invariant genuinely holds and nothing changes), and degrade to a clear warning outside it, explaining why a local artifact-set difference is expected.

**No CI behaviour change.**

## Acceptance

- [ ] `r2:manifest:dry-run` exits 0 locally with an explanatory warning.
- [ ] `CI=true r2:manifest:dry-run` still exits 1 on a genuine mismatch.
- [ ] Reason recorded in-file, including why the committed manifest is not a contributor-buildable artifact.

Issue #8908, opened 2026-07-31
perf(test): public-safety.test.ts spends 43 full-repo scans, 48% of the CI test job


JSONbored:
## Problem

PR CI wall clock is ~16-17 min, set by the `test` job (~970-1000s). Measured across two recent successful runs, the breakdown is not what it looks like:

| Step | Time | Share |
| --- | --- | --- |
| **Test (artifact writers, serial)** | **467-474s** | **48%** |
| Test (parallel, with coverage gate) | 303-315s | 31% |
| Lint + format | 60-64s | 6% |
| ~35 other steps | ~140s | 15% |

The serial pass is not five expensive files — it is **one**:

| File | Local time |
| --- | --- |
| **`tests/public-safety.test.ts`** (48 tests) | **163.8s** |
| `tests/artifacts.test.ts` (23 tests) | 44.0s |
| `tests/validate-error-messages.test.ts` | 2.4s |
| `tests/refresh-build-summary.test.ts` | 0.3s |
| `tests/discovery-artifacts.test.ts` | 0.1s |

`public-safety.test.ts` calls `runScanOutput()` **43 times**. Each call spawns `node scripts/scan-public-safety.ts`, which walks the entire repo. One scan measures **3.79s** locally; 43 x 3.79 = ~163s, matching the observed time exactly.

Together `public-safety` + `artifacts` are **76% of all test execution time across 540 files** (204s of 270s). Third place is 9.6s.

The existing response to this cost was to raise the file's per-test timeout from 15s to 45s (`SCANNER_TEST_TIMEOUT_MS`), which treats the symptom — and the header comment already anticipates the trend: *"its cost scales with repo size"*.

## Why it is safe to batch

Each test writes a fixture to a shared path, scans, and asserts that its path appears (or does not appear) in the output. The scanner reports `<repo-relative path>:<line>: <rule>` and prints **all** findings with no cap (`scripts/scan-public-safety.ts` — `findings` is unbounded), so one planted file's findings are independent of every other's.

So: give each case a uniquely-named file, plant them all once, run **one** scan, and let each test assert against its own path in the shared output. Same assertions, same real full-repo invocation, 43 spawns -> 1.

## Why not just split the job

Splitting the serial pass into its own parallel job would also remove it from the critical path, but it is strictly worse on both axes that matter here:

| | Wall clock | Compute | Runners |
| --- | --- | --- | --- |
[...]

Issue #8834, opened 2026-07-31
fix(mcp): call_rpc dereferences payload.error.code on an upstream 4xx body with no error object


JSONbored:
## Context

`call_rpc` unconditionally dereferences `payload.error.code` on any non-OK proxy response, behind a
comment asserting an invariant that is **false**. The response shape it cannot handle is one the RPC
proxy demonstrably produces.

`src/mcp-server.ts:9390-9398`:

```ts
if (!response.ok) {
  // handleRpcProxyRequest's every error path goes through workers/http.ts's
  // errorResponse(), which always populates error.code/error.message -- no
  // "malformed error body" case exists to guess a fallback for.
  throw toolError(
    (payload as Row).error.code,
    (payload as Row).error.message,
  );
}
```

Not every error path goes through `errorResponse()`. In `workers/request-handlers/rpc-proxy.ts`:

- `classifyUpstreamAttempt` classifies an upstream **4xx** (non-429) as `"fatal"`:
  `workers/request-handlers/rpc-proxy.ts:972` — `if (status >= 400) return "fatal";`
- A `"fatal"` attempt is not retried and not caught by the transient branch (`:1153-1163`); it falls
  through the `if (status < 400)` body-inspection block and reaches
  `workers/request-handlers/rpc-proxy.ts:1248` — `return streamRpcResponse(upstream, endpoint, attempts, status);`
- `streamRpcResponse` (`:1013-1027`) constructs `new Response(upstream.body, { status: status || 502, headers })`
  — **the upstream node's own body and its own 4xx status, forwarded verbatim.** It never touches
  `errorResponse()`.

So `response.ok` is false and `payload` is whatever the third-party node emitted, which need not be a
JSON-RPC error envelope:

- Upstream `403 {"message": "Forbidden"}` → `payload.error` is `undefined` → `TypeError: Cannot read
  properties of undefined (reading 'code')` → swallowed by the generic catch → the agent gets
  `internal_error: The tool failed to complete.` instead of the real upstream rejection.
- Upstream `404 {"error": "not found"}` (a string, not an object) → `payload.error.code` and
  `.message` are both `undefined` → `toolError(undefined, undefined)` → the agent gets the literal
  string `undefined: undefined`, and the `undefined` code is threaded into the `errorCode` analytics
  dimension at `src/mcp-server.ts:11659-11661`, poisoning the failure-cause breakdown that field exists
  to provide.

[...]

Issue #8832, opened 2026-07-31
fix(mcp): get_chain_identity_history returns a success-shaped empty feed for an out-of-range limit


JSONbored:
## Context

`get_chain_identity_history` forwards its `limit` argument to the data Worker without validating it.
The Worker correctly returns a 400 — and the MCP tool converts that 400 into a **successful, empty
result**. An agent asking "which subnets have changed their identity?" with `limit: 500` is told, with
`isError: false`, that no subnet ever has.

The chain, step by step:

1. `src/mcp-server.ts:4103-4109` builds the upstream request from the raw argument:
   `mcpChainIdentityHistoryRequest({ limit: args?.limit })` — no bound check, no coercion. The published
   schema is correct (`schemas-src/mcp-tools/chain-identity-history.ts:12` is
   `z.int().min(1).max(CHAIN_IDENTITY_HISTORY_LIMIT_MAX).optional()`), but nothing enforces it: the MCP
   dispatcher's `validateToolArguments` only checks that args is an object and rejects unknown keys.
2. `workers/data-api.ts:10574-10578` calls `parseLimitParam` with `maxLimit:
   CHAIN_IDENTITY_HISTORY_LIMIT_MAX` (200) and returns a 400 via `queryError`. Correct behavior.
3. `tryPostgresTier` (`workers/postgres-tier.ts:90-99`) treats **any** non-2xx from `DATA_API` as a
   Postgres outage: it logs, captures a PostHog `$exception`, and returns `markPostgresTierFallback()` —
   i.e. `null`.
4. Back at `src/mcp-server.ts:4104-4109`, the `??` fallback then runs
   `buildChainIdentityHistory([], { limit: args?.limit })`, producing
   `{ schema_version: 1, count: 0, subnet_count: 0, changes: [] }` with `isError: false`.

The same silent-empty result comes back for `limit: 0`, `limit: -1`, `limit: 1.5`, and `limit: "abc"`.

Two costs. First, the agent-facing one: a schema-stable empty feed is indistinguishable from "there is
genuinely no identity-change history", so an agent draws a confident false conclusion instead of
retrying with a valid `limit`. The tool's own description (`src/mcp-server.ts:4085-4087`) advertises
"max 200", so an agent that trusts the description and sends 500 has no signal it was wrong. Second,
every such call burns a spurious PostHog `$exception` via `capturePostgresTierFallback`, polluting the
Postgres-outage signal with caller-input errors.

[...]

PR #8921, opened 2026-07-31
chore: release main

### LINKED ISSUE #8368: feat(data): genesis-depth backfill orchestration — account/stake/transfer history from block 0 once the archive reaches tip
## Context

The self-hosted archive node is closing on chain tip (8.03M of 8.71M on 2026-07-27, ~1.4–1.9 blocks/s net). Two pre-existing backfills unblock on their own (#5588 missing block range at height 8,480,000; #5597 hyperparams history) — this issue owns what is NOT yet filed: extending the indexed history that currently starts "when our indexer started" back to genesis for the tables the account/validator/subnet history surfaces read (balances/transfers, stake events, registrations/deregistrations, identity sets).

**Blocked on the archive node reaching chain tip.** Do not start before #5588's unblock condition clears; the same node serves both.

## Requirements

1. Inventory first, in the PR: for each history-backed table the API reads, its current earliest block and which endpoint(s) expose the gap (e.g., account history, position history, stake flows, registration history). The existing backfill binary's sharding pattern is the execution vehicle — this issue extends its job definitions where a table lacks a genesis-capable job, and does NOT invent a second backfill system.
2. Ordering/rate discipline: the archive's RPC is shared with live-follow verification — backfill jobs must respect the same rate-limit posture the existing backfill jobs use (the prior firehose-backlog incident taught this; cite the rate ceiling used in the PR).
3. Idempotence + verification: each job re-runnable; per-table verification queries (count-by-epoch continuity, no duplicate keys) committed as a script so the result is checkable after each shard, not just at the end.
4. TimescaleDB rules apply to any new/altered table: partition column in every unique constraint; integer-now function set before any retention policy.
5. Execution is operational (runs on the data plane, monitored via the ops console) — this repo's PR carries the job definitions, verification scripts, and a runbook doc under `docs/`; it does not carry credentials or host details.
6. Definition of done is data-visible: the account-detail History surfaces (and T4's sibling charts) render pre-indexer history for a known old account, and each verification script passes; paste outputs in the issue before closing.

## Deliverables

- [ ] Gap inventory (table → earliest block → affected endpoints)
- [ ] Backfill job definitions + rate discipline + idempotence
- [ ] Verification scripts + runbook
- [ ] Post-run verification outputs attached; old-account spot-check renders

## Expected outcome

Every history endpoint answers from block 0. "How deep does your data go?" stops having an asterisk.

## Links

#5588, #5597 (unblock together) · `apps/indexer-rs` backfill binary + job patterns · account history endpoints in `public/metagraph/openapi.json`

### LINKED ISSUE #8704: feat(feeds): widen per-subnet feed sources — chain governance events and GitHub releases
Sub-issue of #8701.

## Context

The per-subnet feed (`/api/v1/feeds/subnets/{netuid}`) today carries registry changes,
incidents, and gap reports. A subnet's actual *news* — its chain-governance changes and its
software releases — never appears, even though we already index the former and already talk
to GitHub for the latter:

- **Chain-derived, already in our tiers:** hyperparameter changes
  (`subnet_hyperparams_history`, #4309), ownership transfers (`subnet_ownership_history`),
  lease lifecycle (`SubnetLease*` events, already classified `"governance"` in
  `src/account-events.ts`).
- **GitHub, already fetched per subnet:** `scripts/github-signals.ts` resolves every subnet's
  source repo and calls the GitHub API for commit activity — but does not capture **releases**,
  which are the newsworthy artifact.

This issue turns the existing per-subnet feed into the "intelligently aggregated" feed the
product idea asks for — from primary sources only, no third-party content.

## Requirements

1. **New item kinds** in the per-subnet feed, each with a primary-source link:
   - `hyperparam-change` — param, old→new, block, link to the extrinsic
   - `ownership-change` / `lease-event` — link to the event
   - `release` — tag, name, link to the GitHub release; captured by extending
     `github-signals.ts` (stay REST, per that file's own documented rationale)
2. **Curation rules are code, and they are written down in the module header:**
   - per-source caps per feed window so one noisy repo or one hyperparameter flurry cannot
     flood a feed (the #8611 quiet-channel rule applied to feeds)
   - dedupe: one item per (kind, identity), not one per poll observation
   - every item MUST carry a resolvable primary-source URL — an item that cannot cite its
     source does not ship. This is the line that keeps "aggregated" from becoming "made up".
3. **No new storage system.** Items derive from tiers and captures we already maintain. If an
   item kind would require a new collection pipeline, it is out of scope here — file it.
4. Tag vocabulary extended (`?tag=release`, `?tag=hyperparam`, …) and documented in
   `src/feeds.ts`'s existing header contract.
5. Existing feed consumers unaffected: current item kinds byte-stable (the #8658 lesson —
   prove parity, don't assume it).

## Deliverables — ONE PR (or two: chain-derived, then releases — not more)

- [ ] Chain-derived item kinds with links + tests against captured real rows (fixture rule:
      capture from the producer, never invent — #8646/#8650/#8686)
- [ ] Release capture in `github-signals.ts` + release items
- [ ] Caps + dedupe with tests proving a flood is capped and a re-poll does not duplicate
- [ ] Tag docs updated; existing-item byte-stability test
- [ ] Regenerated artifacts

## Expected outcome

`/api/v1/feeds/subnets/{netuid}.rss` reads like the subnet's changelog: releases, parameter
changes, ownership events, incidents — every line linking to chain or GitHub. Subscribing to
one URL per subnet is genuinely sufficient to follow it.

## Definition of 100% complete

- Every new item kind has a fixture captured from the real producer, not typed from the schema.
- The flood-cap and dedupe tests exist and fail when the caps are removed.
- An item without a primary-source URL is unconstructible (type-level or validated), not just
  discouraged.
- Existing feeds byte-identical for existing kinds, asserted.

### LINKED ISSUE #8717: design-tokens: ratchet-clean apps/ui/src/routes batch 1/4 (5 files, 18 violations)
## Context

`apps/ui`'s Bone & Ink design-guardrail lint rule (`no-restricted-syntax`, `apps/ui/eslint.config.ts`) is warn-tier by default until a directory is verified at 0 violations, at which point it's promoted to error-tier via the `RATCHETED_DIRS` one-way ratchet (#7851/PR #8101). `src/hooks/**`, `src/lib/**`, and `src/components/**` are already ratcheted (the last via #8167–#8171 + #8552). `src/routes/**` is the one remaining un-ratcheted directory in `apps/ui` — confirmed via a live `npx eslint src/routes` run against `origin/main` (`a8baad44b`): **20 files, 59 `no-restricted-syntax` violations**, all either bare arbitrary text sizes, raw spacing outside the 4pt subset, hand-rolled card shells, or raw `target="_blank"` anchors. This issue is batch 1 of 4 covering that directory.

Each violation's own eslint message already names the exact fix (see `apps/ui/CONTRIBUTING.md`'s guardrail table for the full token/primitive reference):
- Bare arbitrary text sizes (`text-[11px]`) → `<SectionLabel>`, `<Chip>`, or the `.mg-type-*` utilities
- Raw spacing outside the 4pt subset → `--mg-space-*` tokens or the Panel/primitives
- Hand-rolled `rounded border bg-card` → `<Panel>` from `@/components/metagraphed/primitives`
- Raw `<a target="_blank">` → `<ExternalLink>` from `@jsonbored/ui-kit`

## Requirements

1. Run `npx eslint <file>` on each file below (or `npx eslint src/routes` for the whole directory) and fix every `no-restricted-syntax` violation listed, to the specific token/component its own eslint message names.
2. Zero visual/behavioral change is the goal — these are one-off values being swapped for the equivalent design-system token, not a redesign. If a file's current one-off value has no matching token (a genuinely new scale point, e.g. a display-size heading above the `mg-type-*` scale's current ceiling), flag it explicitly in the PR body rather than inventing an ad hoc token or silently leaving it unfixed.
3. This batch does **not** get added to `RATCHETED_DIRS` on its own — `src/routes/**` has files outside this batch (see the sibling batches). Leave the ratchet-list update for a follow-up issue once all batches land.

## Deliverables

- [ ] Every file below fixed to 0 `no-restricted-syntax` violations (verify with `npx eslint <file>`), or the specific violation flagged in the PR body with a stated reason if genuinely unfixable without a new token
- [ ] `npx tsc --noEmit` clean
- [ ] `npx prettier --check` clean on touched files
- [ ] `vitest run` — all existing tests still passing (no behavioral change expected)
- [ ] Before/after screenshot table (clickable thumbnails, uploaded not committed) for every touched route — required on any apps/ui PR touching rendered output, per this repo's gate contract, even when the change is visually a no-op (the screenshots are the proof of that)

All of the above ship together in one PR — a PR that fixes only some of these files does not satisfy this issue.

## Files in this batch (5 files, 18 violations)

- `src/routes/-gaps-page.tsx` (7 violations: L262, L809, L832, L862, L908, L913, L1235)
- `src/routes/-index-page.tsx` (5 violations: L321, L427, L438, L536, L544)
- `src/routes/-root-views.tsx` (3 violations: L137, L147, L169)
- `src/routes/-extrinsics-hash-page.tsx` (2 violations: L384, L466)
- `src/routes/-providers-index-page.tsx` (1 violation: L629)

## Expected Outcome

Running `npx eslint <file>` on every file listed above reports 0 `no-restricted-syntax` warnings (or an explicitly-flagged, justified exception per Requirement 2), with no other behavior change — falsifiable directly from a fresh eslint run against the PR's diff.

## Links & Resources

- #7851 (the lint-ratchet mechanism this batch's cleanup feeds into)
- #8167–#8171, #8552 (the precedent batches that cleared `src/components/**`, and the follow-up that promoted it)
- `apps/ui/CONTRIBUTING.md` (the full guardrail token/primitive reference)
- `apps/ui/eslint.config.ts:262` (`RATCHETED_DIRS`) — the list this directory will eventually join

### LINKED ISSUE #8718: design-tokens: ratchet-clean apps/ui/src/routes batch 2/4 (5 files, 15 violations)
## Context

Same effort as the sibling batch issues covering `apps/ui/src/routes/**` — the last un-ratcheted directory in `apps/ui`'s Bone & Ink `no-restricted-syntax` lint rule (see #7851/PR #8101 for the ratchet mechanism, #8167–#8171/#8552 for the precedent that cleared `src/components/**`). Confirmed via a live `npx eslint src/routes` run against `origin/main` (`a8baad44b`): 20 files, 59 violations total across 4 batches. This is batch 2 of 4.

Each violation's own eslint message already names the exact fix (see `apps/ui/CONTRIBUTING.md`'s guardrail table):
- Bare arbitrary text sizes (`text-[11px]`) → `<SectionLabel>`, `<Chip>`, or the `.mg-type-*` utilities
- Raw spacing outside the 4pt subset → `--mg-space-*` tokens or the Panel/primitives

## Requirements

1. Run `npx eslint <file>` on each file below and fix every `no-restricted-syntax` violation listed, to the specific token/component its own eslint message names.
2. Zero visual/behavioral change is the goal. If a file's current one-off value has no matching token, flag it explicitly in the PR body rather than inventing an ad hoc token or silently leaving it unfixed.
3. This batch does **not** get added to `RATCHETED_DIRS` on its own — leave the ratchet-list update for a follow-up issue once all 4 batches land.

## Deliverables

- [ ] Every file below fixed to 0 `no-restricted-syntax` violations (verify with `npx eslint <file>`), or the specific violation flagged in the PR body with a stated reason if genuinely unfixable without a new token
- [ ] `npx tsc --noEmit` clean
- [ ] `npx prettier --check` clean on touched files
- [ ] `vitest run` — all existing tests still passing
- [ ] Before/after screenshot table (clickable thumbnails, uploaded not committed) for every touched route

All of the above ship together in one PR — a PR that fixes only some of these files does not satisfy this issue.

## Files in this batch (5 files, 15 violations)

- `src/routes/-subnets-netuid-page.tsx` (6 violations: L332, L899, L1314, L1321, L2045, L2118)
- `src/routes/-schemas-page.tsx` (4 violations: L205, L267, L557, L562)
- `src/routes/-validators-hotkey-page.tsx` (3 violations: L197, L446, L597)
- `src/routes/-accounts-index-page.tsx` (1 violation: L65)
- `src/routes/-providers-slug-page.tsx` (1 violation: L360)

## Expected Outcome

Running `npx eslint <file>` on every file listed above reports 0 `no-restricted-syntax` warnings (or an explicitly-flagged, justified exception per Requirement 2), with no other behavior change — falsifiable directly from a fresh eslint run against the PR's diff.

## Links & Resources

- #7851, #8167–#8171, #8552 — the ratchet mechanism and precedent batches
- `apps/ui/CONTRIBUTING.md` — the full guardrail token/primitive reference
- `apps/ui/eslint.config.ts:262` (`RATCHETED_DIRS`)

### LINKED ISSUE #8756: feat(data): recover the true runtime-upgrade timeline — chain-derived transition backfill with reconcile
## Problem

#8752 fixed the *disclosure* — `/api/v1/runtime` now admits its gaps. This issue is the
*remedy*: recover the true transition timeline.

The underlying data is worse than #8752 stated. Probing `/api/v1/blocks/{n}` with binary
search (2026-07-30) shows the `blocks` table is coverage **islands**, not a continuum with one
hole:

| range | state |
| --- | --- |
| 0 .. 3,000 | present |
| 182,191 .. 244,200 | present |
| 4,600,000 .. 4,600,299 | present |
| (further small islands at the endpoint's other "transition" blocks) | present |
| **8,599,188 .. tip** | **continuous — coverage starts here** |
| everything else (~98% of the chain) | absent |

Two consequences, both verified against a public archive:

1. **The endpoint's "spec 424 @ 8,599,188" is the coverage start, not a transition.** True
   424 activation is **block 8,513,820 (2026-06-29T19:22Z)** — 85k blocks earlier, inside the
   uncovered region.
2. #5588's "~9,000-block hole" (8,471,001–8,479,999) is the visible corner of a
   ~4,000,000-block missing region (4,600,300 → 8,599,187).

## The tool

`scripts/backfill-runtime-transitions.ts` (sibling PR):

- **Discovery:** binary-searches `state_getRuntimeVersion` over block ranges — cost scales
  with the number of *upgrades* (~200 × log₂(range) ≈ 4,600 RPC calls), not blocks. Runs
  against any public archive; **not** blocked on our own archive node (#2111).
- **Full chain-derived boundary rows:** hash, parent hash, extrinsic count, event count
  (`System.EventCount` at the hash), spec_version, and `observed_at` decoded from the
  `timestamp.set` inherent — supporting both extrinsic format v4 and v5 (the chain switched
  formats mid-history; real fixtures from blocks 1,000,000 and 8,400,000 are in the tests).
  A row whose timestamp cannot be decoded is refused, never fabricated.
- **Reconciles, never trusts** (per owner directive): every existing row at a boundary height
  is compared against chain truth and overwritten on any disagreement — an
  insert/overwrite/in_sync plan is printed before anything is written, and default is dry-run.
- Writes via transactional DELETE+INSERT, not ON CONFLICT — `blocks` keys on
  `(block_number, observed_at)` (Timescale partition column in the PK), so a composite-target
  upsert would *miss* a row whose stored `observed_at` disagrees with chain truth and insert a
  duplicate beside it.

Verified live over [8,400,000 .. 8,737,000]: 11 transitions in 62s, all boundary rows decoded,
including 423 @ 8,486,593 (independently confirmed by the infra investigation) and the
corrected 424 @ 8,513,820.

## Operational runbook (after the PR lands)

```bash
node scripts/backfill-runtime-transitions.ts --out transitions.json          # discovery, no DB
DATABASE_URL=... node scripts/backfill-runtime-transitions.ts                # + reconcile plan (dry)
DATABASE_URL=... node scripts/backfill-runtime-transitions.ts --write        # apply
```

Full-range discovery is ~4,600 sequential RPC calls (≈30–60 min against the public archive).
After `--write`, `/api/v1/runtime` serves the complete ~200-entry timeline with true dates, and
#8753's `coverage_gaps` shrinks to whatever remains genuinely unknown.

## Bonus finding from the live run

Spec **421** activated at **block 8,466,530** — the exact block every `SubnetEmaTaoFlow` value
froze at (#8750). The flow-EMA freeze was caused by the v421 runtime upgrade, not by an
on-chain flag flip. Noted on #8750.

## Acceptance

The complete transition list is recoverable by one command against a public archive; a second
run reports all-in_sync (idempotent); every written row is chain-derived with the block's own
timestamp; existing wrong rows at boundary heights are overwritten, not skipped.

### LINKED ISSUE #8762: fix(data): backfill writes observed_at as an ISO string, but the column is epoch-ms bigint — every --write fails
## Problem

`scripts/backfill-runtime-transitions.ts` writes `blocks.observed_at` as an ISO-8601 string.
The column is a **bigint of epoch milliseconds**, not a timestamp — so **every `--write` run
fails**:

```
PostgresError: invalid input syntax for type bigint: "2023-03-22T17:50:48.001Z"
```

The script as merged (#8757, #8760) cannot complete a write. Discovery and the dry-run
reconcile are unaffected.

## Why the mistake was easy to make

Kanel brands the column as a `string`:

```ts
export type BlocksObservedAt = string & { __brand: 'public.blocks' };
observed_at: BlocksObservedAt;
```

which reads like a date column and is not one. The read side had it right all along —
`src/runtime-versions.ts`'s `toIso(ms)` converts epoch ms to an ISO string for the API response
— but that only tells you the truth if you look at the reader before writing the writer.

## Blast radius

None in production. Postgres's type check rejected it, so nothing was written. **Had the column
been `text` instead of `bigint`, this would have silently corrupted the `observed_at` of every
backfilled row** — the timeline would have rendered with plausible-looking dates that no
downstream consumer could distinguish from correct ones.

## Fix

Pass `r.observed_at_ms` directly. Comment records the type and cites both pieces of evidence
(the reader's `toIso`, and the Postgres error) so it does not get "helpfully" re-converted.

## Verification

End-to-end against production after the fix:

- 134 rows applied
- second run: **0 insert, 0 overwrite, 147 in sync** (idempotent)
- 0 duplicate `block_number`s from the DELETE+INSERT path
- `GET /api/v1/runtime` now serves **147 transitions** (was 23), spec 101 @ genesis → 440 @
  8,713,793, and `coverage_gaps` fell from 6 to 2

## Acceptance

`--write` completes; a second run is fully in-sync; `observed_at` round-trips to the correct
date through the API.

### LINKED ISSUE #8763: De-duplicate OpenAPI worked examples: hoist into components.examples
`public/metagraph/openapi.json` ships every operation with a deterministic worked response example, inlined into the media type. Those examples are a **pure function of the response schema** — `openApiExampleForRoute` → `sampleFromSchema` — so any two routes wrapping the same artifact component sample byte-identically, and the document pays for each copy.

## Measured, on `main` @ 8b85795

- **239** media types carry an example, holding only **211 distinct** values.
- Examples are **178,835 bytes** of a 2,198,245-byte document (**8.1%**).
- The largest single group: the CSV example `"netuid,name\r\n7,Allways"`, written out **21 times**.
- Others: `/api/v1/extrinsics` ≡ `/api/v1/governance/config-changes` ≡ `/api/v1/sudo`; the four `*-history` artifacts (`subnets/{netuid}/history`, `.../concentration/history`, `.../performance/history`, `.../yield/history`); `/api/v1/endpoint-pools` ≡ `/api/v1/rpc/pools`.

The duplication compounds with **#8698 / #8758**, which adds a `{network}` twin per network-addressable route by spreading the base operation (`{...base}`) — including its `responses`. On that branch the same 211 distinct examples are written across **335** media types: **266,658 bytes**, **9.4%** of a 2,836,062-byte document, with **94,474 bytes** of it pure duplicate. That growth is what forced the `openapi.json` budget ceiling up in that PR.

It also lands twice in the generated types: `openapi-typescript` renders each inline `example` as an `@example` JSDoc block, so **`public/metagraph/types.d.ts` and `packages/contract/index.d.ts` are 1,472,252 bytes each**, largely JSDoc'd sample data.

## What to do

Hoist each distinct example into `components.examples` and point the media types at it, using the OpenAPI 3.1 Media Type `examples` map:

```json
"content": {
  "application/json": {
    "schema": { "...": "..." },
    "examples": {
      "CoverageArtifactResponse": { "$ref": "#/components/examples/CoverageArtifactResponse" }
    }
  }
}
```

`examples` is the form OpenAPI 3.1 prefers — singular `example` is deprecated — so this moves toward the standard rather than inventing a local convention.

**Naming must be derived and stable**, not positional: a name that depends on which route is walked first would churn whenever a route is added. Name a JSON example for the artifact component it demonstrates (`CoverageArtifact` → `CoverageArtifactResponse`). A CSV example has no schema to be named after (`type: string` is the whole schema), so group those by content and name them for the alphabetically-first operation that shows them.

Two components whose schemas happen to sample identically should **keep two entries**. Collapsing them files `RpcPoolsArtifact`'s example under `EndpointPoolsArtifactResponse` and makes the document lie about what the reader is looking at — measured at 3,134 bytes, which is not worth buying with wrong documentation.

## Definition of done

- [ ] Every 200 media type carries exactly one `examples` entry, a `$ref` into `components.examples`, and no inline `example`.
- [ ] Every `$ref` resolves; `components.examples` has no orphaned entries.
- [ ] Example names are derived as above and asserted by tests, so route order cannot rename a shipped entry.
- [ ] `scripts/validate-openapi-examples.ts` follows the ref (it reads `media.example` directly today and would report every route as missing an example).
- [ ] `npm run build` run and the regenerated `openapi.json` + generated types committed, per `validate:contract-drift`.
- [ ] Confirm `apps/ui/content/docs/api-reference` still renders examples — it is generated separately (`cd apps/ui && node scripts/generate-openapi-docs.ts`) and is not regenerated by the root build.

## Notes for whoever picks this up

- The docs site renders examples **client-side** from the live spec (`<APIPage/>` in the generated MDX fetches `/metagraph/openapi.json`), so the MDX itself carries no example bodies — the thing to verify is that the renderer resolves the refs, not that the MDX changed.
- Dropping the inline `example` also drops the `@example` JSDoc from the generated `.d.ts` artifacts, since `openapi-typescript` only understands the JSON-Schema `examples` *array*, not the Media Type `examples` *map*. Quantify that in the PR — it is a real, user-visible change to a published artifact and should be a stated decision, not a silent side effect.

### LINKED ISSUE #8766: fix(api): coverage-gap detector reports consecutive spec versions as gaps
## Problem

`detectRuntimeCoverageGaps` (#8753) flags a gap whenever two consecutive transitions are more
than 500,000 blocks apart. That distance heuristic was right for the world it was written in —
the timeline was 90% missing and distance was the only available signal.

After the #8756 backfill made the timeline chain-derived and complete, one of its two remaining
reports is provably wrong:

```
spec 141 @ 1,971,974  ->  spec 142 @ 2,543,779   (571,805 blocks)
```

**141 and 142 are consecutive spec versions.** A hidden upgrade between them would have to be to
a version strictly between 141 and 142, and there is no integer there. The interval is one
long-running runtime — a fact about the chain, not a hole in our data.

## Fix

Skip the distance check entirely when `next.spec_version === prev.spec_version + 1`. This is not
a threshold tweak; it removes a logically impossible case.

The other report survives and should:

```
spec 127 @ 806,973  ->  spec 133 @ 1,404,224   (597,251 blocks, 5 versions skipped)
```

128–132 either never reached mainnet (as 425–431 demonstrably did not) or our coverage is short
there. The payload cannot distinguish those, so flagging it is the honest answer.

## Note on the existing tests

Two threshold tests used `spec 1 -> spec 2` as their fixture — consecutive versions — so they
began failing under the new rule. They were testing the *threshold*, not consecutiveness; their
fixtures are re-pointed to `1 -> 3` so they still exercise what they were written to exercise.

## Acceptance

Consecutive spec versions never produce a gap regardless of block distance; a version skip over
a long span still does; the live endpoint's `coverage_gaps` drops from 2 to 1.

### LINKED ISSUE #8806: fix(api): CORS preflight omits authorization, blocking every browser API-key call
## Context

The CORS preflight handler `corsPreflight` (`workers/api.ts:5926-5958`) builds one fixed
`access-control-allow-headers` value at `workers/api.ts:5950-5955`:

```
content-type, if-none-match, mcp-session-id, mcp-protocol-version,
${WEBHOOK_SECRET_HEADER}, ${WEBHOOK_SUBSCRIPTION_TOKEN_HEADER},
${ALERT_TRIGGER_CREATE_TOKEN_HEADER}, ${ALERT_TRIGGER_OWNER_TOKEN_HEADER}
```

`authorization` is **not in that list**. `Authorization` is not a CORS-safelisted request header, so
any cross-origin request carrying it triggers a preflight, and a preflight whose
`access-control-allow-headers` does not name it fails — the browser never sends the real request.

The UI sends exactly that header. `apps/ui/src/components/metagraphed/api-keys-manager.tsx:53-55`:

```ts
function authHeaders(token: string): HeadersInit {
  return { authorization: `Bearer ${token}` };
}
```

Those calls go from origin `https://metagraph.sh` to `https://api.metagraph.sh` (`DEFAULT_API_BASE`)
— cross-origin, non-safelisted header, preflight required. **The self-serve API-key
generate / list / revoke dashboard cannot work in a browser.**

Three further gaps in the same function compound it, all caused by the default
`methods = "GET, HEAD, OPTIONS"` at `workers/api.ts:5929` applying to routes that are not GET-only:

- `/api/v1/keys` has no branch in the `else if` chain, so the preflight advertises
  `GET, HEAD, OPTIONS`. The route accepts POST (create), GET (list) and DELETE (revoke) —
  `workers/api.ts:1527-1531` (`handleAccountKeysProxy`), dispatched at `workers/api.ts:2149`. So
  `DELETE /api/v1/keys/{prefix}` fails the preflight on **method** as well as on header.
- `/api/v1/auth/wallet/challenge` and `/api/v1/auth/wallet/verify` (dispatched
  `workers/api.ts:2135-2141`) are POST-only and also have no branch. The UI posts JSON to both
  (`apps/ui/src/hooks/use-api-session.ts:71-84`, `content-type: application/json` — non-simple, so it
  preflights). **Signing in is blocked too**, which is the step that produces the token the key
  dashboard then needs.
- `/api/v1/watch/challenges` and `/api/v1/watch/tokens` (dispatched `workers/api.ts:2144-2146`) are
  POST-only with the same shape, called from `apps/ui/src/hooks/use-watch-token.ts:73,81`.

The docs instruct users to do precisely the thing that is blocked:
`apps/ui/content/docs/chain-events.mdx:52` documents presenting
`Authorization: Bearer mg_...` to get the 5x keyed rate-limit tier.

Why this shipped: the only preflight header assertion in the suite,
`tests/api-coverage.test.ts:410-424`, checks only `mcp-session-id` and `mcp-protocol-version`. Nothing
asserts `authorization`, and nothing asserts the advertised methods for any of the POST-only routes
above.

**Mirror the `/mcp` branch inside `corsPreflight` itself — `workers/api.ts:5942-5945`.** That is the
existing, correct shape: one `else if` on `url.pathname` in the same chain, setting `methods` to that
route's real method set, with the route-specific request headers appended to the same single
`access-control-allow-headers` template string at `:5950-5955`.

Every change in this issue belongs inside `corsPreflight`. The following do **not** satisfy it:

- Setting `access-control-allow-headers` anywhere else — `apiHeaders` (`workers/http.ts:63`), a
  wrapper, a middleware, or a new helper module. There is one preflight builder; keep it that way.
- Emitting `access-control-allow-headers: *`. It is not equivalent (it is ignored for credentialed
  requests) and it erases the deliberate documentation value of the explicit list.
- Building a per-route header map, a table-driven route/method registry, or any other new mechanism.
  Extend the existing `else if` chain.
- Inlining the four `*_HEADER` constants as string literals instead of keeping the imported
  identifiers.
- Changing `apps/ui/**` so the client stops sending `authorization`, or adding a same-origin proxy in
  the UI. The Worker is what is wrong. **This PR must not touch `apps/ui/**` at all.**

### 1. BLOCKING FIRST STEP — verify against production before changing any code

A Cloudflare dashboard-level CORS transform rule, which lives outside this repository, could already
be injecting `authorization` in production. You must establish whether it does **before** writing the
fix, and report the result.

Run this and paste the **complete, unedited response headers** into the PR description:

```
curl -isS -X OPTIONS 'https://api.metagraph.sh/api/v1/keys' \
  -H 'Origin: https://metagraph.sh' \
  -H 'Access-Control-Request-Method: DELETE' \
  -H 'Access-Control-Request-Headers: authorization'
```

Then state, in one explicit sentence in the PR description, which of these two cases you observed:

- **`access-control-allow-headers` in the live response does NOT contain `authorization`.** The
  Worker's value is reaching the client unmodified. Implement the full fix below.
- **`access-control-allow-headers` in the live response DOES contain `authorization`.** Something
  outside version control is supplying it. In that case the correct outcome is a *smaller* change,
  not a blind header addition: still add `authorization` to the Worker (so the behaviour is
  version-controlled and survives the rule being deleted), but the PR must also add a code comment
  immediately above the `access-control-allow-headers` construction recording (a) that a
  dashboard-level transform rule was observed in production, (b) the exact
  `access-control-allow-headers` value the live response returned, and (c) the date observed. Also
  report in the PR whether the live `access-control-allow-methods` for `/api/v1/keys` already
  included `DELETE`, since that half may be masked independently.

A PR that does not contain the pasted `OPTIONS` response and the explicit case statement does not
satisfy this issue, regardless of what the diff does.

### 2. Add `authorization` to the allow-headers list

`workers/api.ts:5950-5955`. The resulting comma-separated value must contain **all** of:
`content-type`, `if-none-match`, `authorization`, `mcp-session-id`, `mcp-protocol-version`, and the
four existing constants (`WEBHOOK_SECRET_HEADER`, `WEBHOOK_SUBSCRIPTION_TOKEN_HEADER`,
`ALERT_TRIGGER_CREATE_TOKEN_HEADER`, `ALERT_TRIGGER_OWNER_TOKEN_HEADER`) still referenced as
imported identifiers. All header names stay lowercase. No existing entry may be removed. The value
stays a single template string; do not split it into an array-join or a new exported constant.

### 3. Advertise the real methods for the five POST-capable routes that currently fall through

Add branches to the existing `else if` chain in `corsPreflight`, producing exactly these verdicts:

| path | `access-control-allow-methods` |
| --- | --- |
| `/api/v1/keys` and any subpath (`startsWith`) | `POST, GET, DELETE, OPTIONS` |
| `/api/v1/auth/wallet/challenge` (exact) | `POST, OPTIONS` |
| `/api/v1/auth/wallet/verify` (exact) | `POST, OPTIONS` |
| `/api/v1/watch/challenges` (exact) | `POST, OPTIONS` |
| `/api/v1/watch/tokens` (exact) | `POST, OPTIONS` |

Match the matcher shape the request dispatcher already uses for each route — `startsWith` for
`/api/v1/keys` (`workers/api.ts:2149`), exact equality for the four auth/watch routes
(`workers/api.ts:2135-2146`) — so the preflight and the dispatcher agree by construction.

The verdicts for every path that already has a branch (`/rpc/`, `/api/v1/webhooks/`,
`/api/v1/alerts/triggers`, `/api/v1/watch/triggers`, `/api/v1/graphql`, `/mcp`, `/api/v1/ask`) and
the default `GET, HEAD, OPTIONS` for everything else must be **unchanged**. Note ordering:
`/api/v1/watch/triggers` already has a branch at `:5937`; the new exact-match watch branches must not
shadow or be shadowed by it.

### 4. Regression tests

In `tests/api-coverage.test.ts`, alongside the existing preflight block at `:400-430`:

- A test asserting `access-control-allow-headers` on an `OPTIONS /api/v1/keys` preflight matches
  `/\bauthorization\b/`. This must fail on `main` and pass after the fix.
- A test asserting `access-control-allow-methods` on `OPTIONS /api/v1/keys` equals
  `POST, GET, DELETE, OPTIONS`.
- One test per POST-only route above asserting `access-control-allow-methods` equals `POST, OPTIONS`.
- A test pinning that an unlisted path (e.g. `OPTIONS /api/v1/subnets`) still returns
  `GET, HEAD, OPTIONS`, so a future branch cannot widen the default unnoticed.

Assertions on the allow-headers value must use a substring/regex check for the specific header name,
not a full-string equality against the whole list — an equality assertion makes every future header
addition a test edit and buys nothing.

### 5. CI

`codecov/patch` enforces 99% on changed lines and branches in `workers/**`, branch-counted. Every new
`else if` branch you add is a counted branch and needs a test that takes it. Measure unsharded with
`npm run test:coverage`. All new/changed test files are `.ts`.

## Deliverables

- [ ] PR description contains the pasted, unedited `curl -isS -X OPTIONS` response headers from
      `https://api.metagraph.sh/api/v1/keys` and an explicit statement of which of the two cases in
      Requirement 1 was observed.
- [ ] `authorization` added to the `access-control-allow-headers` template string at
      `workers/api.ts:5950-5955`, with no existing entry removed.
- [ ] `corsPreflight` branch for `/api/v1/keys` advertising `POST, GET, DELETE, OPTIONS`.
- [ ] `corsPreflight` branches for `/api/v1/auth/wallet/challenge`, `/api/v1/auth/wallet/verify`,
      `/api/v1/watch/challenges`, `/api/v1/watch/tokens`, each advertising `POST, OPTIONS`.
- [ ] If and only if a production transform rule was observed: the explanatory code comment described
      in Requirement 1.
- [ ] Regression tests in `tests/api-coverage.test.ts` covering all of Requirement 4, including one
      that fails on `main` (`authorization` in allow-headers on `/api/v1/keys`).
- [ ] No file under `apps/ui/**` is modified.

satisfy this issue.

## Expected Outcome

An `OPTIONS /api/v1/keys` preflight from origin `https://metagraph.sh` returns 204 with
`access-control-allow-headers` containing `authorization` and `access-control-allow-methods` equal to
`POST, GET, DELETE, OPTIONS`; `OPTIONS` on each of `/api/v1/auth/wallet/challenge`,
`/api/v1/auth/wallet/verify`, `/api/v1/watch/challenges`, `/api/v1/watch/tokens` returns
`POST, OPTIONS`; every other path's advertised methods are byte-identical to `main`. A browser at
`https://metagraph.sh` can complete wallet login and then create, list, and revoke an API key against
`https://api.metagraph.sh` without a CORS failure, with no change to `apps/ui/**`.

## Links & Resources

- `workers/api.ts:5926-5958` — `corsPreflight`, the whole surface of this change
- `workers/api.ts:2135-2149` — the dispatcher branches whose method sets the preflight must match
- `workers/api.ts:1527-1566` — `handleAccountKeysProxy`, confirming POST/GET/DELETE
- `apps/ui/src/components/metagraphed/api-keys-manager.tsx:53-55` — the client that sends
  `authorization` (read-only reference; do not modify)
- `apps/ui/src/hooks/use-api-session.ts:71-84`, `apps/ui/src/hooks/use-watch-token.ts:73,81` — the
  POST-only login calls (read-only reference; do not modify)
- `apps/ui/content/docs/chain-events.mdx:52` — the documented `Bearer mg_...` keyed tier
- `tests/api-coverage.test.ts:400-430` — the existing preflight test block to extend

### LINKED ISSUE #8807: fix(api): top-holders net_flow_7d/30d/90d serve a sortable 0 when the rollup row is missing

JSONbored:
---

<details><summary>ui: 0.14.0</summary>

### Features

* **api:** emission-pipeline surface — per-subnet share decomposition and the liquidity/chain-buy TAO split ([#8892](https://github.com/JSONbored/metagraphed/issues/8892)) ([ce9147e](https://github.com/JSONbored/metagraphed/commit/ce9147ed8fb09ae8962ab7727df93a3a1fa684a2))
* **api:** expose action summaries across API surfaces ([#8694](https://github.com/JSONbored/metagraphed/issues/8694)) ([d8b89b5](https://github.com/JSONbored/metagraphed/commit/d8b89b580d46b93498d7ff3826a4101595e32028))
* **api:** express multi-network addressing in OpenAPI and the contract ([#8758](https://github.com/JSONbored/metagraphed/issues/8758)) ([6b54610](https://github.com/JSONbored/metagraphed/commit/6b546104885939300407cc24d122f5d6f5e63105))
* **api:** per-key abuse controls — anomaly signals and a key-level blocklist ([#8695](https://github.com/JSONbored/metagraphed/issues/8695)) ([8123d93](https://github.com/JSONbored/metagraphed/commit/8123d9359e275c490614a9aad68d8f0a68ed13c9))
* **chain:** upgrade radar — track runtime releases from GitHub through testnet to mainnet ([#8737](https://github.com/JSONbored/metagraphed/issues/8737)) ([9e72a6f](https://github.com/JSONbored/metagraphed/commit/9e72a6fbc5f423b755a2cc3848bcd74bb90cc67a))
* **feeds:** make the feed system discoverable — contract, OpenAPI, autodiscovery ([#8738](https://github.com/JSONbored/metagraphed/issues/8738)) ([1b7fae2](https://github.com/JSONbored/metagraphed/commit/1b7fae249847b670cc7f43ca5393634e003b35ce))
* **ui+api:** tenant usage dashboard — quota headroom, throttling, and export ([#8728](https://github.com/JSONbored/metagraphed/issues/8728)) ([ab1cd13](https://github.com/JSONbored/metagraphed/commit/ab1cd13156d680a9b1a99123a6cc048814e45fed))
* **ui:** stop list routes redirecting to fully-expanded query strings ([#8790](https://github.com/JSONbored/metagraphed/issues/8790)) ([eaec728](https://github.com/JSONbored/metagraphed/commit/eaec728cfe3cf8b1bf3596af9487d4ad09621a50))
* **ui:** weekly digest pages at /news, generated from the digest store ([#8782](https://github.com/JSONbored/metagraphed/issues/8782)) ([1ec77de](https://github.com/JSONbored/metagraphed/commit/1ec77de570227f0b59fc4016bd4a39ac798c9bea))

### Bug Fixes

* **api+ui:** disambiguate emission_share as stage 1 of the emission pipeline ([#8789](https://github.com/JSONbored/metagraphed/issues/8789)) ([f0f68a3](https://github.com/JSONbored/metagraphed/commit/f0f68a3f474f88766ed86e1ef909b01f3ee23fd2))
* **api:** exclude unsigned inherents from chain/fees medians and averages ([#8893](https://github.com/JSONbored/metagraphed/issues/8893)) ([133f027](https://github.com/JSONbored/metagraphed/commit/133f027c35c6424398517f6fa9e82fcc4f3e0a3b)), closes [#8809](https://github.com/JSONbored/metagraphed/issues/8809)
* **api:** treat missing top-holders net_flow as null, not zero ([#8851](https://github.com/JSONbored/metagraphed/issues/8851)) ([b722aa6](https://github.com/JSONbored/metagraphed/commit/b722aa6dfb0acbcf790375099408c9d0d2e20375)), closes [#8807](https://github.com/JSONbored/metagraphed/issues/8807)
* **ui:** constrain /apis surface-card ExternalLink so long URLs truncate ([#8726](https://github.com/JSONbored/metagraphed/issues/8726)) ([bfa430c](https://github.com/JSONbored/metagraphed/commit/bfa430c07fee3e9a10fd0437798eaed723970166))
* **ui:** give SectionAnchor title row a shrink/truncate contract ([#8716](https://github.com/JSONbored/metagraphed/issues/8716)) ([5976f9e](https://github.com/JSONbored/metagraphed/commit/5976f9ef8d32d266b1f2f840453b119da166aeb9))
* **ui:** give the /endpoints latency-heatmap scroll container the mg-table-scroll affordance, drop 6 stale overflow-baseline entries ([#8708](https://github.com/JSONbored/metagraphed/issues/8708)) ([a8baad4](https://github.com/JSONbored/metagraphed/commit/a8baad44b8dd6a40b0ce5524a38e4d3e53209169))
* **ui:** keep sub-milli amounts visible in formatNumber ([#8815](https://github.com/JSONbored/metagraphed/issues/8815)) ([#8877](https://github.com/JSONbored/metagraphed/issues/8877)) ([434070f](https://github.com/JSONbored/metagraphed/commit/434070f61a6f2a8355606f02531660361c000028))
* **ui:** ratchet-clean apps/ui routes batch 2/4 (5 files, 15 violations) ([#8727](https://github.com/JSONbored/metagraphed/issues/8727)) ([0418b55](https://github.com/JSONbored/metagraphed/commit/0418b553cc90cb9e649e3e182a6b1012441a38c3)), closes [#8718](https://github.com/JSONbored/metagraphed/issues/8718)
* **ui:** ratchet-clean design-token violations in src/routes batch 1/4 ([#8729](https://github.com/JSONbored/metagraphed/issues/8729)) ([92ef7e6](https://github.com/JSONbored/metagraphed/commit/92ef7e6adb5f8f93f2844e130a6d66ba5db5ee30)), closes [#8717](https://github.com/JSONbored/metagraphed/issues/8717)
* **ui:** right-anchor SankeyMini last-column labels in horizontal mode ([#8725](https://github.com/JSONbored/metagraphed/issues/8725)) ([421c3b7](https://github.com/JSONbored/metagraphed/commit/421c3b7cdaf369e587fcb77ab12fbe5cdcf93c27))
* **ui:** shrink ExternalLink children so long CTA labels truncate ([#8736](https://github.com/JSONbored/metagraphed/issues/8736)) ([49dcfeb](https://github.com/JSONbored/metagraphed/commit/49dcfebb17948b55318f2576cf02d065e6b41850))
* **ui:** stop padding failed AMM exit quotes with un-slipped spot in Simulated exit ([#8907](https://github.com/JSONbored/metagraphed/issues/8907)) ([75ff6a9](https://github.com/JSONbored/metagraphed/commit/75ff6a9fe69dbc16638af346b97834929bd41af8))
* **ui:** stop the subnet operational/curation tiles reporting real data on a failed query ([#8866](https://github.com/JSONbored/metagraphed/issues/8866)) ([1e27452](https://github.com/JSONbored/metagraphed/commit/1e274522ddf01c46797428fb88c71dea8dec928d))
* **ui:** swap raw Bone & Ink one-off values for design tokens in routes batch 3/4 ([#8733](https://github.com/JSONbored/metagraphed/issues/8733)) ([22edc53](https://github.com/JSONbored/metagraphed/commit/22edc53adcf9cc6e24fbf25dda23286ad2883d20))

### Documentation

* **limits:** generate the limits and access page from enforced config ([#8776](https://github.com/JSONbored/metagraphed/issues/8776)) ([a31237c](https://github.com/JSONbored/metagraphed/commit/a31237c5390588f39cf3e03ba852c2b3b82629dd))
</details>

<details><summary>platform: 1.1.0</summary>

### Features

* **api:** derive block emission from issuance instead of the stale storage item ([#8784](https://github.com/JSONbored/metagraphed/issues/8784)) ([0de9e04](https://github.com/JSONbored/metagraphed/commit/0de9e04eb59950992d0c9bf636a2e0f37bc01523))
* **api:** emission-pipeline surface — per-subnet share decomposition and the liquidity/chain-buy TAO split ([#8892](https://github.com/JSONbored/metagraphed/issues/8892)) ([ce9147e](https://github.com/JSONbored/metagraphed/commit/ce9147ed8fb09ae8962ab7727df93a3a1fa684a2))
* **api:** expose action summaries across API surfaces ([#8694](https://github.com/JSONbored/metagraphed/issues/8694)) ([d8b89b5](https://github.com/JSONbored/metagraphed/commit/d8b89b580d46b93498d7ff3826a4101595e32028))
* **api:** express multi-network addressing in OpenAPI and the contract ([#8758](https://github.com/JSONbored/metagraphed/issues/8758)) ([6b54610](https://github.com/JSONbored/metagraphed/commit/6b546104885939300407cc24d122f5d6f5e63105))
* **api:** per-key abuse controls — anomaly signals and a key-level blocklist ([#8695](https://github.com/JSONbored/metagraphed/issues/8695)) ([8123d93](https://github.com/JSONbored/metagraphed/commit/8123d9359e275c490614a9aad68d8f0a68ed13c9))
* **api:** per-tier rate limiting and daily quota enforcement at the Worker edge ([#8687](https://github.com/JSONbored/metagraphed/issues/8687)) ([6ce5cdb](https://github.com/JSONbored/metagraphed/commit/6ce5cdb9bbbe7ad31f0cc86c94b4514c222c9d1e))
* **api:** publish the spec-440 emission gate parameters ([#8787](https://github.com/JSONbored/metagraphed/issues/8787)) ([fbd1ce2](https://github.com/JSONbored/metagraphed/commit/fbd1ce2846c9ca8c96d430df8c6cb08877fcf518))
* **api:** queryable usage + cost rollup across ALL traffic ([#8706](https://github.com/JSONbored/metagraphed/issues/8706)) ([9d1088b](https://github.com/JSONbored/metagraphed/commit/9d1088bdb5903ee0576d69028e8bfea674817b31))
* **chain:** upgrade radar — track runtime releases from GitHub through testnet to mainnet ([#8737](https://github.com/JSONbored/metagraphed/issues/8737)) ([9e72a6f](https://github.com/JSONbored/metagraphed/commit/9e72a6fbc5f423b755a2cc3848bcd74bb90cc67a))
* **data:** box-side emission-gate sampler and dormant flow-path watch ([#8796](https://github.com/JSONbored/metagraphed/issues/8796)) ([8fed48d](https://github.com/JSONbored/metagraphed/commit/8fed48dbebbbe0cc986cf701be631a55ef6bc916))
* **data:** chain-derived runtime-transition backfill with reconcile ([#8757](https://github.com/JSONbored/metagraphed/issues/8757)) ([8b85795](https://github.com/JSONbored/metagraphed/commit/8b85795aaefc57e149b2d4d121ac03611befc3ad)), closes [#8756](https://github.com/JSONbored/metagraphed/issues/8756)
* **data:** emission-gate parameter and enablement change history ([#8793](https://github.com/JSONbored/metagraphed/issues/8793)) ([0daed14](https://github.com/JSONbored/metagraphed/commit/0daed14af8996b7bfb64ce98145d02b1f9f32f0d))
* **data:** ingest the v440 emission-pipeline inputs per subnet ([#8836](https://github.com/JSONbored/metagraphed/issues/8836)) ([49ca303](https://github.com/JSONbored/metagraphed/commit/49ca303e39deb69ae368d43688540656729eb5cb))
* **data:** pure TAO/USD index aggregator ([#8798](https://github.com/JSONbored/metagraphed/issues/8798)) ([457f369](https://github.com/JSONbored/metagraphed/commit/457f369d5c246223972c1b78c38e6a2d01d232a0))
* **data:** store and ingest the TAO/USD index ([#8800](https://github.com/JSONbored/metagraphed/issues/8800)) ([ab58751](https://github.com/JSONbored/metagraphed/commit/ab58751ee1e5d056cdac0d9fda30cc23e9a83fb1))
* **digest:** deterministic grounded draft generation ([#8770](https://github.com/JSONbored/metagraphed/issues/8770)) ([506fd3a](https://github.com/JSONbored/metagraphed/commit/506fd3af9649f7ac5ec25736e8cc3315f14e7d94))
* **digest:** generate and store weekly digests from the published feeds ([#8781](https://github.com/JSONbored/metagraphed/issues/8781)) ([5b7ef41](https://github.com/JSONbored/metagraphed/commit/5b7ef41c0f0b079888fed61027fae50300e6db31))
* **digest:** weekly digest generator spine — ISO weeks, threshold, grounding ([#8755](https://github.com/JSONbored/metagraphed/issues/8755)) ([12c733e](https://github.com/JSONbored/metagraphed/commit/12c733e015903df9526f6714c81709fbdccb4366))
* **feeds:** capture GitHub releases and surface them as per-subnet feed items ([#8754](https://github.com/JSONbored/metagraphed/issues/8754)) ([46bb3e4](https://github.com/JSONbored/metagraphed/commit/46bb3e42805fa6952440773a0842ccf15067c0b5)), closes [#8704](https://github.com/JSONbored/metagraphed/issues/8704)
* **feeds:** make the feed system discoverable — contract, OpenAPI, autodiscovery ([#8738](https://github.com/JSONbored/metagraphed/issues/8738)) ([1b7fae2](https://github.com/JSONbored/metagraphed/commit/1b7fae249847b670cc7f43ca5393634e003b35ce))
* **feeds:** widen per-subnet feed sources — chain governance events ([#8751](https://github.com/JSONbored/metagraphed/issues/8751)) ([99119f0](https://github.com/JSONbored/metagraphed/commit/99119f02c7017c066f65473190e90b52cebb8703)), closes [#8704](https://github.com/JSONbored/metagraphed/issues/8704)
* **registry:** derive machine verification from probe evidence ([#8690](https://github.com/JSONbored/metagraphed/issues/8690)) ([31143de](https://github.com/JSONbored/metagraphed/commit/31143dea3439a246af98c5be2e5d158eb967c08a))
* **ui+api:** tenant usage dashboard — quota headroom, throttling, and export ([#8728](https://github.com/JSONbored/metagraphed/issues/8728)) ([ab1cd13](https://github.com/JSONbored/metagraphed/commit/ab1cd13156d680a9b1a99123a6cc048814e45fed))
* **ui:** stop list routes redirecting to fully-expanded query strings ([#8790](https://github.com/JSONbored/metagraphed/issues/8790)) ([eaec728](https://github.com/JSONbored/metagraphed/commit/eaec728cfe3cf8b1bf3596af9487d4ad09621a50))
* **ui:** weekly digest pages at /news, generated from the digest store ([#8782](https://github.com/JSONbored/metagraphed/issues/8782)) ([1ec77de](https://github.com/JSONbored/metagraphed/commit/1ec77de570227f0b59fc4016bd4a39ac798c9bea))

### Bug Fixes

* **api+ui:** disambiguate emission_share as stage 1 of the emission pipeline ([#8789](https://github.com/JSONbored/metagraphed/issues/8789)) ([f0f68a3](https://github.com/JSONbored/metagraphed/commit/f0f68a3f474f88766ed86e1ef909b01f3ee23fd2))
* **api:** allow authorization and key/auth methods on CORS preflight ([#8831](https://github.com/JSONbored/metagraphed/issues/8831)) ([86634f7](https://github.com/JSONbored/metagraphed/commit/86634f7f00cf79a01041cdf5c7a754c10e14dc19)), closes [#8806](https://github.com/JSONbored/metagraphed/issues/8806)
* **api:** bound the realized_return baseline window so a stale snapshot can't stand in ([#8880](https://github.com/JSONbored/metagraphed/issues/8880)) ([fe78cef](https://github.com/JSONbored/metagraphed/commit/fe78cefab2fca6cdec5bab2201e20eb419d6e429))
* **api:** consecutive spec versions are never a coverage gap ([#8767](https://github.com/JSONbored/metagraphed/issues/8767)) ([4f25ab2](https://github.com/JSONbored/metagraphed/commit/4f25ab2509a5a4b8c2f692bb2c6c121871bd0a24)), closes [#8766](https://github.com/JSONbored/metagraphed/issues/8766)
* **api:** daily-rollup health windows return N calendar days, not N+1 ([#8867](https://github.com/JSONbored/metagraphed/issues/8867)) ([9b1c2bd](https://github.com/JSONbored/metagraphed/commit/9b1c2bd4831f052c4fbf707a852b8a4ef493430c))
* **api:** disclose interior gaps in the runtime-upgrade timeline ([#8753](https://github.com/JSONbored/metagraphed/issues/8753)) ([a6f6de4](https://github.com/JSONbored/metagraphed/commit/a6f6de4bb6d6dd45104a607d4dbafa8abc5c8ddc))
* **api:** exclude unsigned inherents from chain/fees medians and averages ([#8893](https://github.com/JSONbored/metagraphed/issues/8893)) ([133f027](https://github.com/JSONbored/metagraphed/commit/133f027c35c6424398517f6fa9e82fcc4f3e0a3b)), closes [#8809](https://github.com/JSONbored/metagraphed/issues/8809)
* **api:** expose tiered rate-limit headers to cross-origin browsers ([#8865](https://github.com/JSONbored/metagraphed/issues/8865)) ([c95cace](https://github.com/JSONbored/metagraphed/commit/c95caceda59227d2cd3d0203e9f73a66c5dce98f)), closes [#8813](https://github.com/JSONbored/metagraphed/issues/8813)
* **api:** gate handleGithubAccountUpsert with the internal-token check ([#8856](https://github.com/JSONbored/metagraphed/issues/8856)) ([7b0d7b8](https://github.com/JSONbored/metagraphed/commit/7b0d7b8c8b87c64af58c93722ae6e8f8e1c776c4)), closes [#8820](https://github.com/JSONbored/metagraphed/issues/8820)
* **api:** pad incident duration_ms by one probe cadence, publish transient flaps ([#8904](https://github.com/JSONbored/metagraphed/issues/8904)) ([22476f9](https://github.com/JSONbored/metagraphed/commit/22476f9bdb7843c26bbcf927d79c20d87e3eaf37))
* **api:** route /api/v1/watch/push-subscriptions to DATA_API at the edge ([#8835](https://github.com/JSONbored/metagraphed/issues/8835)) ([8539250](https://github.com/JSONbored/metagraphed/commit/8539250e4ca876ae075036f70c081ee300104989))
* **api:** spend daily quota only after the per-minute limiter accepts ([#8847](https://github.com/JSONbored/metagraphed/issues/8847)) ([efc6b81](https://github.com/JSONbored/metagraphed/commit/efc6b814edf41e5ac9dba759b7f0b088d040f8f9)), closes [#8812](https://github.com/JSONbored/metagraphed/issues/8812)
* **api:** treat missing top-holders net_flow as null, not zero ([#8851](https://github.com/JSONbored/metagraphed/issues/8851)) ([b722aa6](https://github.com/JSONbored/metagraphed/commit/b722aa6dfb0acbcf790375099408c9d0d2e20375)), closes [#8807](https://github.com/JSONbored/metagraphed/issues/8807)
* **api:** uncap account activity tx_count and publish modules_called_capped ([#8822](https://github.com/JSONbored/metagraphed/issues/8822)) ([#8890](https://github.com/JSONbored/metagraphed/issues/8890)) ([5ee1c15](https://github.com/JSONbored/metagraphed/commit/5ee1c15922e265283f48391fb6a97af2966ab80f))
* **api:** upsert surface_uptime_daily on stable surface_key ([#8853](https://github.com/JSONbored/metagraphed/issues/8853)) ([96ad929](https://github.com/JSONbored/metagraphed/commit/96ad929d90fe3472416306ed09238c7fff0d7c89)), closes [#8811](https://github.com/JSONbored/metagraphed/issues/8811)
* **build:** make artifact budget warnings mean something again ([#8783](https://github.com/JSONbored/metagraphed/issues/8783)) ([c4d561a](https://github.com/JSONbored/metagraphed/commit/c4d561aeab872d2ec58178f754eb5cd783f33773))
* **build:** regenerate contract artifacts after the signals refresh ([#8775](https://github.com/JSONbored/metagraphed/issues/8775)) ([eca46db](https://github.com/JSONbored/metagraphed/commit/eca46db8682392166b21a1279442129e91d4eac5))
* **ci:** regenerate the digest pages in the weekly sync ([#8785](https://github.com/JSONbored/metagraphed/issues/8785)) ([99d1a4c](https://github.com/JSONbored/metagraphed/commit/99d1a4c1097bf389c897b894d5e86a24216334db))
* **ci:** unblock Postgres migrations after the D1 retirement ([#8788](https://github.com/JSONbored/metagraphed/issues/8788)) ([4978a58](https://github.com/JSONbored/metagraphed/commit/4978a58949919abfeda100142a85d7f8cf772e60))
* **data:** keep the committed GitHub signals capture fresh ([#8768](https://github.com/JSONbored/metagraphed/issues/8768)) ([3870d75](https://github.com/JSONbored/metagraphed/commit/3870d75327c009b54d16079e8479d6d5c461ddf6))
* **data:** runtime-transition backfill — retry/backoff, genesis timestamp, --seed ([#8760](https://github.com/JSONbored/metagraphed/issues/8760)) ([7b88d29](https://github.com/JSONbored/metagraphed/commit/7b88d299758f1738713de1518143533665c9bcc3))
* **data:** write observed_at as epoch ms, not an ISO string ([#8764](https://github.com/JSONbored/metagraphed/issues/8764)) ([26ff4a7](https://github.com/JSONbored/metagraphed/commit/26ff4a72d2d3246932aa10da39f91193cab47269)), closes [#8762](https://github.com/JSONbored/metagraphed/issues/8762)
* **deploy:** give each Worker deploy its own PostHog release identity ([#8797](https://github.com/JSONbored/metagraphed/issues/8797)) ([2518251](https://github.com/JSONbored/metagraphed/commit/2518251db4c32d1ab0c6fdd0d8ee30694938d54e))
* **indexer+postgres:** make the archive-node backfill 172x faster ([#8791](https://github.com/JSONbored/metagraphed/issues/8791)) ([77dcb95](https://github.com/JSONbored/metagraphed/commit/77dcb95b3a78c7670154e0ac2866cfc14059ac92)), closes [#8368](https://github.com/JSONbored/metagraphed/issues/8368)
* **indexer:** let the tail backfill follow the archive node as it syncs ([#8795](https://github.com/JSONbored/metagraphed/issues/8795)) ([b8a85b5](https://github.com/JSONbored/metagraphed/commit/b8a85b5a1303a7a78a1481a8d22b65f53ad307bc))
* **mcp:** bound list_endpoints/list_rpc_endpoints limit and reject over-ceiling ([#8870](https://github.com/JSONbored/metagraphed/issues/8870)) ([52a0932](https://github.com/JSONbored/metagraphed/commit/52a0932c43268cc73688fa106c8f54b88c4dad51)), closes [#8830](https://github.com/JSONbored/metagraphed/issues/8830)
* **mcp:** guard call_rpc's error read against a non-envelope upstream 4xx ([#8863](https://github.com/JSONbored/metagraphed/issues/8863)) ([249d617](https://github.com/JSONbored/metagraphed/commit/249d61737a832a22f5b202e084000fefbbf7e9c9)), closes [#8834](https://github.com/JSONbored/metagraphed/issues/8834)
* **mcp:** publish get_feed netuid dependency in inputSchema ([#8881](https://github.com/JSONbored/metagraphed/issues/8881)) ([aca7a2a](https://github.com/JSONbored/metagraphed/commit/aca7a2afe441d2d9a4a3bcad5757702a71e4da64))
* **mcp:** reject out-of-range get_chain_identity_history limit ([#8854](https://github.com/JSONbored/metagraphed/issues/8854)) ([d0f3454](https://github.com/JSONbored/metagraphed/commit/d0f3454ab194880e1da5e8969109b55dcf1c4495)), closes [#8832](https://github.com/JSONbored/metagraphed/issues/8832)
* **mcp:** reject unrecognised get_account_transfers direction values ([#8850](https://github.com/JSONbored/metagraphed/issues/8850)) ([d6183d6](https://github.com/JSONbored/metagraphed/commit/d6183d69ec5651f16e8018dd4e54ff47c5e4f9ba))
* **mcp:** single-source FEED_KINDS including upgrades for get_feed ([#8861](https://github.com/JSONbored/metagraphed/issues/8861)) ([7650380](https://github.com/JSONbored/metagraphed/commit/76503806343df41b76f036efa47f5002c0e20bf9)), closes [#8828](https://github.com/JSONbored/metagraphed/issues/8828)
* **registry:** stop the callable catalog drifting from the published registry ([#8691](https://github.com/JSONbored/metagraphed/issues/8691)) ([5564166](https://github.com/JSONbored/metagraphed/commit/5564166f089e4fd6b51393cefcc42f3e1ecc8c8d))
* RPC proxies buffer the whole body before enforcing MAX_RPC_BODY_BYTES ([#8872](https://github.com/JSONbored/metagraphed/issues/8872)) ([3a975d8](https://github.com/JSONbored/metagraphed/commit/3a975d8ec9908e24ce7d11a5b9d2457e80510723))
* **schema:** allow positional arrays on chain-events args ([#8864](https://github.com/JSONbored/metagraphed/issues/8864)) ([f9aafcf](https://github.com/JSONbored/metagraphed/commit/f9aafcf93c0dc3514ca8c8a2cde1e8ddf35355f2))
* **schema:** delete hand-edited components shadowed by their Zod owners ([#8897](https://github.com/JSONbored/metagraphed/issues/8897)) ([65baf6e](https://github.com/JSONbored/metagraphed/commit/65baf6e60920b79520e6701fc7ba1481ac5c5f17))
* **schema:** make block_number optional on block-chain-events items ([#8843](https://github.com/JSONbored/metagraphed/issues/8843)) ([de9d23e](https://github.com/JSONbored/metagraphed/commit/de9d23e211fcbae31135416f5cfa6d1208c4ff9d))
* **scripts:** don't fail r2:manifest:dry-run outside CI ([#8916](https://github.com/JSONbored/metagraphed/issues/8916)) ([3cdcbd0](https://github.com/JSONbored/metagraphed/commit/3cdcbd05f88d1498c5b1772d90d8d880ddb684a9)), closes [#8915](https://github.com/JSONbored/metagraphed/issues/8915)
* **ui:** constrain /apis surface-card ExternalLink so long URLs truncate ([#8726](https://github.com/JSONbored/metagraphed/issues/8726)) ([bfa430c](https://github.com/JSONbored/metagraphed/commit/bfa430c07fee3e9a10fd0437798eaed723970166))
* **ui:** give SectionAnchor title row a shrink/truncate contract ([#8716](https://github.com/JSONbored/metagraphed/issues/8716)) ([5976f9e](https://github.com/JSONbored/metagraphed/commit/5976f9ef8d32d266b1f2f840453b119da166aeb9))
* **ui:** give the /endpoints latency-heatmap scroll container the mg-table-scroll affordance, drop 6 stale overflow-baseline entries ([#8708](https://github.com/JSONbored/metagraphed/issues/8708)) ([a8baad4](https://github.com/JSONbored/metagraphed/commit/a8baad44b8dd6a40b0ce5524a38e4d3e53209169))
* **ui:** keep sub-milli amounts visible in formatNumber ([#8815](https://github.com/JSONbored/metagraphed/issues/8815)) ([#8877](https://github.com/JSONbored/metagraphed/issues/8877)) ([434070f](https://github.com/JSONbored/metagraphed/commit/434070f61a6f2a8355606f02531660361c000028))
* **ui:** ratchet-clean apps/ui routes batch 2/4 (5 files, 15 violations) ([#8727](https://github.com/JSONbored/metagraphed/issues/8727)) ([0418b55](https://github.com/JSONbored/metagraphed/commit/0418b553cc90cb9e649e3e182a6b1012441a38c3)), closes [#8718](https://github.com/JSONbored/metagraphed/issues/8718)
* **ui:** ratchet-clean design-token violations in src/routes batch 1/4 ([#8729](https://github.com/JSONbored/metagraphed/issues/8729)) ([92ef7e6](https://github.com/JSONbored/metagraphed/commit/92ef7e6adb5f8f93f2844e130a6d66ba5db5ee30)), closes [#8717](https://github.com/JSONbored/metagraphed/issues/8717)
* **ui:** right-anchor SankeyMini last-column labels in horizontal mode ([#8725](https://github.com/JSONbored/metagraphed/issues/8725)) ([421c3b7](https://github.com/JSONbored/metagraphed/commit/421c3b7cdaf369e587fcb77ab12fbe5cdcf93c27))
* **ui:** shrink ExternalLink children so long CTA labels truncate ([#8736](https://github.com/JSONbored/metagraphed/issues/8736)) ([49dcfeb](https://github.com/JSONbored/metagraphed/commit/49dcfebb17948b55318f2576cf02d065e6b41850))
* **ui:** stop padding failed AMM exit quotes with un-slipped spot in Simulated exit ([#8907](https://github.com/JSONbored/metagraphed/issues/8907)) ([75ff6a9](https://github.com/JSONbored/metagraphed/commit/75ff6a9fe69dbc16638af346b97834929bd41af8))
* **ui:** stop the subnet operational/curation tiles reporting real data on a failed query ([#8866](https://github.com/JSONbored/metagraphed/issues/8866)) ([1e27452](https://github.com/JSONbored/metagraphed/commit/1e274522ddf01c46797428fb88c71dea8dec928d))
* **ui:** swap raw Bone & Ink one-off values for design tokens in routes batch 3/4 ([#8733](https://github.com/JSONbored/metagraphed/issues/8733)) ([22edc53](https://github.com/JSONbored/metagraphed/commit/22edc53adcf9cc6e24fbf25dda23286ad2883d20))

### Performance

* **api:** hoist duplicated OpenAPI examples into components.examples ([#8780](https://github.com/JSONbored/metagraphed/issues/8780)) ([24e101c](https://github.com/JSONbored/metagraphed/commit/24e101ca0a8eb2e06922719804cf4e5e48e86435)), closes [#8763](https://github.com/JSONbored/metagraphed/issues/8763)
* **test:** scan once for public-safety, not once per test ([#8909](https://github.com/JSONbored/metagraphed/issues/8909)) ([ec6a771](https://github.com/JSONbored/metagraphed/commit/ec6a77178bf40d215468b86f140a457e69aa5b97)), closes [#8908](https://github.com/JSONbored/metagraphed/issues/8908)

### Documentation

* **adr:** accept ADR 0022 — data API free, self-hosted surfaces paid ([#8734](https://github.com/JSONbored/metagraphed/issues/8734)) ([f1250ae](https://github.com/JSONbored/metagraphed/commit/f1250ae216f7fe83d3e506de55aef18d0e6a6a92))
* **adr:** record the first-party TAO/USD index posture and methodology ([#8792](https://github.com/JSONbored/metagraphed/issues/8792)) ([b9e0568](https://github.com/JSONbored/metagraphed/commit/b9e0568cc54e0e90c0d9a3867f2a3daf852bb9d3))
* **adr:** record the v440 emission-pipeline publishing posture ([#8779](https://github.com/JSONbored/metagraphed/issues/8779)) ([e98d988](https://github.com/JSONbored/metagraphed/commit/e98d988b16ef74974214847eea72fc056eb2e79f))
* **deploy:** ETH_RPC_URL is a Worker secret, not a var ([#8842](https://github.com/JSONbored/metagraphed/issues/8842)) ([e305ce8](https://github.com/JSONbored/metagraphed/commit/e305ce8f2169c575452208f507a2ada8d6487e49))
* **gardening:** record 2026-07-31 zero-yield contributor-pipeline run ([#8799](https://github.com/JSONbored/metagraphed/issues/8799)) ([f608675](https://github.com/JSONbored/metagraphed/commit/f608675c58054a972e5da8f3ae53cc408638cc0c))
* **limits:** generate the limits and access page from enforced config ([#8776](https://github.com/JSONbored/metagraphed/issues/8776)) ([a31237c](https://github.com/JSONbored/metagraphed/commit/a31237c5390588f39cf3e03ba852c2b3b82629dd))
* TAO/USD venue survey, and an on-chain index superseding ADR 0024 ([#8794](https://github.com/JSONbored/metagraphed/issues/8794)) ([02c52c3](https://github.com/JSONbored/metagraphed/commit/02c52c3edf947bb11578749445ccc9095f96f009))
</details>

---


JSONbored:
🤖 Created releases:

- [ui-v0.14.0](https://github.com/JSONbored/metagraphed/releases/tag/ui-v0.14.0)
- [platform-v1.1.0](https://github.com/JSONbored/metagraphed/releases/tag/platform-v1.1.0)

:sunflower:

Issue #8228, opened 2026-07-26
MCP: add network-awareness to subnet/data tools (currently mainnet-hardcoded)


JSONbored:
## Summary

`src/mcp-server.ts`'s subnet/data tools (`list_subnets`, `get_subnet`, and effectively everything else) read from the bare (mainnet) artifact path with no `network` parameter at all — confirmed by reading the tool implementations directly. The one exception is `call_rpc`, whose `inputSchema` (`CallRpcInputSchema`) already accepts `network: "finney" | "test"` and proxies through the same network-aware RPC handler REST callers use.

Net effect: an agent using metagraphed's MCP tools today always gets mainnet data, with no way to ask about testnet subnets except raw RPC calls.

## Ask

Evaluate adding an optional `network` parameter to the subnet/data tools that mirrors `call_rpc`'s pattern, gated by whatever testnet artifacts actually exist (see the sibling issue on building out more testnet artifacts) — so tools return a clear "not available for this network" error rather than silently mainnet data, and return real testnet data where it exists.

## Non-goals

- Full parity across all ~150 MCP tools in one pass — scope to the subnet/data-lookup tools first

PR #8941, opened 2026-07-31
fix(mcp): validate the network argument before it selects an artifact key (#8804)

### LINKED ISSUE #8804: fix(mcp): network arg is never validated at runtime — MCP silently serves the testnet registry as mainnet
## Context

The MCP server performs **no runtime validation of tool arguments** beyond "is it an object" and
"are there unknown keys". The published Zod schemas are compiled to JSON Schema for the tool
listing and then never enforced. The most damaging consequence today: `network` silently routes
mainnet requests to the **testnet** registry.

### The specific bug

`src/mcp-server.ts:1813-1819`:

```ts
export function networkArtifactPath(
  artifactPath: string,
  network?: "finney" | "test",
): string {
  if (!network || network === "finney") return artifactPath;
  return artifactPath.replace(/^\/metagraph\//, "/metagraph/testnet/");
}
```

The TypeScript signature narrows to `"finney" | "test"`, but the runtime value is unvalidated JSON
off the wire. **Any string that is not exactly `"finney"` takes the testnet branch.** Call sites
pass straight through: `src/mcp-server.ts:3248` (`list_subnets`) and `src/mcp-server.ts:3404`
(`get_subnet_detail`).

The trigger is not adversarial — it is what a well-behaved agent does:

- `get_networks` (`src/mcp-server.ts:7907`) publishes the router's own `NETWORKS` map
  (`workers/api.ts:4129`), which lists **`mainnet`, `finney`, `testnet`, `test`, `local`** — with
  **`mainnet` as the canonical friendly id**.
- An agent reads that, calls `list_subnets {network: "mainnet"}`.
- `"mainnet" !== "finney"` → reads `/metagraph/testnet/subnets.json`, a **real published artifact**
  (`src/network-artifacts.ts:21`) → returns the **testnet registry**.
- `list_subnets`' response projection (`src/mcp-server.ts:3266-3281`) has **no `network` field**, so
  nothing in the response reveals the substitution. `network: "local"` behaves identically.
- `workers/api.ts:4160` states the stakes in the repo's own words: "testnet SN-N is unrelated to
  mainnet SN-N".

`get_subnet_detail` is worse: `src/mcp-server.ts:3413` returns early for any non-`"finney"` network,
so the response also silently loses the live mainnet economics overlay. The agent gets a testnet
subnet's structural record, no economics, and no indication that either happened.

**Only MCP is exposed.** REST canonicalizes via `resolveNetworkPrefix` (`workers/api.ts:4225`)
matching `NETWORK_PREFIX_PATTERN`, so `/api/v1/mainnet/...` resolves correctly. GraphQL declares
`enum Network { finney test }` (`src/graphql-sdl.ts:21`) and graphql-js rejects anything else. MCP
is the one unguarded caller.

### The systemic root cause

`src/mcp-server.ts:2588-2604` is the **only** argument validation at dispatch
(`src/mcp-server.ts:11773`):

```ts
function validateToolArguments(tool: Row, args: Row) {
  if (args === undefined || args === null) return {};
  if (
    typeof args !== "object" ||
    Array.isArray(args) ||
    (tool.inputSchema?.additionalProperties === false &&
      Object.keys(args).some(
        (key) => !Object.hasOwn(tool.inputSchema?.properties ?? {}, key),
      ))
  ) {
    throw toolError("invalid_params", `Invalid arguments for tool ${tool.name}.`);
  }
  return args;
}
```

No type check, no enum check, no required check, no min/max. Every `z.enum(...)`, `z.int().min(1)`,
`.max(200)`, and `required` across all 200+ entries in `MCP_TOOLS` is **decorative at runtime** —
honoured only by clients that validate against the published schema, and by whatever ad-hoc guards
individual handlers happen to have written. `grep` for `optionalEnum(args, "network", ...)` returns
nothing; the helper exists (`src/mcp-server.ts:3060-3070`) and is used for other enums, but
`network` was never wired to it.

**This issue has two parts and a decision.** Part (a) is a bounded fix and is not optional. Part (b)
is a design call that must be answered explicitly in the PR body, not skipped.

**`optionalEnum` at `src/mcp-server.ts:3060-3070` is the existing guard shape — use it verbatim for
`network`, do not write a new helper.**

```ts
function optionalEnum(args: Row, key: string, allowed: readonly string[]) {
  const value = args?.[key];
  if (value === undefined || value === null || value === "") return null;
  if (typeof value !== "string" || !allowed.includes(value)) {
    throw toolError(
      "invalid_params",
      `Argument \`${key}\` must be one of: ${allowed.join(", ")}.`,
    );
  }
  return value;
}
```

It is already used for `sort` and `order` in the very handler that has the bug
(`src/mcp-server.ts:3257-3258`), two lines below the unguarded `network` pass-through.

The published enum to validate against is `McpNetworkSchema` (`schemas-src/shared.ts:25`,
`z.enum(["finney", "test"])`) — the exact schema both `ListSubnetsInputSchema`
(`schemas-src/mcp-tools/list-subnets.ts:61`) and `GetSubnetDetailInputSchema`
(`schemas-src/mcp-tools/get-subnet-detail.ts:18`) already declare and that the runtime ignores.

- A new bespoke `validateNetwork()` helper instead of calling `optionalEnum`.
- Coercing/aliasing `"mainnet"` → `"finney"` and `"testnet"` → `"test"` **in place of** rejecting.
  Aliasing may be added as well (see Requirement 3), but a value outside the published enum and
  outside any documented alias set must still be **rejected**, never silently routed.
- Fixing `list_subnets` only. `get_subnet_detail` has the same defect and a worse blast radius.
- Answering part (b) with silence. If the chosen answer is "no dispatch-level validation", the PR
  body must say so and say why.

### (a) `network` — required, mechanical

1. Both `src/mcp-server.ts:3248` (`list_subnets`) and `src/mcp-server.ts:3404`
   (`get_subnet_detail`) resolve `network` through `optionalEnum(args, "network", ["finney",
   "test"])` **before** it reaches `networkArtifactPath`.
2. Accepted: `"finney"`, `"test"`, `undefined`, `null`, `""` (the last three mean finney).
   Rejected with `toolError("invalid_params", …)`: **every** other value, including `"mainnet"`,
   `"testnet"`, `"local"`, `"FINNEY"`, `0`, `true`, `[]`, `{}`. The error message must name the
   accepted values, matching `optionalEnum`'s existing text.
3. **Resolve the `get_networks` inconsistency.** `get_networks` advertises `mainnet`/`testnet`/
   `local` as usable network ids while these tools accept only `finney`/`test`. Pick one and do it:
   either (i) accept `"mainnet"`/`"testnet"` as documented aliases that canonicalize to
   `finney`/`test` (and reject `"local"`, which has no artifact), or (ii) leave the enum strict and
   amend `get_networks`' description so it does not present ids these tools reject. Do **not** leave
   the two surfaces disagreeing.
4. `networkArtifactPath`'s branch keeps its narrowed signature; the guarantee that only
   `finney`/`test` reach it now holds at runtime, not just in types.

### (b) Dispatch-level validation — a decision that must be made

5. Decide and state, in the PR body, whether `validateToolArguments`
   (`src/mcp-server.ts:2588-2604`) should enforce the published `inputSchema` — types, enums,
   `required`, and numeric bounds — for every tool, rather than each handler re-implementing guards
   by hand. The considerations to weigh and address explicitly:
   - **Behaviour change risk.** Handlers today are deliberately lenient in places (e.g. `clampLimit`
     accepts and clamps values the schema forbids). Schema-driven validation would turn some of
     those into hard `invalid_params`. Enumerate which tools change behaviour, or scope the
     enforcement so none do.
   - **Cost.** `inputSchema` is already `z.toJSONSchema`-derived, so the source Zod object exists —
     validating with the Zod schema directly at dispatch is cheaper than re-walking JSON Schema.
   - **Fallback.** If full enforcement is judged too broad for one PR, the answer may be "no, keep
     per-handler guards" — but then the PR must state that and (a) still ships.
6. If (b) is answered "yes", it ships **in this same PR** with per-tool behaviour changes
   enumerated. Splitting it into a follow-up issue does not satisfy this requirement; deciding "no,
   and here is why" does.

## Deliverables

- [ ] `src/mcp-server.ts` — `network` validated via `optionalEnum` in both `list_subnets` and
      `get_subnet_detail` handlers.
- [ ] The `get_networks` / tool-enum disagreement resolved per Requirement 3 (alias support, or a
      corrected `get_networks` description) — whichever is chosen, implemented, not just noted.
- [ ] A written decision on Requirement 5 in the PR body, with the behaviour-change analysis; if the
      decision is "enforce", the enforcement lands in this PR.
- [ ] `tests/mcp-server.test.ts` — a regression test that **fails before the fix and passes after**: assert
      `list_subnets {network: "mainnet"}` and `get_subnet_detail {netuid: 1, network: "mainnet"}`
      both throw `invalid_params` (or canonicalize to finney if aliasing is chosen) and, critically,
      that **neither reads a `/metagraph/testnet/` artifact path**. Assert the same for `"local"`,
      `"TEST"`, and a non-string value. Assert `network: "test"` still reaches the testnet path so
      the fix does not break legitimate testnet reads.
- [ ] Patch coverage: every changed line and branch in `src/**` / `workers/**` is covered —
      `codecov/patch` enforces `target: 99%, threshold: 0%`. Measure unsharded with
      `npm run test:coverage`. Note that a new defensive `throw` still counts as a changed branch;
      make the guard testable rather than ignoring it.

satisfy this issue.

## Expected Outcome

`list_subnets` and `get_subnet_detail` accept `network` only as `finney` / `test` / absent (plus any
aliases explicitly chosen under Requirement 3), and reject everything else with `invalid_params`
naming the accepted values — matching what GraphQL's `enum Network { finney test }` already enforces
and what REST's `resolveNetworkPrefix` already canonicalizes. `get_subnet_detail {network:
"mainnet"}` can no longer return a testnet subnet with the mainnet economics overlay stripped and no
indication of either. The PR body carries an explicit, reasoned decision on whether
`validateToolArguments` gains schema-driven enforcement for all tools, and implements it if the
answer is yes.

## Links & Resources

- `src/mcp-server.ts:2588-2604` — `validateToolArguments`, the systemic gap.
- `src/mcp-server.ts:11773` — the single dispatch call site.
- `src/mcp-server.ts:1813-1819` — `networkArtifactPath`, the unguarded branch.
- `src/mcp-server.ts:3248`, `:3404`, `:3413` — the two vulnerable handlers and the economics-overlay
  early return.
- `src/mcp-server.ts:3060-3070` — **`optionalEnum`, the guard to mirror**; already used at `:3257`.
- `src/mcp-server.ts:7907` + `workers/api.ts:4129`, `:4160` — `get_networks` and the `NETWORKS` map
  that makes `"mainnet"` the realistic input.
- `src/network-artifacts.ts:21` — the testnet artifacts are real and published.
- `workers/api.ts:4225` — REST's `resolveNetworkPrefix` (unaffected).
- `src/graphql-sdl.ts:21` — GraphQL's `enum Network { finney test }` (unaffected).
- `schemas-src/shared.ts:25` — `McpNetworkSchema`, the published enum the runtime ignores; used by
  `schemas-src/mcp-tools/list-subnets.ts:61` and `schemas-src/mcp-tools/get-subnet-detail.ts:18`.
- `tests/mcp-server.test.ts` — the regression test's home.

JSONbored:
## Summary

- Closes #8804 — MCP's `network` argument was never validated at runtime, so `list_subnets {network:"mainnet"}` and `get_subnet_detail {netuid, network:"mainnet"}` silently served the **testnet** registry, with nothing in either response to reveal it.

### (a) `network` is validated before it selects an artifact key

- `src/mcp-server.ts` — both handlers resolve `network` through the existing **`optionalEnum`** guard (the one already used two lines away for `sort` and `order`), not a new bespoke helper. Accepted: `finney`, `test`, absent, `null`, `""` (the last three mean finney). Everything else throws `toolError("invalid_params", …)` naming the accepted values, via `optionalEnum`'s own message.
- The allowed list is `McpNetworkSchema.options` — **read from the published schema** (`schemas-src/shared.ts`), not hand-copied. The whole failure mode here was the runtime and the advertised `inputSchema` disagreeing; a second literal would let them drift again.
- `optionalEnum` is now generic in its member type (`<T extends string>(…: readonly T[]): T | null`). The guard already proves membership, so this narrows the result to the enum instead of `string` and saves the call sites a cast. No behaviour change; every existing caller keeps inferring `string`.
- `networkArtifactPath` keeps its narrowed `"finney" | "test"` signature — the guarantee now holds at runtime, not only in types.
- `get_subnet_detail`'s economics-overlay early return reads the **validated** value rather than `args?.network`.

### Requirement 3 — the `get_networks` disagreement: chose (ii), keep the enum strict

`get_networks` publishes `mainnet`/`testnet`/`local` while these tools take `finney`/`test`. I took **(ii)** — leave the enum strict and fix the description — rather than (i) aliasing:

- The chain-name spelling is a **deliberate, documented** choice: `schemas-src/shared.ts`'s comment on `McpNetworkSchema` says these are "the same chain-name spelling `call_rpc` already accepts — an agent should not have to learn that the RPC lane says `test` while a data lane says `testnet`". Aliasing `mainnet`/`testnet` into the two data tools would restore exactly that split, in the other direction, because `call_rpc` still takes `finney`/`test` (its pool map has no `mainnet` key). Fixing that consistently would mean changing `call_rpc` too — outside this issue.
- `get_networks`' ids genuinely *are* the REST URL-path alias set (`/api/v1/testnet/...`, `NETWORKS` in `workers/api.ts`). They were never wrong; they were being read as if they were MCP argument values.
- `local` has no artifact on any surface, so an alias set would have needed a third rule anyway.

So `get_networks`' description now says the ids and aliases it lists are REST URL-path segments, that an MCP `network` **argument** takes the chain name (`finney`/`test`, the same spelling `call_rpc` uses), that `mainnet`/`testnet`/`local` are rejected there, and that only `list_subnets`/`get_subnet_detail` take `network` at all. Combined with the `invalid_params` message naming the accepted values, an agent self-corrects in one turn instead of silently reading the wrong chain. No schema change, so no contract regen.

### (b) Dispatch-level schema enforcement — decision: **no**, keep per-handler guards

`validateToolArguments` checks only "is it an object" and "any unknown keys", so every `z.enum`, `.min()`, `.max()` and `required` across the 200+ `MCP_TOOLS` entries is decorative at runtime. That is a real systemic gap and I considered enforcing the published `inputSchema` at dispatch in this PR. **I am not doing it here**, for three reasons:

1. **The behaviour-change surface is not enumerable within this PR.** Handlers are deliberately lenient in places — `clampLimit` accepts and clamps a `limit` the schema forbids, and `optionalEnum` itself accepts `""` where a published enum does not. Turning enforcement on converts an unknown-but-large set of currently-succeeding calls into hard `invalid_params`. The issue asks for those tools to be enumerated or the enforcement scoped so none change; doing that honestly means auditing 206 handlers, and its diff would dwarf and obscure a security fix that should ship now.
2. **Enforcement wants the Zod objects, which dispatch does not have.** `MCP_TOOLS` carries only the `z.toJSONSchema`-derived JSON Schema. Validating with Zod directly (the cheaper option the issue notes) means threading the source schema through all 206 entries — a mechanical refactor, but a wide one, and it is not a bugfix.
3. A scoped variant — enforce only `enum` members, read off `inputSchema.properties[key].enum` — is genuinely tempting and would close this whole bug class. It still changes behaviour for every tool whose handler currently tolerates an out-of-enum value by falling back to a default, which is the same unbounded audit as (1).

Per Requirement 5's own fallback, the answer is recorded here as "no, keep per-handler guards", and (a) ships. I will open a follow-up issue for the systemic work with the audit as its first deliverable — that is a *terminal* no here, not a deferred yes.

## Tests

`tests/mcp-server.test.ts` — a new `network argument validation (#8804)` block, with a `readArtifact` tracer so each case can assert what the handler *asked for*, not just what it returned:

- `mainnet`, `testnet`, `local`, `TEST`, `FINNEY` × `list_subnets` and `get_subnet_detail` — each rejected with `invalid_params` matching `must be one of: finney, test`, **and** no `/metagraph/testnet/` path read at all.
- Non-string values (`0`, `true`, `[]`, `{}`) rejected, not coerced. (Note `0` and `true` were not merely mis-served before — `0` is falsy so it read *mainnet*, while `[]` took the testnet branch. Both are now `invalid_params`.)
- `network: "test"` still reaches `/metagraph/testnet/subnets.json` and `/metagraph/testnet/subnets/1.json` and still suppresses the mainnet economics overlay — the fix does not break legitimate testnet reads.
- `""` / `null` still mean finney and still get the mainnet overlay.

The pre-existing `rejects an unknown network value (#8228)` test is kept and now has company: it only ever passed **by accident of a missing fixture** — `"devnet"` hit the testnet key, the test's `deps` had no artifact there, and the resulting `artifact_not_found` was read as a validation failure. In production, where the testnet artifact does exist, the same call succeeds with the wrong chain's data. That is the bug in one sentence.

Verified: 11 of the new cases fail against the unfixed handler and all pass with it (`git stash` on `src/mcp-server.ts` only).

## Registry Safety

- [x] Links a tracked, currently-open issue (`Closes #8804`) — required.
- [x] Generated artifacts were produced by repo scripts, not hand-edited. (None regenerated — no `schemas-src/` change; the MCP server card is worker-computed.)
- [x] Public API/OpenAPI/schema changes are intentional and documented. (None. The accepted `network` values are unchanged — they are now actually enforced.)

## Validation

- [x] `npm run lint`
- [x] `npm run format:check`
- [x] `npm run typecheck`
- [x] `npm run validate`
- [x] `npx vitest run tests/mcp-server.test.ts` — 1272 passed
- [x] `npx vitest run tests/docs-content-drift.test.ts`
- [x] `git diff --check`

### DIFF
--- a/src/mcp-server.ts
+++ b/src/mcp-server.ts
@@ -44,6 +44,7 @@ import {
   ListSubnetsInputSchema,
   ListSubnetsOutputSchema,
 } from "../schemas-src/mcp-tools/list-subnets.ts";
+import { McpNetworkSchema } from "../schemas-src/shared.ts";
 import {
   GetSubnetInputSchema,
   GetSubnetOutputSchema,
@@ -1822,6 +1823,14 @@ function toolError(code: string, message: string) {
   return error;
 }
 
+// The published `network` enum, read from the schema the tools declare rather
+// than hand-copied, so the runtime guard and the advertised inputSchema cannot
+// drift apart (#8804 — they had). Every handler taking `network` must resolve
+// it through optionalEnum against this list BEFORE it reaches
+// networkArtifactPath, whose non-finney branch is otherwise reachable by any
+// unvalidated string off the wire.
+const MCP_NETWORK_VALUES = McpNetworkSchema.options;
+
 // #8228: rewrite a mainnet artifact path for the requested chain network.
 // Mirrors the Worker's own /metagraph/{prefix}/… partitioning (see
 // NETWORK_KEY_PREFIXES in src/artifact-storage.ts): finney is the unprefixed
@@ -3076,16 +3085,23 @@ const COVERAGE_DEPTH_TIERS = [
 ];
 const COVERAGE_DEPTH_SEVERITIES = ["hard", "missing-data", "needs-review"];
 
-function optionalEnum(args: Row, key: string, allowed: readonly string[]) {
+// Generic in the member type so a caller passing a `readonly ["a","b"]` gets
+// back `"a" | "b" | null` rather than a bare string -- the guard already proves
+// membership, so the narrowing is free and saves every call site a cast.
+function optionalEnum<T extends string>(
+  args: Row,
+  key: string,
+  allowed: readonly T[],
+): T | null {
   const value = args?.[key];
   if (value === undefined || value === null || value === "") return null;
-  if (typeof value !== "string" || !allowed.includes(value)) {
+  if (typeof value !== "string" || !allowed.includes(value as T)) {
     throw toolError(
       "invalid_params",
       `Argument \`${key}\` must be one of: ${allowed.join(", ")}.`,
     );
   }
-  return value;
+  return value as T;
 }
 
 function optionalGapCode(args: Row) {
@@ -3293,9 +3309,13 @@ export const MCP_TOOLS: McpToolDefinition[] = [
       target: "draft-2020-12",
     }),
     async handler(args: z.infer<typeof ListSubnetsInputSchema>, ctx: McpCtx) {
+      // Validated, not trusted: an unrecognised string used to take
+      // networkArtifactPath's testnet branch and silently serve the testnet
+      // registry (#8804). optionalEnum returns null for absent/empty.
+      const network = optionalEnum(args, "network", MCP_NETWORK_VALUES);
       const index = await loadArtifactData(
         ctx,
-        networkArtifactPath("/metagraph/subnets.json", args?.network),
+        networkArtifactPath("/metagraph/subnets.json", network ?? undefined),
       );
       const all = Array.isArray(index.subnets) ? index.subnets : [];
       // Categorical inclusion (status/subnet_type/domain) and exclusion
@@ -3449,9 +3469,16 @@ export const MCP_TOOLS: McpToolDefinition[] = [
       ctx: McpCtx,
     ) {
       const netuid = requireNetuid(args);
+      // See list_subnets: `network` is unvalidated JSON off the wire, and an
+      // unrecognised value used to serve a TESTNET record with the mainnet
+      // economics overlay silently dropped below (#8804).
+      const network = optionalEnum(args, "network", MCP_NETWORK_VALUES);
       const detail = await loadArtifactData(
         ctx,
-        networkArtifactPath(`/metagraph/subnets/${netuid}.json`, args?.network),
+        networkArtifactPath(
+          `/metagraph/subnets/${netuid}.json`,
+          network ?? undefined,
+        ),
       );
       // Same live-economics overlay /api/v1/subnets/{netuid} attaches (#1308):
       // one call carries validator/miner counts, registration, stake, and
@@ -3460,7 +3487,7 @@ export const MCP_TOOLS: McpToolDefinition[] = [
       // record would report mainnet numbers against a testnet subnet (the
       // exact leak tests/network-routing.test.ts guards on the REST side).
       // Testnet carries its own chain economics inside `subnet.economics`.
-      if (args?.network && args.network !== "finney") return detail;
+      if (network && network !== "finney") return detail;
       const { economics } = await loadSubnetEconomics(ctx, netuid);
       return economics ? { ...detail, economics } : detail;
     },
@@ -8058,7 +8085,12 @@ export const MCP_TOOLS: McpToolDefinition[] = [
       "planning a multi-step task against a non-mainnet network: it answers " +
       '"can I get chain data on testnet?" without issuing a request that 404s. ' +
       "The served/unserved split is derived from the router's own routing rules, " +
-      "not a hand-maintained list. Mirrors GET /api/v1/networks.",
+      "not a hand-maintained list. Mirrors GET /api/v1/networks. NOTE: the ids " +
+      "and aliases listed here are REST URL-path segments (/api/v1/testnet/...). " +
+      "An MCP tool's `network` ARGUMENT takes the chain name — `finney` or " +
+      "`test` — the same spelling call_rpc uses; `mainnet`/`testnet`/`local` are " +
+      "rejected there. Only list_subnets and get_subnet_detail take `network` at " +
+      "all; `local` is a per-developer chain with no hosted data on any surface.",
     inputSchema: z.toJSONSchema(GetNetworksInputSchema, {
       target: "draft-2020-12",
     }),
--- a/tests/mcp-server.test.ts
+++ b/tests/mcp-server.test.ts
@@ -8456,6 +8456,169 @@ describe("list_subnets", () => {
     assert.equal(res.body.result.isError, true);
   });
 
+  // #8804: `network` was passed straight into networkArtifactPath, whose
+  // "anything that isn't finney" branch reads /metagraph/testnet/. Any
+  // unrecognised string -- including "mainnet", which get_networks itself
+  // advertises as the canonical id -- therefore served the TESTNET registry
+  // with nothing in the response to reveal the substitution. The pre-existing
+  // "devnet" test above only passed by accident of a missing fixture: it
+  // asserted an artifact_not_found, not a validation error.
+  describe("network argument validation (#8804)", () => {
+    // Records every artifact path the handler asks for, so a test can assert a
+    // rejected network never even attempted a testnet read.
+    function tracingDeps(artifacts: Row = {}) {
+      const base = makeDeps(artifacts);
+      const paths: string[] = [];
+      return {
+        paths,
+        deps: {
+          ...base,
+          readArtifact(env: unknown, path: string) {
+            paths.push(path);
+            return base.readArtifact(env, path);
+          },
+        },
+      };
+    }
+
+    const networkArtifacts = {
+      "/metagraph/subnets.json": {
+        subnets: [{ netuid: 7, slug: "allways", name: "MAINNET Allways" }],
+      },
+      "/metagraph/testnet/subnets.json": {
+        subnets: [{ netuid: 7, slug: "sn-7", name: "TESTNET seven" }],
+      },
+      "/metagraph/subnets/1.json": {
+        schema_version: 1,
+        subnet: { netuid: 1, slug: "apex", name: "MAINNET Apex" },
+      },
+      "/metagraph/testnet/subnets/1.json": {
+        schema_version: 1,
+        subnet: { netuid: 1, slug: "sn-1", name: "TESTNET one" },
+      },
+      "/metagraph/economics.json": {
+        schema_version: 1,
+        summary: { with_economics_count: 1 },
+        subnets: [{ netuid: 1, open_slots: 3 }],
+      },
+    };
+
+    // Every one of these is a value the testnet branch used to accept. The
+    // first two are the realistic ones: get_networks publishes `mainnet` as the
+    // canonical id and `local` as an addressable network.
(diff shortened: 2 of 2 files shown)


JSONbored:
Follow-up for the Requirement 5 "no" is filed as #8942 — the per-tool behaviour-change audit is its first deliverable, so the decision here stays terminal rather than deferred.

PR #8943, opened 2026-07-31
fix(api): value top-holders delegated stake in TAO, not cross-subnet alpha (#8803)

### LINKED ISSUE #8803: fix(api): top-holders delegated_tao/total_tao add per-subnet alpha to real TAO (top account = 71% of max supply)
## Context

`GET /api/v1/accounts/top-holders` (and MCP `get_top_holders`) reports **alpha tokens as if they
were TAO**, then **adds them to real TAO**. The leaderboard's headline column is therefore not a
quantity that exists.

The aggregate at `workers/data-api.ts:6832-6835`:

```sql
FULL OUTER JOIN (
  SELECT np.coldkey, SUM(np.share_fraction * n.stake_tao) AS delegated_tao
  FROM nominator_positions np
  JOIN neurons n ON n.hotkey = np.hotkey AND n.netuid = np.netuid
  GROUP BY np.coldkey) d ON d.coldkey = b.ss58
```

and `src/top-holders.ts:74`:

```ts
total_tao: freeTao + delegatedTao,
```

where `free_tao` (`workers/data-api.ts:6824`) is genuine TAO from a `System::Account` chain-state
scan.

Three premises, each verified in-tree:

1. **`nominator_positions` contains only `netuid != 0` rows.**
   `apps/indexer-rs/src/bin/poller/jobs/validator_nominators.rs:148` —
   `if netuid != 0 && shares.bits > 0 { shares_by_hotkey_netuid... }`, with the comment at `:121-123`
   ("only netuid != 0 and shares > 0 -- root is never share-tracked"). Root/TAO-denominated stake is
   structurally excluded from the table.
2. **Non-root `stake_tao` is denominated in that subnet's alpha, not TAO.**
   `src/metagraph-neurons.ts:941-944` — "Root (netuid 0) stake is TAO-denominated with no AMM/price
   exposure; every other netuid's stake is that subnet's alpha token (#2550)".
3. **The repo's own economics code treats `stake_tao` as alpha.**
   `scripts/lib/economics-artifacts.ts:56-68` `computeAlphaMarketCapTao(alphaPriceTao,
   totalStakeTao) => alphaPriceTao * totalStakeTao`, commented "total_stake_tao is the on-chain
   circulating-alpha proxy". That identity is only dimensionally valid if `total_stake_tao` is alpha.

So `delegated_tao` is 100% alpha and 0% TAO — summed raw across every subnet with **no
`alpha_price_tao` conversion** — and then added to real TAO to make `total_tao`.

**Live production proof** (`GET /api/v1/accounts/top-holders`, 2026-07-31):

| ss58 | reported `delegated_tao` |
| --- | --- |
| `5HSzRtcQ…` (top row) | **14,984,421** |

That single account is reported as holding **71.4% of every TAO that will ever exist**. The top five
rows' `total_tao` sums to **33,036,023** against TAO's **21,000,000** hard cap — **1.57x max
supply**. The numbers are not merely imprecise, they are arithmetically impossible.

Why this is worse than a scaling error: alpha prices currently span roughly 0.002–0.3 TAO, so the
per-account overstatement is **~3x to ~500x and non-uniform**. `total_tao` is the **default sort**
(`DEFAULT_TOP_HOLDERS_SORT`, consumed at `workers/data-api.ts:6846`), so the ordering itself is
wrong — an account whose stake sits in cheap-alpha subnets outranks one holding more real value.
And `src/contracts.ts:2835` advertises this route to agents as carrying "free/delegated/total TAO
columns matching the taostats-style Account/Free/Delegated/Total benchmark", i.e. we explicitly
claim these are comparable to a reference that reports actual TAO.

Two more sites sum the same alpha across subnets. They are **less harmful** (they never add the
result to real TAO) but they carry the same misleading `_tao` name:
`src/account-nominator-positions.ts:84,148` `total_stake_tao` and `src/chain-yield.ts:146,225`
`total_stake_tao`.

**This issue requires a design decision before code.** Two decisions must be made explicitly and
recorded in the PR description; do not pick silently.

- **D1 — conversion source and staleness.** The per-netuid alpha price must come from somewhere,
  and the only in-Worker source is a daily-cadence table (up to ~24h stale). Choose it and say so.
- **D2 — the other two `total_stake_tao` fields.** Either convert them to TAO the same way, or
  rename/relabel them as alpha. Both are acceptable; leaving them ambiguous is not.

**The per-netuid alpha-price join already exists in this file — copy its shape:
`workers/data-api.ts:4287-4290`.**

```sql
SELECT DISTINCT ON (netuid) netuid, alpha_price_tao
  FROM subnet_snapshots
  ORDER BY netuid, snapshot_date DESC
```

That is `handleDeregRiskSnapshot`'s latest-row-per-netuid read of `subnet_snapshots`, and the
comment directly above it (`workers/data-api.ts:4247-4250`) already states the staleness
characteristic you must carry over: "a DAILY cadence table, so `alpha_price_tao` here can lag up to
~24h behind the live economics tier".

**Use `subnet_snapshots.alpha_price_tao` via that exact `DISTINCT ON (netuid) … ORDER BY netuid,
snapshot_date DESC` idiom, joined per-netuid into the `delegated_tao` aggregate.** That resolves D1
and it is the answer this issue expects.

- A new table, a new rollup job, a new cron, or a new artifact to hold alpha prices. Extend the
  existing `subnet_snapshots` read; do not build a parallel price source.
- A single global/average alpha price applied to all subnets. The conversion must be **per-netuid**
  — a flat multiplier preserves the wrong ordering, which is the actual defect.
- Fetching prices from the live economics KV tier or from an outbound HTTP call inside this handler.
- Dropping/hiding `delegated_tao` instead of fixing it.

## Requirements

1. **`delegated_tao` becomes real TAO.** The aggregate at `workers/data-api.ts:6832-6835` joins the
   latest `subnet_snapshots.alpha_price_tao` per `netuid` and sums
   `np.share_fraction * n.stake_tao * s.alpha_price_tao` **per (coldkey, netuid)** before summing to
   the coldkey. Root (`netuid = 0`) rows do not exist in `nominator_positions`, so no special-casing
   is needed — but if a `netuid = 0` row ever appears it must be treated as price 1.0, not dropped.
2. **A missing or non-finite `alpha_price_tao` for a netuid must not silently contribute 0.** That
   netuid's contribution is excluded from `delegated_tao` and the account's `last_updated` /
   response is not otherwise altered — a silent 0 would reintroduce a quieter, harder-to-spot
   version of this same bug. Choose exclusion, and state in a code comment that it is deliberate.
3. **`total_tao = free_tao + delegated_tao` stays as-is** (`src/top-holders.ts:74`) and is now
   dimensionally valid because both addends are TAO. `total_tao` remains the default sort.
4. **Sort correctness.** `sort=delegated_tao` and `sort=total_tao` order by the converted values.
   `sort=free_tao` is unchanged.
5. **Sanity bound as an assertion, not a clamp.** After the fix, no single account's `total_tao` may
   exceed 21,000,000. Do **not** add a runtime clamp that hides a future regression — assert it in
   the test (Deliverable 4) so a regression fails CI instead of being rounded away.
6. **D2 resolved explicitly.** For `src/account-nominator-positions.ts:84,148` and
   `src/chain-yield.ts:146,225`, either (a) apply the same per-netuid conversion so
   `total_stake_tao` is TAO, or (b) rename the field to `total_stake_alpha` and update its schema,
   description, and every consumer in the same PR. Pick one, do it for **both** files consistently,
   and justify the choice in the PR body. Leaving either file untouched does not satisfy this issue.
   If (b) is chosen, the schema rename is a contract change — see Requirement 7.
7. **Schema and contract text.** `schemas-src/routes/top-holders.ts` and
   `schemas-src/mcp-tools/accounts-leaderboards.ts` field descriptions must state that
   `delegated_tao` is TAO-converted from per-subnet alpha at the latest daily
   `subnet_snapshots.alpha_price_tao`, and that it can lag up to ~24h. The route description at
   `src/contracts.ts:2835` must say the same — the current "this account's own total stake positions
   across every hotkey/subnet" wording is what made the bug invisible. Any `schemas-src/` edit
   requires `npm run build` and committing the regenerated `openapi.json` + generated types in the
   **same PR**, or `validate:contract-drift` fails CI.

## Deliverables

- [ ] `workers/data-api.ts` — `delegated_tao` aggregate joins latest-per-netuid
      `subnet_snapshots.alpha_price_tao` using the `DISTINCT ON (netuid)` idiom from
      `workers/data-api.ts:4287-4290`, converting alpha to TAO per (coldkey, netuid).
- [ ] `src/account-nominator-positions.ts` and `src/chain-yield.ts` — D2 applied identically to both
      (convert, or rename to `total_stake_alpha` with all consumers updated).
- [ ] `schemas-src/routes/top-holders.ts`, `schemas-src/mcp-tools/accounts-leaderboards.ts`, and
      `src/contracts.ts` — descriptions state the conversion, its source table, and its ~24h
      staleness. Regenerated `openapi.json` + generated types committed in the same PR.
- [ ] `tests/top-holders.test.ts` — a regression test that **fails before the fix and passes after**:
      feed rows for one coldkey holding stake on two netuids with materially different
      `alpha_price_tao` (e.g. 0.002 and 0.3), assert `delegated_tao` equals the price-weighted sum
      (not the raw sum), assert the two-account ordering **flips** versus the unconverted sum, assert
      a netuid with a null price is excluded rather than counted as 0, and assert no row's
      `total_tao` exceeds 21,000,000.
- [ ] Patch coverage: every changed line and branch in `src/**` / `workers/**` is covered —
      `codecov/patch` enforces `target: 99%, threshold: 0%`. Measure unsharded with
      `npm run test:coverage`.

satisfy this issue.

## Expected Outcome

`GET /api/v1/accounts/top-holders` returns `delegated_tao` and `total_tao` denominated in real TAO,
computed as `SUM(share_fraction * stake_tao * alpha_price_tao)` over the latest per-netuid
`subnet_snapshots` row, so no account reports more than 21,000,000 TAO and the default `total_tao`
ordering reflects TAO value rather than raw cross-subnet alpha count. `total_stake_tao` in
`src/account-nominator-positions.ts` and `src/chain-yield.ts` is either TAO-converted the same way
or renamed to `total_stake_alpha` with its schema updated. `schemas-src` + `src/contracts.ts`
descriptions name the conversion source and its ~24h lag, and the regenerated `openapi.json` is
committed alongside.

## Links & Resources

- `workers/data-api.ts:6820-6849` — the top-holders query (the defect).
- `workers/data-api.ts:4247-4290` — **the alpha-price join to mirror**, plus its staleness comment.
- `src/top-holders.ts:40-82` — `buildTopHoldersEntry` / `total_tao` composition.
- `src/metagraph-neurons.ts:941-944` — non-root stake is alpha (#2550).
- `scripts/lib/economics-artifacts.ts:56-68` — `computeAlphaMarketCapTao`, the alpha identity.
- `apps/indexer-rs/src/bin/poller/jobs/validator_nominators.rs:121-152` — `netuid != 0` ingestion.
- `src/contracts.ts:2835` — the taostats-benchmark claim that must be made true.
- `schemas-src/routes/top-holders.ts`, `schemas-src/mcp-tools/accounts-leaderboards.ts`.
- `tests/top-holders.test.ts` — the regression test's home.

JSONbored:
## Summary

- Closes #8803 — `/api/v1/accounts/top-holders` (and MCP `get_top_holders`) reported per-subnet **alpha** as TAO and added it to real TAO. The headline column was not a quantity that exists: the top row read 14,984,421 "TAO" against a 21,000,000 hard cap, and the top five rows summed to 1.57x max supply.

## D1 — conversion source and staleness

**`subnet_snapshots.alpha_price_tao`, latest row per netuid**, via the exact `SELECT DISTINCT ON (netuid) netuid, alpha_price_tao … ORDER BY netuid, snapshot_date DESC` idiom `handleDeregRiskSnapshot` already uses. No new table, no new rollup, no new cron, no outbound call — the existing read, extended. It is a **daily-cadence** table, so the price can lag up to ~24h behind the live economics tier; that characteristic is now stated in the route contract, in both schemas, and in the SQL's own comment.

### The aggregate (`workers/data-api.ts`)

```sql
SUM(np.share_fraction * n.stake_tao * CASE WHEN np.netuid = 0 THEN 1 ELSE s.alpha_price_tao END)
…
LEFT JOIN (SELECT DISTINCT ON (netuid) netuid, alpha_price_tao
             FROM subnet_snapshots ORDER BY netuid, snapshot_date DESC) s ON s.netuid = np.netuid
WHERE np.netuid = 0
   OR (s.alpha_price_tao >= 0 AND s.alpha_price_tao < 'Infinity'::numeric)
```

- **Per-netuid, not flat.** A single global/average price would have preserved the wrong ordering, which is the actual defect.
- **Missing or non-finite price ⇒ excluded, never a silent 0.** The `WHERE` drops that netuid's rows from the sum. `>= 0` rejects NULL and negatives; `< 'Infinity'::numeric` rejects both `Infinity` **and** `NaN` — in Postgres numeric ordering `NaN` sorts above every value, so `NaN >= 0` is true and only the upper bound catches it. A `COALESCE(…, 0)` here would have reintroduced a quieter, harder-to-spot version of this same bug, and the code comment says so.
- **`netuid = 0` converts at 1.0.** It cannot appear today (`validator_nominators.rs`: "root is never share-tracked"), but root stake is TAO-denominated, so if it ever is tracked it must convert at par rather than be dropped for having no `subnet_snapshots` row.
- **One `SUM`, not per-(coldkey, netuid) subtotals summed afterwards.** Multiplication distributes over the sum, so the two are algebraically identical and this is one aggregation pass instead of two. Same result, one less `GROUP BY`.
- `total_tao = free_tao + delegated_tao` is **unchanged** in `src/top-holders.ts` and is now dimensionally valid, and remains the default sort. `sort=delegated_tao` / `sort=total_tao` order by the converted values for free (they sort the response rows); `sort=free_tao` is untouched.
- **No runtime clamp.** The 21M bound is asserted in the tests so a regression fails CI, rather than being rounded away in production.

### D2 — chose (b), rename, for both files

`src/account-nominator-positions.ts` → `total_stake_alpha`; `src/chain-yield.ts` → `total_stake_alpha` **and** `total_emission_alpha`.

Rationale, since the issue asks for it explicitly:

1. **Neither field is ever added to real TAO.** That addition is what made top-holders produce impossible numbers. What is left in these two is purely a misleading name, and a rename fixes exactly the defect that exists without silently changing a published number's value under an existing consumer.
2. **Converting `chain/yield` would break it.** `network_yield` / `validator_yield` / `miner_yield` are emission÷stake, and the emission is alpha too. Converting the denominator alone makes the ratios meaningless; converting both would silently change three published yield metrics on a route this issue is not about. As alpha÷alpha the ratios stay dimensionally coherent and their values unchanged — which is why `total_emission_tao` is renamed alongside its denominator rather than left behind to imply a mismatched ratio.
3. **`account-positions` sums alpha the same way top-holders did**, but the number stands alone rather than being blended into a TAO figure, so naming it correctly is sufficient. `positions[].stake_tao` keeps the on-chain column name every neurons-tier route uses; renaming *that* repo-wide is its own change, and I flag below why.

**Honest note on scope.** While implementing this I found `total_stake_tao` is used across ~25 schema files for the same alpha quantity (validators, subnet history, economics, portfolio, …). The two renamed here are the ones #8803 names; the rest are a repo-wide naming pass that deserves its own issue rather than being smuggled into a bugfix. I'm filing that separately rather than leaving it implicit.

### Contract + descriptions (Requirement 7)

- `schemas-src/routes/top-holders.ts` and `schemas-src/mcp-tools/accounts-leaderboards.ts` — `free_tao` / `delegated_tao` / `total_tao` gain descriptions naming the conversion, its source table, and its ~24h staleness.
- `src/contracts.ts` — the route description's "this account's own total stake positions across every hotkey/subnet" (the wording that made the bug invisible while the same paragraph claimed parity with a taostats-style Account/Free/Delegated/Total benchmark) now states the conversion, the daily source, the ~24h lag, the exclusion rule, and why `total_tao` is a valid sum.
- `schemas-src/routes/{account-positions,chain-yield}.ts` and `src/graphql-sdl.ts` — the renamed fields carry descriptions saying they are alpha, not TAO, and why.
- Regenerated and committed in this PR: `public/metagraph/openapi.json`, `public/metagraph/api-index.json`, `public/metagraph/types.d.ts`, `packages/contract/index.d.ts`, `generated/graphql/types.ts`. `validate:contract-drift` passes. `r2-manifest.json` / `schemas/index.json` are not in the diff (the build auto-reverted them, as designed).

### Consumers of the rename

`src/graphql.ts` resolvers, `src/graphql-sdl.ts`, and `apps/ui/src/lib/metagraphed/{types,queries}.ts` + their two test files. The `apps/ui` part is **`lib/` only and renders nothing new** — no component or route reads either field (`ChainYield` and `AccountPositions` are not surfaced in any panel today), so there is no visual change and no screenshot table applies.

## Tests

**Fails before the fix, passes after** — `tests/data-api.test.ts`, where this repo already asserts emitted SQL for this exact route (`queryText()`): the query must no longer contain `SUM(np.share_fraction * n.stake_tao) AS delegated_tao`, must multiply by `alpha_price_tao` per netuid with the `netuid = 0 THEN 1` arm, must use the `DISTINCT ON (netuid) … ORDER BY netuid, snapshot_date DESC` read of `subnet_snapshots`, must carry the exclusion predicate, and must **not** contain `COALESCE(s.alpha_price_tao`. Verified red on the unfixed handler (`git stash` on `workers/data-api.ts` alone), green with it.

The conversion itself is an aggregate in Postgres — correctly, since folding it in the Worker would mean shipping every (coldkey, netuid) position row out of the database on each request — so the SQL is what a test can pin. The **consequences** are pinned separately in `tests/top-holders.test.ts` with a two-netuid fixture at the real price spread (0.002 and 0.3 TAO):

- the price-weighted sum is not the raw sum (25,000,000 alpha at 0.002 → 50,000 TAO; 200,000 alpha at 0.3 → 60,000 TAO);
- the default `total_tao` **ordering flips** — the cheap-alpha whale outranks the richer holder before, and loses after, which is why a flat multiplier would have fixed nothing;
- a netuid with no usable price is **excluded**, leaving the account's priced positions intact rather than zeroing it;
- **no account's `total_tao` exceeds 21,000,000**, with a sanity assertion that the same fixture through the old raw-alpha path really does break the cap — so the bound is a live check, not a vacuous one. (25M alpha on one subnet is an unremarkable balance and an impossible TAO one: a subnet's alpha is a different token, not bounded by TAO's cap. That is precisely how the old query exceeded max supply.)

Also updated for the rename: `tests/{chain-yield,account-nominator-positions,graphql,mcp-server,data-api}.test.ts` and the two `apps/ui` query tests.

## Registry Safety

- [x] Links a tracked, currently-open issue (`Closes #8803`) — required.
- [x] Generated artifacts were produced by repo scripts (`npm run build`), not hand-edited.
- [x] R2-only/high-churn detail artifacts are not committed (`r2-manifest.json` / `schemas/index.json` auto-reverted by the build).
- [x] Public API/OpenAPI/schema changes are intentional and documented — `delegated_tao`/`total_tao` change **value** (they become real TAO); `total_stake_tao` changes **name** on two routes. Both are described above and in the regenerated contract.

## Validation

- [x] `npm run build` (contract artifacts regenerated + committed)
- [x] `npm run validate`
- [x] `npm run validate:contract-drift`
- [x] `npm run lint` · `npm run format:check` · `npm run typecheck`
- [x] `npm run lint --workspace=apps/ui` · `format:check` · `typecheck --workspace=apps/ui`
- [x] `npx vitest run` over `top-holders`, `data-api`, `chain-yield`, `account-nominator-positions`, `graphql`, `zod-schemas`, `contracts`, `docs-content-drift`, `db-row-types`, `mcp-server`, `api-coverage` — all green
- [x] `npm test --workspace=apps/ui` for the two changed query tests
- [x] `git diff --check`


JSONbored:
The repo-wide `*_tao`-holds-alpha naming pass this PR deliberately scoped out is filed as #8945.

Issue #8928, opened 2026-07-31
perf(ci): the e2e overflow check runs at 2 Playwright workers on a 4-core runner


JSONbored:
## Where the time is

The `ui` job is ~502s, and the responsive-overflow e2e step is **241s — 48% of it**, now the largest single block in CI after #8909/#8925 (`test` is ~639s and falling).

| ui step | time |
| --- | --- |
| **Responsive overflow check (e2e)** | **241s** |
| Build | 70s |
| Test | 46s |
| Lint / Typecheck / Install | ~85s |
| 7 drift checks combined | 17s |

The check is 26 routes x 4 viewports = **104 tests**, ~2.3s each.

### 1. It runs against `npm run dev` — while the same job already builds production

`apps/ui/playwright.config.ts` starts `webServer: { command: "npm run dev" }`. Vite's dev server compiles each route on demand, so a 26-route sweep pays first-hit compilation 26 times. Meanwhile the **same job** already runs a full production `Build` (70s) as a separate step, whose output is then thrown away for the e2e run.

Serving the e2e from a preview of that existing build should remove the per-route compile entirely.

**This changes what is tested** (dev bundle -> production bundle), which is why it needs a decision rather than a quiet swap:
- *For*: the production bundle is what actually ships, and this would catch prod-only breakage the dev server hides.
- *Against*: `.claude/skills/metagraphed/SKILL.md` Phase C2 deliberately points the check at the dev server so it matches the contributor screenshot workflow. That intent would need revisiting or restating.

### 2. `workers` is unset

`fullyParallel: true` is set but `workers` is not, so Playwright falls back to `os.cpus().length / 2` — **2 workers on a 4-core GitHub runner**. Raising it explicitly is a one-line lever, but each worker drives a browser and rendering is CPU-bound, so the gain is likely sublinear and could introduce timeout flake. Must be measured on CI, not locally: a dev laptop's core count makes the result non-transferable.

### 3. `retries: 1` on CI

Any flaky test silently doubles its own cost.

## Why this is filed rather than shipped

Neither lever can be validated cheaply off-CI. A local measurement would run on a different core count against a different server, so it would not transfer — and this job gates a bot that auto-closes contributor PRs, so an unvalidated parallelism bump is the wrong kind of guess. Same standard applied to #8922.

## Suggested order

1. Decide dev-vs-production for the e2e target (the semantic question, and the bigger win).
[...]

PR #8947, opened 2026-07-31
perf(ci): pin the e2e worker count instead of inheriting the runner's core count (#8928)

JSONbored:
## Summary

- Refs #8928 — **finding #2 only**, the bounded one. Playwright's `workers` was unset, so it fell back to `os.cpus().length / 2` = **2 workers on the 4-core `ubuntu-latest` runner**. Pinned to 4.

This deliberately does **not** touch finding #1 (dev-vs-production e2e target) or #3 (`retries: 1`) — #8928's own suggested order treats the parallelism lever as a separate change, and #1 turns out to be much larger than the issue assumes (see below).

## Baseline, measured on CI

Five recent completed `ui` jobs, "Responsive overflow check (e2e)" step wall time:

| run | duration |
| --- | --- |
| 1 | 232s |
| 2 | 245s |
| 3 | 247s |
| 4 | 260s |
| 5 | 263s |

**Mean ~249s, spread 232–263s (~13%).** That spread is the thing to keep in mind reviewing this: a single run landing at, say, 235s would be *inside existing noise* and would prove nothing. A credible win has to clear ~230s and repeat.

I will post this PR's own measurements as a comment once its runs land, and **close it unmeasured-or-worse** rather than merge a parallelism bump on a hopeful single sample. The issue is explicit that this job gates a bot which auto-closes contributor PRs, so an unvalidated guess here is the wrong kind of cheap.

## What to watch besides wall time

Each worker drives its own Chromium and rendering is CPU-bound, so the gain is expected to be **sublinear** — 4 workers on 4 cores is not 2x. The specific risk is that oversubscription pushes individual tests toward their timeout, and `retries: 1` on CI would absorb that as *silently doubled cost* rather than a visible failure. So the review criteria are:

1. Step wall time clears the ~230s noise floor, repeatably.
2. No new retries. A run that gets faster while retrying is not faster.

## On finding #1 (dev vs production) — one correction and one blocker

I looked at this first since the issue calls it the bigger win. Two things the issue's framing gets wrong in opposite directions:

**The "against" argument is weaker than stated.** #8928 says Phase C2 "deliberately points the check at the dev server so it matches the contributor screenshot workflow", and `playwright.config.ts`'s comment says the same. But the screenshot workflow does not use this config at all — `tests/e2e/capture-pr-screenshots.ts` spawns its **own** `npm run dev` processes per worktree. The two are already fully decoupled machinery; the parity claim is about rendered output, not shared setup. So switching this check's target would not disturb the screenshot workflow.

**But it is not a `vite preview` swap either.** `apps/ui` is a TanStack Start SSR app built through Nitro with the **cloudflare-module** preset — production output is a Worker bundle (`dist/server/wrangler.json`), not a static directory `vite preview` can serve. Serving it means standing up wrangler, and the `ui` job currently runs `Build` **after** the e2e step, so the steps would also need reordering to reuse that output.

That makes finding #1 a real change to the job that gates contributor auto-close, not a one-liner — worth doing, worth its own PR, and worth doing after this smaller lever is measured so the two effects don't confound each other. I'd suggest updating #8928's finding #1 to record both points so the next attempt doesn't start from the `vite preview` assumption.

## Registry Safety

- [x] References a tracked, currently-open issue (`Refs #8928`) — partial by design, so not `Closes`.
- [x] Generated artifacts were produced by repo scripts, not hand-edited. (None.)
- [x] Public API/OpenAPI/schema changes are intentional and documented. (None — CI-only.)

## Validation

- [x] `npx prettier --check` + `npx eslint` on the changed file
- [x] Diff is one file, 10 lines, all but one of them the comment explaining why the value is pinned
- [ ] **CI measurement — the actual acceptance criterion, posted as a comment below when the runs land**

### DIFF
--- a/apps/ui/playwright.config.ts
+++ b/apps/ui/playwright.config.ts
@@ -8,6 +8,16 @@ const PORT = 8080;
 export default defineConfig({
   testDir: "./tests/e2e",
   fullyParallel: true,
+  // #8928: pinned, because Playwright's default is os.cpus().length / 2 --
+  // 2 workers on the 4-core ubuntu-latest runner, leaving half the machine
+  // idle for a check that is 26 routes x 4 viewports = 104 browser tests and
+  // the single largest block in the `ui` job. Rendering is CPU-bound so the
+  // gain is sublinear, not 2x. Pinned rather than left implicit for a second
+  // reason: the default tracks the MACHINE, so the same suite silently runs
+  // at a different width on a dev laptop than in CI, which makes a local
+  // timing measurement non-transferable -- exactly why this issue was filed
+  // rather than shipped on a local benchmark. Measured on CI; see the PR.
+  workers: 4,
   forbidOnly: !!process.env.CI,
   retries: process.env.CI ? 1 : 0,
   reporter: process.env.CI ? "github" : "list",


JSONbored:
## Measurement: `workers: 4` is faster and **breaks the check**. Not mergeable.

Two runs, both on this branch, plus the baseline I collected before opening it.

| config | e2e step | result |
| --- | --- | --- |
| `workers` unset (= 2 on a 4-core runner) | 232s / 245s / 247s / 260s / 263s — **mean ~249s** | pass, 10 of the 10 most recent completed runs |
| `workers: 4` — attempt 1 | **186s** | **fail** — 3 tests failed, 1 still failed after retry |
| `workers: 4` — attempt 2 | **203s** | **fail** — same 3 tests |

**~22% faster (195s mean vs 249s), and it does not pass.** The speedup is real and outside the baseline's 232–263s noise band. It is also irrelevant, because the same three tests failed on both runs:

- `tests/e2e/multisig-related-error.spec.ts:64` — a failed lookup shows an error state
- `tests/e2e/multisig-related-error.spec.ts:85` — a successful empty lookup still shows the empty-result copy
- `tests/e2e/evidence-deep-link.spec.ts:62` — `#evidence` resolves to the About tab

with `expect(locator).toBeVisible() failed / element(s) not found` and `expect(page).toHaveURL(/[?&]tab=about/)` never settling off `…/subnets/1#evidence`.

**Same three tests, both runs — deterministic, not flake.** Ruled out as pre-existing: the e2e step succeeded on all ten most recent completed runs at the default 2 workers. The failure arrived with this change.

`retries: 1` is doing exactly what finding #3 of #8928 warns about here: 3 tests failed, 2 passed on retry, so the job reports "1 failed" and quietly paid for five extra test executions. Without that retry this would read as a 3-test failure.

## Closing, per the criterion in the PR body

This job gates a bot that auto-closes contributor PRs. A gate that fails deterministically is worse than a gate that takes an extra minute, so a 22% saving does not buy it. Closing rather than merging or leaving it to rot — the branch stays if either follow-up below wants it.

## What the failures actually say

**None of the three are in the overflow sweep.** They are two unrelated spec files that merely share `testDir` and now run concurrently with the 104-test sweep instead of being partly serialized behind it. So this is not "the overflow check can't parallelize" — it is "two timing-sensitive navigation specs lose their races when the machine is saturated."

That points at a more surgical fix than backing the global number off: keep the sweep parallel and constrain the two specs (`test.describe.configure({ mode: "serial" })`, or split them into their own Playwright project with its own worker count). Recorded on #8928 along with the numbers above, so the next attempt starts from data rather than from the default.

Issue #9009, opened 2026-08-02
auth(mcp): migrate auth_required surface credentials out of tool arguments


JSONbored:
Split out of #8967 / ADR 0027, which decided the access model but deliberately left this alone.

## Problem

`auth_required` subnet surfaces take a caller-supplied `credential` **inside tool arguments** (`src/mcp-server.ts:11161`, `:11240-11363`).

That means a secret belonging to a third-party subnet API travels as a tool parameter. Consequences:

- It lands in whatever the client logs, and in the MCP conversation transcript.
- It is subject to `$mcp_parameters` capture — redacted by `redactMcpSensitiveFields`, but the safe design is for the value never to be in the argument object at all.
- It is the pattern OAuth-aware MCP clients are actively moving away from.

## Why it was not done in #8967

Two reasons, both still true:

1. **It changes published tool schemas** for real callers. Removing or deprecating a documented argument is a breaking change for anyone already passing it, and needs a deprecation path rather than a swap.
2. **It needs somewhere for the credential to live.** A session-bound credential store keyed on the authenticated identity does not exist yet. ADR 0027 established that `/mcp` has a real authenticated identity to bind to (`mg_` key account, or the OAuth GitHub identity on `ctx.props`), which is the prerequisite — but the store itself is unbuilt.

## Shape of the fix

- A session-bound credential lookup: an authenticated caller registers a surface credential once, and subsequent `call_subnet_surface` invocations resolve it from the session rather than the arguments.
- The in-band `credential` argument **survives for anonymous callers**, per ADR 0027's Model B — otherwise anonymous access to `auth_required` surfaces would be removed as a side effect of a security cleanup, which is a product decision this issue should not make on its own.
- A deprecation window on the argument for authenticated callers before it is rejected outright.

## Prerequisite worth confirming first

`$mcp_auth_tier` (landed with #8967) will show what share of `call_subnet_surface` traffic on `auth_required` surfaces is authenticated at all. If it is ~zero, the session-bound path has no users yet and the sequencing question ("build the store, or first give people a reason to authenticate") answers itself.

Do not start this without checking that number.

PR #9076, opened 2026-08-02
feat(mcp): publish which tools require authentication in tools/list

### LINKED ISSUE #9072: auth(mcp): tools/list does not say which tools require authentication — clients discover it only by being refused
## Problem

Three MCP tools refuse anonymous callers — `store_surface_credential`, `list_surface_credentials`, `delete_surface_credential` (#9009's credential store, which needs an identity to bind a stored secret to). **Nothing published says so.**

The requirement lives only inside `requireCredentialStore` (`src/mcp-server.ts`). `tools/list` advertises all three exactly like the 207 public ones. So the only way for a client to learn the precondition is to call a tool and be refused — and an agent that has to fail to discover a precondition usually just stops, rather than going off to authenticate.

This is the same defect class ADR 0027 was written about: the server card advertised `authentication: "none"` while the endpoint was an OAuth 2.1 protected resource. A discovery document that disagrees with the endpoint is a defect, not cosmetics.

It also matters for ADR 0027 clause 3 ("authentication buys throughput, not reach"), which #9009 made partly false. There is now reach behind authentication, and nothing tells anyone.

## Fix

Publish the flag in each tool's `_meta` as `metagraph.sh/auth_required`.

`_meta` rather than `annotations` deliberately: `annotations` is the MCP spec's own **closed** vocabulary (`readOnlyHint`, `destructiveHint`, `idempotentHint`, `openWorldHint`). Inventing a key inside it would be publishing a claim the spec does not define. `_meta` is the sanctioned extension point.

## On hand-maintenance

The declaration is a set of three names, not something derived from the handlers. That is the honest shape — "does this code path eventually reach an auth check" is not a property to infer from a function body and then publish as a **contract**.

What makes it safe is that it cannot go stale. A test probes **every** tool anonymously and fails if the declared set and the enforced set differ in either direction:

- a tool that refuses anonymous callers but is not declared → a real precondition no client can discover;
- a tool that is declared but serves them → the card tells clients to authenticate for nothing.

Both directions must be covered, or the check is worth half of nothing.

## Definition of complete

- [ ] `tools/list` publishes the auth requirement for exactly the tools that enforce it.
- [ ] A test proves declaration == enforcement in **both** directions, by probing rather than by reading the list.
- [ ] That test is verified to actually fail when the declaration is wrong (mutation, both directions).
- [ ] The flag is not smuggled into the spec's fixed `annotations` vocabulary.

JSONbored:
Closes #9072

## What was wrong

Three tools refuse anonymous callers — `store_surface_credential`, `list_surface_credentials`, `delete_surface_credential` (#9009's credential store, which needs an identity to bind a stored secret to). **Nothing published that.** `tools/list` advertised them exactly like the 207 public ones, so a client could only learn the precondition by calling a tool and being refused — and an agent that has to fail to discover a precondition usually stops rather than going off to authenticate.

Same defect class ADR 0027 was written about: a discovery document that disagrees with the endpoint. It also bears on that ADR's clause 3 (*"authentication buys throughput, not reach"*), which #9009 made partly false — there is now reach behind authentication, and nothing said so.

## `_meta`, not `annotations`

Published as `_meta["metagraph.sh/auth_required"]`. `annotations` is the MCP spec's **closed** vocabulary (`readOnlyHint`, `destructiveHint`, `idempotentHint`, `openWorldHint`) — inventing a key inside it would publish a claim the spec doesn't define, and a spec-validating client would be right to reject it. `_meta` is the sanctioned extension point, and the namespaced key says whose extension it is.

Omitted entirely rather than `false` on public tools: an absent key is the natural "no requirement", and emitting it on 207 tools is noise. A test asserts that absence.

## The declaration is hand-written — and can't go stale

A set of three names, deliberately. "Does this code path eventually reach an auth check" is not a property worth inferring from a function body and then publishing as a **contract**.

What makes it safe is `tests/mcp-auth-required.test.ts`, which does **not** read the declaration and compare it to another list. It **probes every advertised tool anonymously** and compares what the server actually did against what it advertised, in both directions:

- **enforced but not declared** → a real precondition no client can discover;
- **declared but not enforced** → the card tells clients to authenticate for nothing.

### Verified by mutation, both directions

| mutation | result |
| --- | --- |
| remove `delete_surface_credential` from the set | ❌ *"these tools refuse anonymous callers but tools/list does not say so"* |
| add `get_subnet` to the set | ❌ *"these tools are advertised as requiring auth but serve anonymous callers"* |
| restored | ✅ 2 passed |

Two details that keep the probe honest:

- It runs against a **provisioned-but-anonymous** env. `requireCredentialStore` checks identity *before* store configuration, so an unprovisioned env would fail with `surface_credential_store_unavailable` and hide the very requirement under test.
- It asserts at least one tool enforced auth, so the sweep silently degrading to "nothing enforces anything" can't pass both assertions vacuously.
- `globalThis.fetch` is stubbed for the sweep, so a 200-plus-tool probe never leaves the process.

## Verification

`npx vitest run` — **13,947 passing, 0 failing**. `typecheck`, `lint`, and `validate:mcp` (210 tools, full lifecycle) all clean.

The pre-existing `listToolDefinitions` key-allowlist test caught the new `_meta` key on the first run — working exactly as designed; `_meta` is now allowlisted there, with a pointer to the test that pins *which* tools carry it.

### DIFF
--- a/src/mcp-server.ts
+++ b/src/mcp-server.ts
@@ -11897,6 +11897,37 @@ const TOOL_OUTPUT_SCHEMAS = {
   ),
 };
 
+/** The `_meta` key carrying the auth requirement on an advertised tool.
+ *
+ * `_meta` and NOT `annotations` (#9072): `annotations` is the MCP spec's own
+ * CLOSED vocabulary -- readOnlyHint / destructiveHint / idempotentHint /
+ * openWorldHint. Inventing a key inside it would publish a claim the spec does
+ * not define, and a client validating against the spec would be right to
+ * reject it. `_meta` is the sanctioned extension point, and the namespaced key
+ * says whose extension it is. */
+export const MCP_AUTH_REQUIRED_META_KEY = "metagraph.sh/auth_required";
+
+/**
+ * The tools that refuse an anonymous caller. Since #9009 the surface-credential
+ * store needs an identity to bind a stored secret to, so all three of its tools
+ * fail closed for anonymous callers -- and until #9072 nothing published said
+ * so, leaving a client to discover the precondition only by being refused.
+ *
+ * This is a DECLARED set, not one derived from the handlers. "Does this code
+ * path eventually reach an auth check" is not a property worth inferring from a
+ * function body and then publishing as a contract. What keeps it honest is that
+ * it cannot silently go stale: tests/mcp-auth-required.test.ts probes EVERY
+ * tool anonymously and fails if the declared set and the enforced set differ in
+ * either direction -- an undeclared tool that refuses anonymous callers is a
+ * precondition no client can discover, and a declared tool that serves them
+ * tells clients to authenticate for nothing.
+ */
+export const AUTH_REQUIRED_TOOL_NAMES: ReadonlySet<string> = new Set([
+  "store_surface_credential",
+  "list_surface_credentials",
+  "delete_surface_credential",
+]);
+
 export function listToolDefinitions() {
   return MCP_TOOLS.map((tool: Row) => {
     const outputSchema =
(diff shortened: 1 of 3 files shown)


JSONbored:
Closing as a duplicate. #9070 landed the same feature on main while this was in CI, and its implementation is equivalent or better — it probes every tool anonymously in both directions, guards against a vacuous pass, asserts `auth_required` stays out of the spec's closed `annotations` vocabulary, and provisions the credential store so an anonymous probe is distinguishable from an unprovisioned one (the same subtlety this PR called out).

Nothing here is additive, so there's no reason to spend a review cycle on it. #9072 is closed by #9070.

PR #9088, opened 2026-08-02
feat(api): `fields=` projection on the three neuron routes

### LINKED ISSUE #9082: feat(api): `fields=` projection on the neuron routes — an agent asking "is my miner registered" pays 24k tokens for 256 full rows
## Problem

`GET /api/v1/subnets/1/metagraph` returns **95,424 bytes** — 256 neurons × 17 fields. Measured against production, 2026-08-02.

For an MCP caller that is roughly **24,000 tokens**, spent every time, whatever the question was. "Is my miner registered on subnet 1" needs `uid` and `hotkey`. Those two fields are 18,448 bytes:

| `fields=` | bytes | ~tokens | reduction |
| --- | --- | --- | --- |
| _(all 17, today)_ | 95,424 | 23,856 | 1.0× |
| `uid,hotkey,coldkey,axon,active` | 41,015 | 10,253 | 2.3× |
| `uid,hotkey,stake_tao,validator_permit` | 28,686 | 7,171 | 3.3× |
| `uid,hotkey` | 18,448 | 4,612 | 5.2× |
| `uid,incentive,emission_tao` | 11,174 | 2,793 | 8.5× |

The catalogue already has this primitive. `?fields=` is implemented in `workers/list-query.ts:627` and works on every list route. The three neuron routes are not list routes, so they never got it:

- `GET /api/v1/subnets/{netuid}/metagraph` — `get_subnet_metagraph`
- `GET /api/v1/subnets/{netuid}/validators` — `list_subnet_validators`
- `GET /api/v1/subnets/{netuid}/neurons/{uid}` — `get_neuron`

All three project the same row shape (`NeuronSchema`, `schemas-src/routes/subnet-metagraph.ts:28`), so one change covers all three.

GraphQL does not rescue a caller here either: `subnet_metagraph` is declared `JSON` and documented "opaque JSON passed through verbatim" (`src/graphql-sdl.ts:311`), so a GraphQL selection set cannot narrow it.

## Scope — and what this deliberately does not claim

`fields=` is applied to the built rows, not pushed into the `SELECT`. The neuron builder derives fields from other columns — `immunity_expires_at_block` needs `registered_at_block`, `block_number`, `captured_at` and `is_immunity_period` together (`src/metagraph-neurons.ts:196`) — so narrowing the SQL would silently drop derivations rather than just columns.

So this is a **payload and caller-token** change, not a latency fix. `get_subnet_metagraph`'s 2,204 ms p95 is the backing tier read and stays where it is (measured live: 1.53 s cold, 0.26 s warm for the same route). The p95 tail belongs with the storage migration, not here.

## Non-hand-maintained: derive the field set, don't declare it

The allowed field set must come from the published schema. `NeuronSchema`'s own shape is the contract, so a field added to it is projectable the same day, and a projection map that could drift out of step with the contract never exists.

That is also strictly better than what `list-query.ts` does today. It derives the allowed set from the **union of keys present in the returned rows** (`workers/list-query.ts:652`), which is correct for heterogeneous collections but wrong for a schema-optional field: `immunity_expires_at_block` is emitted only for rows currently inside their immunity window, so on a subnet where nobody is, `?fields=immunity_expires_at_block` would be rejected as "unsupported" when it is a perfectly valid field of the contract.

## One primitive, not two

Extract the projection into a shared module and have `list-query.ts` use it, passing its existing row-union set so its behaviour is unchanged. Callers get one parameter with one syntax and one error idiom across every route that has it; the repo gets one implementation to reason about.

The pieces: parse and validate `fields`, resolve it against an allowed set (schema-derived here, row-union there), project rows, and echo the applied projection in `meta` the way list routes already do.

MCP takes it as a `fields` tool input — the same shape `list_subnet_validators` already uses for its MCP-only `limit` / `min_stake_tao` post-filters (`src/mcp-server.ts:6263`), applied after the tier returns.

## Definition of complete

- [ ] `fields=` works on all three neuron routes, on REST and MCP.
- [ ] The allowed set is derived from `NeuronSchema`, with no second list to maintain.
- [ ] A test proves derived == schema in both directions, and is verified to fail when they diverge (mutation, both directions).
- [ ] An unknown field is a 400 naming the offending field, matching list routes' existing message.
- [ ] Requesting a schema-optional field absent from every row in this response succeeds and returns it where present — the case `list-query.ts`'s row-union derivation gets wrong.
- [ ] `list-query.ts` uses the shared primitive, with its behaviour unchanged.
- [ ] Omitting `fields` returns exactly today's shape, byte for byte.
- [ ] `npm run build` regenerated + committed, and `apps/ui`'s api-reference docs regenerated.

JSONbored:
Closes #9082

`GET /api/v1/subnets/1/metagraph` returns ~95 KB — 256 neurons × 17 fields, roughly **24k tokens** for an MCP caller, spent whatever the question was. *"Is my miner registered"* needs `uid` and `hotkey`: **5.2× less**.

`fields=` already existed in `workers/list-query.ts` and worked on every list route. The three neuron routes aren't list routes, so they never got it. All three project the same `NeuronSchema` shape, so one change covers metagraph, validators, and neuron-detail — REST **and** MCP.

## One primitive, not two

The projection moves to `src/field-projection.ts`, and `list-query.ts` now calls it while passing its **existing row-union rule**. Its behaviour is unchanged — its 100 tests pass untouched — and callers get one syntax and one error idiom everywhere.

## Why the neuron routes resolve against the schema instead

Not merely a different source — a **strictly better one here**. `immunity_expires_at_block` is emitted only for a neuron currently inside its immunity window, so a row-union check rejects:

```
?fields=uid,immunity_expires_at_block   → 400 unsupported
```

on any subnet where nobody is — for a field the published contract plainly has. Verified against the live handler, it now returns **200**.

Deriving from `NeuronSchema.shape` also means a field added to the schema is projectable the same day, with **no second list to drift**. And `resolveFieldProjection` takes a *predicate* rather than a materialised Set, so `list-query.ts` keeps its lazy scan (~1 row on a valid request, not ~1160).

## Payload change, not a latency one

Applied to the **built rows**, not pushed into the `SELECT`: the neuron builder derives fields from other columns (`immunity_expires_at_block` needs `registered_at_block` + `block_number` + `captured_at` + `is_immunity_period` together), so narrowing the SQL would silently drop *derivations* rather than just columns. `get_subnet_metagraph`'s p95 belongs with the storage migration, exactly as the issue says.

## Two ordering details that would otherwise be silent bugs

- On `list_subnet_validators` the projection runs **after** `min_stake_tao` — otherwise `fields=uid` drops the very column the filter reads.
- Neuron-detail uses `projectRow`, so a cold store's `neuron: null` **stays null** instead of becoming `{}`.

`meta.fields` echoes the applied projection, as the list routes already do.

## Verification

Live behaviour on the metagraph route:

| request | result |
| --- | --- |
| _(none)_ | 200, full rows |
| `?fields=uid,hotkey` | 200, `meta.fields: uid,hotkey` |
| `?fields=uid,immunity_expires_at_block` | **200** — the row-union case |
| `?fields=uid,bogus_field` | 400 `unsupported field for neurons: bogus_field` |

**13,960 passing, 0 failing.** `typecheck`, `lint`, and `validate:api/mcp/contract-drift/openapi/types/schemas/schema-enums/openapi-examples/docs` all clean. 11 new tests cover the primitive, including the schema-optional case, absent-not-null omission, and the null-neuron shape.

### DIFF
--- a/schemas-src/mcp-tools/get-subnet-metagraph.ts
+++ b/schemas-src/mcp-tools/get-subnet-metagraph.ts
@@ -9,6 +9,13 @@ export const GetSubnetMetagraphInputSchema = z
   .object({
     netuid: z.int().min(0),
     validator_permit: z.boolean().optional(),
+    /** #9082: narrow each returned neuron row to these fields. The rows are
+     * 17+ fields wide and a full subnet is ~24k tokens; "is my miner
+     * registered" needs two of them. Applied after the tier returns, like
+     * this family's other MCP-only post-filters. Field names come from the
+     * published neuron schema; an unknown one is an invalid_params error
+     * rather than a silent no-op. */
+    fields: z.array(z.string()).min(1).optional(),
   })
   .strict();
 export type GetSubnetMetagraphInput = z.infer<
--- a/schemas-src/mcp-tools/neurons.ts
+++ b/schemas-src/mcp-tools/neurons.ts
@@ -10,6 +10,13 @@ export const GetNeuronInputSchema = z
   .object({
     netuid: z.int().min(0),
     uid: z.int().min(0),
+    /** #9082: narrow each returned neuron row to these fields. The rows are
+     * 17+ fields wide and a full subnet is ~24k tokens; "is my miner
+     * registered" needs two of them. Applied after the tier returns, like
+     * this family's other MCP-only post-filters. Field names come from the
+     * published neuron schema; an unknown one is an invalid_params error
+     * rather than a silent no-op. */
+    fields: z.array(z.string()).min(1).optional(),
   })
   .strict();
 export type GetNeuronInput = z.infer<typeof GetNeuronInputSchema>;
--- a/schemas-src/mcp-tools/validators.ts
+++ b/schemas-src/mcp-tools/validators.ts
@@ -18,6 +18,13 @@ export const ListSubnetValidatorsInputSchema = z
     netuid: z.int().min(0),
     limit: z.int().min(1).optional(),
     min_stake_tao: z.number().min(0).optional(),
+    /** #9082: narrow each returned neuron row to these fields. The rows are
+     * 17+ fields wide and a full subnet is ~24k tokens; "is my miner
+     * registered" needs two of them. Applied after the tier returns, like
+     * this family's other MCP-only post-filters. Field names come from the
+     * published neuron schema; an unknown one is an invalid_params error
+     * rather than a silent no-op. */
+    fields: z.array(z.string()).min(1).optional(),
   })
   .strict();
 export type ListSubnetValidatorsInput = z.infer<
--- a/schemas-src/routes/subnet-metagraph.ts
+++ b/schemas-src/routes/subnet-metagraph.ts
@@ -24,7 +24,7 @@
 import { z } from "zod";
 import { successEnvelopeSchema } from "../envelope.ts";
 
-const NeuronSchema = z
+export const NeuronSchema = z
   .object({
     uid: z.int().min(0),
     hotkey: z.string().nullable(),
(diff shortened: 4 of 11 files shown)


JSONbored:
Closing as a duplicate — #9087 landed the same feature and closed #9082 while this was in CI.

I checked theirs against the three subtleties I'd found rather than assuming, and it handles all of them: the schema-derived allowed set (so a schema-optional field like `immunity_expires_at_block` isn't wrongly rejected), projection applied **after** `min_stake_tao` on `list_subnet_validators` — with a comment saying exactly why — and all three routes wired on both REST and MCP.

Nothing here is additive. My apologies for the duplicated effort; that's on me for not checking whether main already implemented an open issue before starting.

Issue #8251, opened 2026-07-26
feat(ui): validators rebuild — full virtualized directory, concentration bar, deduped detail


JSONbored:
Phase 2 · Core pages. **Validators rebuild — full directory, calmer concentration story, deduped detail.**

## Context (measured)

Index shows only 20 of 1,016 validators; below the table sits a full-bleed flat-mint treemap (the loudest visual element on the site) + a stake-intensity heatmap matrix. Est. APY column includes uncapped 242%/180% outliers (Phase 0 caps them; this issue keeps the rule). Detail page: 8 stat cards, then three identical "2.26%" APY cards (7d/30d/90d), then a 117-row per-subnet table with a Stake button per row (11,476 px on mobile).

## Requirements

- **Index:** full 1,016 rows virtualized; ⭐ watchlist column (shared store); columns Operator · Take · Est. APY (capped) · Active subnets · Nominators · Dominance · Total stake · 30d Δ; "how to evaluate" explainer stays collapsed. Concentration: top-10 stacked horizontal bar (one accent moment); treemap retires; stake-intensity matrix behind a "Concentration detail" toggle or moves to the detail pages.
- **Detail:** identity header (name, hotkey/coldkey copyable, identity links) → KPI band of 6 (Total stake · Est. APY (one card, 7/30/90 toggle) · Take · Active subnets · Nominators · Avg val-trust) → tabs: Per-subnet performance (default top-20 by stake, search + expand-all) · Nominators · History. ONE Stake/Delegate CTA in the header (partner-validator + disclosure rules unchanged); per-row buttons die.
- Cross-links: subnet ↔ validator ↔ account everywhere entities appear.
- Mobile: KPI 2×3; per-subnet table becomes rows of Subnet · Stake · Dividends with expandable detail.

## Acceptance

- [ ] All 1,016 searchable/sortable without pagination stalls.
- [ ] Exactly one APY value block on detail; toggle switches window without duplicate cards.
- [ ] No full-bleed accent fills anywhere on either page.

PR #9130, opened 2026-08-02
fix(contract): declare /api/v1/validators limit max 2000, guard against drift (#9127)

### LINKED ISSUE #9127: fix(contract): /api/v1/validators declares max limit 100 but enforces 2000
## The bug

`/api/v1/validators` enforces a maximum `limit` of **2000**. Our published OpenAPI declares **100**.

```
GET /api/v1/validators?limit=2000  ->  200 OK, 1035 rows
```

Per the spec we publish, that request is invalid. It is also **the request our own site makes** — `apps/ui/src/server.ts:255` fetches `?limit=2000` on the SSR path, and `ALL_VALIDATORS_LIMIT = 2000` in `-validators-index-page.tsx:48`. A consumer generating a client from `openapi.json` gets one that rejects, at build time, the exact call the validators directory depends on.

## Why it drifted

#8251 deliberately raised the runtime ceiling 100 → 2000 so the directory could fetch the full validator set (~1,014 live) in one request for client-side virtualization. `GLOBAL_VALIDATOR_LIMIT_MAX` (`src/metagraph-neurons.ts:129`) carries that change with its measurements. The contract was never updated with it.

**The runtime is right; the contract is stale.** Nothing lowers to 100 — that would break the directory.

## Three stale copies of one number

| Where | Says | Should say |
| --- | ---: | ---: |
| `src/contracts.ts` — `/api/v1/validators` `limit.schema.maximum` | 100 | 2000 |
| `src/contracts.ts` — route description, *"limit caps the list (default 20, max 100)"* | 100 | 2000 |
| `src/mcp-server.ts` — `list_global_validators` description, *"(default 20, max 100)"* | 100 | 2000 |

The MCP tool itself already clamps to `GLOBAL_VALIDATOR_LIMIT_MAX`, so only its prose is wrong — but that prose is what a model reads to decide what to send.

## Scope: it is exactly this one route

Every runtime limit constant was checked against its declared maximum. All the others agree:

| Constant | Runtime | Contract | |
| --- | ---: | ---: | --- |
| `GLOBAL_VALIDATOR_LIMIT_MAX` | 2000 | 100 | ❌ |
| `MOVERS_LIMIT_MAX` | 100 | 100 | ✅ |
| `CHAIN_TURNOVER_LIMIT_MAX` | 100 | 100 | ✅ |
| `TOP_HOLDERS_LIMIT_MAX` | 100 | 100 | ✅ |
| `ACCOUNTS_LIST_LIMIT_MAX` | 100 | 100 | ✅ |
| `SUBNET_EVENT_SUMMARY_RECENT_LIMIT_MAX` | 50 | 50 | ✅ |
| `CHAIN_IDENTITY_HISTORY_LIMIT_MAX` | 200 | 200 | ✅ |

So this is a one-line correction plus a regen — and a guard, because the thing that failed here is not the number.

## The actual fix

The ceiling is **declared twice, in two files, with nothing tying them together** — the same shape as the presence-filter problem in #9118, where two fields describing one fact were free to disagree. A number that is correct today and hand-maintained in two places is a number that will drift again.

1. Correct all three copies; `npm run build` to regenerate `openapi.json` + types; regenerate the `apps/ui` api-reference docs.
2. **Add a derived guard**: for each route whose `limit` bound is a named runtime constant, import both the constant and the contract's declared `maximum` and assert they are equal. The pairing (which constant backs which route) is a declared judgement; both *values* are read from source, so neither side can move alone. Mutation-test it — changing either number must fail with the route named.

## Acceptance

- [ ] `/api/v1/validators` declares `maximum: 2000`, matching what it enforces
- [ ] Both descriptions say max 2000
- [ ] A test fails if `GLOBAL_VALIDATOR_LIMIT_MAX` and the declared maximum ever diverge again, for this and every other constant-backed limit
- [ ] `openapi.json`, generated types, and the `apps/ui` api-reference docs regenerated and committed

## Out of scope

`/api/v1/accounts/{ss58}/extrinsics` and `.../transfers` declare `maximum: 1000` but the serving tier caps pages at 200 — so `?limit=1000&offset=1000` silently skips rows 200–999. That cap is upstream in the Postgres tier, which is mid-migration to Cloudflare; filed separately rather than papered over here.

shin-core:
## Summary

`/api/v1/validators` enforces a `limit` ceiling of **2000** (`GLOBAL_VALIDATOR_LIMIT_MAX`, `src/metagraph-neurons.ts` — raised 100 → 2000 by #8251 so the validators directory can fetch the full live set in one request) while the published contract still declared **100**. A consumer generating a client from `openapi.json` got one that rejects, at build time, the exact `?limit=2000` call the directory depends on (`apps/ui/src/server.ts` SSR path). The runtime is right; the contract was stale.

## What changed

- **Corrected all three stale copies of the number**, per the issue's inventory:
  - `src/contracts.ts` — `/api/v1/validators` `limit.schema.maximum`: `100` → `2000`
  - `src/contracts.ts` — route description: *"default 20, max 100"* → *"max 2000"*
  - `src/mcp-server.ts` — `list_global_validators` description prose (the tool itself already clamps to the constant; only its prose was wrong)
- **Regenerated + committed** `openapi.json`, generated `types.d.ts`, `api-index.json`, `packages/contract/index.d.ts`, and the `apps/ui` api-reference doc (`content/docs/api-reference/validators/global-validators.mdx`).
- **Added a derived guard** (`tests/invariants-contracts.ts`): for every route whose `limit` maximum is a named runtime constant, it reads *both* the constant (from its module) and the declared `maximum` (from `API_ROUTES`) and asserts they are equal. The pairing (which constant backs which route) is a declared judgement; both values are read from source, so neither side can move alone. Covers all seven constant-backed limits (`GLOBAL_VALIDATOR`, `ACCOUNTS_LIST`, `TOP_HOLDERS`, `MOVERS`, `CHAIN_TURNOVER`, `CHAIN_IDENTITY_HISTORY`, `SUBNET_EVENT_SUMMARY_RECENT`).

## Acceptance

- [x] `/api/v1/validators` declares `maximum: 2000`, matching what it enforces
- [x] Both descriptions (route + MCP tool) say max 2000
- [x] A test fails if `GLOBAL_VALIDATOR_LIMIT_MAX` and the declared maximum ever diverge again — for this and every other constant-backed limit. Mutation-tested both directions: reverting the contract to 100 **and** bumping the constant to 2001 each fail with the route named.
- [x] `openapi.json`, generated types, and the `apps/ui` api-reference doc regenerated and committed

## Validation

- `npm run build` (artifacts regenerated; no deploy-owned drift committed)
- `npm run typecheck` · `npx eslint` · `npx prettier --check` — clean
- `npm run validate` · `npm run validate:contract-drift` · `npm run validate:openapi` · `npm run validate:mcp` — green
- `npx vitest run tests/invariants-contracts.test.ts tests/mcp-server.test.ts tests/contracts.test.ts tests/contracts-mcp.test.ts tests/contract-change-summary.test.ts` — all pass
- `npm run test:coverage` — patch coverage satisfied

Closes #9127


JSONbored (commented):
Maintainer-only / already completed, thank you though.

Issue #9146, opened 2026-08-02
Post-wipe parity: port the remaining Postgres-flagged route families to Cloudflare-native serving


JSONbored:
The box's Postgres is retired (#9145 flips the flags at wipe). Every family below has its **data already verified row-for-row in the lakehouse** — what is missing is Cloudflare-native serving. Until each port lands, its routes serve schema-stable empties (never errors, never wrong data).

In priority order:

## 1. neurons / metagraph family (`METAGRAPH_NEURONS_SOURCE`, ~92 call sites)
Subnet metagraph, validators, concentration/performance/yield/turnover/movers. This is the **only live-refreshed family**: a sync posts chain state through `/api/v1/internal/neurons-sync` into Postgres. Store is small (neurons 30,333 + neuron_daily 726,485 rows) — fits D1 comfortably. Port = repoint the write path at D1 (`createD1Sql` precedent from the user-state port) + port the family's reads in data-api. Restores **live** data, not just history.

## 2. Small-table detail routes → lakehouse cold-tier readers
Same explicit-reader pattern as `src/blocks-cold-tier.ts` / `extrinsics` / `events` (shared formatters, decline-don't-degrade, data-api cursor tokens):
- `subnet_hyperparams(+history)`, `subnet_identity_history`, `subnet_ownership(+history)` 
- `account_identity(+history)`, `account_balances`, `nominator_positions`, `validator_nominator_counts`
- `rpc_proxy_events` (rpc-usage analytics), `self_health_daily`
These are frozen-at-export snapshots; serving them beats serving nothing while their refresh lanes get rebuilt.

## 3. Heavy aggregates → precomputed projections
Stake flow, transfers, OHLC, alpha volume, chain performance/yield, top-holders. Request-time scans of the 894M-row event tables are the wrong shape for R2 SQL (~1–2s/query, no indexes); the right shape is scheduled projection artifacts (R2/D1) built by Worker crons over the lakehouse. R2 SQL dialect measured 2026-08-02: GROUP BY/JOIN/CTE/window functions/tuple comparison all supported; only OFFSET and UNION missing.

## Constraints that already bit
- R2 SQL has **no bound parameters** — every literal goes through the safe-literal guards in `src/r2-sql.ts`; filters decline rather than degrade.
- Under a burst of concurrent cold queries R2 SQL can transiently fail; the failure-generation guard keeps those responses out of the edge cache (observed + confirmed self-healing in the 2026-08-02 failover rehearsal).
[...]

Issue #8945, opened 2026-07-31
fix(api): *_tao field names across ~25 schemas hold cross-subnet alpha, not TAO


JSONbored:
## Context

`total_stake_tao` is used across roughly **25 schema files** for a quantity that is, in most of them,
not TAO.

A non-root neuron's `stake_tao` is that subnet's **alpha** token, not TAO (`src/metagraph-neurons.ts`,
#2550). Anything that sums `stake_tao` across subnets and calls the result `*_tao` is therefore
publishing a cross-subnet alpha count under a TAO name. The repo's own economics code already relies
on this: `scripts/lib/economics-artifacts.ts`'s `computeAlphaMarketCapTao(alphaPriceTao,
totalStakeTao) => alphaPriceTao * totalStakeTao` is only dimensionally valid *because*
`total_stake_tao` is alpha.

#8803 fixed the one place where this actually produced impossible numbers — `/accounts/top-holders`
added the alpha sum to genuine `System::Account` TAO, so the top account reported 71% of a supply
hard-capped at 21M — and renamed the two fields that issue named
(`/accounts/{ss58}/positions` and `/chain/yield`) to `total_stake_alpha` / `total_emission_alpha`.

**The rest were left alone on purpose.** Renaming two of ~25 identically-named fields is defensible
when those two are the ones an issue names; renaming a moving subset of the remainder ad hoc is how
you end up with the same concept spelled two ways depending on which route you ask. The rest need one
deliberate pass, or a deliberate decision to leave them.

## Why this is maintainer-only

Every option here is a breaking change to public route shapes, MCP tool outputs, and the GraphQL
schema at once, or an explicit decision to keep a known-misleading name. It also needs a judgement
this repo has not yet made in one place: which of these fields should be **renamed** (they are alpha
counts and always will be) versus **converted** (they are meant to be a value, and should be priced
like #8803's `delegated_tao`).

## Requirements

1. **Inventory first.** Every field spelled `*_tao` whose value is or may be alpha, with its route,
   its schema file, and what the number is used for. Start from
   `grep -rn "total_stake_tao\|total_emission_tao\|stake_tao" schemas-src/ src/ workers/`.
2. **Classify each one as rename, convert, or leave**, with the reason:
   - **rename** when the field is a count and nothing downstream treats it as a value (e.g. a yield
     denominator — `/chain/yield` is the precedent, where converting the denominator alone would
     have broken three published ratios);
[...]

PR #9164, opened 2026-08-02
feat(serving): read the neurons family from D1 when Postgres misses

JSONbored:
The read half of #9157. Without it, D1 fills up and the routes still serve empty.

#9157 landed the neurons **write** path onto D1 and `0007_neurons.sql` landed the tables — but nothing reads them. Every neurons route still asks the Postgres tier first and, on a miss, builds its payload from an empty array. Once the box is wiped that miss is permanent, so the metagraph, neuron-detail and per-subnet validator routes serve a schema-stable empty forever, along with the MCP and GraphQL surfaces over them.

## Rows, not payloads

`tryPostgresTier` returns an already-built payload from `DATA_API`, but the builders that produced it are pure and **already imported at every call site** — they're exactly what the current `buildSubnetMetagraph([], netuid)` fallback calls. So the reader returns rows and each call site runs the builder it already runs.

One formatter per payload instead of two kept in step by hand, which is also why a D1-served response is byte-indistinguishable from a Postgres-served one.

## A bug the tests caught before it shipped

Validators get their **own** reader rather than a filter over the subnet read. `buildSubnetValidators` does *not* filter — it formats whatever rows it's handed — and the Postgres tier applied `WHERE validator_permit` upstream in SQL. My first version reused the unfiltered read, which would have **served every miner as a validator**. The test asserts the filter is present in the query rather than accepting a pre-filtered stub, so a future refactor can't quietly reintroduce it.

## Deliberately not wired

`/api/v1/validators` and `/api/v1/accounts`. Those builders denominate their `*_tao` columns from a `priceByNetuid` map, and the empty fallback passes `NO_ALPHA_PRICES` — harmless *only* because it has no rows to price. Feeding them real D1 rows without the price join would publish alpha-denominated numbers under `_tao` field names: the exact defect #8803 fixed for top-holders and #8945 still tracks. Serving those empty is better than serving them wrong. Recorded as a comment in the module so the omission reads as a decision, not an oversight.

## Failure contract

Follows `src/analytics-live.ts` deliberately: a failed read degrades to zero rows **and bumps a generation counter**, so a handler can distinguish "D1 answered with nothing" from "D1 did not answer" and refuse to edge-cache the latter. A transient D1 blip must not pin zeros into the edge as fresh. Both the JSON and CSV return paths carry the marking — CSV is a separate return in two of these handlers and is easy to miss.

## Verification

23 tests. **Patch coverage is complete** — zero uncovered statements and zero uncovered branches across both changed files, measured by intersecting the diff's added lines with v8's uncovered set, not by whole-file percentage:

```
src/neurons-d1-read.ts:                uncovered_stmt=[] uncovered_branch=[]
workers/request-handlers/entities.ts:  uncovered_stmt=[] uncovered_branch=[]
```

Covers both result shapes D1 and the test fakes return, an unrecognised shape, no binding, read failure vs empty-but-successful (only the former bumps the generation), the `?validator_permit=true` filter, and degraded marking on all five return paths. `tsc`, `lint` and `prettier` pass; neighbouring suites (`metagraph-neurons`, `neurons-d1-schema`) still green at 125 tests.

## Sequencing

This is safe to merge now — it only changes behaviour when the Postgres tier *misses*, which is currently never. It becomes load-bearing the moment the box goes, and `METAGRAPH_NEURONS_SOURCE` can flip to `d1` once the container (#9162) is populating the table.

Refs #9146

### DIFF
--- a/src/neurons-d1-read.ts
+++ b/src/neurons-d1-read.ts
@@ -1,0 +1,153 @@
+// The neurons read tier, against D1 (#9146) — the read half of #9157's write
+// path.
+//
+// #9157 landed the writer and 0007_neurons.sql landed the tables, so the
+// neurons family has a home on Cloudflare. Nothing reads it yet: every neurons
+// route asks the Postgres tier first and, when that misses, builds its payload
+// from an empty array. Once the indexer box is wiped that miss becomes
+// permanent, so ~26 REST routes plus the MCP and GraphQL surfaces over them
+// would serve a schema-stable empty forever. This is what they fall back to
+// instead.
+//
+// SHAPE, AND WHY IT RETURNS ROWS RATHER THAN PAYLOADS. `tryPostgresTier`
+// returns an already-built payload from the DATA_API service binding, but the
+// builders that produced it (buildSubnetMetagraph / buildNeuronDetail /
+// buildSubnetValidators, src/metagraph-neurons.ts) are pure and already
+// imported by every call site — they are exactly what the current
+// `buildSubnetMetagraph([], netuid)` fallback calls. So this module returns
+// ROWS and lets each call site run the builder it already runs. One formatter
+// per payload rather than two kept in step by hand, which is also why a
+// D1-served response is indistinguishable from a Postgres-served one.
+//
+// The column list is `NEURON_INSERT_COLUMNS` — the same list the writer binds
+// and the migration declares — so a column added to one without the others
+// fails tests/neurons-d1-schema.test.ts rather than silently reading NULLs.
+//
+// Failure contract is `src/analytics-live.ts`'s, deliberately: a failed read
+// degrades to zero rows and bumps a generation counter, so a handler can tell
+// "D1 answered with nothing" from "D1 did not answer" and refuse to edge-cache
+// the latter as fresh. A transient D1 blip must not pin zeros into the edge.
+
+import { NEURON_INSERT_COLUMNS } from "./metagraph-neurons.ts";
+import { registerModuleStateReset } from "./module-state-registry.ts";
+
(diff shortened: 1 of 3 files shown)


JSONbored:
Closing — superseded by #9160/#9165, which landed while this was in review and solve it a better way.

I built this as a Worker-side D1 fallback in `entities.ts`: `tryPostgresTier` misses, and the handler reads D1 directly. #9160 instead gave every neurons read in the data Worker a D1 twin, and #9165 taught `tryPostgresTier` to *forward* a `"d1"` flag to `DATA_API` rather than return null for it:

```ts
const DATA_API_D1_FLAGS = new Set<string>(["METAGRAPH_NEURONS_SOURCE"]);
const forwards =
  value === "postgres" ||
  (value === "d1" && DATA_API_D1_FLAGS.has(flagName as string));
```

With `METAGRAPH_NEURONS_SOURCE: "d1"`, `tryPostgresTier` now returns a payload rather than null, so the fallback this PR adds is unreachable. Merging it would ship ~230 lines of dead code plus three call-site branches that can never be taken.

Their approach is also the better one: keeping the read next to the write in `data-api.ts` means one place owns the neurons SQL, instead of a second implementation in the request Worker that has to be kept in step by hand.

One thing worth carrying across, since it was a real bug I hit here and it applies to any neurons read: **`buildSubnetValidators` does not filter by `validator_permit`** — it formats whatever rows it's handed, and the Postgres tier applied `WHERE validator_permit` upstream in SQL. Any D1 twin of the per-subnet validators query needs that predicate explicitly, or it serves every miner as a validator. Worth a test that asserts the filter is in the query rather than accepting a pre-filtered fixture.

PR #9166, opened 2026-08-02
feat(infra): run the chain-state poller as a Cloudflare Container

JSONbored:
The last producer still tied to the indexer box. #9162 made a Postgres-free `POLLER_ONLY` set startable; this gives it somewhere to run.

## No producer rewrite

The `metagraph` job holds **no Postgres client** — it POSTs to the existing `neurons-sync` route, which #9157 moved onto D1 and #9160/#9165 now serve reads from. So this is the same binary that has written these rows for months, in a container instead of a systemd unit.

The alternatives both carried real risk. A GitHub Action (#9158) puts the producer back outside Cloudflare. A hand-written TypeScript metagraph reader — which I started and threw away — would have had to reproduce two columns that **aren't on chain at all** (`rank` has no storage item in dTAO; all 239 `SubtensorModule` items checked, and the Python producer's header claiming otherwise is stale — it derives rank by incentive position) plus get alpha-vs-TAO denomination right, with the only verification path disappearing when the box does.

## Metagraph only, deliberately

Three poller jobs are Postgres-free, but only one still has readers. As of #9145, `METAGRAPH_SUBNET_HYPERPARAMS_SOURCE` and `METAGRAPH_ACCOUNT_IDENTITY_SOURCE` are `"retired"` — those routes serve a schema-stable empty by decision. Running their jobs here would burn chain reads writing rows nothing will read, and imply a liveness the API no longer claims. `METAGRAPH_NEURONS_SOURCE` is `"d1"` and live, so that is the one lane still needing a producer.

`POLLER_ONLY` is pinned **in the image** rather than left to deployment: `main.rs` only skips its `DATABASE_URL` requirement when every enabled job is Postgres-free, so an override that adds a Postgres-backed job fails loudly at startup instead of running a job with nowhere to write.

## Verified against a real build, not just config

Built with OrbStack and run with image defaults — no `DATABASE_URL`, no `POLLER_ONLY` override:

```
poller: starting jobs (each connects its own chain client)
poller: POLLER_ONLY set, running only: metagraph
poller: no Postgres-backed job enabled, DATABASE_URL not required
connect_chain: OnlineClient ready
metagraph: 6250 delegate take(s), 129 netuid(s) of validator trust, fetching all metagraphs
```

It reaches the live chain and starts a real tick. `wrangler deploy --dry-run` registers the container cleanly, and caught two config errors first: `instance_type: "standard"` is deprecated in favour of `"standard-1"`, and the class needed exporting from `api.entry.ts`.

## The singleton constraint

`max_instances: 1`, and the DO is always addressed as `idFromName("global")` — with a test asserting the name, because this is correctness rather than cost. Two schedulers POSTing the same snapshots would race the neurons route's per-netuid prune, which keys on `captured_at`: an older instance's snapshot landing after a newer one's would **delete UIDs the newer one just wrote**. The job is idempotent against itself, not against a second scheduler.

## Keep-alive

Containers sleep when idle and the poller is a forever-loop, so the existing raw-capture cron pings `ensurePollerRunning()` to reset the idle timer — piggy-backing on a live trigger rather than adding one whose only job is to exist. `sleepAfter` is 1h against a 15-minute ping, so three consecutive missed ticks still cannot reap a healthy poller. The ping is reported, never thrown: one failed lane must not stop the others in the same handler.

## Two packaging traps this hit

**`@cloudflare/containers` breaks plain Node.** Its dist entry imports `./lib/container` with no file extension, which Node ESM cannot resolve. Re-exporting the class through `api.ts` therefore killed `validate:committed-seed` and every other Node script that imports the handler. The class is now exported from `api.entry.ts` directly — only wrangler bundles that file, so the dependency stays on the deployed path and off every script's.

**The same import breaks vitest.** `ensurePollerRunning` lives in its own module (`workers/poller-container-control.ts`) importing nothing from the Container runtime, so all five branches stay covered instead of hiding behind a v8 ignore.

## Verification

5 tests, `tsc`, `lint`, `prettier`, `validate:committed-seed`, `wrangler deploy --dry-run`, and the live container run above.

Refs #9146

### DIFF
--- a/deploy/poller.Dockerfile
+++ b/deploy/poller.Dockerfile
@@ -1,0 +1,66 @@
+# The chain-state poller, as a Cloudflare Container (#9146).
+#
+# SAME BINARY, NEW HOME. apps/indexer-rs/Dockerfile builds both targets and
+# ships `poller` already, but its CMD is backfill-rs's entrypoint -- on the box
+# the poller was selected by the Ansible role's `command:`, not by the image.
+# There is no role any more, so this image exists to make `poller` the command.
+#
+# WHY A CONTAINER RATHER THAN A REWRITE. The three jobs this runs
+# (metagraph, subnet-hyperparams, account-identity) hold no Postgres client:
+# each POSTs to an existing Worker sync route, and those routes now persist to
+# D1. So the producer needs no code change at all -- it is the same binary that
+# has written these rows for months. The alternative was hand-writing a
+# MetagraphInfo SCALE decoder in the Worker, which would have to reproduce two
+# columns that are not on chain (`rank` has no storage item in dTAO; `trust` is
+# a constant 0) and get alpha-vs-TAO denomination right, with no way to verify
+# any of it once the box it would be checked against is gone.
+#
+# Build context is apps/indexer-rs/ (see wrangler.jsonc's containers entry), so
+# the COPY paths below are relative to that crate, matching its own Dockerfile.
+FROM rust:1-bookworm AS build
+WORKDIR /app
+COPY Cargo.toml Cargo.lock ./
+COPY src ./src
+# --locked so the image cannot silently resolve a different dependency tree
+# than the box-built binary this replaces.
+RUN cargo build --release --locked --bin poller
+
+FROM debian:bookworm-slim
+# ca-certificates: subxt uses rustls and the sync routes are HTTPS, so without
+# these every chain connection and every POST fails at TLS.
+#
+# curl is NOT optional. metagraph.rs's post_sync shells out to it via
+# tokio::process::Command (piping the body through stdin rather than argv, so a
+# 30k-row payload never hits an arg-length limit) instead of pulling in an HTTP
+# client crate -- the same reason apps/indexer-rs/Dockerfile installs it.
+# Without it the job does all its chain work, resolves every subnet, and then
+# dies at the last step with "spawn curl: No such file or directory", which
+# looks like a network fault rather than a missing binary. Caught by running
+# this image against production, not by reading it.
+RUN apt-get update \
+ && apt-get install -y --no-install-recommends ca-certificates curl \
+ && rm -rf /var/lib/apt/lists/*
+WORKDIR /app
+COPY --from=build /app/target/release/poller /app/poller
+
+# METAGRAPH ONLY, and the narrowness is the point.
+#
+# Three poller jobs are Postgres-free (metagraph, subnet-hyperparams,
+# account-identity), but only ONE of them still has readers: as of #9145,
+# METAGRAPH_SUBNET_HYPERPARAMS_SOURCE and METAGRAPH_ACCOUNT_IDENTITY_SOURCE are
+# "retired", so those routes serve a schema-stable empty by decision. Running
+# their jobs here would burn chain reads writing rows nothing will ever read,
+# and would imply a liveness the API no longer claims.
+#
+# METAGRAPH_NEURONS_SOURCE is "d1" -- live, read through DATA_API's D1 twin
+# (#9160/#9165). That is the one lane still needing a producer.
+#
+# Set in the image rather than left to the deployment: main.rs only skips its
+# DATABASE_URL requirement when every enabled job is Postgres-free, so an
+# override that adds a Postgres-backed job fails loudly at startup instead of
+# running a job with nowhere to write.
+ENV POLLER_ONLY=metagraph
+# The public archive endpoint. The private fullnode the box used is gone.
+ENV EVENTS_RPC_URL=wss://archive.chain.opentensor.ai:443
+
+CMD ["/app/poller"]
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,6 +16,7 @@
         "packages/ui-kit"
       ],
       "dependencies": {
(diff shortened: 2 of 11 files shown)


JSONbored:
## Blocker found while verifying this end-to-end — and it is NOT in this PR

This container works. Running it against production, it connects, reads the chain, and produces the full snapshot:

```
metagraph: 6250 delegate take(s), 129 netuid(s) of validator trust, fetching all metagraphs
metagraph: resolved 129 subnet(s), 30333 rows
metagraph: tick failed (sync POST failed: curl: (22) The requested URL returned error: 502
```

**`POST /api/v1/internal/neurons-sync` returns 502 `{"error":"d1 write failed"}` for any payload above 2 rows.** So the neurons lane cannot advance regardless of where the producer runs — the Actions variant (#9158) would hit exactly the same wall.

### Reproduction

Authenticated POSTs of identical, valid rows differing only in count:

| rows | result |
| ---: | --- |
| 1 | 200 |
| 2 | 200 |
| 5 | 502 `d1 write failed` |
| 100 / 1200 / 2000 | 502 |

### What I ruled out

- **Schema drift.** Deployed `neurons`, `neuron_daily` and `account_position_daily` all exist, and `neuron_daily`'s 22 columns match `NEURON_DAILY_COLUMNS` exactly, in order.
- **The SQL shape.** A hand-written 5-row multi-row upsert with `ON CONFLICT DO UPDATE` succeeds against production D1 for `neurons` *and* for `account_position_daily` — including the duplicate-`account` case that real coldkeys produce, which I expected to be the culprit and is not.
- **Parameter budget.** 5 rows is ~285 bound parameters across the batch; `D1_PARAM_BUDGET` is 900 per statement and the module's own comment records 1,200 executing fine.
- **Auth, payload size, and row content** — 1 and 2 rows of the same generated rows succeed.

So it is something in `writeNeuronSnapshotToD1`'s batch as assembled by the handler, not the individual statements. `db.batch()` there mixes the three upserts plus one `DELETE` per netuid; my next step would be logging the failing statement index inside that catch, since `d1 write failed` currently swallows the underlying SQLite message.

### Note on the units trap

Live `neurons.captured_at` is **epoch milliseconds** (`1785657051538`). My first probe sent seconds, which made the per-netuid prune (`captured_at < ?`) a no-op — lucky, because a correct-looking probe with a *newer* millisecond timestamp would have pruned every real row for that netuid. Anyone testing this route against production should send a timestamp OLDER than the live max, or use a sentinel netuid. I verified D1 is intact afterwards: 30,333 rows across 129 subnets, and I removed the sentinel rows I created.

### Where this leaves things

D1 currently holds a complete snapshot frozen at `2026-08-02T07:50:51Z` and serves it correctly. Merging this container gets the producer onto Cloudflare and off the box, but the metagraph will not actually advance until the sync route's D1 write is fixed.

JSONbored:
Closing — #9170 moved `apps/indexer-rs` to the private infra repo, and this PR builds its image from `./apps/indexer-rs`, a path that no longer exists here. The container belongs on the other side of that boundary now.

Everything it verified still holds and is worth carrying over, so recording it here rather than losing it with the branch:

**The producer needs no rewrite.** `metagraph.rs` already POSTs to `/api/v1/internal/neurons-sync` and holds no Postgres client. Built and run against production, `POLLER_ONLY=metagraph` with no `DATABASE_URL` starts cleanly and does real work:

```
poller: POLLER_ONLY set, running only: metagraph
poller: no Postgres-backed job enabled, DATABASE_URL not required
metagraph: 6250 delegate take(s), 129 netuid(s) of validator trust, fetching all metagraphs
metagraph: resolved 129 subnet(s), 30333 rows
```

**Four things that will bite whoever wires this up in infra:**

1. **`curl` is required in the runtime image.** `post_sync` shells out to it (piping the body through stdin so a 30k-row payload never hits an arg limit). A slim image without it does all the chain work, resolves every subnet, then dies at the final POST with `spawn curl: No such file or directory` — which reads like a network fault, not a missing binary.
2. **The instance must be a singleton.** Two schedulers race the neurons route's per-netuid prune, which keys on `captured_at`: an older snapshot landing after a newer one deletes UIDs the newer one just wrote.
3. **`NEURONS_SYNC_SECRET` is now rotated.** It did not exist on the main Worker (only data-api), so I generated a new value and set it on both. Data-api uses Worker versions, so it needed `wrangler versions secret put` plus a deploy — done, after confirming the live version was exactly main's HEAD, so the deploy shipped identical code with only the secret changed.
4. **`@cloudflare/containers` breaks plain Node.** Its dist entry imports `./lib/container` with no extension, so re-exporting the Container class through `workers/api.ts` kills every Node script that imports the handler (`validate:committed-seed` first). It has to be exported from the wrangler-only entrypoint. Same import breaks vitest, so control logic needs to sit in a module that does not import the Container runtime.

**And the reason none of this was enough:** the sync route itself was rejecting every real payload. `POST /api/v1/internal/neurons-sync` returned 502 `d1 write failed` for anything above 4 rows, because D1's Workers binding caps a query at 100 bound parameters while `D1_PARAM_BUDGET` was 900. Fixed in #9175 — without that, the container would have run perfectly and written nothing.

The `DATABASE_URL` change this depended on (#9162) is already merged and travelled with the crate.

PR #9169, opened 2026-08-02
feat(poller): run the chain-state poller as a Cloudflare Container

JSONbored:
#209's poller half, deployed and running: the box's `indexer-rs-poller` died with the box, freezing the metagraph-state family at its last poll. This revives it on Cloudflare Containers in the Postgres-free shape #9162 made possible.

## Shape

- `apps/indexer-rs/Dockerfile.poller` — same build as the main Dockerfile, but bakes `POLLER_ONLY=metagraph,subnet-hyperparams,account-identity` and the poller CMD (Containers configure an image, not a command, so the selection must live in the image; the main Dockerfile keeps backfill's CMD untouched)
- `workers/poller-container.ts` — a minimal Worker keeping ONE instance alive. **Liveness is by cron, not by request**: the poller serves no HTTP, so a 5-minute scheduled ping revives it after eviction/crash/deploy, and the 15-minute metagraph cadence bounds a worst-case restart at one missed tick
- `wrangler.poller.jsonc` — the `metagraphed-poller` worker: container binding, singleton DO, cron

The three lanes read the chain over public RPC and POST to the api worker's token-authed sync routes, which persist to D1 (#9157 + siblings) — no database access anywhere in the container.

## Already live

Deployed ahead of merge (this config deploys via `wrangler deploy -c wrangler.poller.jsonc`, not Workers Builds): application created, secrets set, cron armed. Freshness is verifiable as `neurons.captured_at` advancing in D1 every ~15 min.

## Explicitly out of scope

The five Postgres-backed lanes (ownership, balances, nominators, self-stake, self-health) stay disabled until they have a Cloudflare-native sink, and the block-stream decode Containers indexer is separate follow-up — both tracked in #9146.

Refs #9146.

### DIFF
--- a/apps/indexer-rs/Dockerfile.poller
+++ b/apps/indexer-rs/Dockerfile.poller
@@ -1,0 +1,26 @@
+# The POLLER image for the Cloudflare Container deployment (#209 /
+# metagraphed#9146): identical build to ./Dockerfile, but with the poller as
+# the image CMD. The main Dockerfile keeps backfill's entrypoint as CMD and
+# selects the poller via docker_container `command:` on the box -- Cloudflare
+# Containers configure an image, not a command, so the selection has to live
+# in the image itself here. Two Dockerfiles sharing the build stanza verbatim
+# is the cost; a divergence would surface as a build failure, not a silent
+# behavioral difference, since both COPY the same two binaries.
+FROM rust:1-bookworm AS build
+WORKDIR /app
+COPY Cargo.toml Cargo.lock ./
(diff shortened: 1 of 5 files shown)


JSONbored:
Superseded by #9171 + the private-repo move: the poller container's source and deploy config now live in metagraphed-infra (services/indexer-rs) per the infrastructure/backend boundary decision. The deployed metagraphed-poller worker is unaffected and stays live — future deploys run from the private checkout.

PR #9175, opened 2026-08-03
fix(data-api): size the D1 write to the binding's parameter cap, not the HTTP API's

JSONbored:
> [!IMPORTANT]
> **Production is broken right now.** Every neurons sync above 4 rows returns 502 `d1 write failed`, so the metagraph has not advanced since `2026-08-02T07:50:51Z`. The write path landed in #9157/#9160, but nothing it is given can actually be written.

## Root cause

D1's **Workers binding** rejects a query binding more than **100 parameters**. `D1_PARAM_BUDGET` was `900`, justified in its own comment by a measurement that 1,200 parameters "execute fine" against production D1.

They do — through `wrangler d1 execute`, which goes over the **HTTP API** and has a far higher ceiling. The binding does not. The two paths look identical from the repo, and nothing in the test suite could tell them apart because the tests asserted `D1_PARAM_BUDGET`'s own value rather than the platform limit it exists to respect.

## The arithmetic is exact

`neuron_daily` is the widest table at 22 columns, so five rows is 110 bound parameters — the first chunk to cross 100.

Confirmed against production:

| rows | widest chunk | result |
| ---: | ---: | --- |
| 3 | 66 params | **200** |
| 4 | 88 params | **200** |
| 5 | 110 params | **502** |
| 100 / 1200 / 2000 | — | 502 |

## Why this looked like something else

Every statement the writer emits executes fine on its own via `wrangler d1 execute` — including all four in a single multi-statement transaction. The deployed schema matches `NEURON_DAILY_COLUMNS` exactly, in order. Duplicate-`account` upserts (the shape real coldkeys produce) succeed. So it presented as a batch problem, or a schema problem, or a size problem — and it is none of those. It is a per-query cap that only exists on the binding.

## The fix

`D1_PARAM_BUDGET = 100`, still **derived** per table so adding a column re-sizes the chunk rather than silently crossing the limit: 5 rows/statement for `neurons`, 4 for `neuron_daily`, 6 for `account_position_daily`.

That makes a full 30k-row snapshot ~18,700 statements, which one `db.batch()` will not carry — so the batch is chunked too, at `src/observations-d1.ts`'s already-proven 50.

**That trades one implicit transaction for several, and the ordering is what makes it safe.** Callers append every upsert *before* any prune, and `runBatches` preserves that order, so a mid-run failure leaves stale rows the next tick removes. The inverse — prune first, then fail — would delete live UIDs with nothing to replace them. The order is a contract, documented as one, not an accident of how the array was built.

## Tests that assert the limit, not the constant

Three new tests, and I verified they fail for the right reason: **all three fail at 900 and pass at 100.**

```
× no emitted statement exceeds the binding's parameter limit
× the budget admits no chunk wider than the limit, for any table
× statements are split across bounded batches
```

They pin `D1_BINDING_PARAM_LIMIT = 100` independently of `D1_PARAM_BUDGET`, so raising the budget again can only fail here. That is the gap that let this ship: the previous tests were written from the writer's own assumption, so they were green throughout.

Refs #9146


JSONbored:
Closing — #9173 landed the same diagnosis and fix independently while this was open. Confirmed against production just now: 5, 50 and 500-row payloads all return 200, so the lane is unblocked.

We converged on the same root cause from different directions, which is reassuring: the Workers binding caps a query at 100 bound parameters, the 900 budget was measured through the wrangler/HTTP path where 1,200 genuinely do execute, and the batch needed slicing once the budget dropped. #9173's 90 is the safer number of the two.

Its tests also pin the 100 limit independently of the constant, which was the gap I most wanted closed — so there is nothing here worth carrying over.

**One small thing worth a follow-up:** `tests/neurons-d1-write.test.ts` still carries the pre-fix rationale at the `buildUpsert` chunk-size test:

> Measured against production D1: 1,200 bound parameters execute fine, so the 900 budget has real headroom.

That comment is now the opposite of the finding, and it sits directly above an assertion about the bound-parameter budget — exactly where someone re-tuning this would read it and conclude there is headroom that does not exist on the binding.

For the record, the reproduction that isolated it, in case the boundary is ever re-tested: `neuron_daily` is the widest table at 22 columns, so 5 rows = 110 params is the first chunk to cross 100 — 3 rows (66) and 4 rows (88) returned 200 while 5 returned 502. Every statement executes fine alone via `wrangler d1 execute`, including all four in one multi-statement transaction, which is exactly why it presented as a batch or schema problem rather than a per-query cap.

PR #9201, opened 2026-08-03
fix(projections): an empty lane result is an answer, not a failure

JSONbored:
## Summary

**`main`'s `Validate` has been red since #9195.** `handleScheduled`'s projection tick reported `ok: false` and the two `api-coverage` tests covering it failed. A red main blocks the gate for every PR, so this goes first.

## The bug

`computeBlocksSummary` declined on `rows.length === 0` — conflating *"the lakehouse returned nothing"* with *"the lakehouse could not answer"*.

The `ProjectionLane` contract is explicit that `null` means the latter:

> `compute(env)` … or **null when the lakehouse could not answer EVERY query** — a partial answer is refused

Every sibling lane follows it, returning `null` only when a query itself came back `null`. The runner cannot tell the two apart, so one lane's empty scan marked the **whole tick** failed.

`buildBlocksSummary` already has an explicit zero-block branch, so the empty body was always well-formed — nothing had to be invented to stop declining.

**The failure posture is unchanged where it matters:** a query that genuinely cannot be answered still returns `null`, so a failed compute still never overwrites a good artifact.

## The test is also why this was missable

Its expected `lanes` map and artifact-key list were **written out by hand**, so registering a lane broke an assertion that had nothing to say about the new lane. The test, not the code, is what turned main red first.

Both are now derived from `PROJECTION_LANES`. The key list is derived from each lane's **name** rather than from `artifactKey` — deriving it from the same field the runner reads would move both sides together and catch nothing. Every current lane's key is exactly `metagraph/projections/<name>.json`, so a key that stops matching its name is either a typo or a writer/reader pair that has drifted.

I also added the case neither direction had: an empty result must produce a body with `row_count: 0`, **and** an unanswerable query must still decline.

## Mutations

| Mutation | Result |
| --- | --- |
| restore the empty-is-failure conflation (main's bug) | **2 fail** |
| stop declining on an unanswerable query | `TypeError: Cannot read properties of null` — the decline path is real |
| point a lane at the wrong artifact key | *"every registered lane must write the artifact key its name implies"* |

The second one is worth noting: my first attempt at that mutation silently patched a **different** lane's null check — `perl` replaced only the first match in the file — and passed. Targeting `computeBlocksSummary` specifically is what proved the assertion has teeth.

## Validation

```
npm run typecheck / lint / format:check    clean
tests/api-coverage.test.ts (lane cron)     4 passed
```

Locally the full file fails 51 on this branch and **53 on clean `origin/main`** — the difference is exactly the two CI reports, and the remainder is the known local-only class (they do not appear in CI, where only the 2 lane tests failed).

Fixes the failing `test` job on `main`.

### DIFF
--- a/src/projection-lanes.ts
+++ b/src/projection-lanes.ts
@@ -785,7 +785,12 @@ async function computeBlocksSummary(
     `SELECT ${BLOCKS_SUMMARY_READ_COLUMNS} FROM chain.blocks ` +
       `ORDER BY block_number DESC LIMIT ${BLOCKS_SUMMARY_SCAN_CAP}`,
   );
-  if (rows === null || rows.length === 0) return null;
+  // null ONLY when the lakehouse could not answer -- the contract every other
+  // lane follows. An empty result is an ANSWER, not a failure: treating it as
+  // one made the whole tick report ok:false, because the runner cannot tell a
+  // lane that declined from a lane that broke. buildBlocksSummary already has
+  // an explicit zero-block branch, so the empty body is well-formed.
+  if (rows === null) return null;
   return {
     schema_version: 1,
     generated_at: new Date(generatedAt).toISOString(),
--- a/tests/api-coverage.test.ts
+++ b/tests/api-coverage.test.ts
@@ -3388,33 +3388,60 @@ describe("handleScheduled PROJECTION_LANES_CRON", () => {
         },
       },
     });
+    // DERIVED from the registry, not hand-listed. The previous literal map
+    // went stale the moment a lane was added -- #9195 registered
+    // blocks-summary and this assertion, not the lane, is what turned main
+    // red. Deriving it means a new lane is covered the day it lands, and the
+    // property actually worth asserting (every registered lane ran, reported a
+    // count, and wrote its own artifact key) is what gets checked.
+    const { PROJECTION_LANES } = await import("../src/projection-lanes.ts");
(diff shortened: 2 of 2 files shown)


JSONbored:
Closing — this was the wrong fix, and #9200 has the right one.

I read `computeBlocksSummary`'s `rows.length === 0` decline as a contract violation, because every sibling lane returns null only when a *query* returns null and the interface comment says null means "could not answer EVERY query". Making it return `row_count: 0` turned the tick green.

It also would have introduced the exact hazard the module's failure posture exists to prevent. The runner logs `compute declined; previous artifact left in place` — a decline **preserves** the last good artifact. So on any empty read my version would have overwritten a good block-production card with a zeroed one, silently. #9200's comment names this directly: the lane "refuses to overwrite a good card with a zeroed one." That is a deliberate per-lane choice, not an inconsistency, and `chain.blocks` is never empty in production anyway.

I verified the runner's behaviour only after seeing #9200 — had I checked what `null` actually causes before assuming what it meant, I would not have opened this.

The one piece here worth keeping is independent of lane semantics: the test's expected `lanes` map and artifact-key list are written out by hand, so registering a lane breaks an assertion that has nothing to say about the new lane — the test, not the code, is what turned main red first. #9200 fixed the immediate failure but left both literals in place, so the next lane will do it again. I'll raise that separately as a test-only change.

PR #9205, opened 2026-08-03
test(projections): teach the api-coverage lane tests about blocks-summary

JSONbored:
## Main is red

#9195 registered the `blocks-summary` projection lane and updated the whole-tick tests in `tests/projection-lanes.test.ts` — but `tests/api-coverage.test.ts` carries **its own** pair of `PROJECTION_LANES_CRON` tests with the same assumption, and those were missed.

It landed red because #9195's own test run was **cancelled** rather than completed at merge, so nothing caught it before it hit main.

## Why they fail

Both tests fake the lakehouse as returning zero rows for every query. That makes `blocks-summary` **decline by design**: it stores a shaped card, and `buildBlocksSummary([])` asserts a chain with zero blocks, zero authors and no spec version — which is false, not an honest empty. Persisting it would overwrite real numbers. So the tick writes one fewer artifact than the assertions expect.

## The fix

Both fakes now feed a row to the blocks scan only, matched on `ORDER BY block_number DESC` rather than the table name — `chain-activity` aggregates over `chain.blocks` too, and feeding *it* rows would change what that lane computes. This mirrors the helper already added to `tests/projection-lanes.test.ts`.

Test-only; no source change. `tests/api-coverage.test.ts`: **312 tests green**.

### DIFF
--- a/tests/api-coverage.test.ts
+++ b/tests/api-coverage.test.ts
@@ -3372,14 +3372,48 @@ describe("handleScheduled PROJECTION_LANES_CRON", () => {
     });
   });
 
-  test("recomputes and writes every registered lane's artifact on one tick", async () => {
-    const puts: string[] = [];
-    globalThis.fetch = (async () =>
-      ({
+  /**
+   * A lakehouse fake for a whole projection tick: every lane sees an empty
+   * result EXCEPT the blocks scan, which gets one row.
+   *
+   * blocks-summary stores a shaped card rather than rows, so an empty scan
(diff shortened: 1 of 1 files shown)


JSONbored:
Superseded by #9200, which landed the same fix — including the `tests/api-coverage.test.ts` half — and matches blocks-summary's read specifically rather than every `chain.blocks` query. Closing rather than resolving the conflict; nothing here is left to carry over.

PR #9222, opened 2026-08-03
feat(chain): serve the registration leaderboard from a scheduled projection

### LINKED ISSUE #9221: fix(chain): /chain/registrations serves an empty leaderboard — no Cloudflare-native source
`GET /api/v1/chain/registrations` (plus the `get_chain_registrations` MCP tool and GraphQL's `chain_registrations`) returns a schema-stable empty block: `subnet_count: 0`, `subnets: []`.

Verified live 2026-08-03 — `?limit=7` returns `{'subnets': 0}`.

## Cause

The D1 loader was retired in #4909 (`account_events`' D1 table was dropped in #4772), so all three callers do `tryPostgresTier() ?? buildChainRegistrations([])`. With Postgres eliminated the tier always misses and there is no Cloudflare-native source behind it.

## The data exists

`chain.account_events` in the R2 lakehouse holds **1,071,405** `NeuronRegistered` rows, genesis to head. A live 7d aggregate returns 6,023 distinct registrants and a real per-subnet leaderboard (netuid 5: 463 registrations / 415 registrants).

## Shape

This is a chain-wide aggregate with no selective predicate — the per-subnet grouping alone scans **186 MB** (~2.8s). Per #9146 and `src/account-feeds-cold-tier.ts`'s own header, chain-wide aggregates belong in a **scheduled projection lane**, not a request-time read.

Note the network rollup is a separate `COUNT(DISTINCT hotkey)` and is *not* derivable from the per-subnet rows: one hotkey registering on three subnets is three subnet-level registrants but one network-wide distinct registrant.

Part of #9146.

JSONbored:
## The bug is worse than "empty"

`/chain/registrations` served **real 7d numbers and an empty 30d block**. That asymmetry is not a data gap — R2 SQL *refuses* the query the request-time rollup issues:

```
40015: scan budget exceeded: scanning too much data for count(DISTINCT) with
GROUP BY. Reduce data with WHERE clauses, use approximate functions
(APPROX_DISTINCT, ...), add a GROUP BY to distribute the aggregation, or
narrow the column selection.
```

Verified live. Narrowing with a `block_number >=` floor does **not** help — the budget is spent on the DISTINCT, not the row range. So the request-time shape structurally cannot answer 30d, and no amount of edge caching fixes that.

## The fix

A projection lane that distributes the aggregation, per the engine's own remedy: `GROUP BY netuid, hotkey` — which carries no `COUNT(DISTINCT)` at all — reduced writer-side.

This is **exact, not approximate**. I deliberately did not reach for `APPROX_DISTINCT`: `distinct_registrants` is a published figure, and silently approximating it would be a data defect dressed up as a performance choice.

### Proven equivalent, not assumed

Both forms run against the same fixed 7d cutoff; the distributed reduction reproduces the `COUNT(DISTINCT)` ground truth field for field:

| netuid | ground truth | distributed |
|---|---|---|
| 5 | 463 / 415 | 463 / 415 |
| 15 | 392 / 368 | 392 / 368 |
| 68 | 391 / 116 | 391 / 116 |
| 33 | 390 / 248 | 390 / 248 |
| 79 | 390 / 344 | 390 / 344 |
| **network distinct** | **6,023** | **6,023** |

### Why a projection and not a cold-tier read

Measured on the live engine: 30d is **22,597 `(netuid, hotkey)` pairs over 32,883 events, 494 MB scanned**. Fine once per tick; impossible per request. `src/account-feeds-cold-tier.ts`'s header already draws this line — account-scoped feeds get request-time reads because one address is a selective predicate; chain-wide aggregates go to projections.

### The rollup is not a sum of the rows

The network figure is computed over the pair set, not summed from the per-subnet rows: one hotkey registering on three subnets is three subnet-level registrants but **one** network-wide distinct registrant. A lane test pins this — hotkey `A` on two subnets yields per-subnet counts of 1 and 1 but a network count of 1, not 2.

Only the reduced rows are stored (one per netuid), never the 22,597 raw pairs.

## Supersedes the request-time read for this route

This replaces the `loadChainEventRollup` call for `/chain/registrations` only. The helper and `CHAIN_SERVING_ROLLUP` stay — `/chain/serving` still uses them, and its now-unused import here is removed.

**`/chain/serving` has the identical 30d defect** (7d: 2,933 · 30d: 0), from the same query shape. Filed as #9227 rather than widened into this change.

## Tests

10 lane tests + 17 reader tests, including the failure modes that produce a *plausible* number rather than an error: a null hotkey counted as a registrant, a non-numeric count turning a subnet's total into `NaN`, the network rollup recomputed from rows, and a half-failure where the cheap stamp succeeds and the heavy read fails.

Affected suites: **4,182 passing**. Patch coverage by diff-intersection against v8's uncovered set: **0 uncovered statements, 0 uncovered branches**.

Closes #9221
Refs #9146

### DIFF
--- a/src/chain-registrations-artifact.ts
+++ b/src/chain-registrations-artifact.ts
@@ -1,0 +1,89 @@
+// Network-wide registration activity served from a SCHEDULED PROJECTION
+// artifact when the Postgres tier misses (#9146).
+//
+// A PROJECTION, NOT A REQUEST-TIME READ, and that is the whole design point.
+// This is a chain-wide aggregate with no selective predicate: a COUNT(*) plus
+// two COUNT(DISTINCT hotkey) over every NeuronRegistered row in the window.
+// Measured on the live engine (2026-08-03) the per-subnet grouping alone reads
+// **186 MB** at ~67 MB/s -- roughly 2.8s. The account-scoped feeds can afford
+// a request-time lakehouse read because one address is a selective predicate
+// against a big table; this is not that shape. src/account-feeds-cold-tier.ts's
+// header draws the same line: chain-wide aggregates moved to scheduled
+// projections.
+//
+// Rows are stored VERBATIM, per the chain-* lane convention: the route's
+// ?limit= only slices the leaderboard, and buildChainRegistrations owns that
+// slice plus the network rollup and the intensity distribution. One artifact
+// therefore serves every window/limit combination the route accepts.
+
+import {
+  buildChainRegistrations,
+  CHAIN_REGISTRATIONS_WINDOWS,
+  DEFAULT_CHAIN_REGISTRATIONS_WINDOW,
+} from "./chain-registrations.ts";
+
+export const CHAIN_REGISTRATIONS_PROJECTION_KEY =
+  "metagraph/projections/chain-registrations.json";
+
+interface ArtifactBucket {
+  get(key: string): Promise<{ json(): Promise<unknown> } | null>;
+}
+
+/**
+ * The projected registration leaderboard for one window, or null when the
+ * artifact store cannot answer FAITHFULLY (unbound, missing object,
+ * unrecognized body, a window the lane did not precompute) so the caller keeps
+ * its schema-stable empty. Decline, never approximate.
+ */
+export async function loadChainRegistrationsFromArtifact(
+  env: Env | null | undefined,
+  query: { window?: string | null; limit?: number },
+): Promise<ReturnType<typeof buildChainRegistrations> | null> {
+  const bucket = (env as { METAGRAPH_ARCHIVE?: ArtifactBucket } | null)
+    ?.METAGRAPH_ARCHIVE;
+  if (!bucket?.get) return null;
+  try {
+    const object = await bucket.get(CHAIN_REGISTRATIONS_PROJECTION_KEY);
+    if (!object) return null;
+    const body = (await object.json()) as {
+      schema_version?: unknown;
+      windows?: unknown;
+    } | null;
+    // A body that is not the artifact the lane wrote is a decline, not a
+    // guess -- same contract as the sibling chain-* readers.
+    if (
+      body?.schema_version !== 1 ||
+      typeof body.windows !== "object" ||
+      body.windows === null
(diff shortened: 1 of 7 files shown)


JSONbored:
Heads-up on an overlap, since this and #9216 landed close together.

**#9216 merged a cold-tier read for `/api/v1/chain/registrations`** (and `/chain/serving`) about an hour ago — `src/chain-event-rollup-cold-tier.ts`, wired in `workers/request-handlers/analytics.ts`. So the REST route this PR describes as empty is now served, and this branch will conflict there.

Worth reconciling deliberately rather than by merge order, because the two approaches differ in a way that matters:

- **Cold-tier read (merged):** a per-netuid `GROUP BY` at request time. `NeuronRegistered` groups to ~129 rows, so it fits a request. Fresh by construction, no artifact to go stale.
- **Scheduled projection (this PR):** precomputed artifact, cheaper per request, but adds a lane that must run for the route to have anything.

I picked the cold tier only because the builders already expect pre-aggregated rows (`buildChainRegistrations`' own comment: *"the SQL loader GROUPs BY netuid, so production rows are unique per subnet"*) and ~129 rows is small. If the projection is preferred for cost or latency, superseding the cold-tier path is reasonable — it is one call site to remove.

**The genuinely unclosed gap is the one this PR gets right and #9216 did not: MCP and GraphQL.** I wired only the REST handler, so `get_chain_registrations` and `chain_registrations` still return empty on main. That was my omission.

I am not touching registrations further to avoid colliding with you — I am closing the same MCP/GraphQL parity gap for `/chain/serving` only, which this PR does not cover.

Two things measured live while building #9216 that apply here either way:

- The network-wide distinct **cannot be summed** from the per-subnet rows — one hotkey registering on five subnets is five rows and one registrant. It needs its own ungrouped query (2,933 network-wide servers for `AxonServed` in 7d, well under the per-subnet sum).
- `COUNT(DISTINCT)` **is** supported by R2 SQL; `count_if` is not. An early probe of mine returned 0 from a supported function purely because `hotkey` was null on that event kind, which is worth knowing before trusting any zero.

PR #9238, opened 2026-08-03
fix(api): answer /chain/serving from the lakehouse on MCP and GraphQL too

### LINKED ISSUE #9229: fix(api): /chain/serving is served from the lakehouse on REST but not MCP or GraphQL
## The gap

#9216 wired the lakehouse rollup into the **REST handler alone**. So on main today:

| Surface | `/chain/serving` |
| --- | --- |
| `GET /api/v1/chain/serving` | real numbers — 20 subnets, 2,933 network-wide servers |
| `get_chain_serving` (MCP) | **zeroed card** |
| `chain_serving` (GraphQL) | **zeroed card** |

Three surfaces answering the identical question, two of them wrong, and a caller cannot tell which one is lying. That was my omission in #9216, not a pre-existing gap.

## Why a shared loader rather than three call sites

Each surface would otherwise repeat "load the rollup, hand it to the builder, fall back on decline". Three copies is three chances to drift — the same reason `src/rpc-usage-loader.ts` exists and says *"for REST + MCP parity"* in its first line.

The loader returns **null** on a lakehouse miss rather than the zeroed card, because each surface's fallback is its own contract: GraphQL deliberately answers with a schema-stable card rather than an error, and that decision belongs at the call site.

## Acceptance

- [ ] All three surfaces call one loader; none re-implements the rollup
- [ ] The network-wide distinct still comes from its own ungrouped query — it cannot be summed from the per-subnet rows, since one hotkey serving five subnets is five rows and one server
- [ ] An unknown window falls back to 7d rather than silently widening the range
- [ ] A lakehouse miss declines, leaving each caller's existing fallback intact
- [ ] Mutation-tested: unwiring MCP, unwiring GraphQL, and deriving the network total from the rows must each fail
- [ ] Zero uncovered changed lines or branches

## Out of scope

`/chain/registrations` has the same parity gap, but #9222 is open against it with a scheduled-projection approach. Left alone deliberately to avoid colliding — see the note there.

Refs #9146, #9215

JSONbored:
## Summary

- `#9216` wired the lakehouse rollup into the REST handler alone, so two of the three surfaces answering `/chain/serving` returned a zeroed card.

Measured on production before this change:

| Surface | 7d |
| --- | --- |
| `GET /api/v1/chain/serving` | 20 subnets, 2,931 servers, 7,561 announcements |
| `get_chain_serving` (MCP) | **zeroed card** |
| `chain_serving` (GraphQL) | **zeroed card** |

## What Changed

- New `src/chain-serving-loader.ts` — one loader, mirroring `src/rpc-usage-loader.ts`'s "for REST + MCP parity" precedent.
- REST, MCP and GraphQL all call it; none re-implements the rollup.

Two bugs the REST call site had been casting past, now resolved in the loader:

- **window** is normalized against `ANALYTICS_WINDOWS` *before* the scan. An unrecognized value narrows to 7d instead of sending `windowDays: undefined` and silently widening the range — and the normalized label is carried into the card, so a response can never report a window its numbers did not come from.
- **limit** is resolved once for both halves. `parseLimitParam` types its result `number | undefined`, and left to their own defaults the scan caps at 200 while the builder caps at 20 — scanning ten times the rows the response can carry. The old `as unknown as Parameters<typeof buildChainServing>[1]` casts hid this; they are gone.

The loader declines with **null**, never the zeroed card — each surface's fallback is its own contract (GraphQL deliberately answers with a schema-stable card rather than an error), and returning the empty card here would make all three fallbacks identical and unreachable.

## Acceptance

- [x] All three surfaces call one loader; none re-implements the rollup
- [x] The network-wide distinct still comes from its own ungrouped query
- [x] An unknown window falls back to 7d rather than silently widening the range
- [x] A lakehouse miss declines, leaving each caller's existing fallback intact
- [x] Mutation-tested — see below
- [x] Zero uncovered changed lines or branches

## Mutation testing

Parity is asserted by driving all three entry points against **one** fake lakehouse and comparing their output, rather than three unit tests that could each pass while the surfaces disagreed. The fixture's per-subnet distincts sum to 12 against a network block of 7, so the gap is the detector.

| Mutation | Result |
| --- | --- |
| Unwire MCP | ✗ fails (`MCP subnet_count`) |
| Unwire GraphQL | ✗ fails (`GraphQL subnet_count`) |
| Derive the network total by summing the rows | ✗ fails (3 tests) |

## Registry Safety

- [x] Links a tracked, currently-open issue (`Closes #9229`) — required.
- [x] Public API changes are intentional — MCP and GraphQL now return the numbers REST already did; no schema change, `validate:contract-drift` green.

## Validation

- [x] `npm run typecheck` · `npm run lint` · `npm run format:check` · `git diff --check`
- [x] `npm run validate` · `npm run validate:contract-drift` · `npm run scan:public-safety`
- [x] `npm run worker:test`
- [x] `npx vitest run` — **624 files / 14,778 tests, exit 0**
- [x] The touched suites run deterministically — `chain-serving-loader`, `chain-serving`, `chain-event-rollup-cold-tier`, `graphql`, `mcp-server`, `analytics` × 3 consecutive runs, 2,507 passed each time
- [x] Patch coverage: diff lines ∩ v8 uncovered set = **∅** across all four files (loader 83, graphql 7, mcp-server 7, analytics 12), statements and branches

> Separately: the full local suite showed load-dependent flakes in the subprocess-spawning suites (`script-utils`, `r2-upload`, `lib-helpers`), which have no declared timeout and so get vitest's 5s default outside `test:ci`. Unrelated to this change — fixing in a follow-up.

### DIFF
--- a/src/chain-serving-loader.ts
+++ b/src/chain-serving-loader.ts
@@ -1,0 +1,83 @@
+// Shared /chain/serving lakehouse rollup loader for REST + MCP + GraphQL parity.
+//
+// #9216 wired the lakehouse rollup into the REST handler alone, so the same
+// question answered three ways gave real numbers on `GET /api/v1/chain/serving`
+// and a zeroed card on `get_chain_serving` and `chain_serving` -- with nothing
+// in any of the three payloads letting a caller tell which was lying. Same
+// reason src/rpc-usage-loader.ts exists: the alternative is each surface
+// repeating "load the rollup, hand it to the builder, decline on a miss", which
+// is three places to drift.
+//
+// DECLINES WITH null, never with the zeroed card. Each surface's fallback is
+// its own published contract -- GraphQL deliberately answers with a
+// schema-stable card rather than an error, REST falls through to its empty
+// payload -- and that choice belongs at the call site, not here. Returning the
+// empty card from this function would quietly make every caller's fallback
+// identical and unreachable.
+import {
+  ANALYTICS_WINDOWS,
+  DEFAULT_ANALYTICS_WINDOW,
+} from "../workers/config.ts";
+import {
+  CHAIN_SERVING_ROLLUP,
+  loadChainEventRollup,
+} from "./chain-event-rollup-cold-tier.ts";
+import {
+  buildChainServing,
+  CHAIN_SERVING_LIMIT_DEFAULT,
+  type ChainServingResult,
+} from "./chain-serving.ts";
+
+/**
+ * The lakehouse rollup for one window, built into the published card.
+ *
+ * `window` is normalized against ANALYTICS_WINDOWS rather than trusted. All
+ * three callers validate it before getting here, so this is belt-and-braces --
+ * but the failure it prevents is not a 4xx, it is a `windowDays: undefined`
+ * reaching the scan and silently widening it past the range the caller asked
+ * for. The label is then carried into the built card, so a fallback can never
+ * report a window the numbers did not come from.
+ */
+export async function loadChainServingRollup(
+  env: Parameters<typeof loadChainEventRollup>[0],
+  {
+    window,
+    limit,
+    now,
+    query,
+  }: {
+    window?: unknown;
+    /**
+     * Optional because parseLimitParam types its result `number | undefined`
+     * even when a defaultLimit is supplied. Resolved to ONE number here and
+     * used for both halves: the scan cap and the builder's row cap defaulted
+     * separately (200 vs 20) if each were left to its own, which would scan ten
+     * times the rows the response can carry.
+     */
+    limit?: number;
+    now?: number;
+    query?: Parameters<typeof loadChainEventRollup>[2]["query"];
+  },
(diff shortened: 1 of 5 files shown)


JSONbored:
Duplicate — #9230 landed the same fix while this was in flight, and it is the one on main. Closing in favour of it.

Two small things this branch handled that #9230's `loadChainServingColdTier` still does, both latent rather than live (every caller validates before calling, so neither is reachable today):

1. **The fallback label does not match the fallback scan.** `windowDays` resolves via `?? 7`, but the original `window` string is handed to `buildChainServing` — so an unrecognised label scans 7 days and publishes a card claiming that label. The doc comment's promise ("instead of quietly widening the range") holds for the scan but not for what the response says it is.
2. **`limit` left undefined defaults twice, differently.** `loadChainEventRollup` caps at 200 and `buildChainServing` at 20, so an omitted limit scans ten times the rows the response can carry.

Filing those as their own issue rather than reopening this.

PR #9246, opened 2026-08-03
feat(api): serve the per-account serving and registrations footprints

### LINKED ISSUE #9245: feat(api): serve the per-account serving and registrations footprints from the lakehouse
## The gap

`/api/v1/accounts/{ss58}/serving` and `/api/v1/accounts/{ss58}/registrations` answer schema-stable zeros for every account:

```
{ "subnet_count": 0, "subnets": [] }
```

They are the two `account_events` account feeds with **no cold-tier reader**. `makeAccountEventHandler` already takes a `coldTier` hook and `src/account-feeds-cold-tier.ts` already serves the sibling feeds — these two were simply never wired.

The data is there: `AxonServed` (2,791,121 rows) and `NeuronRegistered` (1,071,405), both with a **populated hotkey on every row**.

## The predicate is the whole design question

The existing weight-setters reader resolves `(netuid, uid)` slots against the neurons table, because `WeightsSet` has a NULL hotkey on all 50,890,747 rows and the chain event only emits `[netuid, uid]`.

**These two must not borrow that path.** They carry a real hotkey, so the account *is* the hotkey and the predicate is a single equality. Attributing by neuron slot instead would credit an account with rows from whoever holds that slot **now** — a subtly wrong answer that still looks plausible, since slots are reassigned after deregistration.

## Acceptance

- [ ] Both routes serve real per-subnet footprints with first/last timestamps
- [ ] Each filters by `hotkey`, and neither uses the `(netuid, uid)` slot workaround
- [ ] Each counts under its own alias — the builders read `announcements` and `registrations` respectively, and a shared alias leaves one reading `undefined` and reporting an empty footprint rather than failing
- [ ] An unparseable address declines **without querying**: R2 SQL has no bound parameters, so a rejected address must never reach the string
- [ ] An empty footprint is an answer (zeros with a real `generated_at`), not a decline
- [ ] The wiring itself is tested through the handler — a unit test of the loader cannot reach the `coldTier` lambda, and an unwired feed answers zeros indistinguishable from a cold store
- [ ] Zero uncovered changed lines or branches

## Not in scope

`/accounts/{ss58}/axon-removals` and `/deregistrations` stay empty and correctly so: `AxonInfoRemoved` and `NeuronDeregistered` appear nowhere in `SubtensorModule`'s event vocabulary — the chain never emits them.

Refs #9146

JSONbored:
## Summary

`/api/v1/accounts/{ss58}/serving` and `/registrations` answered schema-stable zeros for **every** account. They were the two `account_events` feeds with no cold-tier reader — `makeAccountEventHandler` already takes a `coldTier` hook and `src/account-feeds-cold-tier.ts` already serves their siblings, so these were simply never wired.

The data was there all along: `AxonServed` (2,791,121 rows) and `NeuronRegistered` (1,071,405), both with a **populated hotkey on every row**.

## The predicate is the whole design question

The existing weight-setters reader resolves `(netuid, uid)` slots against the neurons table, because `WeightsSet` has a NULL hotkey on all 50,890,747 rows — the chain event only emits `[netuid, uid]`.

**These two must not borrow that path.** They carry a real hotkey, so the account *is* the hotkey and the predicate is a single equality. Attributing by neuron slot would credit an account with whatever the slot's **current** holder did — a subtly wrong answer that still looks plausible, since slots are reassigned after deregistration. A test asserts the slot form is absent.

## Details that decide correctness

- **Each feed counts under its own alias.** The builders read `row.announcements` and `row.registrations`; a shared alias leaves one reading `undefined` and reporting an empty footprint rather than failing.
- **An empty footprint is an answer, not a decline.** An account that served nothing really did serve nothing — declining falls through to the same zeros but loses the `generated_at` the caller reports.
- **An unparseable address declines without querying.** R2 SQL has no bound parameters, so a rejected address must never reach the string.

## The wiring is tested through the handler

The `coldTier` lambda is exactly the part a loader unit test cannot reach, and an unwired feed answers zeros **indistinguishable from a cold store**.

| Mutation | Result |
| --- | --- |
| unwire the serving cold tier | *"the serving route did not query AxonServed — its cold tier is unwired, so it answers zeros indistinguishable from a cold store"* |
| drop the hotkey predicate | fails — would return every account's rows |
| reuse one alias for both feeds | *"did not match /AS registrations/"* |
| skip the address guard | fails |

That first mutation initially *passed* because my `perl` pattern didn't match after prettier reflowed the line — worth stating, since a mutation that silently fails to apply looks exactly like a test with teeth.

## Correctly left empty

`/accounts/{ss58}/axon-removals` and `/deregistrations` stay zeroed: `AxonInfoRemoved` and `NeuronDeregistered` appear **nowhere** in `SubtensorModule`'s event vocabulary — the chain never emits them.

## Validation

```
npm run typecheck / lint / format:check    clean
account routes + cold-tier suites          65 passed
```

**Patch coverage: zero uncovered changed lines or branches.**

Closes #9245
Refs #9146

### DIFF
--- a/src/account-feeds-cold-tier.ts
+++ b/src/account-feeds-cold-tier.ts
@@ -59,6 +59,16 @@ import {
   COUNTERPARTIES_SCAN_CAP,
   type CounterpartyRelationshipResult,
 } from "./counterparties.ts";
+import {
+  buildAccountServing,
+  DEFAULT_SERVING_WINDOW,
+  SERVING_WINDOWS,
+} from "./account-serving.ts";
+import {
+  buildAccountRegistrations,
+  DEFAULT_REGISTRATION_WINDOW,
+  REGISTRATION_WINDOWS,
+} from "./account-registrations.ts";
 import { decodeCursor, encodeCursor } from "./cursor.ts";
 import { r2SqlQuery, safeBlockNumber, safeSs58Literal } from "./r2-sql.ts";
 import { OFFSET_EMULATION_CAP } from "./r2-sql-blocks.ts";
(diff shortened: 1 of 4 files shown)


JSONbored:
Closing as a duplicate — #9244 shipped the same two feeds while this was in review, and its version is the better one.

I checked rather than assuming, because the routes still returned zeros after it merged. Two things worth recording:

**Their predicate is correct.** Both readers key on `hotkey = '<addr>'`, which is the right attribution for `AxonServed` and `NeuronRegistered` — those carry a real hotkey on every row, unlike `WeightsSet`. Borrowing the `(netuid, uid)` slot path that weight-setters needs would have credited an account with whatever the slot's *current* holder did, since slots are reassigned after deregistration. #9244 avoids that.

**It also wires MCP, which this PR did not.** That is the same parity gap I had to fix separately in #9230 for `/chain/serving`, so #9244 is strictly ahead here.

**The zeros are deploy lag, not a bug.** The live Worker is version `5d8c0f9d` created **10:57:41Z**; #9244 merged at **11:06:03Z**. The deployed code predates it by nine minutes. I confirmed the underlying query is sound by running #9244's exact SQL against the lakehouse for the busiest `AxonServed` hotkey in the window:

```
netuid 7 · 748 announcements · 2026-07-16 → 2026-08-02 · 242 MB scanned · 3.3s
```

Well inside the 15s query timeout, so it will serve real numbers once the deploy lands. Worth a re-check afterwards — Workers Builds has stalled without failing before.

The only thing here not in #9244 is a handler-level test asserting the `coldTier` lambda is actually reached (a loader unit test cannot reach it, and an unwired feed answers zeros indistinguishable from a cold store). If that is wanted I can raise it as a test-only follow-up rather than carry this branch.

Issue #9254, opened 2026-08-03
fix(accounts): the account summary card reads all zeros while its own detail routes read real rows


JSONbored:
## What

`GET /api/v1/accounts/{ss58}` returns an all-zero card while the **same account's own detail routes read real rows from the same table**.

Measured 2026-08-03 for `5E2LP6EnZ54m3wS8s1yPvD5c3xo71kQroBw7aUVK32TKeZ5u` (probed sequentially — this API declines under concurrent probing):

```
/api/v1/accounts/{ss58}                -> event_count 0, subnet_count 0, event_kinds 0,
                                          registrations 0, recent_events 0
/api/v1/accounts/{ss58}/events         -> 6 events
/api/v1/accounts/{ss58}/registrations  -> 146
/api/v1/accounts/{ss58}/serving        -> 63
```

## Why

`handleAccountSummary` has exactly one tier: `tryPostgresTier(METAGRAPH_ACCOUNT_EVENTS_SOURCE) ?? buildAccountSummary(ss58, {})`. The box's Postgres is gone, so that read always misses and **every field on the card goes to zero at once** — while the sibling routes each got a lakehouse reader in `src/account-feeds-cold-tier.ts`.

The card is not a lower-fidelity view of those routes; it is the account landing surface, so this is the most visible instance of the pattern.

## Scope

`buildAccountSummary` composes **three** sources, and it is null-safe per field:

| field group | source | status |
| --- | --- | --- |
| `event_count`, `subnet_count`, `event_scan_capped`, `first_*`/`last_*`, `event_kinds`, `recent_events` | `chain.account_events` | **portable now** — same table the sibling readers already use |
| `registrations` | `neurons` (D1, live) | needs its own reader |
| `activity` (`tx_count`, `total_fee_tao`, `modules_called`) | extrinsics tier, by signer | needs its own reader |

Porting the `account_events` half is the bulk of the card and is self-contained. The other two stay empty rather than being faked from the first — they are genuinely different sources with different attribution (`neurons` by hotkey, extrinsics by signer, `account_events` by hotkey OR coldkey).

## Two things an implementation must get right

1. **Attribution must match the feed.** `loadAccountEventsColdTier` uses `(hotkey = X OR coldkey = X)`. A card that attributed differently would reintroduce the same disagreement in a subtler form.
[...]

PR #9283, opened 2026-08-03
fix(accounts): take subnet_count from a GROUP BY subquery, not COUNT(DISTINCT)

### LINKED ISSUE #9281: fix(accounts): the account summary card still serves zeros — its aggregate query is rejected by R2 SQL
## What

#9254 (PR #9257) merged, deployed, and **the card still serves zeros**. Verified 33 minutes after merge, three sequential probes:

```
/api/v1/accounts/{ss58}  ->  event_count 0, event_kinds 0, recent_events 0
```

Deploys were landing normally over the same window — `/chain/weights/setters` (#9251) began serving data in that period — so this is not deploy lag. It is my bug.

## Cause

The aggregate read I shipped uses `count(DISTINCT netuid)`:

```sql
WITH scan AS (SELECT … ORDER BY block_number DESC LIMIT 5000)
SELECT count(*) AS c, count(DISTINCT netuid) AS sc, min(block_number) AS fb, … FROM scan
```

R2 SQL **rejects an ungrouped `count(DISTINCT)` at this scale**:

```
40015: scan budget exceeded: scanning too much data for count(DISTINCT)
without GROUP BY
```

This is the identical failure the #9251 correction found and fixed in `loadChainEventIdentityRollup`. I wrote this reader before that correction landed and reasoned that the CTE's `LIMIT 5000` made the DISTINCT cheap. It does not: **the planner still evaluates the DISTINCT against the underlying scan**, so being inside a capped CTE saves nothing.

A rejected query returns null, the loader declines by design, and the handler falls through to `buildAccountSummary(ss58, {})` — the all-zero card. The decline path worked exactly as intended; the query never should have been sent.

## Fix

Take `subnet_count` from its own `GROUP BY` subquery, the pattern #9251 settled on:

```sql
WITH scan AS (…) SELECT count(*) AS sc FROM (SELECT netuid FROM scan GROUP BY netuid)
```

`count(*)` stays ungrouped — a plain row count has no DISTINCT in it, so neither the budget rejection nor the wrong-question problem applies.

## Guard

A test asserting **no read in this loader uses `COUNT(DISTINCT)` at all**, mirroring the one added alongside the #9251 fix. Confirmed it bites by reintroducing the original query.

The deeper lesson, worth stating once: a lakehouse query that typechecks, passes every local test against a fake engine, and reads correctly can still be rejected by the engine at production scale. **The fake-engine tests cannot see this class of failure** — only a production probe after deploy can, which is how this was found.

JSONbored:
## Summary

- #9254 (PR #9257) merged, deployed, and **the card still serves zeros**. This is my bug, found by probing production after the merge.

Verified 33 minutes post-merge, three sequential probes:

```
/api/v1/accounts/{ss58}  ->  event_count 0, event_kinds 0, recent_events 0
```

Not deploy lag — `/chain/weights/setters` (#9251) began serving data over the same window.

## Cause

The aggregate read used `count(DISTINCT netuid)`. R2 SQL **rejects an ungrouped `count(DISTINCT)` at this scale**:

```
40015: scan budget exceeded: scanning too much data for count(DISTINCT) without GROUP BY
```

Identical to the failure the #9251 correction found. I wrote this reader before that landed and reasoned the CTE's `LIMIT 5000` made the DISTINCT cheap. **It does not** — the planner still evaluates the DISTINCT against the underlying scan, so being inside a capped CTE saves nothing.

The rejected query returned null, the loader declined exactly as designed, and the handler fell through to `buildAccountSummary(ss58, {})`. The decline path was correct; the query should never have been sent.

## Fix

`subnet_count` now comes from its own `GROUP BY` subquery — the shape #9251 settled on:

```sql
WITH scan AS (…) SELECT count(*) AS sc FROM (SELECT netuid FROM scan GROUP BY netuid)
```

`count(*)` stays ungrouped: a plain row count carries no DISTINCT, so neither the budget rejection nor the wrong-question problem applies to it.

## Guard

A test asserting **no read in this loader uses `COUNT(DISTINCT)` at all**, mirroring the one added alongside the #9251 fix. Confirmed it bites by reintroducing the original query (`1 failed | 8 passed`).

## Worth recording

This class of failure is **invisible to the fake-engine tests**. The query typechecks, every local assertion passes, and it reads correctly — only a production probe after deploy shows the engine refusing it. Both instances of this bug (mine here, and #9251's) were found that way, not by CI.

## Registry Safety

- [x] Links a tracked, currently-open issue (`Closes #9281`) — required.
- [x] No schema change; `validate:contract-drift` green.

## Validation

- [x] `npm run typecheck` · `npm run lint` · `npm run format:check`
- [x] `npm run validate` · `npm run validate:contract-drift` · `npm run scan:public-safety`
- [x] `npm run worker:test`
- [x] `npx vitest run` — **15,087 tests passing**
- [x] Patch coverage: diff lines ∩ v8 uncovered set = **∅** (48 changed lines)

### DIFF
--- a/src/account-feeds-cold-tier.ts
+++ b/src/account-feeds-cold-tier.ts
@@ -804,38 +804,59 @@ export async function loadAccountSummaryColdTier(
     `FROM chain.account_events WHERE ${where} ` +
     `ORDER BY block_number DESC LIMIT ${ACCOUNT_EVENT_SUMMARY_SCAN_CAP}`;
 
-  const [aggRows, kindRows, probeRows, recentRows] = await Promise.all([
-    query(
-      env,
-      `WITH scan AS (${scan}) SELECT count(*) AS c, count(DISTINCT netuid) AS sc, ` +
-        `min(block_number) AS fb, max(block_number) AS lb, ` +
-        `min(observed_at) AS fo, max(observed_at) AS lo FROM scan`,
-    ),
-    query(
-      env,
-      `WITH scan AS (${scan}) SELECT event_kind AS kind, count(*) AS count ` +
-        `FROM scan GROUP BY event_kind`,
-    ),
-    // CAP + 1 so "exactly CAP" is distinguishable from "more than CAP".
-    query(
-      env,
-      `SELECT count(*) AS c FROM (SELECT block_number FROM chain.account_events ` +
-        `WHERE ${where} LIMIT ${ACCOUNT_EVENT_SUMMARY_SCAN_CAP + 1})`,
-    ),
-    query(
-      env,
-      `SELECT ${EVENT_COLUMNS} FROM chain.account_events WHERE ${where} ` +
-        `${FEED_ORDER} LIMIT ${limit}`,
-    ),
-  ]);
+  const [aggRows, subnetRows, kindRows, probeRows, recentRows] =
+    await Promise.all([
+      query(
+        env,
+        `WITH scan AS (${scan}) SELECT count(*) AS c, ` +
+          `min(block_number) AS fb, max(block_number) AS lb, ` +
+          `min(observed_at) AS fo, max(observed_at) AS lo FROM scan`,
+      ),
+      // subnet_count is its own GROUP BY subquery, NOT count(DISTINCT netuid).
+      // The engine rejects an ungrouped count(DISTINCT) at this scale -- the
+      // `40015: scan budget exceeded ... without GROUP BY` failure #9251 hit --
+      // and being inside a capped CTE does not save it, because the planner
+      // still evaluates the DISTINCT against the underlying scan. That
+      // rejection made this whole read decline, which is why the card shipped
+      // in #9254 still served zeros in production.
+      query(
+        env,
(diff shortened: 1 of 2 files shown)


JSONbored:
Superseded by #9282, which landed the same fix on main — the distinct subnet count is now its own `GROUP BY` subquery over the capped scan window (`src/account-feeds-cold-tier.ts`), so the ungrouped `count(DISTINCT netuid)` that R2 SQL rejects with `40015: scan budget exceeded` is gone.

Verified in production for the address this was reported against:

```
event_count   : 5000     (was 0)
subnet_count  : 2        (was 0)
registrations : 1        (netuid 46, uid 232 — matches /subnets exactly)
```

Three sessions independently found this same bug and wrote the same remedy today; #9287 was closed for the same reason. Nothing here is lost — the branch now conflicts against a main that already contains its change.

One piece of fallout worth noting: #9282 and #9285 were each green against the main they branched from, but together the test stubs in #9285 kept matching on `count(DISTINCT netuid)` — a marker #9282 had removed — which turned main red until #9295 fixed the stubs to match each read's actual clause.

Issue #9263, opened 2026-08-03
fix(accounts): the account summary serves empty event aggregates while /events serves 100 for the same account


JSONbored:
## Measured 2026-08-03

Same account (the chain's top extrinsic signer), two routes, same underlying stream:

| Route | Result |
|---|---|
| `/api/v1/accounts/{ss58}/events` | **events: 100** |
| `/api/v1/accounts/{ss58}` (summary) | **`recent_events: 0`, `event_kinds: 0`, `registrations: 0`, `labels: 0`** |

The account demonstrably has activity — `/extrinsics` returns 100, `/stake-flow` returns a subnet, `/portfolio` returns a position. The summary's aggregates are the only thing empty.

## Why it matters

The account page is usually the **first** thing a user opens for an address, and it currently claims the account has no event history and no event kinds. That is not a degraded answer, it is a **wrong** one: the data is right there, and the sibling route returns it.

`meta.source` reads `chain-events`, so the route believes it answered — it is not declining, it is reporting empty aggregates as fact.

## Likely cause

#9178 wired account-scoped cold-tier readers for `/transfers`, `/stake-flow`, `/stake-moves`, `/weight-setters` and `/counterparties`, but the **summary** handler's event-aggregation leg (`event_kinds`, `recent_events`, `registrations`) still reads the retired Postgres tier and falls through to a zeroed card rather than to the lakehouse.

## Requirements

1. Point the summary's aggregates at the **same** cold-tier reader `/accounts/{ss58}/events` already uses — one reader, one decoder, so the two routes cannot disagree about the same account again.
2. **The two routes must agree.** Add a test asserting that for an account with events, the summary's `recent_events` is non-empty and its `event_kinds` covers what `/events` returns. The defect here is precisely that two views of one stream disagreed and nothing caught it.
3. Keep decline-don't-widen: if the cold read fails, decline rather than emitting a zeroed card that reads as fact.
4. Check `registrations` and `labels` on the same route — `labels` comes from the entities artifact join and may be a separate miss.
5. 99% patch coverage branch-counted on `src/**`/`workers/**`.

Found during the post-port completeness audit, alongside #9260 (`/blocks/{n}/chain-events` empty for all 8.76M historical blocks). Both are the same class: **a route that reports empty as fact where a reader is simply missing.**

Refs #9146.

PR #9287, opened 2026-08-03
fix(accounts): keep the summary card's SQL inside R2 SQL's scan budget

### LINKED ISSUE #9286: fix(r2-sql): a rejected query logs only its status, discarding the engine's own explanation
## Measured in production 2026-08-03

The chain's top signer, `5Fv5t8frGG3MKtahp4WafKPmT5xZDbqWf8aFZpXyvjHTgzzx`, on three routes at the same moment:

| Route | Result |
|---|---|
| `/api/v1/accounts/{ss58}/events` | **5 events**, newest at block 8763529 (near head) |
| `/api/v1/accounts/{ss58}/subnets` | **1 registration** — netuid 46, uid 232 |
| `/api/v1/accounts/{ss58}` (the summary card) | `event_count: 0`, `subnet_count: 0`, `registrations: []`, `recent_events: []`, `event_kinds: []` |

`wrangler tail metagraphed` shows **exactly one** `[r2-sql] r2 sql: HTTP 422` per summary request — three of the four queries succeed, one is rejected, and `loadAccountSummaryColdTier` correctly declines as a whole rather than publishing a half-measured card. The route then serves the schema-stable zero card, which is indistinguishable from an account that has genuinely never done anything.

## Root cause, with in-repo precedent

`src/account-feeds-cold-tier.ts:810`:

```sql
WITH scan AS (...) SELECT count(*) AS c, count(DISTINCT netuid) AS sc, ... FROM scan
```

That is an **ungrouped `count(DISTINCT)`**, and this repo has already measured what R2 SQL does with one at this scale. From `src/chain-event-rollup-cold-tier.ts:313`:

> `40015: scan budget exceeded: scanning too much data for count(DISTINCT) without GROUP BY` — which made the whole rollup decline and served this route a card of zeros.

Same engine, same table (`chain.account_events`), same symptom, same wording. The rollup was fixed by rewriting the distinct count as a `GROUP BY` subquery; the account summary was never given the same treatment.

## Fix

Apply the established pattern: keep `count(*)` ungrouped (a plain row count has no distinct in it), and move the distinct netuid count into its own `GROUP BY` subquery, exactly as `chain-event-rollup-cold-tier.ts` does.

## The second half — this must not need a live tail again

`src/r2-sql.ts:135` throws `r2 sql: HTTP ${status}` and **discards the response body**. R2 SQL returns a numbered, self-explaining error (`40015: scan budget exceeded…`); throwing it away is what made a one-line diagnosis into a tail-plus-archaeology session. It should carry the engine's own message, bounded — the same change #9284 makes for the Analytics Engine client, which had the identical blindness and cost the identical investigation on the same day.

## Verification

Not "tests pass": `/api/v1/accounts/5Fv5t8frGG3MKtahp4WafKPmT5xZDbqWf8aFZpXyvjHTgzzx` must report a non-zero `event_count` and a `recent_events` list consistent with what `/events` returns for the same account.

Related: #9263 (the same card's `registrations` leg, fixed in #9285 via D1 — that fix is necessary but not sufficient, because the event half fails here in the lakehouse), #9284.

JSONbored:
## Summary

`/api/v1/accounts/{ss58}` served an **all-zero card for every account**. Measured today on the chain's top signer, `5Fv5t8frGG3MKtahp4WafKPmT5xZDbqWf8aFZpXyvjHTgzzx`, across three routes at the same moment:

| Route | Result |
|---|---|
| `/accounts/{ss58}/events` | **5 events**, newest at block 8763529 (near head) |
| `/accounts/{ss58}/subnets` | **1 registration** — netuid 46, uid 232 |
| `/accounts/{ss58}` | `event_count: 0`, `subnet_count: 0`, `registrations: []`, `recent_events: []` |

`wrangler tail` showed **exactly one** `[r2-sql] r2 sql: HTTP 422` per summary request — three of four reads fine, one refused. Declining as a whole is correct; the problem is that the route then publishes the schema-stable zero card, which is indistinguishable from an account that has genuinely never done anything.

## Root cause, with in-repo precedent rather than a guess

The refused read selected an **ungrouped `count(DISTINCT netuid)`**. `src/chain-event-rollup-cold-tier.ts:313` already records what R2 SQL does with one of those over `chain.account_events`:

> `40015: scan budget exceeded: scanning too much data for count(DISTINCT) without GROUP BY` — which made the whole rollup decline and **served this route a card of zeros**.

Same engine, same table, same symptom, same wording. The rollup was fixed by rewriting its distinct count as a `GROUP BY` subquery; the summary was never given the same treatment.

## The fix

The distinct netuid count becomes its own `GROUP BY` subquery over the **same capped scan window** (asserted in tests — reading a wider span would count subnets over more events than the card reports). `count(*)` stays ungrouped: a plain row count has no distinct in it, so the budget rule does not reach it. The count rides back inside `agg`, so the `sc` field callers already read is unchanged.

An empty subnet result is treated as **zero, not a decline** — an account with no events in the window has no netuid rows either, and that is a measured zero rather than a refused query. Keeping those distinct is the whole point of the bug.

## Why it needed a live tail, and what changes

`src/r2-sql.ts` threw `r2 sql: HTTP ${status}` on the non-2xx arm and **discarded the body**, while the `success: false` arm right beside it has always surfaced the engine's message. R2 SQL names and numbers what it refused; that asymmetry is what turned a one-line diagnosis into an investigation. The status now carries the engine's own explanation, bounded to 300 chars.

This is the same correction #9284 makes to the Analytics Engine client, which had the identical blindness and cost the identical investigation on the same day — two engines, two silent 422s, two routes quietly serving stale or empty data.

## Tests

`tests/account-summary-cold-tier.test.ts` (12 passed) — including a guard that **no** read asks for an ungrouped `count(DISTINCT)`, that the distinct count is a `GROUP BY` subquery over the capped window, and that an empty result is zero rather than a decline.

`tests/r2-sql.test.ts` (20 passed) — the engine's numbered message reaches the logged error; an empty or unreadable body still reports the status; a long body is bounded.

Patch coverage by **diff-intersection** (changed lines ∩ v8 uncovered set):

| file | added | uncovered | untaken branch arms |
|---|---:|---:|---:|
| `src/account-feeds-cold-tier.ts` | 23 | 0 | 0 |
| `src/r2-sql.ts` | 31 | 0 | 0 |
| **TOTAL** | **54** | **0** | **0** |

`npm run typecheck`, `eslint`, `prettier` clean. No schema change.

**Post-merge check that settles it:** `/api/v1/accounts/5Fv5t8frGG3MKtahp4WafKPmT5xZDbqWf8aFZpXyvjHTgzzx` must report a non-zero `event_count` consistent with what `/events` returns for the same account.

## Relationship to #9285

#9285 fixes the same card's `registrations` leg via D1 (#9263). That fix is necessary but **not sufficient** — the event half fails here, in the lakehouse, for a different reason. The two are independent and touch disjoint files.

Closes #9286

### DIFF
--- a/src/account-feeds-cold-tier.ts
+++ b/src/account-feeds-cold-tier.ts
@@ -804,41 +804,67 @@ export async function loadAccountSummaryColdTier(
     `FROM chain.account_events WHERE ${where} ` +
     `ORDER BY block_number DESC LIMIT ${ACCOUNT_EVENT_SUMMARY_SCAN_CAP}`;
 
-  const [aggRows, kindRows, probeRows, recentRows] = await Promise.all([
-    query(
-      env,
-      `WITH scan AS (${scan}) SELECT count(*) AS c, count(DISTINCT netuid) AS sc, ` +
-        `min(block_number) AS fb, max(block_number) AS lb, ` +
-        `min(observed_at) AS fo, max(observed_at) AS lo FROM scan`,
-    ),
-    query(
-      env,
-      `WITH scan AS (${scan}) SELECT event_kind AS kind, count(*) AS count ` +
-        `FROM scan GROUP BY event_kind`,
-    ),
-    // CAP + 1 so "exactly CAP" is distinguishable from "more than CAP".
-    query(
-      env,
-      `SELECT count(*) AS c FROM (SELECT block_number FROM chain.account_events ` +
-        `WHERE ${where} LIMIT ${ACCOUNT_EVENT_SUMMARY_SCAN_CAP + 1})`,
-    ),
-    query(
-      env,
-      `SELECT ${EVENT_COLUMNS} FROM chain.account_events WHERE ${where} ` +
-        `${FEED_ORDER} LIMIT ${limit}`,
-    ),
-  ]);
+  const [aggRows, subnetRows, kindRows, probeRows, recentRows] =
+    await Promise.all([
+      query(
+        env,
+        `WITH scan AS (${scan}) SELECT count(*) AS c, ` +
+          `min(block_number) AS fb, max(block_number) AS lb, ` +
+          `min(observed_at) AS fo, max(observed_at) AS lo FROM scan`,
+      ),
+      // The distinct netuid count, as a GROUP BY subquery rather than the
+      // ungrouped `count(DISTINCT netuid)` this used to select alongside the
+      // aggregates above.
+      //
+      // R2 SQL REJECTS the ungrouped form over this table -- `40015: scan
+      // budget exceeded: scanning too much data for count(DISTINCT) without
+      // GROUP BY` -- and because one rejected query declines the whole loader,
+      // the summary card served zeros for every account, including accounts
+      // whose own /events route was returning rows from the same table at the
+      // same moment. src/chain-event-rollup-cold-tier.ts hit this first and
+      // this is its remedy, applied here rather than rediscovered.
+      //
+      // `count(*)` above stays ungrouped: a plain row count has no distinct in
+      // it, so the budget rule does not apply to it.
+      query(
+        env,
+        `WITH scan AS (${scan}) SELECT count(*) AS sc ` +
+          `FROM (SELECT netuid FROM scan GROUP BY netuid)`,
+      ),
+      query(
+        env,
+        `WITH scan AS (${scan}) SELECT event_kind AS kind, count(*) AS count ` +
+          `FROM scan GROUP BY event_kind`,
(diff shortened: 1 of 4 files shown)


JSONbored:
Superseded. The query fix in this PR duplicates #9282, which landed on main independently and is verified working in production (`event_count` 0 -> 5000 for the address in #9286).

Reopening the still-needed half — the R2 SQL client discarding the engine's rejection message — as a focused PR.

PR #9422, opened 2026-08-04
feat(testnet): serve the thirteen projection-backed chain routes on testnet

### LINKED ISSUE #9412: feat(testnet): network-dimension the projection lanes — 13 chain-data route families are gated to mainnet only by a hardcoded FROM clause
Part of #8697. The last mainnet-only chain-data family after #9411.

Thirteen route families are still gated to mainnet purely because the projection lanes that feed
them hardcode `FROM chain.*`. Every one of them reads `blocks`, `extrinsics` or `account_events` —
tables the decode lane already fills for **both** networks (`chain_testnet` is decoded through
7,700,842 and advancing hourly). Nothing about these routes is mainnet-specific except the
namespace in their `FROM` clause.

## What is gated, and why

| route family | lane reads |
| --- | --- |
| `/blocks/summary` | `chain.blocks` |
| `/chain/registrations`, `/chain/deregistrations` | `chain.account_events` |
| `/chain/transfers`, `/chain/transfer-pairs` | `chain.account_events` |
| `/chain/stake-flow`, `/chain/stake-moves`, `/chain/stake-transfers` | `chain.account_events` |
| `/chain/alpha-volume` | `chain.account_events` |
| `/chain/activity`, `/chain/calls`, `/chain/fees`, `/chain/signers` | `chain.extrinsics` / `chain.blocks` |

All 13 are in `MAINNET_ONLY_ROUTE_PATHS`, so a testnet caller gets the documented 404 rather than
an empty — correct today, and exactly the thing to fix rather than work around.

## Why a projection and not a request-time reader

`src/blocks-summary-artifact.ts`'s header states the rule: request-time scans of the large chain
tables are the wrong shape for R2 SQL (~1-2s/query, no indexes), and scheduled projections are the
right one. So this is **not** "add a cold-tier reader per route" — it is giving the existing lanes a
network dimension and letting the same cron fill both.

## Scope

- `ProjectionLane.compute(env)` becomes `compute(env, network)`, and every `FROM chain.X` goes
  through `chainTable("X", network)` — the function that already exists for exactly this, so a lane
  cannot be network-aware in its `WHERE` clause and mainnet-only in its `FROM`.
- Artifact keys get network-scoped the way `networkKvKey` scopes KV: **mainnet unchanged**
  (`metagraph/projections/...`), testnet prefixed. Every already-written artifact stays readable and
  a mainnet request keeps hitting the exact key it hits today.
- `runProjectionLanes` iterates networks × lanes. A network whose lakehouse cannot answer must not
  write a zeroed card over a good one — the all-or-nothing contract each lane already states.
- The 14 artifact readers take the network; the router dispatches these routes with it, from a
  shared table like `dispatchChainHistoryRoute` rather than a second copy of the branches.
- The 13 come off `MAINNET_ONLY_ROUTE_PATHS` only once their testnet artifact is **verified
  written** — a route answering an empty forever reads as broken rather than unsupported, which is
  the rule #8700 set for itself.

## Cost

Doubling the lane count doubles the cron's R2 SQL work in principle, but `chain_testnet` currently
holds ~840 blocks against mainnet's 8.77M, and R2 SQL bills on bytes scanned. The testnet half is
rounding error until testnet coverage grows — at which point it is worth re-measuring per lane the
way #9411's feed window was.

JSONbored:
Closes #9412.

The reader and route half. #9415 gave the **lanes** a network dimension, so
`metagraph/projections/testnet/*` is being written; this serves it, which closes the issue.

Thirteen route families were gated to mainnet for exactly one reason: the projection lanes that
feed them hardcoded `FROM chain.*`. Every one reads `blocks`, `extrinsics` or `account_events` —
tables the decode lane already fills for both chains. Nothing about them was mainnet-specific except
a namespace and, now, a key.

Now servable on testnet: `/blocks/summary`, `/chain/activity`, `/chain/alpha-volume`,
`/chain/calls`, `/chain/deregistrations`, `/chain/fees`, `/chain/registrations`, `/chain/signers`,
`/chain/stake-flow`, `/chain/stake-moves`, `/chain/stake-transfers`, `/chain/transfer-pairs`,
`/chain/transfers`.

## What changed

- **The 15 artifact readers take the network** and resolve their key through `projectionKey()`.
  **Mainnet is unchanged** — every reader asks for the exact object it asked for before, asserted
  per reader in both directions (it reaches for testnet's key off mainnet, and never for
  mainnet's).
- **`dispatchProjectionRoute` is the third shared table**, beside the live-chain and chain-history
  ones, so the bare path and the `/{network}/` path resolve to the same handler with a different
  network and nothing else. It replaces thirteen hand-written branches on the bare path.
  `/blocks/summary` moves out of the history table into it — the comment there said opening it
  needed the network threaded *first*, which is what this does.
- **The edge cache is scoped per network.** The `/{network}/` prefix is stripped before dispatch,
  so both chains reached these handlers with byte-identical paths — and their card shapes are
  identical too, so a testnet card served to a mainnet caller would have looked completely
  well-formed. `edgeCacheScope` leaves mainnet's label bare so warm entries survive.
- **`PROJECTION_ROUTE_PATHS` is the third derived list**, feeding the capability matrix.
  `tests/projection-networks.test.ts` holds it and the router together in both directions, and
  asserts one route per lane so a lane cannot fill an artifact nobody serves.

## Two things worth a second look

**An import cycle, found the hard way.** `projectionKey` started life in `projection-lanes.ts`,
which imports every reader's key constant — so a reader importing it back formed a cycle that
surfaced as `artifactKey` reading `undefined` at module init, not as an error. It now lives in
`chain-network.ts` beside `networkKvKey` and `chainTable`, which is where it belonged anyway: the
same scoping rule for the third store.

**Three test files used a route this change opens as their "still gated" example.** Each moved to
`/api/v1/chain/weights`, which reads the neurons tier and has no testnet twin — the example moved,
the property did not. Worth confirming you agree those tests still guard what they were written to
guard.

## Verification

- `npx vitest run` — **15,936 passed**.
- `validate:openapi` now covers **106 network variants** (was 92); `validate:contract-drift`,
  `validate:types`, `validate:ui-docs-drift` (343 pages), `validate:mcp` (214 tools),
  `validate:openapi-examples` (189/189) all green.
- Patch coverage **37/37 lines, 70/70 branches**, by diff intersection against the lcov report.
- Every new assertion is paired with its positive — the reader actually read, the query was
  actually issued — because "it does not reach for mainnet's key" passes perfectly on a reader that
  reaches for nothing.

## After this merges

The lanes have been filling `metagraph/projections/testnet/*` since #9415 (cron `11,41 * * * *`).
I'll confirm each of the thirteen serves real testnet numbers rather than its zeroed floor once
this deploys, and report anything that comes back empty.


JSONbored:
**Premise verified in production.** All thirteen testnet projection artifacts are written and fresh, measured at 16:45–16:46 UTC on the first lane tick after #9415 deployed:

| artifact | generated_at | rows |
| --- | --- | --- |
| blocks-summary | 16:45:09 | 1963 (blocks 7,700,000 → 7,701,962) |
| chain-activity | 16:45:44 | 4 |
| chain-alpha-volume | 16:46:25 | 0 |
| chain-calls | 16:45:58 | 60 |
| chain-deregistrations | 16:45:25 | 0 |
| chain-fees | 16:46:08 | 158 |
| chain-registrations | 16:45:11 | 2 |
| chain-signers | 16:46:19 | 312 |
| chain-stake-flow | 16:45:41 | 42 |
| chain-stake-moves | 16:46:37 | 0 |
| chain-stake-transfers | 16:46:26 | 8 |
| chain-transfer-pairs | 16:46:31 | 400 |
| chain-transfers | 16:45:28 | 400 |

The three zeros are *measured* zeros, not failures: a lane that cannot read declines and writes nothing at all, so a written artifact with `row_count: 0` means the query ran and the window genuinely holds no rows — expected on a chain with ~5.5 hours of decoded history.

Unrelated to this PR, the run surfaced a **pre-existing mainnet defect**, filed separately: `chain-stake-moves` and `chain-stake-transfers` have not been rewritten since 2026-08-03T09:13, and `/api/v1/chain/stake-moves` is publishing `newest_observed: 2026-08-02T22:08` under a `7d` label with no degradation marker. Notably testnet's copy of the same lane wrote fine here — same query shape, small table — which points at scan volume rather than syntax.

Issue #9455, opened 2026-08-05
mcp: swept all 215 tools — three published fields that can never carry a value, one 74h-stale snapshot


JSONbored:
## What was done

Every registered MCP tool was called at least once against production (`https://api.metagraph.sh/mcp`), driven by each tool's own `inputSchema.required` against a fixture map:

```
# summary: ok=191 errored=10 untestable=14 total=215
```

The 10 errors and 14 untestable are sweep-harness limitations, not defects — wrong fixture shapes (`get_api_schema` with a subnet slug), correctly-rejected inputs (`get_webhook_subscription` wants a UUID v4), correct auth refusals (`list_surface_credentials` → `auth_required`), and tools needing fixtures the harness has no safe value for (`store_surface_credential`, `call_rpc`, `decode_evm_call`).

Below are the findings that survived verification. Each is a **published field that can never carry a value**, or a snapshot far older than its tool implies — the class the `alpha_price_change_1d` bug (#9449) belonged to.

## 1. `get_top_holders` is 74 hours stale, and its net-flow fields are always null

```
captured_at = 2026-08-02T00:05:06.441Z   (age 74.2h at time of check)
account_count = 2965
net_flow_7d / net_flow_30d / net_flow_90d = null on 20/20 returned accounts
```

Two separate problems in one response: the snapshot itself has not refreshed in three days, and three declared fields are null for every account. Nothing in the tool description says the data is a multi-day-old snapshot, so a caller reasonably reads it as current.

## 2. `list_blocks.author` is always null — while `get_blocks_summary` computes author decentralisation from the same data

```
list_blocks(limit=5)  -> author null 5/5, spec_version null 2/5
get_blocks_summary()  -> {"holders":20,"total":5000,"gini":0.0014,
                          "nakamoto_coefficient":10, ...}
```

This is the strongest signal in the sweep, because the two disagree about whether the data exists. `get_blocks_summary` reports author concentration over 5,000 blocks across 20 distinct authors — so authors ARE recorded. `list_blocks` returns `null` for every one of them.

Note `spec_version` is null on only 2 of 5, so that column is partially populated; `author` is not partially populated, it is uniformly absent. Whatever `list_blocks` projects from, it is not the source `get_blocks_summary` reads.

## 3. `list_global_validators.realized_return_1m` is null for every validator

```
list_global_validators(limit=50) -> realized_return_1m null on 50/50
```

[...]

Issue #9193, opened 2026-08-03
chore(workers): delete the dead Postgres tier now that no Hyperdrive binding exists


JSONbored:
Data residency follow-up: **all data now lives in Cloudflare** (D1, R2/R2-SQL lakehouse, KV) and every refresh lane runs on Cloudflare compute (Worker crons + the poller Container). What is left is dead code, not a dead data path.

## State

`#9186` removed the `HYPERDRIVE` binding from `wrangler.data.jsonc`, and no wrangler config declares one anymore. But two Workers still import the Postgres client and still carry their full Postgres branches:

- `workers/data-api.ts` — 13,595 lines, **27 `HYPERDRIVE` gate sites**, `import postgres from "postgres"`
- `workers/hyperdrive-sync-retry.ts` — 70 lines, only caller is data-api

Every one of those branches gates on `env.HYPERDRIVE?.connectionString`, which is now permanently `undefined`, so they are unreachable. They were deliberately left in place at the wipe as a single-flag revert path (`#9145`), and that revert path is now moot — the Postgres they would revert to was destroyed.

## Why it is worth doing

- The `postgres` driver ships in two Worker bundles that can never call it.
- 27 dead gate sites are 27 places a future reader has to reason about a tier that does not exist, and 27 places a new dual-write can be written against the wrong store by pattern-matching.
- Coverage and tests currently carry the dead branches too.

## Requirements

1. Delete the Postgres branches and the `postgres` import from both Workers; delete `hyperdrive-sync-retry.ts` if nothing survives it.
2. Where a handler was a **dual-write** (D1 + Postgres), the D1 path becomes the only path — keep its existing failure semantics (D1 required, 503 unbound / 502 on failure with the capture labels), do not loosen them.
3. Where a read did `tryPostgresTier` → cold tier, the forward gate and `DATA_API_D1_FLAGS` behaviour must not change: routes still decline to the lakehouse/D1 tiers exactly as today. This is a deletion, not a re-plumb.
4. Drop the now-unreachable `"postgres"` values from the `METAGRAPH_*_SOURCE` flag unions if nothing can set them anymore, and say so in the PR.
5. Full suite green; `codecov/patch` 99% branch-counted on whatever lines do change.

**Sequencing:** do this when no other PR is editing `workers/data-api.ts` — it is the most-touched file in the repo and this diff is large by construction.

Refs #9146.

PR #9463, opened 2026-08-05
fix(api): withdraw the top-holders net-flow fields whose rollup left with Postgres

### LINKED ISSUE #9461: top-holders: withdraw net_flow_7d/30d/90d and their sort keys — wallet_flow_daily did not survive the Postgres removal
Split out of #9455, which carries three findings. This issue is only the first
one, and only the half that is actionable now: withdrawing the dead fields.
Rebuilding the rollup is deliberately **not** in scope here.

## The fields cannot be filled, and the sorts are worse than the fields

`/api/v1/accounts/top-holders`, `get_top_holders` and the GraphQL `topHolders`
field publish `net_flow_7d`, `net_flow_30d` and `net_flow_90d`, and accept
those same three names as `?sort=` keys.

`src/top-holders.ts:19-26` names the source: `wallet_flow_daily`, a daily
coldkey-keyed rollup of `account_events`. That table was **Postgres**.
`workers/data-api.ts:995-998` writes the rollup handler up in the past tense —
"RETIRED (#9193): the Postgres tables this wrote were destroyed with the box …
What follows describes what it DID". There is no D1 migration for it:
`migrations/d1/` has no `wallet_flow_daily`, and on the production D1

```sql
SELECT name FROM sqlite_master WHERE type='table' AND name LIKE '%flow%'
-- emission_flow_watch      (the only match)
```

So the fields are advertised, sortable, and structurally unfillable.

## The sort options are the real defect

A null column does not fail loudly when you sort by it. `compareTopHoldersSort`
places non-numbers last and tiebreaks by `ss58`, so *every* row ties and the
whole "ranking" collapses to lexicographic ss58 order. Live, right now:

```
get_top_holders(sort=net_flow_30d, limit=5)
  5C4jr9gTxHFbDycY8JvP5tSiVgqkGdhkKgYCGBZ7r5jcy3D4
  5C4stSNCjgyxqbTJa7TTCFFRAJxKrkKhvfnHNs5BT23TZEqz
  5C4zv89Vra1KpRxyaTMVtwQvDNLLSKWCD1K6qFB2mG5hgr2H
  5C523K1nqhGWUrpvRtJxSCdiWLiRxR1RefjtcHbQc8fQ84hk
  5C5HuoidEheJizKJYQmo3HxGA2vVnZjYWPgWwNZ4MQudEKAR
```

That is sorted by ss58, not by flow, and the response says `"sort":
"net_flow_30d"` while doing it. A caller cannot tell a real flow ranking from
this. A null field is merely useless; a sort key that silently answers with a
different ordering than the one it names is wrong.

## Scope

Withdraw the three fields and the three sort options. Nothing else.

There are several **unrelated and working** `net_flow` fields in this repo —
`/subnets/{netuid}/stake-flow`, `/accounts/{ss58}/stake-flow`,
`/chain/stake-flow`, and the alpha-volume surfaces. Those are sourced from the
lakehouse, not `wallet_flow_daily`, and must not be touched. Grep for the
suffixed names (`net_flow_7d` / `net_flow_30d` / `net_flow_90d`), never bare
`net_flow`.

Touch points:

- `schemas-src/routes/top-holders.ts` — response schema + sort enum
- `schemas-src/mcp-tools/accounts-leaderboards.ts` — MCP tool schema
- `src/top-holders.ts` — `TOP_HOLDERS_SORTS`, the row mapper, the signed-number
  guard that exists only for these fields, the header comment
- `workers/request-handlers/entities.ts` — `TOP_HOLDERS_CSV_COLUMNS`
- `src/contracts.ts` — the artifact + route descriptions
- `src/mcp-server.ts` — the tool description
- `docs/backend-artifact-contracts.md`
- Regenerate `openapi.json` / types / `packages/contract` (Path B), plus
  `apps/ui/content/docs/api-reference/accounts/top-holders.mdx` from `apps/ui`

## Is this breaking?

`docs/api-stability.md` says existing fields are not removed within `v1`. The
precedent says the rule is about fields that carry data: #9066 dropped
`unpriced_stake_alpha` / `unpriced_emission_alpha` from five route schemas
outright, in `v1`, with no `CONTRACT_VERSION` bump and no deprecation window,
on exactly this reasoning — publishing our own incompleteness is not a contract
worth keeping. A field that has only ever been `null` and can only ever be
`null` is the same case: removing it removes `null`, not data.

The `?sort=` enum is the one place a caller can observe a behaviour change —
`?sort=net_flow_7d` goes from 200-with-ss58-order to `400 invalid_query`. That
is the correct direction: a request whose stated intent cannot be honoured
should be refused, not answered with a different ordering. Note the published
OpenAPI **parameter** enum (`src/contracts.ts`) already lists only
`total_tao|free_tao|delegated_tao`, so a client generated from `openapi.json`
could not ask for these sorts in the first place.

## Not in scope

The underlying events survive in the lakehouse (`chain.account_events`), so the
rollup is rebuildable — as a feature project with real R2 SQL scan cost to size
first, not as part of this withdrawal. #9455 stays open for the other findings
it carries.

JSONbored:
Closes #9461

`net_flow_7d`, `net_flow_30d` and `net_flow_90d` came from `wallet_flow_daily`
(`src/top-holders.ts:19-26`), a daily coldkey-keyed rollup of `account_events`.
That table was **Postgres**. It went with the box, and nothing replaced it —
`migrations/d1/` has no `wallet_flow_daily`, and on the production D1:

```sql
SELECT name FROM sqlite_master WHERE type='table' AND name LIKE '%flow%'
-- emission_flow_watch      (the only match)
```

`workers/data-api.ts` already says as much about the rollup that fed it, in the
past tense: "RETIRED (#9193): the Postgres tables this wrote were destroyed with
the box … What follows describes what it DID."

So the three fields have been advertised, sortable, and structurally unfillable.

## The sort keys were the actual defect

A null field is useless. A **sort key** over an all-null column is wrong,
because it does not fail — `compareTopHoldersSort` placed non-numbers last and
tiebroke by `ss58`, so every row tied and the ranking collapsed to lexicographic
ss58 order while the response still echoed `"sort": "net_flow_30d"`. Live before
this change:

```
get_top_holders(sort=net_flow_30d, limit=5)
  5C4jr9gTxHFbDycY8JvP5tSiVgqkGdhkKgYCGBZ7r5jcy3D4
  5C4stSNCjgyxqbTJa7TTCFFRAJxKrkKhvfnHNs5BT23TZEqz
  5C4zv89Vra1KpRxyaTMVtwQvDNLLSKWCD1K6qFB2mG5hgr2H
  5C523K1nqhGWUrpvRtJxSCdiWLiRxR1RefjtcHbQc8fQ84hk
  5C5HuoidEheJizKJYQmo3HxGA2vVnZjYWPgWwNZ4MQudEKAR
```

Sorted by ss58, offered as a flow ranking. `?sort=net_flow_7d` now answers
`400 invalid_query` instead — refusing an intent we cannot honour is the right
answer, and it is the only externally visible behaviour change here.

## Is removing a response field breaking?

**No, and it ships in `v1` — with precedent.**

`docs/api-stability.md` says "Existing fields are not removed or retyped …
within `v1`." That rule is about fields that carry data. #9066 removed
`unpriced_stake_alpha` / `unpriced_emission_alpha` from five route schemas
outright, inside `v1`, with no `CONTRACT_VERSION` bump, no `schema_version`
bump, no `/api/v2` and no deprecation window, on exactly this reasoning:
publishing our own incompleteness is not a contract worth honouring. This is the
same case, only stronger — those fields could at least in principle hold a
number. These three could only ever be `null`, so removing them removes `null`,
not data. No consumer can have a working code path that reads them.

Accordingly: no `CONTRACT_VERSION` bump (it has not moved for any of the
comparable removals), no `schema_version` bump (that is reserved for a breaking
*envelope* change, and the envelope is untouched), no `v2`.

One detail worth recording: the published OpenAPI **query parameter** enum in
`src/contracts.ts` already listed only `total_tao|free_tao|delegated_tao`, so a
client generated from `openapi.json` could never have asked for these sorts. The
`net_flow_*` sorts were reachable only via the MCP tool schema and a hand-built
REST URL. This PR makes the parameter enum, the response-body `sort` enum, the
MCP tool schema and `TOP_HOLDERS_SORTS` all agree on three — they did not agree
before.

## Scope

Only the top-holders fields. This repo has several **working** `net_flow`
fields — `/subnets/{netuid}/stake-flow`, `/accounts/{ss58}/stake-flow`,
`/chain/stake-flow`, alpha-volume — all lakehouse-sourced and untouched. Every
edit here was found by grepping the suffixed names, never bare `net_flow`.

- `schemas-src/routes/top-holders.ts`, `schemas-src/mcp-tools/accounts-leaderboards.ts` — the two Zod owners
- `src/top-holders.ts` — `TOP_HOLDERS_SORTS`, the entry type + row mapper, and `numberOrNullSigned`, the signed-number guard that existed only for these fields
- `workers/request-handlers/entities.ts` — `TOP_HOLDERS_CSV_COLUMNS`
- `src/contracts.ts`, `src/mcp-server.ts`, `docs/backend-artifact-contracts.md` — the prose that documented them
- regenerated `openapi.json`, `types.d.ts`, `api-index.json`, `contracts.json`, `packages/contract`, and `apps/ui/content/docs/api-reference/accounts/top-holders.mdx`

`compareTopHoldersSort` loses its null-last branch. All three surviving sort
keys come out of `numberOrZero` (or their sum), so both sides are always finite
numbers — the branch was unreachable the moment the nullable keys left, and its
own comment named the fields that justified it. Keeping it would have meant
keeping tests that construct states the mapper cannot produce.

## Not in scope

The events still exist in the lakehouse (`chain.account_events`), so the rollup
is rebuildable — as its own project with real R2 SQL scan cost to size first.
#9455 stays open for the other findings it carries.

## Verification

`npm run lint`, `format:check`, `typecheck`, `validate`, `validate:contract-drift`,
`validate:openapi`, `validate:types`, `validate:mcp`, `validate:schema-enums`,
`validate:docs`, `validate:generated-client`, `validate:client-sdk-sync`,
`validate:openapi-examples`, `validate:ui-docs-drift`, `validate:api`,
`validate:artifact-budgets`, `validate:surface` — all green. Full `vitest run`
green.

New assertions pin the withdrawal rather than just deleting the old ones: an
entry never carries the keys even when the source row still has them, and both
REST and MCP reject each withdrawn sort instead of answering in ss58 order.


JSONbored:
Superseded. Withdrawing the fields was the wrong call: `chain.account_events` is current to the chain head and carries coldkey/event_kind/amount_tao, and the StakeAdded-minus-StakeRemoved computation already runs live against it (`/chain/stake-flow`, `/accounts/{ss58}/stake-flow`). What died is the precomputed `wallet_flow_daily` table, not the data or the math — "structurally unfillable" in the description above is wrong.

`src/account-history-cold-tier.ts` is the precedent that settles it: it faced the identical loss of `account_events_daily` and recomputed live from `chain.account_events` rather than withdrawing, on the grounds that "a live answer beats a smaller diff".

The route's holdings columns are on the same footing — they are served from a frozen artifact of the decommissioned box (`captured_at` 2026-08-02), which is #9455's staleness finding. So this is one job, not two: give `/api/v1/accounts/top-holders` a live source for all of it. Tracking that in a new issue.

Issue #7510, opened 2026-07-21
Epic: TypeScript migration — eliminate hand-maintained .mjs/.js across the backend/tooling stack


JSONbored:
## Context

The repo is split in half. `apps/ui`, `packages/ui-kit`, and `packages/client` are already 100% TypeScript, each with its own `tsconfig.json` and a pinned `typescript` devDependency. Everything else — the actual backend — is not:

| Area | Files | LOC (approx) |
|---|---|---|
| `src/` | 171 `.mjs`, 0 `.ts` | 71,400 |
| `workers/` | 25 `.mjs` | 28,700 |
| `scripts/` (+ `scripts/lib/`) | 109 + 11 `.mjs` | 38,800 |
| `tests/` | 471 `.mjs` | 205,000 |
| `deploy/wss-lb/` | 11 `.mjs` | 1,600 |

811 hand-maintained `.mjs` files, ~346k lines, with **zero static type checking today**. There is no root `tsconfig.json`, no `typescript` devDependency at the root, and the root ESLint config ([eslint.config.mjs](../eslint.config.mjs)) is deliberately JS-only. `npm run validate:types` does not type-check anything — it only regenerates the OpenAPI `.d.ts` and diffs it for drift.

## Goal

This is not an incremental JSDoc bolt-on. The goal is **full elimination** of hand-maintained `.mjs`/`.js` in this codebase in favor of `.ts`, so that:
- Contributors default to `.ts` for all new work going forward — no more hand-maintained type drift, no more silently-wrong shapes.
- The compiler catches what ~338k lines of untyped JS currently rely on runtime tests and manual review to catch.
- LOC/file count actually goes **down** — JSDoc blocks, manual runtime shape-guards, and redundant validation collapse once real types exist and the compiler enforces them.

## Explicit non-goal: generated artifacts stay generated

Do **not** hand-convert or otherwise touch:
- `packages/contract/index.d.ts`
- `public/metagraph/types.d.ts`
- `generated/metagraphed-client.ts`
- anything under `registry/**/generated/**` or `packages/client/dist/**`

These are `openapi-typescript`/build output, already correct-by-construction. Their **generator scripts** (e.g. `scripts/generate-types.mjs`) are in scope for conversion like any other script — their **output** is never hand-edited, migration or not.

## Why this is unusually tractable here (verified, not assumed)

[...]

PR #9468, opened 2026-08-05
chore(scripts): commit the live MCP tool smoke-sweep as npm run smoke:mcp

### LINKED ISSUE #9462: mcp: commit the live tool smoke-sweep harness as a script
The tool sweep behind #9455 — call every registered MCP tool, then look at the
*shape* of what each one returned — found two real defects that hand-testing had
missed (`list_blocks.author` null while `get_blocks_summary` computes author
decentralisation from the same data; `list_global_validators.realized_return_1m`
null since Postgres was removed in #9429).

It only ever existed as a scratch file. That means the next person who wants to
ask "does every tool still return something?" has to rewrite it, and the
coverage it established rots silently as tools are added.

## What to build

`scripts/mcp-smoke-sweep.ts`, wired to `npm run smoke:mcp`:

1. `tools/list` against the MCP endpoint (default `https://api.metagraph.sh/mcp`,
   overridable). The endpoint may answer as SSE — parse the last `data:` line.
2. Per tool, build arguments from its **own** `inputSchema.required` against a
   fixture map, rather than a hand-maintained per-tool argument table.
3. A required field with no fixture reports **UNTESTED** — never a silent skip.
   That is the part that stops coverage rotting as tools are added.
4. `tools/call`, classified as ERROR (`isError`, with the structured error code),
   RPCERR (JSON-RPC envelope), or OK.
5. On OK, walk for the first array-of-objects and flag `EMPTY`,
   `ALL_NULL(<fields>)` (every row null), `UNIFORM(<fields>)` (one value shared
   across >= 5 rows).

## Constraints

- **The output is a triage list, not a verdict.** The heuristic is deliberately
  noisy — `UNIFORM(netuid)` on a tool filtered *by* netuid is expected. This has
  to be documented in the script itself, because treating it as a verdict already
  produced two false bug reports during the original run.
- **Never fails on a flag.** Non-zero exit on transport failure only; a noisy
  heuristic must not be able to red-line anything.
- **Not in CI.** One run is ~215 live production calls — explicit opt-in
  invocation, never something a PR triggers.
- `.ts`, not `.mjs`/`.js` (`validate:no-hand-written-mjs`).
- `scripts/` sits outside `codecov/patch`'s scope, but the pure helpers (the
  flagging heuristic, the argument builder, the SSE parser) carry the real logic
  and need unit tests.

JSONbored:
## Summary

The tool sweep behind #9455 found two real defects that hand-testing had missed, and then existed only as a scratch file. This commits it as `scripts/mcp-smoke-sweep.ts`, wired to `npm run smoke:mcp`.

It enumerates via `tools/list` (parsing the last `data:` line when the endpoint answers as SSE), builds each call from the tool's **own** `inputSchema.required` against a fixture map, classifies the result (OK / ERROR / RPCERR), and on OK walks for the first array-of-objects to flag `EMPTY`, `ALL_NULL(...)`, `UNIFORM(...)`.

A required field with no fixture reports **UNTESTED** rather than being skipped — that is the part that stops coverage rotting as tools are added.

**The output is a triage list, not a verdict.** The heuristic is deliberately noisy (`UNIFORM(netuid)` on a tool filtered *by* netuid is expected), and treating it as a verdict already produced two false bug reports during the original run. That framing is documented in the file header, in `--help`, and printed under every run. A SUSPECT flag can never fail the command — only a transport failure sets a non-zero exit code.

Not wired into CI: one run is ~215 live production calls, so it stays an explicit opt-in invocation.

## Two things the first production run forced

Running it against production was the only way to find these, and both are the kind of defect that would have made the harness untrustworthy:

- **Serial by default.** At concurrency 2, the endpoint's per-client rate limit turned **34 healthy tools** into `-32600 Too many MCP requests` — a parallel sweep manufacturing failures indistinguishable from real ones. Throttled calls are now retried with backoff rather than reported, and `isRateLimited` matches on both code *and* message so a genuine `-32600` still surfaces.
- **Per-tool fixture overrides.** `ref` is a block ref to `get_block` but a composite `<block>-<index>` to `get_extrinsic` — sent the wrong one, `get_extrinsic` returned an empty row set and the heuristic flagged a perfectly healthy tool `EMPTY`. A fixture bug wearing a defect's clothes. `slug` (provider vs adapter) and `query` (search term vs GraphQL) are polymorphic the same way.

## Verification against production

```
# summary: ok=205 errored=5 untestable=5 total=215 suspect=97   (exit 0)
```

Better than the prototype's 191/10/14, and the exit-code contract holds with 97 flags outstanding. The remaining 5 errors are all legitimate: 2 correct `auth_required` refusals on the credential store, 2 tools whose real contract is "one of these two" but declare no `required` field (a documented limit of driving the sweep off `required`), and 1 `invalid_graphql_query`. The 5 UNTESTED are the fields with no safe value — `credential`, `owner_token`, `method`, `to`/`input`, and a subscription UUID — deliberately withheld rather than invented.

All three findings from #9455 reproduce:

```
OK  list_blocks             rows=50  SUSPECT ALL_NULL(author)
OK  get_top_holders         rows=20  SUSPECT ALL_NULL(net_flow_7d,net_flow_30d,net_flow_90d)
OK  list_global_validators  rows=20  SUSPECT ALL_NULL(realized_return_1m)
```

## Tests

`scripts/**` is outside `codecov/patch`'s scope, but the pure helpers carry the real logic, so they get unit tests: the flagging heuristic, the argument builder, fixture resolution, the SSE parser, rate-limit detection, and the worker pool — 46 tests in `tests/mcp-smoke-sweep.test.ts`. Importing the module never hits the network (direct-execution guard, same pattern as `smoke-live-api.ts`).

## Validation

- `npm run lint` — clean
- `npm run format:check` — clean
- `npm run typecheck` — clean
- `npx vitest run` — 670 files / 15,985 tests passed
- `npm run validate:no-hand-written-mjs` — clean (`.ts`, per #7510)
- `npm run smoke:mcp` — full run against production, exit 0

Note the full suite needs `npm run build` first; without it 139 reader tests fail on served artifacts that only exist after a build.

Closes #9462

### DIFF
--- a/package.json
+++ b/package.json
@@ -125,6 +125,7 @@
     "deploy:wss-lb": "scripts/deploy-worker-with-sourcemaps.sh wrangler.wss-lb.jsonc",
     "deploy:wss-lb:preview": "scripts/deploy-worker-with-sourcemaps.sh wrangler.wss-lb.jsonc --preview",
     "smoke:live": "node scripts/smoke-live-api.ts",
+    "smoke:mcp": "node scripts/mcp-smoke-sweep.ts",
     "lint": "eslint .",
     "format:check": "prettier --check .",
     "pretypecheck": "npm run build --workspace=packages/chain-summaries",
--- a/scripts/mcp-smoke-sweep.ts
+++ b/scripts/mcp-smoke-sweep.ts
@@ -96,7 +96,7 @@ export function buildFixtures(now: Date = new Date()): Record<string, unknown> {
     netuid: 64,
     netuids: [7, 8],
     uid: 0,
-    // Top-of-leaderboard coldkey -- an account with balance, stake positions and
+    // Top-of-leaderboard account -- carries balance, stake positions and
     // history, so account tools return populated rows rather than a valid-but-
     // empty account.
     ss58: "5EYCAe5jLQhn6ofDSvqF6iY53erXNkwhyE1aCEgvi1NNs91F",
--- a/scripts/mcp-smoke-sweep.ts
+++ b/scripts/mcp-smoke-sweep.ts
@@ -1,0 +1,617 @@
+// Live smoke sweep across every registered MCP tool (#9455).
+//
+// Enumerates tools from the server's own `tools/list`, builds arguments for each
+// tool from its own `inputSchema.required` against a fixture map, calls it, and
+// reports the shape of what came back. It exists to catch the class of defect
+// hand-testing misses: a published field that can never carry a value, or a
+// snapshot far older than its tool implies.
+//
+// THE OUTPUT IS A TRIAGE LIST, NOT A VERDICT. The EMPTY / ALL_NULL / UNIFORM
+// heuristic is deliberately noisy, because a quiet heuristic would miss the
+// bugs this is for. `UNIFORM(netuid)` on a tool you filtered BY netuid is
+// expected, not a finding; so is `ALL_NULL` on a field that is genuinely null
+// for the one fixture account. Every SUSPECT line has to be confirmed by hand
+// against the tool's contract before it becomes a bug report -- during the
+// original sweep, two flags were written up as defects before that step and
+// were wrong. For the same reason a SUSPECT flag NEVER fails this script: only
+// a transport failure sets a non-zero exit code.
+//
+// Not wired into CI: one run makes ~215 live production calls. It is an
+// explicit, opt-in invocation (`npm run smoke:mcp`), never something a PR
+// triggers.
+//
+// A known limit of driving the sweep off `required`: a handful of tools
+// (how_do_i_call, verify_integration) declare no required field but reject an
+// empty argument set, because their real contract is "one of these two". They
+// report ERROR with that message rather than OK, which is honest -- covering
+// them would mean hand-maintaining the per-tool argument table this design
+// exists to avoid.
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+import { flagValue } from "./lib.ts";
+
+// Live MCP JSON-RPC payloads. Every field below is read for reporting only, and
+// an unexpected shape is exactly what this sweep exists to surface -- typing
+// each hop through `unknown` would force a cast at every `?.` for no real
+// safety gain. Mirrors the `Row` precedent in smoke-live-api.ts and lib.ts.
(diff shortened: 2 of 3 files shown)


JSONbored:
## Follow-up: fixed a false positive the harness produced against production

Re-reviewing the full sweep output, `get_adapter` was reported `SUSPECT EMPTY` while its payload was healthy — the exact failure mode this PR is meant to prevent, shipped in the harness itself.

**Root cause:** `[].every(...)` is `true`, so an empty array of *anything* is structurally indistinguishable from an empty row set. The walk picked whichever empty array it reached first:

- `get_lineage` was judged on `broken_links: []` — an empty **good-news** array — instead of the 43 rows in `links`.
- `get_adapter` was judged on `snapshot.dimensions.crown.shape.array_fields`, a schema descriptor five levels down; it has no row set at all.

My first attempt (breadth-first, shallowest wins) was wrong: it didn't fix `get_adapter` and it *regressed* `get_extrinsic` onto an empty `events` array. Fixing it properly needed the actual payload shapes, not a guess.

**The fix, two rules:**

1. A non-empty row array outranks an empty one. An empty array is accepted only when nothing populated is in reach — which is exactly when `EMPTY` is the honest answer, so a genuinely empty response still reports it.
2. Nothing deeper than `MAX_ROW_ARRAY_DEPTH` (3) counts. Real row sets are shallow (`data.boards.healthiest` at the deepest); schema descriptors are not.

**Measured across a full production sweep, before → after:**

| | before | after |
|---|---|---|
| `SUSPECT EMPTY` | 30 | 27 |
| spurious EMPTY removed | — | `get_adapter`, `get_lineage` |
| new EMPTY flags (regressions) | — | **0** |
| summary | `ok=205 errored=5 suspect=97` | `ok=206 errored=4 suspect=95` |

All three #9455 findings still reproduce. (A third tool, `get_account_counterparties`, also stopped flagging, but it kept the same `path=` — that one is live-data variance, not this fix.)

Also now documented: only **one** row set per response is examined, so a defect confined to a later board of a multi-board response (`get_registry_leaderboards`) is out of this harness's reach. The reported `path=` names which one was read.

Gates re-run: `lint`, `format:check`, `typecheck` clean; `npx vitest run` 670 files / 15,989 tests passed; full `npm run smoke:mcp` exit 0.

PR #9497, opened 2026-08-05
feat(serving): recompute top-holders from live stores, so every column has a producer

### LINKED ISSUE #9469: top-holders is served from a frozen snapshot: give every column a live source, starting with the account_balances sink that no longer exists
`/api/v1/accounts/top-holders` is served entirely from a frozen artifact — a
one-time snapshot of the decommissioned Postgres box. `captured_at` has read
`2026-08-02T00:05:06.441Z` since the box died and cannot advance:
`handleTopHoldersList` tries `tryPostgresTier` (gone), then falls through to
`loadTopHoldersFromArtifact`, whose own header says "the route's inputs are
frozen snapshots of the decommissioned box, so the query's one-time result IS
the live result."

This supersedes #9461, which proposed withdrawing the net-flow fields. That was
the wrong call: the data is alive and the computation already runs elsewhere.
Every column needs a live source, and only one of them is genuinely missing a
producer.

## Column-by-column

| Column | Source | Status |
|---|---|---|
| `delegated_tao` | `nominator_positions` x `neurons` x `subnet_snapshots` | **Live in D1 today** — 126,508 rows / 24,121 coldkeys, `captured_at` current |
| `net_flow_7d/30d/90d` | `chain.account_events` | **Live in the lakehouse today**, current to chain head |
| `free_tao` | `account_balances` | **No sink anywhere** — see below |
| `total_tao` | `free_tao + delegated_tao` | Blocked on `free_tao` |

## `free_tao`: the producer is writing to a database that no longer exists

The chain of events, all verified:

1. `scripts/fetch-account-balances.py` POSTed to
   `/api/v1/internal/account-balances-sync`. Deleted from this repo in #7012.
2. The poller job that replaced it
   (`services/indexer-rs/src/bin/poller/jobs/account_balances.rs`, infra)
   deliberately dropped the HTTP round-trip for a **direct Postgres write** —
   `tokio_postgres`, COPY-to-staging, `INSERT INTO account_balances`. It sits in
   `POSTGRES_JOBS` in the poller's `main.rs` and is spawned with `db_url`.
3. `account_balances_sync_secret` was then **pruned from the vault**
   (metagraphed-infra#150, 2026-07-20) as "the one vault secret the retirement
   made fully dead."
4. Postgres was destroyed (#9193). `handleAccountBalancesSync`
   (`workers/data-api.ts`) now stops at its auth gate.

So the producer still runs and still scans the chain correctly — it just writes
into nothing. `account_balances` is not among D1's 42 tables.

Worth recording: even before that, `workers/api.ts` notes the public-facing
proxy for this route was missing entirely, so "every real data-refresh-cron run
since #6742 shipped 405'd on this exact route, meaning `account_balances` has
never received a row from any caller." The proxy exists now; the sink does not.

## The fix is a migration this repo has already done twice

`validator_nominators.rs` and `subnet_hyperparams.rs` are spawned **without**
`db_url` and POST to their internal sync routes instead. That is exactly why
`nominator_positions` is live in D1 while `account_balances` is not. Apply the
same shape:

**metagraphed-infra**
- Convert `account_balances.rs` from the Postgres COPY/upsert to chunked HTTP
  POST at `/api/v1/internal/account-balances-sync`, mirroring
  `validator_nominators.rs`. Scale: the job's own header measures **542,618**
  `System::Account` entries, so chunking is mandatory — the sync route's sibling
  caps are 25,000 rows / 8 MB per request (~22 requests per pass).
- Drop `"account-balances"` from `POSTGRES_JOBS` and stop passing `db_url`.
- Re-provision `account_balances_sync_secret` in the vault and as the Worker
  secret. **Human step — not something an agent should do.**

**metagraphed**
- New D1 migration for `account_balances` (ss58, free_tao, reserved_tao,
  captured_at). Migrations here are applied by hand — merging does not create
  the table.
- Revive `handleAccountBalancesSync` against D1 + `src/account-balances-d1-write.ts`,
  following `handleNominatorPositionsSync` / `src/nominator-positions-d1-write.ts`
  from #9273/#9292 (upsert-only, no prune — an account whose balance drops to
  zero is skipped, not deleted, so the table is "every account that has ever
  held a balance").
- `src/top-holders-hot-tier.ts`: compose the leaderboard from the three live
  sources, replacing `loadTopHoldersFromArtifact`. Keep the schema-stable empty
  posture, and follow the hot-then-cold switch #9292 established — decline while
  the table is empty so the frozen artifact still answers, rather than gating on
  a deploy flag.
- Restore `net_flow_*` to a real ranking. **Measure first**: `GROUP BY coldkey`
  with no selective predicate is the high-cardinality shape R2 SQL rejects with
  `40015`, so this likely needs a precomputed daily rollup (the shape
  `wallet_flow_daily` had) rather than a per-request scan. Price the lane in GB
  before committing to it.
- Staleness watchdog + cron, mirroring
  `src/nominator-positions-staleness-watchdog.ts`.
- Regenerate `openapi.json`/types/`packages/contract`, and the `apps/ui`
  api-reference page from inside `apps/ui`.

## Done when

`captured_at` advances on its own, `free_tao` matches a live
`System::Account` read for a spot-checked account, and `?sort=net_flow_30d`
returns an actual flow ranking rather than ss58 order.

JSONbored:
Closes #9469. The last piece: #9483 gave `free_tao` a D1 sink and metagraphed-infra#311 revived the producer that fills it — this rebuilds the artifact so **every column the route publishes has a live source again**.

## Where each column now comes from

| Column | Source | Was |
|---|---|---|
| `free_tao` | D1 `account_balances` — the System::Account state scan | frozen 2026-08-02 |
| `delegated_tao` | D1 `nominator_positions` × `neurons` × `subnet_snapshots` | frozen 2026-08-02 |
| `net_flow_7d/30d/90d` | lakehouse `account_events` stake stream | **null for every account, always** |
| `total_tao` | derived | blocked on `free_tao` |

`wallet_flow_daily`, the rollup the original query read for net flow, exists only in prose — its cron retired with the box and no such D1 table was ever created. That is why those three columns have been uniformly null, which is the defect #9455 flagged alongside the staleness.

`free_tao` is the one column that genuinely **cannot** come from the lakehouse: that store holds events, and reconstructing a balance from transfer/fee/stake replay is exactly the drift a state scan exists to avoid.

The lane writes the **same R2 key** the reader has served since #9155 — which anticipated this handover in its own header — so the read path is untouched.

## Why it is not a `PROJECTION_LANES` entry

I tried that first. It needed three special cases to fit: it reads D1 as well as the lakehouse; its key is not under `metagraph/projections/` (the registry's own test enforces that prefix, and the frozen artifact still sits at the old key as the fallback); and it is mainnet-only by construction. Three exceptions to join a uniform registry is the registry telling you the lane belongs elsewhere — so it follows the `live-economics-refresh` shape instead: its own cron string, its own dispatch branch.

**Six-hourly**, not the fleet's half-hour tick: `free_tao` refreshes on the poller's `ACCOUNT_BALANCES_POLL_SECS` (21600) and the lane's other read is a 90-day lakehouse scan. A finer cadence would pay R2 SQL forty-eight times a day to restate the same holdings. The three windows come from **one** 90-day scan via conditional aggregation rather than three separately-windowed queries — the same `CASE`-over-`event_kind` form `src/account-feeds-cold-tier.ts` already runs against this table.

## The two refusals that carry the design

**A cold balance tier declines.** `free_tao` dominates `total_tao` for exactly the accounts a top-holder list is about, so publishing an empty balance table would reorder the ranking and drop real whales off it — not merely age it. The runner writes nothing and the reader keeps serving whatever is at the key. This is also what let the lane ship before its producer: it self-activates the moment rows land, no flag and no second deploy.

**Net flow does not decline.** It is a secondary sort key that is null for every account in the artifact this replaces, and losing it must not leave `free_tao` stale another six hours. Instead the degradation is *reported*: one exception on `net_flow_available: false`, because "the lakehouse declined" and "nobody moved any stake" are indistinguishable in the response otherwise.

## Descriptions

Every published surface claimed the route was "computed live from `account_balances`… joined with `nominator_positions` × `neurons` and `wallet_flow_daily`" — tables that did not exist. The MCP tool, REST route + artifact contracts and GraphQL SDL now state the real six-hour cadence, that `captured_at` is the balance scan's own stamp (so it trails the present by up to one cycle rather than tracking request time), and that a null `net_flow_*` means no recorded movement rather than zero.

## Verification

`lint`, `format:check`, `typecheck`, `validate` clean; `npx vitest run` — **670 files, 16,103 tests passing**. Regenerated `openapi.json` + types are in the commit.

The producer is confirmed writing in production: `account_balances` went from 0 rows to streaming 25k-row chunks (top free balance 5,458,444.81 τ, `captured_at` advancing) after metagraphed-infra#311 was merged and the poller container redeployed. Note that the infra repo has **no deploy workflow** — merging there does not ship the poller; it needs a manual `wrangler deploy -c services/indexer-rs/wrangler.poller.jsonc`, which is what actually started the lane.


JSONbored:
Closing — superseded by #9492 on one half, and **measurably wrong** on the other.

**net_flow_\*** — #9492 shipped this while I was building, and did it better: it priced the scan against production before writing the lane (1.65 GB / 7.1 s for the 90-day conditional aggregation, so a *daily* cadence for 1.65 GB/day rather than 79 GB/day at the shared twice-hourly tick). My version used the same one-scan conditional-aggregation shape but at 6-hourly and unpriced.

**delegated_tao** — this is the part worth recording, because my PR would have published a confidently wrong column. I rebuilt it from `nominator_positions × neurons × subnet_snapshots` on the strength of the route's own contract text, without measuring the join. #9492 measured it, and I have now confirmed the numbers independently against production D1:

```
positions priced: 28,902/126,508 = 22.8%
coldkeys priced:   6,673/24,121  = 27.7%
```

`share_fraction` is a fraction of each **(hotkey, netuid)** alpha pool, but the only stake source in D1 is `neurons.stake_tao`, which exists solely for hotkeys holding a UID **on that exact subnet** — while a delegate accrues alpha on every subnet staked to it, not just the ones it is registered in. So the join silently drops ~77% of position rows. #9492 reports it ranking an account the frozen snapshot puts at 81,185 τ as **0**, and dropping another out of the payload entirely.

That is exactly the "published field that is confidently wrong" class this whole effort exists to fix, so it should not ship behind an unmeasured assumption. `delegated_tao` needs a real per-(hotkey, netuid) stake source before it can be recomputed, which is its own piece of work.

The lane's decline-on-cold-balance-tier posture and the free_tao composition are still wanted, but they belong in a change scoped to `free_tao` alone now that the producer is live — not bundled with a broken `delegated_tao`.

PR #9542, opened 2026-08-05
fix(events): export price_at_tx and price_basis in the CSV projection

### LINKED ISSUE #9537: events CSV: price_at_tx and price_basis are dropped from both ?format=csv exports
The CSV variants of the account- and block-event feeds silently drop two fields the JSON contract publishes.

`workers/request-handlers/entities.ts:647-659`:

```ts
const EVENTS_CSV_COLUMNS = [
  "block_number", "event_index", "event_kind", "hotkey", "coldkey",
  "netuid", "uid", "amount_tao", "alpha_amount", "observed_at",
  "extrinsic_index",
];
```

Eleven columns. `price_at_tx` and `price_basis` are absent, though both are declared on the shared `AccountEvent` component (`schemas-src/routes/subnet-events.ts:37`, `:42`), published in `openapi.json`, and returned by `formatAccountEvent` on every JSON response.

Both consumers of the list are affected — `handleAccountEvents` (`:4738-4745`) and `handleBlockEvents` (`:6465-6472`).

## Measured

The same event, two formats:

```
get_account_events(ss58=5E2LP6…, kind="StakeAdded", limit=1)
  price_at_tx: 0.028416706
  price_basis: "trade_exact"
```

The `?format=csv` export of that row carries neither column.

## Why it matters here

`price_at_tx` is the per-row execution price — the thing a CSV consumer doing cost-basis or fill analysis actually wants, and the reason `src/price-at-tx.ts` exists. `price_basis` is what tells them whether the number is a real trade (`trade_exact`) or the root-subnet special case (`root_no_pool`); without it a null price is unattributable.

## Fix

Append both to `EVENTS_CSV_COLUMNS`, in the `formatAccountEvent` field order the comment above the list already promises ("stable so a CSV consumer's columns never shift" — appending preserves that for existing columns). Add a test asserting the CSV header covers every field the JSON row publishes, so the two cannot drift again.

Found while triaging the always-null sweep (#9455); not something the sweep itself flags, since it only reads JSON.

JSONbored:
## Summary

`EVENTS_CSV_COLUMNS` (`workers/request-handlers/entities.ts:647`) listed 11 columns and omitted both price fields. It is the only declaration of the CSV projection, and all three event feeds share it — `handleAccountEvents` (`:4744`), subnet events (`:5534`), `handleBlockEvents` (`:6473`) — so every `?format=csv` export silently dropped data the JSON contract publishes on every row.

## Measured

```
get_account_events(ss58=5E2LP6…, kind="StakeAdded", limit=1)
  price_at_tx: 0.028416706
  price_basis: "trade_exact"
```

Neither column appeared in the CSV of that same row.

## Why these two specifically

`price_at_tx` is the per-row execution price — the field a CSV consumer doing cost-basis or fill analysis actually wants, and the reason `src/price-at-tx.ts` exists. `price_basis` is what tells them whether a null price means "no trade leg" or the root-subnet special case (`root_no_pool`); without it a null is unattributable.

Appended rather than interleaved, so existing consumers' column positions keep the stability the list's own comment promises.

## The OpenAPI example moved with it

Its sample row now carries a real `alpha_amount`. A `StakeAdded` with an empty alpha leg is not a shape production emits, and it left the two derived columns un-illustratable. The price shown is arithmetically true for that row — `12.5 / 440` at rao precision, matching `resolvePriceAtTx` → `roundUnit` — not a placeholder.

Regenerated and committed: `openapi.json`, `packages/contract/index.d.ts`, `public/metagraph/types.d.ts`. `validate:contract-drift` passes.

## The test derives, it does not restate

A restated column list would have to be updated by the same hand that forgot the first time. The new assertion in `tests/account-events.test.ts` compares `EVENTS_CSV_COLUMNS` against `formatAccountEvent`'s own output keys, **in both directions**:

- a field the row emits and the CSV drops → fails
- a column with no source field → fails (it would write a blank cell forever and read as "always null" to a consumer, which is the defect class this whole sweep is about)

Verified it fails on the unfixed code with the precise cause:

```
× the events CSV projection covers every field formatAccountEvent emits
  AssertionError: formatAccountEvent emits field(s) the CSV export drops: price_at_tx, price_basis
```

Two other test files (`tests/subnet-events.test.ts`, `tests/block-csv.test.ts`) had each hand-copied the header as a literal. **Those literals agreed with the bug**, which is why nothing caught it — both now derive from the exported constant. `tests/worker-runtime.test.ts`'s `CHAIN_EVENTS_CSV_HEADER` is a different feed (pallet/method shape) and is untouched.

## Validation

```
npm run lint            npm run format:check          npm run typecheck
npm run validate        npm run validate:contract-drift
npm run scan:public-safety
npm run build
npx vitest run tests/account-events.test.ts tests/subnet-events.test.ts \
  tests/block-csv.test.ts tests/block-events.test.ts tests/worker-runtime.test.ts
  -> 146 passed
```

Closes #9537

### DIFF
--- a/packages/contract/index.d.ts
+++ b/packages/contract/index.d.ts
@@ -29209,8 +29209,8 @@ export interface operations {
                         data?: components["schemas"]["AccountEventsArtifact"];
                     };
                     /**
-                     * @example block_number,event_index,event_kind,hotkey,coldkey,netuid,uid,amount_tao,alpha_amount,observed_at,extrinsic_index
-                     *     8454388,3,StakeAdded,5Hotkey_sample,5Coldkey_sample,7,3,12.5,,2026-07-03T00:00:00.000Z,2
+                     * @example block_number,event_index,event_kind,hotkey,coldkey,netuid,uid,amount_tao,alpha_amount,observed_at,extrinsic_index,price_at_tx,price_basis
+                     *     8454388,3,StakeAdded,5Hotkey_sample,5Coldkey_sample,7,3,12.5,440,2026-07-03T00:00:00.000Z,2,0.028409091,trade_exact
                      */
                     "text/csv": string;
                 };
@@ -48143,8 +48143,8 @@ export interface operations {
                         data?: components["schemas"]["SubnetEventsArtifact"];
                     };
                     /**
-                     * @example block_number,event_index,event_kind,hotkey,coldkey,netuid,uid,amount_tao,alpha_amount,observed_at,extrinsic_index
-                     *     8454388,3,StakeAdded,5Hotkey_sample,5Coldkey_sample,7,3,12.5,,2026-07-03T00:00:00.000Z,2
+                     * @example block_number,event_index,event_kind,hotkey,coldkey,netuid,uid,amount_tao,alpha_amount,observed_at,extrinsic_index,price_at_tx,price_basis
+                     *     8454388,3,StakeAdded,5Hotkey_sample,5Coldkey_sample,7,3,12.5,440,2026-07-03T00:00:00.000Z,2,0.028409091,trade_exact
                      */
                     "text/csv": string;
                 };
--- a/public/metagraph/openapi.json
+++ b/public/metagraph/openapi.json
@@ -359,7 +359,7 @@
         }
       },
       "AccountEventsCsv": {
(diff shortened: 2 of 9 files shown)


JSONbored:
CI caught a fourth hand-written copy of this header — `tests/account-routes.test.ts:85`, which my earlier sweep missed. Fixed the same way as the other three: derived from the exported `EVENTS_CSV_COLUMNS` rather than restated.

That makes the count **four** literals, all of which asserted the wrong thing in unison — which is precisely why nothing caught the missing columns in the first place. An exhaustive repo-wide sweep for the header string now returns only `packages/client/dist` (untracked/gitignored, regenerated on build), so there is no fifth.

Re-verified against the **full** suite this time rather than scoped files: `npx vitest run` → **682 files, 16,358 tests, 0 failures**.

Issue #9549, opened 2026-08-05
D1 sync lanes spend one bound parameter per column per row, so a full pass is tens of thousands of statements


JSONbored:
Every D1 sync lane builds its statements through `chunkStatements`, which spends **one bound parameter per column per row**. The Workers binding caps a statement at 100 parameters (`D1_PARAM_BUDGET` is 90), so the chunk size is `budget / columnCount`:

| lane | columns | rows/statement | statements per full pass |
|---|---|---|---|
| `account_balances` | 4 | 22 | ~14,600 (364k rows) |
| `hotkey_alpha` | 4 | 22 | ~34,000 (762k pools) |
| `neurons` | 21 | 4 | ~16,000 (4.5k rows tested; far more live) |

The producer's HTTP request has **60 seconds** (`post_sync_json` gives curl `-m 60`). It does not finish. `account_balances` published 48% of a pass that way; metagraphed-infra#325 responded by shrinking the chunk from 25,000 to 5,000 rows, which moved the failure from request 8 to request 29 rather than removing it. Production is still landing partial passes today (`expected 364,284, received 140,000, completed NULL`).

## The parameter budget is not a law about rows

It is a law about **parameters**, and a whole chunk fits in one if it travels as JSON:

```sql
INSERT INTO t (a, b, c) SELECT json_extract(value,'$[0]'), json_extract(value,'$[1]'), json_extract(value,'$[2]')
FROM json_each(?1) WHERE true
ON CONFLICT (a) DO UPDATE SET ... WHERE t.captured_at <= excluded.captured_at
```

Same upsert, same staleness guard, **one statement per chunk**. Verified against production D1: SQLite's JSON1 is enabled, `ON CONFLICT` parses after a `SELECT` given the `WHERE true`, and a replayed older capture is still refused by the guard.

## Done when

- `chunkStatements` binds one parameter per statement regardless of column count, and every lane that shares it inherits that.
- Chunking is bounded by serialized payload size rather than parameter count.
- The byte budget is set from a **documented** D1 limit, not from an experiment run through `wrangler d1 execute` — that CLI path permits 1,200 bound parameters while the binding permits 100, which is the confusion that cost the first 15 production neurons syncs.
- The tests assert the new invariant (one parameter) rather than the old arithmetic, and prove end-to-end against real SQLite that every row still lands in the right column.

Issue #9541, opened 2026-08-05
contracts/docs: self-health text describes the retired box, and rank/axon/take semantics are undocumented


JSONbored:
Three pieces of published or in-repo text describe infrastructure that no longer exists, or omit a field the contract declares. All found while triaging the always-null sweep (#9455) — each one cost time during that triage, and the consumer-facing ones are actively misleading.

## 1. `self-health`'s contract text describes a decommissioned box (consumer-facing)

`src/contracts.ts:1936` and `:4512` both publish:

> …written by the indexer box's self-health poller rather than by a Worker sharing a failure domain with what it checks. … **Computed live from the self_health_daily/self_health_checks Postgres tier** at /api/v1/self-health (no static file).

Neither clause is true. The box is decommissioned and `METAGRAPH_SELF_HEALTH_SOURCE` is `"retired"` (`wrangler.jsonc:275`), so the Postgres tier short-circuits and the **lakehouse daily rollup** answers — `loadSelfHealthColdTier` (`src/self-health-cold-tier.ts:58-84`), which reads `chain.self_health_daily` only.

This is why the sweep flagged `current_ok`/`http_status`/`latency_ms`/`checked_at`/`note` as always-null: the per-check tier died with the box, and the cold tier passes `[]` as latest rows deliberately (`:83`). Null is the honest "unmeasured" encoding — but a consumer reading this contract text has every reason to expect a live probe state, and no way to learn otherwise.

The text should say what actually serves it and that the per-check fields are null while no live tier exists — the same treatment `alpha_price_change_1h` and `realized_return_1m` (#9503) already got.

## 2. `workers/api.ts` claims the health cron writes `self_health_checks`

`workers/api.ts:1590` states the 15-minute health-probe cron is what writes `self_health_checks`. It is not: `src/health-prober.ts:1-15` writes `surface_checks` / `surface_status` for third-party subnet surfaces. Nothing in this repo has ever written `self_health_checks` — a tree-wide grep returns only comments and tests. The same claim is duplicated in `tests/health-prober.test.ts:783-786`.

## 3. `get_neuron_history` does not mention `take`

`src/mcp-server.ts:7433` and `src/graphql-sdl.ts:314` both enumerate the series as "…validator permit, and axon", while `NeuronSchema` declares `take` and the route now serves it (#9523). A caller reading the description would not know to ask for it.

## 4. `rank` and `axon` null semantics are undocumented

[...]

Issue #9540, opened 2026-08-05
graphql: 39 resolvers on retired tiers serve confident zeros — most of the block-explorer surface is empty in production


JSONbored:
GraphQL's `chain_alpha_volume` never consults the projection tier, so it serves a permanently zeroed card while REST and MCP serve real rows from the same lane.

`src/graphql.ts:7523-7528`:

```ts
const data =
  ((await tryPostgresTier(
    context.env,
    postgresTierRequest(context, "/api/v1/chain/alpha-volume", params),
    "METAGRAPH_ACCOUNT_EVENTS_SOURCE",
  )) as Row | null) ?? buildChainAlphaVolume([], { limit: safeLimit });
```

`METAGRAPH_ACCOUNT_EVENTS_SOURCE` is `"retired"` (`wrangler.jsonc:242`), so the first leg always returns null and the resolver falls straight to the empty builder. `loadChainAlphaVolumeFromArtifact` is never called.

Both sibling surfaces do call it:

- MCP — `src/mcp-server.ts:5484`
- REST — `workers/request-handlers/analytics.ts:1928`

## Measured

```
REST/MCP  get_chain_alpha_volume()  -> 20 subnets, real buy/sell volume
GraphQL   { chain_alpha_volume { subnet_count } }  -> subnet_count: 0, empty leaderboard
```

The resolver's own comment claims parity it does not have — "Same tryPostgresTier … fallback contract REST's handleChainAlphaVolume uses" — but REST gained the projection tier in #9146 and this resolver was not updated with it.

## Why it matters

This is the failure mode recorded as "no surface owns the tier cascade": three surfaces each assemble their own ladder over the same lane, so one can silently lose a rung. A GraphQL consumer gets a schema-valid, confidently-zeroed answer with no error and no degraded marker.

## Fix

Insert `loadChainAlphaVolumeFromArtifact` between the retired tier and the empty floor, matching MCP and REST, and correct the stale comment. The market-cap index (#9526) should be threaded in the same call so the ratio is not null on GraphQL alone.

Worth checking the other resolvers in this file for the same missing rung while here — this one was found only because #9526 touched the lane.

JSONbored:
## Rescoping: this is not one resolver, it is most of the GraphQL surface

[...]

Issue #9538, opened 2026-08-05
rpc endpoints: the probe computes an error string that neither the KV row nor the live overlay carries


JSONbored:
`list_rpc_endpoints.error` is null on every row. That is currently *correct* — all 9 served endpoints are healthy — but it would stay null during a real outage too, because the value never reaches the served row through the live path.

This was ruled a false positive during the #9455 sweep triage, on the grounds that `error` is only set on a failed probe. That ruling holds for the observed nulls. The latent gap behind it does not.

## The prober computes it and then loses it

`src/health-prober.ts:551` sets it on a probe throw:

```ts
error: (error as Error)?.message || "probe threw",
```

and it is persisted to D1. But the KV rpc-pool row projection — the object the live overlay actually reads — omits it (`src/health-prober.ts:709-730`): the row carries `status, classification, latency_ms, latest_block, archive_support, last_ok, consecutive_failures, pool_eligible, reliability_*` and no `error`.

The serving overlay drops it too. `mergeRpcEndpoints` (`src/health-serving.ts:259-275`) spreads the static row and overlays `status`, `classification`, `latency_ms`, `archive_support`, `health_source`, `health_stale`, `observed_at` — not `error`.

So the served `error` is always whatever the **static build** baked, and the committed artifact's rows are all `health_source: "missing-probe"`. A live failing endpoint shows `error: null` alongside `status: "failed"`.

## Why this is worth closing rather than documenting

It is the same shape as the defect the sweep exists to catch: a published field whose value cannot reach the response, so a consumer cannot distinguish "no errors" from "errors dropped". The difference from the genuine false positives (`get_self_health`, `superseded_by`) is that those are null *by construction* and this one is null *by omission* — the producer has the value in hand.

## Fix

Carry `error` on the KV rpc-pool row and overlay it in `mergeRpcEndpoints`, alongside the `status`/`classification` it belongs with. It must be overlaid unconditionally (including back to `null`), or a recovered endpoint would keep a stale error string from a previous sweep — the same drift `by_status` was recomputed to avoid in that function's own comment.

Health stays probe-derived: this moves a probe-computed value to the surface, it does not author one.

JSONbored:
Correction to the trace above, on my own claim: **`error` is not persisted to D1 at all.** `migrations/d1/0002_observations.sql` has no `error` column on either `surface_checks` or `surface_status`.

[...]

PR #9553, opened 2026-08-05
chore: release main

### LINKED ISSUE #9386: fix(accounts): /accounts/{ss58} 503s ~50% of the time for a high-activity coldkey, and the decline carries no reason
## `/api/v1/accounts/{ss58}` 503s about half the time for a high-activity coldkey

Measured against production twice, ~25 minutes apart, **sequentially** with 1–2s spacing (so this is not a self-inflicted concurrent-probe artefact):

```
coldkey 5GsbTgfvgCH4xdqSkiPb7EaBBFLHjWH5vfEALhJaewSFpZX9
  run 1:  200 503 503 503 503 200 200 200      -> 4/8 failed
  run 2:  200 503 503 503 503 200 200 200      -> 4/8 failed
```

The identical shape both times is itself notable — it is not random jitter.

**Control, same route, same conditions:**

```
hotkey 5E2LP6EnZ54m3wS8s1yPvD5c3xo71kQroBw7aUVK32TKeZ5u
  200 200 200 200 200 200 200 200               -> 8/8 ok
```

These two addresses are the hotkey and coldkey of the **same** validator — the largest by stake (1.84M TAO, 2,474 nominators). So this is not "an obscure account"; it is one of the most looked-up accounts on the network, failing half the time.

The failure is a typed decline, not a crash:

```json
{"error": {"code": "account_summary_unavailable"}}
```

## The decline itself is correct — the failure behind it is not

`src/account-summary-card.ts` is deliberately built to decline rather than publish an all-zero card (#9305, and the reasoning in `answerAccountSummary` is right). This issue is **not** a request to change that. It is that the leg behind it fails, for this account, half the time.

## What the route does per request

`loadAccountSummaryColdTier` issues **five concurrent R2 SQL queries** under one `Promise.all`, and `answerAccountSummary` declines if *any* of them returns null. So the route's success probability is the product of five, and its cost is five scans of `chain.account_events` per request.

Those scans are not cheap. Measured directly against the lakehouse, a single

```sql
SELECT count(*) FROM chain.account_events WHERE hotkey = '5Gsb…'
```

reports `r2_requests_count: 3390`, `files_scanned` in the thousands — to return `c: 172`. `account_events` is not partitioned on hotkey/coldkey, so every one of these is a broad scan, and the route does five of them at once.

## What I could not determine from outside

Which of the three plausible mechanisms actually fires — `QUERY_TIMEOUT_MS` (15s) being hit, the R2 SQL scan budget rejecting, or a request-rate limit — needs Worker-side visibility. Probing from a workstation is not conclusive: my own token started returning `403 Forbidden` after issuing the five queries a few times, which is consistent with a rate/quota limit but is *my* client's limit, not the Worker's.

So the mechanism is stated as a hypothesis, deliberately. The **user-visible failure rate is measured**, and that is what needs fixing.

## Suggested direction

1. **Record why the leg declined.** `loadAccountSummaryColdTier` returns bare `null` for every failure mode, and `r2SqlQuery` already distinguishes an HTTP status, an engine rejection (with its numbered message) and a timeout. Threading that reason into the 503 would turn this from a guess into a log line — and would have answered the previous paragraph in one request.
2. **Stop paying five scans.** Four of the five queries are aggregates over the *same* `scan` CTE. If the engine will serve them as one multi-aggregate query, the cost and the five-way failure product both collapse.
3. **Reconsider the concurrency.** Whatever the mechanism, five simultaneous broad scans of the same table from one request is the shape most likely to trip any of the three limits at once.

## Related

The `positions` route on the same account correctly reports `degraded: {reason: "positions_unpriceable"}`, so the degraded-marker convention is working elsewhere — this route just has no equivalent detail on its decline.

### LINKED ISSUE #9424: retired Postgres-era scripts read as live work, and a workflow comment points at a migration that will never happen
The estate is Cloudflare-only — every paid server is decommissioned and Postgres is gone. Three scripts and one workflow comment have not caught up, and both failure modes send a reader (or an agent) after work that does not exist.

## The comment

`.github/workflows/check-emission-drift.yml` says:

> The other three (emission-gate-sample, reconcile-neurons, export-parquet) all require `DATABASE_URL` and are **blocked on the chain-data tier's own migration**.

That migration is done. Nothing is blocked; the jobs are retired. And the list is wrong on its own terms — `emission-gate-sample` needs no database and **already runs**, as `EMISSION_GATE_SAMPLE_CRON` in `workers/config.ts`.

## The scripts

Three cannot run and never will again:

| Script | Why |
|---|---|
| `scripts/chain-firehose-relay.ts` | polled the indexer box's local Postgres; the head is a Durable Object firehose now |
| `scripts/reconcile-neurons.ts` | neurons are D1-native with a **single** producer by design — a second writer loses UIDs to the prune |
| `scripts/export-parquet.ts` | read both databases via DuckDB's postgres extension on the box that hosted them; that move is complete |

Each still opens with setup instructions written as if you could follow them.

## Scope

- Correct the workflow comment to say what actually became of each job.
- Mark the three scripts retired at the top, including *why*, so the reason survives.
- Keep them in the tree. A future validator box may want some of this shape back, and deleting them buys nothing that leaving them costs.

## Non-goals

Removing the `postgres`/`pg` dependencies. They are still imported by these files, and dropping them is a separate change with its own blast radius.

## Acceptance

No file in the repo describes a Postgres job as pending, and anyone opening one of the three scripts learns it is retired before they learn how it used to be invoked.

### LINKED ISSUE #9434: observability: the API/MCP capture layer drops uncaught throws, storms, and 95% of the MCP protocol surface
## Problem

An audit of the PostHog instrumentation across REST, MCP and GraphQL found five defects in the capture layer itself. Each was verified against both the code and 7 days of live event data in project 220683.

### 1. An uncaught REST throw produces no `$exception` at all

`withUsageTelemetry` (`workers/api.ts`) is `try`/`finally` with **no `catch`**, and `workers/api.entry.ts` removed Sentry's `withSentry()` wrap in #7766 without replacing it. So an uncaught throw anywhere in the REST pipeline yields a `usage_event` with `ok:false` and nothing else — no stack, no message, no Issue. The most severe class of failure this Worker can have is also the least diagnosable.

`src/usage-telemetry.ts` already documents the hole rather than closing it:

> a truly uncaught throw has no dedicated `$exception` capture of its own ... just without a stack trace

### 2. `$exception` has no storm guard, and a storm already happened

Sampling protects `usage_event` and deliberately never touches failures, which leaves `$exception` with nothing in front of it. One fault duly spent a month of quota in four days: `wallet-auth-keys` (`relation "api_key_usage_daily" does not exist`) captured **871,649 events** between 2026-07-29 and 08-02 — a single Issue, one occurrence per request, **87% of the 1M/month allowance**.

Because exhausting the tier *drops* events indiscriminately, the real cost was the error inbox and the alerts that exist to report the outage causing it — not the events.

### 3. 95% of the MCP protocol surface is sampled out by accident

`resolveUsageSampleRate` exempts events carrying `mcpTool`, which is only ever `tools/call`. Every other protocol method — `ping`, `resources/read`, `prompts/get`, the notifications — is emitted by `scheduleMcpProtocolUsageEvent` with `route: "mcp:<method>"` and **no** `mcpTool`, so it falls through to the REST default of `0.05`.

This contradicts the module's own stated contract, which lists MCP tool calls as never sampled because they are "the product's core signal". The whole MCP surface is ~2K events/day, so the exemption costs nothing sampling was introduced to save.

### 4. `query_graphql` reports its failures as successes

`dispatchTool` derives `isError` from whether the handler threw. The `query_graphql` handler returns `{data, errors}` normally in every case, so a query that resolved to **nothing but errors** emitted `$mcp_tool_call{$mcp_is_error:false}` and `usage_event{ok:true}`.

This tool's failure rate is structurally unmeasurable while every other tool's is not — and the same query over REST is correctly counted as a failure, because `withUsageTelemetry` keys `ok:false` on the `x-metagraph-error-code` header GraphQL already sets. The two surfaces disagree about whether the identical query failed.

### 5. No event carries an environment or release

Two consequences, both observed:

- A local `wrangler dev` session captures into the **same** PostHog project as production. Error Tracking currently shows issues whose frames resolve to `.wrangler/tmp/dev-*` paths inside a local worktree, interleaved with real production faults and counted in the same Issue. Nothing in the payload separates them.
- A regression cannot be pinned to the deploy that introduced it — which `$mcp_server_version` already buys on the MCP events, and no other surface ever got.

Cloudflare's `version_metadata` binding (`CF_VERSION_METADATA`) has been declared in `wrangler.jsonc` since setup and is read by nothing.

### Also: three dead route labels

`AUTH_SURFACE_ROUTES` declares labels for `/.well-known/oauth-authorization-server` and both `/.well-known/oauth-protected-resource` paths. `workers/api.entry.ts` hands everything except a bare no-token `/mcp` to `oauthProvider.fetch`, and `@cloudflare/workers-oauth-provider` answers its own discovery documents internally — so those requests never reach `withUsageTelemetry`. #8996's stated goal ("does anyone actually authenticate?") is answered with three permanent zeroes. `tests/request-usage-telemetry.test.ts` asserted them by calling `usageRouteLabel` directly, which is why nothing caught it.

## Proposed fix

- Capture uncaught REST throws as `$exception` (rethrown unchanged; observe, don't handle).
- Add a per-fingerprint `$exception` throttle carrying the suppressed count, so a storm still reports continuously and still says how much it stood for. Configured by var and off by default, matching the precedent both existing sampling knobs set for test determinism.
- Exempt the whole `mcp:` route namespace from usage sampling, with the prefix single-sourced so the label and the exemption cannot diverge.
- Raise a tool error from `query_graphql` on a rejected request (caller error) and on a genuine resolver fault (internal), reusing GraphQL's own `x-metagraph-error-code` verdict so REST and MCP agree.
- Stamp `environment` + `release` on every captured event, derived from `CF_VERSION_METADATA`.
- Delete the three dead labels.

## Out of scope

`auth_tier` on REST `usage_event`s. It is declared on `UsageEvent` and populated only on the MCP path; the tier is resolved inside `applyTieredRateLimit` per route family, so plumbing it to the global wrapper is a cross-cutting refactor that does not belong with these fixes.

### LINKED ISSUE #9440: observability: three watchdogs and the alerting path compute a verdict and report it to nobody
## Problem

Three watchdogs and the alerting path compute the right answer and report it to nobody. Separately, two Worker entries swallow uncaught faults the same way `workers/api.ts` did before #9434.

### 1. Three watchdogs report on no channel at all

`workers/api.entry.ts` discards what `scheduled` returns — the Workers runtime ignores it. So a verdict that exists only in a cron's return value has been computed and dropped.

| Watchdog | Notifies | Durable record |
| --- | --- | --- |
| `chain-detail` / `neurons` / `nominator-positions` / `validator-nominator-counts` | yes (#9330) | yes (#9340) |
| `projection-staleness` | yes | yes, as of #9431 |
| **`freshness`** (`workers/api.ts` `runFreshnessWatchdog`) | partial — see below | **no** |
| **`lakehouse-seam`** (`src/lakehouse-seam-watchdog.ts`) | **no** | **no** |
| **`safe-mode`** (`src/safe-mode-watchdog.ts`) | **no** | **no** |

`lakehouse-seam` and `safe-mode` are the stark cases: they compute `reasons` correctly and return them into the void. A drifted decode seam, or **the chain entering SafeMode**, is detected every tick and told to no one.

`freshness` does emit a `usage_event`, but only when `shouldReport` decides the verdict is new — and that function deliberately suppresses a repeat of the same signature. A lane critical for six hours notifies once and then goes quiet. Correct for a notification, useless as a record. Which sources were stale, and how many, live only in the discarded return value.

### 2. A watchdog that cannot run is indistinguishable from a healthy one

Every one of these returns `{ok: false, reason: "unreachable"}` rather than throwing (deliberately — a cron that throws is a cron nobody can read the result of). But `handleScheduled`'s `$exception` capture only fires on a throw, so a broken watchdog produces one `usage_event` and nothing else. Its silence looks exactly like "everything is fine", which is the one thing a watchdog must never do.

`lane_health`'s vocabulary already has the right word for this and nothing uses it: `unknown`, documented in `schemas-src/routes/self-health.ts` as explicitly *not* a synonym for `ok`.

### 3. Alerting can stop end-to-end with no signal on either side

- `AlerterHub.refreshTriggers` swallows both a thrown refresh and a non-2xx (`workers/alerter-hub.ts`). A hub whose trigger cache stops refreshing keeps evaluating every chain event against whatever it last loaded, forever, and looks healthy doing it — deliveries still fire, so even the failures-only delivery event (#8997) stays quiet. Triggers created, edited or deactivated after the failure never take effect.
- `AlerterHub` imports no exception capture at all, so `evaluate()` throwing is unrecorded.
- `ChainFirehoseHub`'s `/evaluate` ping is a bare `catch {}`, so the other side of that boundary swallows the evidence too. A hub that is unreachable, timing out, or throwing before it reaches a delivery produces **nothing anywhere**.

### 4. Two Worker entries still swallow uncaught faults

`workers/data-api.ts` and `workers/registry-sync-api.ts` wrap their dispatchers only to emit a trace span — and tracing is off by default (`src/tracing.ts` defaults the rate to 0; neither config sets one). So the unsampled path, which is every request in production, returns the dispatcher's promise unwrapped. Each Worker already has a capture helper; neither is reachable from its entry. This is the same defect #9434 fixed on the main Worker.

## Proposed fix

- `freshness` and `lakehouse-seam` write a `lane_health` row every tick, including `unknown` when they cannot evaluate.
- `lakehouse-seam` and `safe-mode` notify on their alarm condition, matching the sibling watchdogs.
- `safe-mode` deliberately gets **no** `lane_health` row: that table backs the public self-health card, whose consumers read a non-ok lane as "our data is behind". SafeMode is a chain condition, not a staleness of ours — filing it there would publish a false statement about our own freshness.
- `AlerterHub` reports refresh failures (naturally bounded by the 5-minute cache TTL) and captures `evaluate()` faults; `ChainFirehoseHub` reports an unreachable hub behind a rate window, since `broadcast()` runs once per chain event.
- `data-api` and `registry-sync-api` capture uncaught faults at their entries, on both the sampled and unsampled paths.

## Out of scope

Enabling `POSTHOG_TRACES_SAMPLE_RATE` on `data-api`/`registry-sync-api` — their span code is currently unreachable, but choosing a rate has quota consequences on a project that already exhausted its August allowance, and should be a deliberate decision with a measured number behind it rather than a side effect of this fix.

### LINKED ISSUE #9442: /validators lost every operator name when Postgres was removed — the D1 dispatcher never joins account_identity
Verified live 2026-08-04: GET /api/v1/validators returns coldkey_identity: {has_identity:false, name:null, ...} on all 100 of the top 100 rows, so the /validators UI shows bare keys with no operator names. The UI is rendering correctly — the API stopped serving identities.

Root cause: workers/data-api.ts's /api/v1/validators and /api/v1/validators/:hotkey handlers pass identityByColdkey: new Map() with a comment saying the identity join 'keeps its Postgres loader's degraded values (no D1 home)' — and src/metagraph-neurons.ts's loadGlobalValidators/loadValidatorDetail carry the same #5234-era posture ('the Postgres-backed route is what actually joins'). When Postgres was removed, the D1 path became the only path and the join silently disappeared.

The premise is stale: account_identity IS actively written in D1 (the account-identity-sync path; verified live: 470 rows, 369 named, newest capture today). The join just needs to read it — whole-table (a few hundred rows), so no IN list trips the 100-parameter bind cap.

### LINKED ISSUE #9444: /validators repeats a multi-key operator's name at every rank its keys land on
Teams that run several validator keys (Ventura Labs, Yuma, etc.) appear as duplicate entries scattered through the /validators directory — the same operator name at every rank one of its hotkeys reaches. With the identity join restored (#9442), the duplication becomes the dominant visual noise on the page.

Wanted: a clean, organized grouping — one entry per operator, without inventing aggregate rows that would blend per-key take/APY/stake into misleading numbers.

Design: cluster an operator's keys adjacent under its best-ranked row in the current sort. The anchor row carries the name plus a ×N chip; the operator's other keys render as indented continuation rows (icon + hotkey, no repeated name), each still a real per-key row linking to its own detail page. Grouping is by the coldkey identity's declared name; rows without an identity stay individual. On by default with a 'Group by operator' toggle, shareable via search param.

### LINKED ISSUE #9446: observability: six recorders miss the deployment dimensions, REST still has no auth_tier, and two DOs capture nothing
## Problem

Three gaps left by the preceding observability work, one of them introduced by it.

### 1. Six of eight recorders never got the deployment dimensions

#9434 added `environment` / `release` by editing `recordUsageEvent` and `recordExceptionEvent`. The module has **eight** recorders. The other six — `$mcp_tool_call`, `$mcp_initialize`, `$mcp_tools_list`, `ai_degraded`, `$ai_embedding`, `$ai_generation` — shipped with no deployment dimension at all.

Confirmed against live data after that deploy:

| event | environment | release |
| --- | --- | --- |
| `usage_event` | `production` | `45f4e203-…` |
| `$exception` | `production` | `45f4e203-…` |
| `$mcp_tool_call` | *(null)* | *(null)* |

So a breakdown by environment silently drops the entire MCP and AI families — exactly the separation the dimension was added to provide.

### 2. `auth_tier` is still absent on REST

Declared on `UsageEvent` since #8993 and populated only on the MCP path. REST carries ~99% of traffic, so the question the tier system exists to answer — what share of usage is authenticated, and on which tier — has no data behind it. `applyTieredRateLimit` already resolves the tier and returns it; all four call sites discard it.

### 3. Two Durable Objects still capture nothing

- `SubnetStatusHub` imports only `recordUsageEvent`, and that event is emitted *after* a handler returns — so a handler that threw produced neither a usage event nor an `$exception`. A hub rejecting every subscribe looks identical to one nobody is subscribing to.
- `McpSessionHub.alarm()` has no capture. A DO alarm that throws is retried by the platform and recorded nowhere: no request to fail, no caller to notice. An alarm stuck in a retry loop is indistinguishable from a hub with no sessions to expire, which is its normal state.

## Proposed fix

- Stamp the deployment dimensions in all eight recorders, the same way `assignMcpAttribution` is applied per-recorder across the `$mcp_*` family, plus a table-driven test asserting every exported recorder carries them — so a ninth cannot silently miss it.
- Carry the resolved tier from the gates to the usage-event chokepoint and record it as `auth_tier`, omitted (not defaulted to "anonymous") for routes with no tiered gate, since those never checked a key.
- Capture uncaught faults in both Durable Objects, rethrowing unchanged.

### LINKED ISSUE #9451: Error tracking still exceeds the free tier after #9430: two Workers run unguarded and the web client force-records replays
## What's wrong

The July PostHog bill included real overage dollars, and \$exception is one of the drivers that is **still** over budget after #9430's storm guard landed:

- Measured 2026-08-03/04 (post-#9430, post-sampling): ~4.6K \$exception/day ≈ **140K/month**, over the 100K/month free error-tracking tier. Top fingerprints: `r2-sql:Error` (~2.0K/day), `r2-sql:AbortError` (~1.8K/day), plus ~20 `projection:*` lane fingerprints each admitted once per minute per isolate.
- #9430's `POSTHOG_EXCEPTION_STORM_WINDOW_MS` is declared **only in wrangler.jsonc**, and `vars` are per-config: `metagraphed-data-api` (the Worker where the 871,649-event `wallet-auth-keys` storm actually happened) and `metagraphed-registry-sync-api` capture with **no guard at all**.
- The web client (`apps/ui/src/lib/analytics.ts`) has no client-side rate limit on `captureException`, and #7761's "always-on replay for sessions with an exception" **force-starts a session recording on every exception** (`startSessionRecording(true)`) — an exception storm in a render loop converts directly into billable recordings (5K/month free).

## Fix

1. Widen the storm window 60s → 300s in `wrangler.jsonc` (60s × ~25 steady-state fingerprints × isolates still sums over the tier; 12 reports/hour/fingerprint/isolate is ample for diagnosis and the suppressed count is already carried on the next admitted event).
2. Declare `POSTHOG_EXCEPTION_STORM_WINDOW_MS` in `wrangler.data.jsonc` and `wrangler.registry.jsonc` at the same value.
3. `apps/ui`: throttle `captureException` client-side (per-error-signature window + hard per-pageload cap) and drop the `startSessionRecording(true)` force-record.

Part of the 2026-08-04 billing-overage response (July invoice ~\$650; free-tier posture is the goal).

### LINKED ISSUE #9452: observability: a whole day of subnet_snapshots went missing and nothing reported it
## Problem

**2026-08-01 is missing entirely from `subnet_snapshots`** — all 129 subnets, a whole day of price/economics history that does not exist. Nothing reported it. It was found four days later by reading the table by hand while investigating #9449.

Daily coverage over the last 21 days, from production D1:

```
2026-08-05  129 subnets
2026-08-04  129
2026-08-03  129
2026-08-02  129
2026-08-01  ← absent
2026-07-31  129
...every other day: 129
```

## Why nothing noticed

The snapshot write path reports its outcome by **returning it**, and every caller discards the return value.

- `writeSubnetSnapshot` (`src/health-prober.ts:1207-1216`) returns `{ok:false, reason:"no_rows"}` when no profiles carry a netuid, and `{ok:false, reason: syncResult.reason}` when the sync itself declined.
- `syncSubnetSnapshotToPostgres` (`src/health-prober.ts:1097`) returns `{synced:false, reason:"no_rows"}` before it ever reaches a write, when the economics source produced nothing to write.
- Neither path captures an exception, writes a `lane_health` row, or emits a usage event. The verdict goes to the cron dispatcher, and `workers/api.entry.ts` discards what `scheduled` returns.

So "the economics source was degraded, so today's history was never written" and "everything is fine" produce byte-identical telemetry: silence.

Note the D1 write itself is NOT the gap — `upsertSubnetSnapshotsToD1` (`src/observations-d1.ts`) does capture on a thrown write (`route: "observations-d1-snapshots"`). There were zero such captures on Aug 1, which is what points at the earlier, silent return paths rather than a failed insert.

## What Aug 1 actually looked like

Aug 1 was the day of the `wallet-auth-keys` storm. `$exception` by route for that day:

```
wallet-auth-keys                    97,181
realized-return-baseline-query       2,466
tao-usd-index                        1,196
/api/v1/chain-events/stats             327
...
subnet-snapshot-sync                    20   ← the snapshot writer
```

`subnet-snapshot-sync` threw 20 times. That is the Postgres-side POST, and it was captured — but the D1 side of the dual write left no trace either way, and the day is missing from D1, so the captured POST failures are a symptom rather than the whole story.

## Impact beyond a missing row

This is the series `alpha_price_change_*` reads (#9449) and that `/api/v1/subnets/{netuid}/trajectory` serves. A hole in it silently changes what every window computes over — the elapsed-time fix in #9449 is robust to gaps, but the underlying history is still wrong for that day and always will be.

## Proposed fix

Report the verdict instead of returning it, matching what #9441 established for the watchdogs:

- Capture an `$exception` when the snapshot write declines for a reason that is not "nothing to do" — `no_rows` from a degraded economics source is a failure, not a quiet success.
- Record the outcome durably (a `lane_health` row) on every run, so "the snapshot lane did not write today" is answerable without reading the table by hand.
- Distinguish, in the reason, between "the economics source gave us nothing" and "we had rows and the write failed" — they have different causes and different fixes.

## Out of scope

Backfilling 2026-08-01 itself. The economics values for that day are gone unless they can be reconstructed from another tier; whether that is worth doing is a separate call.

### LINKED ISSUE #9457: $exception autocapture bypasses the client throttle, and DO code-update resets are captured as faults
Follow-up to #9451 / #9453, from reading the live PostHog error inbox.

## 1. Exception autocapture bypasses #9453's client-side throttle

The project has exception autocapture enabled (`autocapture_exceptions_opt_in: true`). posthog-js sends those from its own `window.onerror` / `unhandledrejection` handlers — they never pass through this app's `captureException()`, so #9453's throttle does not see them at all.

This is not hypothetical. The `Invalid regular expression: /^*/settings/*$/: Nothing to repeat` storm (a malformed replay-blocklist rule, since corrected project-side) arrived exactly this way — `handled: false`, once per pageview, across every route, for 3 real users.

**Fix:** move the throttle into posthog-js's `before_send` hook, which is the one gate every `$exception` passes through regardless of origin. Returning `null` drops the event client-side, so it is never ingested and never billed.

Also collapses a double-spend introduced in #9453: `captureException()` consumed a budget slot on the way in and then met the throttle *again* in `before_send`, where a second signature miss could drop an event that had already been admitted. One gate, one budget.

## 2. "Durable Object reset because its code was updated" is captured as an exception

43 occurrences in the last week. This is the runtime reporting that a deploy landed while a DO was live: expected on every deploy, self-healing, no action available to a reader.

`workers/api.ts`'s firehose-ingest handler already documents it as "expected/occasional, not a real fault" and swallows it at that one call site — but the DOs' own alarm loops capture it, and their per-isolate change-detectors (e.g. `ChainFirehoseHub.lastCapturedHeadPollerError`) **cannot** suppress it: the reset is what wipes the isolate holding the memo, so every occurrence looks new.

**Fix:** suppress centrally in `recordExceptionEvent` via an exact-prefix list of runtime-owned strings — all four DOs can raise it, so a local guard in one of them would leave the others. Runs *before* the storm guard so a dropped platform message doesn't consume a real fingerprint's window.

## Context

Measured 2026-08-04: `$exception` is running ~1.5-4.6K/day ≈ **~90K/month against the 100K free tier** — i.e. borderline every month before these two fixes. Product analytics sits at ~450K/month against 1M (the 0.05 sampling is collapsing ~270K real requests/day into ~13K ingested events).

Verified already-fixed while triaging, no action needed: the malformed replay regex (corrected in project config 2026-08-04, last error 14:20 that day) and `D1_ERROR: too many SQL variables` (the 15 occurrences on 08-02 are exactly the "first 15 production syncs" that `D1_PARAM_BUDGET = 90` was introduced to fix).

JSONbored:
---

<details><summary>ui: 0.16.0</summary>

### Features

* **accounts:** compose top-holders' free_tao from the balance ledger ([#9504](https://github.com/JSONbored/metagraphed/issues/9504)) ([f3efe4e](https://github.com/JSONbored/metagraphed/commit/f3efe4e070d26dd6a3e0cb61c753ba20a86bf00f)), closes [#9501](https://github.com/JSONbored/metagraphed/issues/9501)
* **accounts:** give top-holders' net_flow_* sorts a live ranking again ([#9492](https://github.com/JSONbored/metagraphed/issues/9492)) ([af50eeb](https://github.com/JSONbored/metagraphed/commit/af50eeb448d8def7961bcbda89fb37d610ace781)), closes [#9469](https://github.com/JSONbored/metagraphed/issues/9469)
* **burn:** every subnet's live registration cost in one call, ranked cheapest-first ([#9403](https://github.com/JSONbored/metagraphed/issues/9403)) ([a1d6f40](https://github.com/JSONbored/metagraphed/commit/a1d6f40db452ec074d6f87728f3c452192ed068c))
* **burn:** record a registration-cost series, so "is this subnet getting more expensive" is answerable ([#9407](https://github.com/JSONbored/metagraphed/issues/9407)) ([9932f9e](https://github.com/JSONbored/metagraphed/commit/9932f9e53bb6a6fa5f7b05c2da4e746ad1dc787e))
* **chain-events:** serve the family on testnet, and stop forwarding to a store that no longer exists ([#9411](https://github.com/JSONbored/metagraphed/issues/9411)) ([4484144](https://github.com/JSONbored/metagraphed/commit/4484144d571bc31fffb47ae345f7ea197a902f30))
* **testnet:** serve the thirteen projection-backed chain routes on testnet ([#9422](https://github.com/JSONbored/metagraphed/issues/9422)) ([2161160](https://github.com/JSONbored/metagraphed/commit/216116078e59b9e487313c752d526883a5241aa9))
* **ui:** make every table column filterable, and surface the fields the API was already returning ([#9460](https://github.com/JSONbored/metagraphed/issues/9460)) ([05234d4](https://github.com/JSONbored/metagraphed/commit/05234d4d4294d08e77aa28dda93ea257a4f71ad3))
* **validators:** group a multi-key operator's rows under one entry ([#9445](https://github.com/JSONbored/metagraphed/issues/9445)) ([4de675c](https://github.com/JSONbored/metagraphed/commit/4de675cbe563097dd68201155f3325885d6735cd)), closes [#9444](https://github.com/JSONbored/metagraphed/issues/9444)

### Bug Fixes

* **api:** recent blocks reported zero events, and eight more correctness gaps alongside it ([#9414](https://github.com/JSONbored/metagraphed/issues/9414)) ([4761416](https://github.com/JSONbored/metagraphed/commit/47614165d4c42362cc6b15cf528738b584fc461f))
* **networks:** stop telling callers testnet has no chain history, when it has ([#9432](https://github.com/JSONbored/metagraphed/issues/9432)) ([a96170a](https://github.com/JSONbored/metagraphed/commit/a96170a466d40083a54dee19bbf8ffee7d77d515))
* **observability:** say the top-holders leaderboard is a frozen snapshot, and watch it ([#9467](https://github.com/JSONbored/metagraphed/issues/9467)) ([c511903](https://github.com/JSONbored/metagraphed/commit/c511903644c6240e1219606acde2d7a4069e7126))
* **telemetry:** guard every $exception emitter, not just the main Worker's ([#9453](https://github.com/JSONbored/metagraphed/issues/9453)) ([811a463](https://github.com/JSONbored/metagraphed/commit/811a4636f57131a61c1bfe7a221845527c3fe946)), closes [#9451](https://github.com/JSONbored/metagraphed/issues/9451)
* **telemetry:** put the exception budget where autocaptured faults pass ([#9458](https://github.com/JSONbored/metagraphed/issues/9458)) ([f6e9f6e](https://github.com/JSONbored/metagraphed/commit/f6e9f6ee96e8ceba6a41b5d65aef8688523b0479)), closes [#9457](https://github.com/JSONbored/metagraphed/issues/9457)
* **ui:** pin every table header to its own scrollport, not the page offset ([#9413](https://github.com/JSONbored/metagraphed/issues/9413)) ([9f4dc1d](https://github.com/JSONbored/metagraphed/commit/9f4dc1d5eeeb0fd7872b7d2ede5083c291aee1de))
* **ui:** pin table column widths so they stop shifting as virtualized rows mount ([#9456](https://github.com/JSONbored/metagraphed/issues/9456)) ([e4f6324](https://github.com/JSONbored/metagraphed/commit/e4f6324c983326f81b919c4218d95d7e11313a23))
* **ui:** stop list tables dragging their pinned header off-screen, and cover /subnets ([#9418](https://github.com/JSONbored/metagraphed/issues/9418)) ([3961bc6](https://github.com/JSONbored/metagraphed/commit/3961bc6daed729de8077b0c43e4f4426b50f9cd9))
* **ui:** stop the new "did this route render?" check turning main red ([#9439](https://github.com/JSONbored/metagraphed/issues/9439)) ([54ce02c](https://github.com/JSONbored/metagraphed/commit/54ce02c9166217b398ef3b22285c12d89d721630))
* **ui:** stop virtualized table rows drifting as you scroll, and put the scroll affordance on the element that scrolls ([#9448](https://github.com/JSONbored/metagraphed/issues/9448)) ([113aa3a](https://github.com/JSONbored/metagraphed/commit/113aa3aef97ce3eb881d6f1df4038f857d2cd197))

### Documentation

* **contracts:** correct text that describes the decommissioned box ([#9547](https://github.com/JSONbored/metagraphed/issues/9547)) ([426d1aa](https://github.com/JSONbored/metagraphed/commit/426d1aacc6dc560803c7b3cde798ad7d9cb48fd3)), closes [#9541](https://github.com/JSONbored/metagraphed/issues/9541)
</details>

<details><summary>platform: 1.3.0</summary>

### Features

* **accounts:** compose top-holders' free_tao from the balance ledger ([#9504](https://github.com/JSONbored/metagraphed/issues/9504)) ([f3efe4e](https://github.com/JSONbored/metagraphed/commit/f3efe4e070d26dd6a3e0cb61c753ba20a86bf00f)), closes [#9501](https://github.com/JSONbored/metagraphed/issues/9501)
* **accounts:** give top-holders' net_flow_* sorts a live ranking again ([#9492](https://github.com/JSONbored/metagraphed/issues/9492)) ([af50eeb](https://github.com/JSONbored/metagraphed/commit/af50eeb448d8def7961bcbda89fb37d610ace781)), closes [#9469](https://github.com/JSONbored/metagraphed/issues/9469)
* **api:** give hotkey_alpha a D1 sink so delegated_tao has a pool total ([#9512](https://github.com/JSONbored/metagraphed/issues/9512)) ([4bda09b](https://github.com/JSONbored/metagraphed/commit/4bda09b84803d4839feccd2982144a0f0a54987e)), closes [#9502](https://github.com/JSONbored/metagraphed/issues/9502)
* **blocks:** derive the block author at the head instead of publishing null ([#9481](https://github.com/JSONbored/metagraphed/issues/9481)) ([afe5a1b](https://github.com/JSONbored/metagraphed/commit/afe5a1b662667920fceb38491b5c6319b9a20490)), closes [#9480](https://github.com/JSONbored/metagraphed/issues/9480)
* **burn:** every subnet's live registration cost in one call, ranked cheapest-first ([#9403](https://github.com/JSONbored/metagraphed/issues/9403)) ([a1d6f40](https://github.com/JSONbored/metagraphed/commit/a1d6f40db452ec074d6f87728f3c452192ed068c))
* **burn:** record a registration-cost series, so "is this subnet getting more expensive" is answerable ([#9407](https://github.com/JSONbored/metagraphed/issues/9407)) ([9932f9e](https://github.com/JSONbored/metagraphed/commit/9932f9e53bb6a6fa5f7b05c2da4e746ad1dc787e))
* **chain-events:** serve the family on testnet, and stop forwarding to a store that no longer exists ([#9411](https://github.com/JSONbored/metagraphed/issues/9411)) ([4484144](https://github.com/JSONbored/metagraphed/commit/4484144d571bc31fffb47ae345f7ea197a902f30))
* **chain-summaries:** give the governance feeds a summary sentence ([#9535](https://github.com/JSONbored/metagraphed/issues/9535)) ([4f0a352](https://github.com/JSONbored/metagraphed/commit/4f0a3527f1f4044fd2555c365cd7eca09e52e3cb)), closes [#9525](https://github.com/JSONbored/metagraphed/issues/9525)
* **chain:** read each block's event count in the head poller, closing the tip gap ([#9420](https://github.com/JSONbored/metagraphed/issues/9420)) ([28c84e2](https://github.com/JSONbored/metagraphed/commit/28c84e20dc5d78532fd1d55813e1e04d16168d12))
* **profiles:** resolve team from the subnet's subnet-team provider ([#9531](https://github.com/JSONbored/metagraphed/issues/9531)) ([0bf7fa0](https://github.com/JSONbored/metagraphed/commit/0bf7fa0ebe93627ed4a3d1cc877a132c0915cfef)), closes [#9524](https://github.com/JSONbored/metagraphed/issues/9524)
* **projections:** give the projection lanes a network dimension, so testnet gets its own artifacts ([#9415](https://github.com/JSONbored/metagraphed/issues/9415)) ([e0b638d](https://github.com/JSONbored/metagraphed/commit/e0b638d1bb55815b03ddab5b434a23f4ae2fab67))
* **testnet:** serve the thirteen projection-backed chain routes on testnet ([#9422](https://github.com/JSONbored/metagraphed/issues/9422)) ([2161160](https://github.com/JSONbored/metagraphed/commit/216116078e59b9e487313c752d526883a5241aa9))
* **ui:** make every table column filterable, and surface the fields the API was already returning ([#9460](https://github.com/JSONbored/metagraphed/issues/9460)) ([05234d4](https://github.com/JSONbored/metagraphed/commit/05234d4d4294d08e77aa28dda93ea257a4f71ad3))
* **validators:** group a multi-key operator's rows under one entry ([#9445](https://github.com/JSONbored/metagraphed/issues/9445)) ([4de675c](https://github.com/JSONbored/metagraphed/commit/4de675cbe563097dd68201155f3325885d6735cd)), closes [#9444](https://github.com/JSONbored/metagraphed/issues/9444)

### Bug Fixes

* **accounts:** collapse the summary's five concurrent scans to three, and say why a decline happened ([#9406](https://github.com/JSONbored/metagraphed/issues/9406)) ([83aaab1](https://github.com/JSONbored/metagraphed/commit/83aaab17f5e0d91f800ae12df77d95a9ead49a2a)), closes [#9386](https://github.com/JSONbored/metagraphed/issues/9386)
* **accounts:** refuse to rank top-holders off an account_balances load that is not provably complete ([#9517](https://github.com/JSONbored/metagraphed/issues/9517)) ([e349199](https://github.com/JSONbored/metagraphed/commit/e349199e6e45f18c2eb99fe2f06d65b960672d55)), closes [#9511](https://github.com/JSONbored/metagraphed/issues/9511)
* **api:** answer 404 for a path that is not a route, instead of 405 for everything ([#9521](https://github.com/JSONbored/metagraphed/issues/9521)) ([816529b](https://github.com/JSONbored/metagraphed/commit/816529b0e597ced77b818ba952af7aba019084f7))
* **api:** answer an external consumer's seven interface defects across REST, MCP and the permit model ([#9499](https://github.com/JSONbored/metagraphed/issues/9499)) ([ff0bfb4](https://github.com/JSONbored/metagraphed/commit/ff0bfb48dcfa7921bf34fb7e2e46802c04b7bd8e))
* **api:** give account_balances a D1 sink so free_tao has a live producer again ([#9483](https://github.com/JSONbored/metagraphed/issues/9483)) ([5bbdec0](https://github.com/JSONbored/metagraphed/commit/5bbdec024f296feaf4d162c897170cb9b2e5efd1)), closes [#9478](https://github.com/JSONbored/metagraphed/issues/9478)
* **api:** recent blocks reported zero events, and eight more correctness gaps alongside it ([#9414](https://github.com/JSONbored/metagraphed/issues/9414)) ([4761416](https://github.com/JSONbored/metagraphed/commit/47614165d4c42362cc6b15cf528738b584fc461f))
* **artifacts:** stop spending an R2 miss to learn what the contract already knew ([#9486](https://github.com/JSONbored/metagraphed/issues/9486)) ([fbcfb7e](https://github.com/JSONbored/metagraphed/commit/fbcfb7e94be055c3cfd2766c429b8a29119451e8)), closes [#9485](https://github.com/JSONbored/metagraphed/issues/9485)
* **chain-alpha-volume:** give vol_mcap_ratio the denominator it needs ([#9532](https://github.com/JSONbored/metagraphed/issues/9532)) ([ffe64df](https://github.com/JSONbored/metagraphed/commit/ffe64df2b7ecae6b433daa748a52f368c0235b99)), closes [#9526](https://github.com/JSONbored/metagraphed/issues/9526)
* **chain-events:** serve the block-hash form instead of 503ing it ([#9484](https://github.com/JSONbored/metagraphed/issues/9484)) ([fc0a844](https://github.com/JSONbored/metagraphed/commit/fc0a8441976e550b81472f1e4d2b33ce07833530)), closes [#9482](https://github.com/JSONbored/metagraphed/issues/9482)
* **ci:** report Actions script failures to PostHog instead of dropping them ([#9508](https://github.com/JSONbored/metagraphed/issues/9508)) ([1673524](https://github.com/JSONbored/metagraphed/commit/167352414dbdc881929b0d1ef16451c4a9017905)), closes [#9507](https://github.com/JSONbored/metagraphed/issues/9507)
* **ci:** route Dependabot PRs down the tokenless Codecov path ([#9405](https://github.com/JSONbored/metagraphed/issues/9405)) ([23ead93](https://github.com/JSONbored/metagraphed/commit/23ead93457ff04afbf7cf96da1b29b07316edba2))
* **d1:** carry a sync chunk in one bound parameter, not one per column per row ([#9550](https://github.com/JSONbored/metagraphed/issues/9550)) ([cdb95da](https://github.com/JSONbored/metagraphed/commit/cdb95da32b5e6c2389c729aa5b1d4bce346a917f)), closes [#9549](https://github.com/JSONbored/metagraphed/issues/9549)
* **data-api:** read take and validator_trust back out of neuron_daily ([#9529](https://github.com/JSONbored/metagraphed/issues/9529)) ([f8f0bf4](https://github.com/JSONbored/metagraphed/commit/f8f0bf46b3f1f5f075de80741401d9845fa2137e)), closes [#9523](https://github.com/JSONbored/metagraphed/issues/9523)
* **economics:** measure the price-change windows in time, not in dates ([#9450](https://github.com/JSONbored/metagraphed/issues/9450)) ([a8914ec](https://github.com/JSONbored/metagraphed/commit/a8914ec98757d334078ec672050c8170974deda1))
* **economics:** publish spot price, and say that alpha_price_tao is the moving average ([#9409](https://github.com/JSONbored/metagraphed/issues/9409)) ([3ce2667](https://github.com/JSONbored/metagraphed/commit/3ce266748d29c68f7007458b600c2bdb6ef33e1f))
* **economics:** serve spot_price_tao on the full economics blob's rows ([#9438](https://github.com/JSONbored/metagraphed/issues/9438)) ([18ad0c4](https://github.com/JSONbored/metagraphed/commit/18ad0c4d19c1ed3b22e9f2324830429baf7038f0))
* **emission:** read the chain in batches, so the sampler stops 429ing itself dead ([#9479](https://github.com/JSONbored/metagraphed/issues/9479)) ([0628c8d](https://github.com/JSONbored/metagraphed/commit/0628c8da8a1f19e8d06e98f0f26951b8c14fafe1)), closes [#9477](https://github.com/JSONbored/metagraphed/issues/9477)
* **events:** export price_at_tx and price_basis in the CSV projection ([#9542](https://github.com/JSONbored/metagraphed/issues/9542)) ([5fbe8a6](https://github.com/JSONbored/metagraphed/commit/5fbe8a676563775a7fce7071ff5222492c0b38e5)), closes [#9537](https://github.com/JSONbored/metagraphed/issues/9537)
* **firehose:** release the SSE slot when the subscriber is provably gone ([#9490](https://github.com/JSONbored/metagraphed/issues/9490)) ([0dfa972](https://github.com/JSONbored/metagraphed/commit/0dfa97296410cfd5414f57ab0a470a2868007775)), closes [#9489](https://github.com/JSONbored/metagraphed/issues/9489)
* **graphql:** give nine resolvers the projection tier they were missing ([#9546](https://github.com/JSONbored/metagraphed/issues/9546)) ([8a4ba4d](https://github.com/JSONbored/metagraphed/commit/8a4ba4d88d16b0678c5a38c22d7f55ee2ec7d952)), closes [#9540](https://github.com/JSONbored/metagraphed/issues/9540)
* **health:** answer the prober's continuity read so last_ok survives a run ([#9534](https://github.com/JSONbored/metagraphed/issues/9534)) ([ad35b9a](https://github.com/JSONbored/metagraphed/commit/ad35b9a4990be79cbca3c1e77d146018a4fbccff)), closes [#9522](https://github.com/JSONbored/metagraphed/issues/9522)
* **health:** carry a probe's error onto the served rpc endpoint row ([#9544](https://github.com/JSONbored/metagraphed/issues/9544)) ([71b9843](https://github.com/JSONbored/metagraphed/commit/71b9843ffce575cea163afe8ffbe049dc664510b)), closes [#9538](https://github.com/JSONbored/metagraphed/issues/9538)
* **mcp:** make the push channel real from initialize, and serve /mcp/ ([#9514](https://github.com/JSONbored/metagraphed/issues/9514)) ([a1d300b](https://github.com/JSONbored/metagraphed/commit/a1d300be6952769395b70f819e6a0b2a77fadc99))
* **mcp:** say when a call_module scope was declined, not measured ([#9548](https://github.com/JSONbored/metagraphed/issues/9548)) ([276932b](https://github.com/JSONbored/metagraphed/commit/276932bec4d2e1d4fc901feef32e358b15f44db7)), closes [#9536](https://github.com/JSONbored/metagraphed/issues/9536)
* **mcp:** stop answering /mcp/ with an OAuth challenge nobody can complete ([#9506](https://github.com/JSONbored/metagraphed/issues/9506)) ([e167cae](https://github.com/JSONbored/metagraphed/commit/e167cae0b9cabc7a07dca938b7647fca10121b4a))
* **networks:** stop telling callers testnet has no chain history, when it has ([#9432](https://github.com/JSONbored/metagraphed/issues/9432)) ([a96170a](https://github.com/JSONbored/metagraphed/commit/a96170a466d40083a54dee19bbf8ffee7d77d515))
* **observability:** let the top-holders watchdog alert on the state it was built for ([#9476](https://github.com/JSONbored/metagraphed/issues/9476)) ([7dcd503](https://github.com/JSONbored/metagraphed/commit/7dcd5039ae41f027395f82ef21059d7434c521e2)), closes [#9475](https://github.com/JSONbored/metagraphed/issues/9475)
* **observability:** report the verdicts and faults that currently reach nobody ([#9441](https://github.com/JSONbored/metagraphed/issues/9441)) ([a78711d](https://github.com/JSONbored/metagraphed/commit/a78711df3562c0424a9f399448155e9aaef19a19)), closes [#9440](https://github.com/JSONbored/metagraphed/issues/9440)
* **observability:** report when a day's subnet snapshot never lands ([#9454](https://github.com/JSONbored/metagraphed/issues/9454)) ([5bbb9bd](https://github.com/JSONbored/metagraphed/commit/5bbb9bdcf2a293c9d930ae60a14cd54ca37ba051)), closes [#9452](https://github.com/JSONbored/metagraphed/issues/9452)
* **observability:** say the top-holders leaderboard is a frozen snapshot, and watch it ([#9467](https://github.com/JSONbored/metagraphed/issues/9467)) ([c511903](https://github.com/JSONbored/metagraphed/commit/c511903644c6240e1219606acde2d7a4069e7126))
* **portfolio:** mark positions at spot, not at the chain's moving average ([#9419](https://github.com/JSONbored/metagraphed/issues/9419)) ([871c5d3](https://github.com/JSONbored/metagraphed/commit/871c5d3f3a0af993b4f7a9b3da49bcbaf11f5f00))
* **projections:** record the staleness watchdog's verdict durably, not only when it alerts ([#9431](https://github.com/JSONbored/metagraphed/issues/9431)) ([a632d88](https://github.com/JSONbored/metagraphed/commit/a632d88cceb226ac364dda112c7ea6cae18da125))
* **projections:** split the aggregate that 422s, and notice when a lane stops writing ([#9427](https://github.com/JSONbored/metagraphed/issues/9427)) ([95d3d9d](https://github.com/JSONbored/metagraphed/commit/95d3d9d747b5d434e988264341ce526be420a8b4))
* **r2-sql:** back off when the account rate limit rejects a query ([#9472](https://github.com/JSONbored/metagraphed/issues/9472)) ([bd3b006](https://github.com/JSONbored/metagraphed/commit/bd3b006b645c94431407f587a5271a309202ab4e)), closes [#9465](https://github.com/JSONbored/metagraphed/issues/9465)
* **repo:** ignore Wrangler's .dev.vars so local secrets cannot be committed ([#9509](https://github.com/JSONbored/metagraphed/issues/9509)) ([#9510](https://github.com/JSONbored/metagraphed/issues/9510)) ([0440eee](https://github.com/JSONbored/metagraphed/commit/0440eeea2ea1b622a1ffe7b2a6d31d2351a0b520))
* **telemetry:** capture what the API and MCP surfaces silently drop ([#9435](https://github.com/JSONbored/metagraphed/issues/9435)) ([194b34e](https://github.com/JSONbored/metagraphed/commit/194b34e24d6b6a47229f71e7814816f99a924304)), closes [#9434](https://github.com/JSONbored/metagraphed/issues/9434)
* **telemetry:** finish the dimensions the last two passes left half-applied ([#9447](https://github.com/JSONbored/metagraphed/issues/9447)) ([d5beff8](https://github.com/JSONbored/metagraphed/commit/d5beff861e8b00497072cd03dc2827c8aa79c646)), closes [#9446](https://github.com/JSONbored/metagraphed/issues/9446)
* **telemetry:** give data-api and registry-sync-api a trace rate they can emit at ([#9470](https://github.com/JSONbored/metagraphed/issues/9470)) ([14a88bb](https://github.com/JSONbored/metagraphed/commit/14a88bb323a01eef4df112d8c4411a2411d56c60)), closes [#9466](https://github.com/JSONbored/metagraphed/issues/9466)
* **telemetry:** guard every $exception emitter, not just the main Worker's ([#9453](https://github.com/JSONbored/metagraphed/issues/9453)) ([811a463](https://github.com/JSONbored/metagraphed/commit/811a4636f57131a61c1bfe7a221845527c3fe946)), closes [#9451](https://github.com/JSONbored/metagraphed/issues/9451)
* **telemetry:** put the exception budget where autocaptured faults pass ([#9458](https://github.com/JSONbored/metagraphed/issues/9458)) ([f6e9f6e](https://github.com/JSONbored/metagraphed/commit/f6e9f6ee96e8ceba6a41b5d65aef8688523b0479)), closes [#9457](https://github.com/JSONbored/metagraphed/issues/9457)
* **telemetry:** say which r2-sql query failed, without a new fingerprint ([#9471](https://github.com/JSONbored/metagraphed/issues/9471)) ([17dec65](https://github.com/JSONbored/metagraphed/commit/17dec659b7569c579bad0e4d64cec095cba455da)), closes [#9459](https://github.com/JSONbored/metagraphed/issues/9459)
* **ui:** pin every table header to its own scrollport, not the page offset ([#9413](https://github.com/JSONbored/metagraphed/issues/9413)) ([9f4dc1d](https://github.com/JSONbored/metagraphed/commit/9f4dc1d5eeeb0fd7872b7d2ede5083c291aee1de))
* **ui:** pin table column widths so they stop shifting as virtualized rows mount ([#9456](https://github.com/JSONbored/metagraphed/issues/9456)) ([e4f6324](https://github.com/JSONbored/metagraphed/commit/e4f6324c983326f81b919c4218d95d7e11313a23))
* **ui:** stop list tables dragging their pinned header off-screen, and cover /subnets ([#9418](https://github.com/JSONbored/metagraphed/issues/9418)) ([3961bc6](https://github.com/JSONbored/metagraphed/commit/3961bc6daed729de8077b0c43e4f4426b50f9cd9))
* **ui:** stop the new "did this route render?" check turning main red ([#9439](https://github.com/JSONbored/metagraphed/issues/9439)) ([54ce02c](https://github.com/JSONbored/metagraphed/commit/54ce02c9166217b398ef3b22285c12d89d721630))
* **ui:** stop virtualized table rows drifting as you scroll, and put the scroll affordance on the element that scrolls ([#9448](https://github.com/JSONbored/metagraphed/issues/9448)) ([113aa3a](https://github.com/JSONbored/metagraphed/commit/113aa3aef97ce3eb881d6f1df4038f857d2cd197))
* **validators:** restore the operator-identity join the Postgres removal severed ([#9443](https://github.com/JSONbored/metagraphed/issues/9443)) ([9f4e2f5](https://github.com/JSONbored/metagraphed/commit/9f4e2f5e124503f1a9a106e270642be01b00032a)), closes [#9442](https://github.com/JSONbored/metagraphed/issues/9442)
* **validators:** serve the burn-derived costs and childkey ratio in the ranking ([#9488](https://github.com/JSONbored/metagraphed/issues/9488)) ([c46bc85](https://github.com/JSONbored/metagraphed/commit/c46bc85597c2367353e141cb7fa987c471ad16e6)), closes [#9487](https://github.com/JSONbored/metagraphed/issues/9487)
* **watchdog:** count coldkeys covered on the nominator-positions lane ([#9539](https://github.com/JSONbored/metagraphed/issues/9539)) ([be18998](https://github.com/JSONbored/metagraphed/commit/be189986304bd095d73f8677fac9882a28bf4b1a))
* **watchdog:** count what an account-balances pass covered, not just when it landed ([#9530](https://github.com/JSONbored/metagraphed/issues/9530)) ([aeec5f9](https://github.com/JSONbored/metagraphed/commit/aeec5f934f101eaff23f20254e755d1fad4bbec3))
* **watchdog:** give the neurons and validator-nominator-counts lanes a coverage signal ([#9533](https://github.com/JSONbored/metagraphed/issues/9533)) ([6bfb2c9](https://github.com/JSONbored/metagraphed/commit/6bfb2c9420beff3f4d490d861305e7699ccfa48d))

### Performance

* **capture:** flush in chunks, pace the reads, run the lanes concurrently ([#9436](https://github.com/JSONbored/metagraphed/issues/9436)) ([d012819](https://github.com/JSONbored/metagraphed/commit/d01281981a9eeaac086c6f13897765e0dcf617c0))

### Documentation

* **contracts:** correct text that describes the decommissioned box ([#9547](https://github.com/JSONbored/metagraphed/issues/9547)) ([426d1aa](https://github.com/JSONbored/metagraphed/commit/426d1aacc6dc560803c7b3cde798ad7d9cb48fd3)), closes [#9541](https://github.com/JSONbored/metagraphed/issues/9541)
* **scripts:** say that the Postgres-era jobs are retired, not blocked ([#9425](https://github.com/JSONbored/metagraphed/issues/9425)) ([51f4ca6](https://github.com/JSONbored/metagraphed/commit/51f4ca63c38b85fc16ab77c1c7b97f3a185f223a)), closes [#9424](https://github.com/JSONbored/metagraphed/issues/9424)
* **validators:** say why realized_return_1m reads null before its history exists ([#9503](https://github.com/JSONbored/metagraphed/issues/9503)) ([09b9937](https://github.com/JSONbored/metagraphed/commit/09b9937c31c5f93540c424687c6c78c4114107c1)), closes [#9500](https://github.com/JSONbored/metagraphed/issues/9500)
* **workers:** point HOTKEY_ALPHA_SYNC_SECRET at the one Worker that reads it ([#9520](https://github.com/JSONbored/metagraphed/issues/9520)) ([4d84326](https://github.com/JSONbored/metagraphed/commit/4d843265259268c67482184d856d0904964c5204)), closes [#9519](https://github.com/JSONbored/metagraphed/issues/9519)
</details>

---


JSONbored:
🤖 Created releases:

- [ui-v0.16.0](https://github.com/JSONbored/metagraphed/releases/tag/ui-v0.16.0)
- [platform-v1.3.0](https://github.com/JSONbored/metagraphed/releases/tag/platform-v1.3.0)

:sunflower:

PR #9555, opened 2026-08-05
fix(graphql): give the block and extrinsic reads their tiers

JSONbored:
## Summary

Five more resolvers on the retired `BLOCKS`/`EXTRINSICS` flags reached only the empty builder, so GraphQL served block 8776109 as *"no such block"* while REST and MCP returned it.

Continues #9546. `PENDING_WIRING` drops **20 → 15**.

## Wired

| Resolver | Tier |
| --- | --- |
| `blocks` | `loadBlockFeedColdTier` |
| `block` | `loadBlockColdTier` |
| `block_extrinsics` | `answerBlockDetail` (hot → cold) |
| `block_events` | `answerBlockDetail` (hot → cold) |
| `extrinsic` | `answerExtrinsicDetail` (hot → cold) |

`blocks` forwards **every filter**, which matters more than the rows alone: the empty builder ignores them, so reaching it dropped a filtered query's *meaning* as well as its result.

## The gap case is why three of these aren't plain loaders

A block ref can land in the **gap** between the decoded seam and the hot window. That is a decline, not an absence — the data exists and this deployment cannot currently read it, and the decode lane closes it within the hour.

- REST answers it `503 block_detail_unavailable` (`chainDetailGapResponse`)
- MCP throws a tool error (`chainDetailAnswerOrThrow`)

Returning the empty builder there would state "this block has no events" — the exact confident zero this whole effort is removing. `gapAwareBlockDetail` raises a `GraphQLError` with `BLOCK_DETAIL_UNAVAILABLE` plus `block_number`/`decoded_through`, so the condition is diagnosable from the response.

A genuine `miss` still degrades to the schema-stable empty — this surface's existing convention, unchanged.

## Pre-existing failure, called out

`tests/economics-surface-parity.test.ts` fails 3 tests on **clean `origin/main`** too, identically, before and after this change. Verified by hard-resetting to `origin/main`, running a full `npm run build`, and re-running: same 3.

It's an ordering dependency on the gitignored R2 fixture tree that surfaced when main's test-file count changed (682 → 685 files shifts alphabetical order, and the fixtures are populated as a side effect of an earlier-sorting file). CI shards differently and stays green. **Not introduced here** — flagged separately rather than folded into this PR.

## Validation

```
npm run lint   npm run format:check   npm run typecheck
npm run validate   npm run scan:public-safety
npx vitest run
  -> 684 passed / 685 files; the single failing file is the pre-existing one above,
     byte-identical to clean main
```

Refs #9540

### DIFF
--- a/src/graphql.ts
+++ b/src/graphql.ts
@@ -26,6 +26,23 @@ import { resolveMarketCapIndex } from "./market-cap-index.ts";
 import { loadChainCallsFromArtifact } from "./chain-calls-artifact.ts";
 import { loadChainFeesFromArtifact } from "./chain-fees-artifact.ts";
 import { loadExtrinsicFeedColdTier } from "./extrinsics-cold-tier.ts";
+import { loadExtrinsicColdTier } from "./extrinsics-cold-tier.ts";
+import {
+  loadBlockColdTier,
+  loadBlockFeedColdTier,
+} from "./blocks-cold-tier.ts";
+import { loadBlockExtrinsicsColdTier } from "./extrinsics-cold-tier.ts";
+import { loadBlockEventsColdTier } from "./events-cold-tier.ts";
+import {
+  answerBlockDetail,
+  answerExtrinsicDetail,
+  chainDetailGapMessage,
+  isEmptyEventPayload,
+  isEmptyExtrinsicPayload,
+  loadBlockEventsHotTier,
+  loadBlockExtrinsicsHotTier,
+  type ChainDetailAnswer,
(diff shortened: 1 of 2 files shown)


JSONbored:
Folded into a single PR that closes #9540 outright, rather than a chain of partial ones. The block/extrinsic work here is the first commit on `fix/graphql-tier-cascade`, unchanged; the remaining 15 resolvers, the guard corrections and the empty PENDING_WIRING land alongside it.

PR #9734, opened 2026-08-07
fix(mcp): default the page on every list tool, not just the one measured

### LINKED ISSUE #9730: 17 MCP tools return >100 KB with no arguments; list_endpoints returns 9 MB
#9701 fixed `list_candidates` returning 7.5 MB unbounded. It was not the only one — it was the one that got measured.

Called every zero-required-argument tool that declares a `limit`, against production, with `arguments: {}`:

| Tool | Bytes | Rows |
|---|---:|---:|
| `list_endpoints` | **9,059,868** | 3,492 |
| `list_surfaces` | **7,820,280** | 3,492 |
| `list_evidence` | **3,623,240** | 3,871 |
| `list_search` | **2,881,186** | 3,757 |
| `list_search_index` | **2,208,844** | 3,757 |
| `list_profiles` | **1,287,854** | 129 |
| `list_review_enrichment_targets` | 511,240 | 138 |
| `list_enrichment_queue` | 511,014 | 129 |
| `list_profile_completeness` | 366,116 | 129 |
| `get_economics` | 307,610 | 129 |
| `list_curation` | 306,330 | 129 |
| `list_adapter_candidates` | 254,334 | 127 |
| `list_enrichment_evidence` | 221,068 | 129 |
| `list_endpoint_incidents` | 204,464 | 137 |
| `list_gaps` | 173,114 | 129 |
| `list_extrinsics` | 172,330 | 50 |
| `list_providers` | 163,332 | 136 |

`list_endpoints` at 9 MB is roughly 2.3M tokens — more than ten times a 200K context window, and it takes no required arguments, so it is a plausible first move for an agent exploring the server.

88 tools declare a `limit`; **9** declare a default.

## Cause — one line, one seam

`paginateRows` in `workers/list-query.ts`:

```js
const shouldPage = requestedLimit !== null || requestedCursor !== null;
const limit = shouldPage ? Math.min(...) : rows.length;
...
rows: shouldPage ? rows.slice(cursor, next) : rows,
```

Omit both `limit` and `cursor` and **every row is returned**. `DEFAULT_LIMIT` exists but is only reachable once the caller has already opted into paging, so it never applies to the call that needs it most.

32 modules under `src/*-mcp.ts` reach this through `applyQueryFilters`, which is why the fix belongs at the seam rather than in 17 handlers.

## Not a bug for REST

The same engine serves the REST routes, and unbounded is the right REST default — a browser can stream 9 MB. #9701 already established the asymmetry: **the MCP surface carries the default, the REST route does not**, because the context window is the hard constraint. `total` and `next_cursor` ride in the envelope either way, so the full set stays reachable by paging; it stops being reachable by accident.

The `get_chain_*` family is already fine (all under 20 KB with no arguments) because those handlers apply their own defaults — the gap is specific to the artifact-backed `list_*` loaders that forward no limit at all.

## Fix

Thread an explicit `defaultLimit` through `applyQueryFilters` → `applyListTransform` → `paginateRows`, have every MCP loader pass it, and gate it so a tool added later cannot reintroduce the hole.

JSONbored:
## Summary

#9701 fixed `list_candidates` returning 7.5 MB unbounded. It was not the only one — it was the one that got measured.

Called every zero-required-argument tool that declares a `limit`, against production, with `arguments: {}`:

| Tool | Bytes | Rows |
|---|---:|---:|
| `list_endpoints` | **9,059,868** | 3,492 |
| `list_surfaces` | **7,820,280** | 3,492 |
| `list_evidence` | **3,623,240** | 3,871 |
| `list_search` | **2,881,186** | 3,757 |
| `list_search_index` | **2,208,844** | 3,757 |
| `list_profiles` | **1,287,854** | 129 |
| …11 more over 100 KB | | |

`list_endpoints` at 9 MB is roughly 2.3M tokens — ten times a 200K context window — from a tool that takes **no required arguments** and is therefore a plausible *first* call from an agent exploring the server. 88 tools declare a `limit`; **9** declared a default.

## One line explains all of them

`paginateRows` in `workers/list-query.ts`:

```js
const shouldPage = requestedLimit !== null || requestedCursor !== null;
...
rows: shouldPage ? rows.slice(cursor, next) : rows,
```

Omit both `limit` and `cursor` and every row is returned — with `DEFAULT_LIMIT` unreachable, because it applies only once the caller has already opted into paging.

## Why the seam, not seventeen handlers

The hole is **invisible at the call site**. Nothing about writing a new loader suggests that passing no limit means passing no limit to anything, which is why #9701's per-tool fix left sixteen more live. A dedicated `src/mcp-list-query.ts` wraps the engine with the default, the 31 loaders import that instead, and a **structural gate** asserts no `src/*-mcp.ts` reaches past it — so a loader written tonight is covered tonight rather than being the next one somebody measures.

## Measured end to end, on production rows

| Tool | Before | After | |
|---|---:|---:|---|
| `list_search` | 1,347,700 | 4,390 | 307× |
| `list_surfaces` | 3,750,533 | 15,138 | 248× |
| `list_endpoints` | 4,267,337 | 25,968 | 164× |
| `list_curation` | 147,683 | 16,853 | 9× |
| `list_gaps` | 82,660 | 9,816 | 8× |

(These are the same rows the REST routes serve today, run through both code paths — so the "before" column is the behaviour on `main`, not an estimate.)

## What is deliberately untouched

**REST stays unbounded.** A browser can stream 9 MB and a context window cannot, so the surface with the hard constraint carries the default — #9701's asymmetry, generalised. `total` and `next_cursor` still ride in the envelope, so the full collection remains reachable **by paging**; it is no longer reachable **by accident**.

**GraphQL stays unbounded too**, and that is a decision rather than an oversight: a GraphQL query states its own page size and an invalid one is an error rather than a substituted default, so a hidden default there would contradict the schema clients read. The gate's allow-list names both exemptions and says why.

## Cleanup

`list_candidates`' own copy of this default comes out with it — the hole was never specific to candidates, only measured there first, and two mechanisms setting the same limit on the same call is one more than can be kept honest. `CANDIDATES_LIMIT_MAX` stays; only the redundant default goes.

## Two tests changed, not worked around

- `candidates-mcp` asserted the query **string** carried the default. It now asserts the caller **receives** twenty rows — the fact that actually mattered, and one a refactor cannot satisfy by accident.
- `search-mcp`'s REST/MCP parity loop compared param sets with no `limit`, which are no longer equivalent questions. Every set now carries one, and the deliberate divergence is asserted explicitly at the end of that test, on both sides, so a later change that silently aligned them fails.

## Validation run

```
npm run build                     # 7/7
npm run validate                  # 129 subnets, 3442 surfaces
npm run validate:api              # 200 routes
npm run validate:contract-drift   # passed
npx tsx scripts/validate-mcp.ts   # 224 tools
npm run lint && npm run format:check && npx tsc --noEmit
```

Tests: **3,595 passed** across the 43 suites touching the list engine, the MCP server, GraphQL and every `*-mcp` loader — plus 5 new cases in `tests/mcp-list-query-default.test.ts` covering the structural gate, the default, an explicit limit winning in **both** directions (the default must not have become a ceiling), a bare `cursor` still paging, and a loader raising its own ceiling explicitly.

Closes #9730

### DIFF
--- a/src/adapter-candidates-mcp.ts
+++ b/src/adapter-candidates-mcp.ts
@@ -3,7 +3,7 @@
 // /metagraph/review/adapter-candidates.json artifact.
 
 import { z } from "zod";
-import { applyQueryFilters, type Row } from "../workers/list-query.ts";
+import { applyMcpQueryFilters, type Row } from "./mcp-list-query.ts";
 import type { StorageReadResult } from "../workers/storage.ts";
 import { API_QUERY_COLLECTIONS, QUERY_ENUMS } from "./contracts.ts";
 import {
@@ -185,7 +185,7 @@ export async function loadAdapterCandidatesList(
       "Adapter candidates snapshot unavailable.",
     );
   }
-  const transformed = applyQueryFilters(
+  const transformed = applyMcpQueryFilters(
     blob as Record<string, unknown>,
     queryUrl,
     "adapter-candidates",
--- a/src/candidates-mcp.ts
+++ b/src/candidates-mcp.ts
@@ -2,12 +2,9 @@
 // Applies the same list-query transforms as the REST route over the baked
 // /metagraph/candidates.json artifact (mirrors gaps-mcp.ts for GET /api/v1/gaps).
 
-import {
-  CANDIDATES_LIMIT_DEFAULT,
-  CANDIDATES_LIMIT_MAX,
-} from "./route-limits.ts";
+import { CANDIDATES_LIMIT_MAX } from "./route-limits.ts";
 import { z } from "zod";
-import { applyQueryFilters, type Row } from "../workers/list-query.ts";
+import { applyMcpQueryFilters, type Row } from "./mcp-list-query.ts";
 import type { StorageReadResult } from "../workers/storage.ts";
 import { API_QUERY_COLLECTIONS, QUERY_ENUMS } from "./contracts.ts";
 import {
@@ -112,15 +109,13 @@ export function candidatesQueryUrl(
       );
     }
     url.searchParams.set("limit", String(limit));
-  } else {
-    // DEFAULTED, because absent used to mean unbounded (#9700). This tool
-    // takes no required arguments, so `list_candidates {}` is the obvious
-    // first call -- and it returned 7,537,056 bytes, ~1.9M tokens, roughly ten
-    // context windows. Not a big answer: no usable answer. The envelope still
-    // carries `total` and `next_cursor`, so the whole catalog is reachable by
-    // paging rather than by accident.
-    url.searchParams.set("limit", String(CANDIDATES_LIMIT_DEFAULT));
   }
+  // No `else` any more. This tool's own default (#9701, when `list_candidates
+  // {}` returned 7,537,056 bytes) is now applyMcpQueryFilters' default for
+  // EVERY loader that shares the seam (#9730) -- it was never specific to
+  // candidates, it was specific to the tool that got measured first. Two
+  // mechanisms setting the same limit on the same call is one more than can be
+  // kept honest.
   if (args?.cursor !== undefined) {
     const cursor = args.cursor;
     if (typeof cursor !== "number" || !Number.isInteger(cursor) || cursor < 0) {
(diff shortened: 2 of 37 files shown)


JSONbored:
Superseded by a branch rebuilt cleanly on current main. The rebase produced a both-added conflict in `src/route-limits.ts` where the shared `/**` opener sat above the conflict marker, so any textual resolution would have detached a doc block from its constant. Rebuilt rather than patched — see the replacement PR.

PR #9757, opened 2026-08-07
fix(queue): refuse chain-detail the transport it measurably does not fit

JSONbored:
Closes metagraphed-infra#359. Follow-up: metagraphed-infra#383.

Adding a lane to `SYNC_QUEUE_LANES` is deliberately a **one-word deploy-time change**. For `chain-detail` that one word would take out the highest-cadence lane on the platform, and nothing in the code said so.

## The measurement

`chain-detail`'s indivisible unit is **one block**. Its four families are posted together precisely so a block and its extrinsics cannot land separately, and `packMultiFamilyMessage` refuses to split them for that reason. So the question was never "how many blocks per message" — the batch is already one block. It is whether **one** block fits.

Built from the real rows of the busiest block captured, #8790494 (35 extrinsics, 694 chain events, 515 account events — 1,245 rows):

| | |
|---|---|
| raw JSON | **476.6 KiB** |
| queue per-message cap | 128 KiB |
| | **3.7× over** |

The producer's own doc comment estimated "350–662 KiB per two blocks" — right about the order, low on the per-block figure.

## Why a guard rather than a comment

The failure mode is not degradation. Every tick would 502 on an oversize `send()`, and **the producer advances its cursor only on a POST that succeeded** — so the lane does not fall behind gracefully, it **wedges**, retrying the same oversized batch forever. That is the cheapest possible mistake causing the most expensive possible outcome, and `chain-detail` is also the largest D1 writer here: ~1,245 rows every 12 seconds is **~9M rows/day**, against `account-balances`' ~1.5M.

`syncLaneUsesQueue` now refuses a lane in `QUEUE_INELIGIBLE_LANES` regardless of the flag. A co-named healthy lane still routes — there is a test for both directions.

## This is not "the lane cannot move"

The payload is enormously repetitive — the same ss58 addresses, pallet names and event kinds, hundreds of times over. gzip on those exact bytes:

| | |
|---|---|
| gzip -9 | **40.5 KiB** |
| ratio | **11.8×** |
| headroom under the cap | 87.5 KiB |
| blocks per message | ~3 |

So the answer is a wire-encoding change, and it is decisive rather than marginal. It needs `CompressionStream` at the route, `DecompressionStream` at the consumer, the byte budget measuring the **compressed** size, and a loud failure for a block that still overflows — a real design with its own failure modes, which is why it is metagraphed-infra#383 and not smuggled in here.

The constant carries the measurement and the ratio, so the next person to look does not have to re-derive either.

## Nothing else changes

The multi-family message shape, `packMultiFamilyMessage`, and the consumer's family writer are correct, tested, and untouched. Compression drops in behind them.

## Validation

```
npm run typecheck    # clean
npm run lint         # clean
npx prettier --check # clean
npx vitest run tests/sync-batch-queue.test.ts tests/data-api-sync-queue-consumer.test.ts \
  tests/chain-detail-sync-route.test.ts   # 81 passed
```

Verified the guard test catches its own removal: deleting the `QUEUE_INELIGIBLE_LANES` line from `syncLaneUsesQueue` fails `refuses a lane that cannot fit the transport, even when named`.

Patch coverage by intersecting the diff with a v8 report: **0 uncovered statements, 0 uncovered branches** (58 changed lines).

## Template Used

- Backend/code change

### DIFF
--- a/src/sync-batch-queue.ts
+++ b/src/sync-batch-queue.ts
@@ -245,6 +245,58 @@ const ENVELOPE_HEADROOM = 2 * 1024;
  */
 export const MULTI_FAMILY_LANES: readonly string[] = ["chain-detail"];
 
+/**
+ * Lanes refused this transport **as it is currently encoded**.
+ *
+ * MEASURED, not assumed (metagraphed-infra#359). `chain-detail`'s indivisible
+ * unit is one block: its four families are posted together precisely so a block
+ * and its extrinsics cannot land separately, and `packMultiFamilyMessage`
+ * refuses to split them for that reason. So the question is never "how many
+ * blocks per message" — it is whether ONE block fits.
+ *
+ * As raw JSON it does not. Built from the real rows of the busiest block
+ * captured (#8790494 — 35 extrinsics, 694 chain events, 515 account events,
+ * 1,245 rows):
+ *
+ *   raw JSON     476.6 KiB      against a 128 KiB per-message cap — 3.7x over
+ *
+ * No producer setting reaches that, because the batch is already one block.
+ *
+ * THE FIX IS COMPRESSION, AND IT IS NOT CLOSE. The payload is enormously
+ * repetitive — the same ss58 addresses, pallet names and event kinds over and
+ * over — and gzip gets **11.8x** on exactly these bytes:
+ *
+ *   gzip -9       40.5 KiB      87.5 KiB of headroom, ~3 blocks per message
+ *
+ * That is a real design change (`CompressionStream` at the route,
+ * `DecompressionStream` at the consumer, a measured ceiling, and a loud failure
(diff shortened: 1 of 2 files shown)


JSONbored:
**Merge order note.** #9765 removes this guard, because it fixes the reason the guard exists: the message is compressed, one chain-detail block goes 476.6 KiB → 40.5 KiB, and the lane fits.

This PR is still worth landing on its own — it is the safety net while the compression change is reviewed, and the trap it closes is live today (one word in `SYNC_QUEUE_LANES` wedges the highest-cadence lane).

**If #9765 merges first, close this instead of merging it** — re-adding the guard afterwards would refuse a lane that now fits. #9765's branch already contains this commit, so its net diff is the compression alone.

Issue #9737, opened 2026-08-07
SN53 is tracked as EfficientFrontier/SignalPlus; on-chain and first-party evidence now both say engy


JSONbored:
`registry/subnets/efficientfrontier.json` records SN53 as **EfficientFrontier by SignalPlus**, with `source_repo` pointing at `oxylok/53-EfficientFrontier`.

A maintainer pass on **2026-07-16** already saw the on-chain identity say `engy` and **deliberately declined to rename**, recording why in `curation.gap_notes`:

> *"…no GitHub presence for 'engy' was found… Deliberately NOT renamed given zero infrastructure corroboration for the new on-chain identity; flagging for close attention on the next pass."*

That was the right call on the evidence available. The evidence has since changed — the operator's repo was pushed publicly after that pass. Re-verified today against primary sources:

## On-chain

`/api/v1/subnets/53` description: **"Verified inference for frontier open models."**

## github.com/hanlinai/engy — exists, active

```
full_name:   hanlinai/engy
pushed_at:   2026-07-29T16:37:29Z
description: "Verified inference for frontier open models. Bittensor subnet 53."
```

The repo description is **character-for-character** the on-chain description plus the netuid. Its README states plainly: *"Bittensor subnet 53"*, *"Subnet: netuid 53 on Bittensor"*, *"Company: hanlin.ai"*, *"API and pricing: engy.ai"*, contact `ning@engy.ai`. That is the corroboration the July pass looked for and could not find.

## The tracked source_repo is stale

```
oxylok/53-EfficientFrontier   pushed_at: 2025-10-18T12:24:26Z
```

Nine months without a push, while the subnet ships. It is not dead (200, not archived), but it is not the subnet's development anymore.

## First-party API is live — four endpoints

| | |
|---|---|
| `GET provider.engy.ai/api/subnet/v1/epoch/latest` | 200 — `{"v":1,"netuid":53,"epoch_index":7,…}` |
| `GET provider.engy.ai/api/subnet/v1/epoch/{index}` | 200 — `{"epoch_index":1,"netuid":53,…,"status":"finalized"}` |
| `GET provider.engy.ai/api/subnet/v1/models` | 200 — per-model pricing + TOPLOC params |
| `GET provider.engy.ai/api/subnet/v1/miners` | 200 — miner roster |

Both `engy.ai` and `hanlin.ai` return 200.

[...]

PR #9771, opened 2026-08-07
feat(registry): flag surfaces whose URL names a moving target

### LINKED ISSUE #9746: 26 tracked surfaces end in /latest; a richer parameterized sibling is invisible to the capture
Found while fixing SN53 (#9737). Its first-party API exposes both `GET /api/subnet/v1/epoch/latest` and `GET /api/subnet/v1/epoch/{index}`. The second is strictly more capable — same shape, any finalized epoch — and a capture that records only the first cannot see it, because nothing about `/latest` announces that `/{index}` exists.

The registry now tracks both for SN53, with the rationale recorded on the surface: *"they are different capabilities — 'what is happening now' against 'what happened then' — and a probe of the first proves nothing about the second."*

## It generalizes — measured across the whole registry

Scanning all 129 `registry/subnets/*.json` for callable surfaces (`subnet-api`, `openapi`, `sse`, `data-artifact`):

| | |
|---|---:|
| callable surfaces tracked | 2,278 |
| **ending in a `latest`/`current`/`newest`/`head` terminal** | **26**, across **13 subnets** |
| of those, with any tracked sibling under the same path prefix | 17 of 26 |
| surfaces tracking a parameterized path anywhere | 684 |

The 26:

```
SN1    apex.api.macrocosmos.ai/dashboard/competitions/{id}/submissions/{hotkey}/latest
SN7    api.all-ways.io/events/latest
SN7    api.all-ways.io/miners/{hotkey}/scores/current
SN15   api.oroagents.com/v1/public/suites/current
SN15   api.oroagents.com/v1/public/races/current
SN15   api.oroagents.com/v1/admin/races/current
SN19   api.blockmachine.io/epochs/current
SN36   api.eirel.ai/v1/runs/current
SN36   api.eirel.ai/v1/admin/runs/current
SN36   api.eirel.ai/v1/internal/checkpoints/{thread_id}/latest
SN36   api.eirel.ai/v1/serving/releases/current
SN36   api.eirel.ai/v1/submissions/current
SN38   api.chronollm.com/rounds/current
SN49   tournament-api.nepher.ai/api/v1/weight-commits/latest
SN50   api.synthdata.co/validation/scores/latest
SN50   api.synthdata.co/v2/meta-leaderboard/latest
SN50   api.synthdata.co/v2/leaderboard/latest
SN50   api.synthdata.co/v2/prediction/latest
SN99   api.leoma.ai/tasks/latest
SN100  chain.joinbase.ai/v1/weights/latest
SN110  api.green-compute.com/platform/builds/{build_id}/jobs/latest
SN114  platform.thesoma.ai/api/private/frontend/competition/timeframe/current
SN114  platform.thesoma.ai/api/public/frontend-key/competition/timeframe/current
SN120  api.affine.io/api/v1/rank/current
SN120  api.affine.io/api/v1/scores/latest
SN120  api.affine.io/api/v1/scores/weights/latest
```

`api.blockmachine.io/epochs/current` and `chain.joinbase.ai/v1/weights/latest` are the same shape as SN53's, on subnets where nobody has looked.

## What this is NOT

**Not "probe `/1` and see".** Guessing sibling paths against third-party APIs is unsolicited traffic against endpoints we do not own, and a 200 on a guessed path is weak evidence that the operator publishes it as a surface — the registry's whole standard is that a `source_url` independently proves the subnet publishes the thing.

The SN53 sibling was found the right way: it is **documented in the operator's own repo** (`docs/superpowers/specs/2026-07-18-provider-endpoint-design.md`), which also carries a dated amendment renaming `/weights/latest` to `/epoch/latest` and stating the old path was deleted — which is exactly why a capture predating it points at a 404.

## Worth designing before building

Two directions, and I do not think the choice is obvious:

1. **Flag, don't fetch.** Surface the 26 as an enrichment signal — "this subnet publishes a `/latest`; its docs may describe a parameterized sibling" — and let the existing evidence-backed enrichment path resolve each one against the operator's own documentation. Cheap, no unsolicited traffic, no invented surfaces.
2. **Teach the capture to read the docs it already has.** Several of these subnets have a tracked `docs` or `openapi` surface; a sibling declared in an OpenAPI document the registry already fetches is evidence, not a guess. Narrower, but only helps where a spec exists.

Option 1 is what SN53's fix effectively did by hand. Filing rather than implementing because the choice affects what the registry is willing to assert about surfaces nobody has confirmed, which is a policy question rather than a mechanical one.

JSONbored:
## Summary

Found while fixing SN53 (#9737): its first-party API exposes both `/api/subnet/v1/epoch/latest` **and** `/api/subnet/v1/epoch/{index}`. The second is strictly more capable — same shape, any finalized epoch — and a capture that records only the first cannot see it, because nothing about `/latest` announces that `/{index}` exists.

**It generalises.** 27 callable surfaces across 14 subnets end in a `latest`/`current`/`newest`/`head` terminal, including `api.blockmachine.io/epochs/current` and `chain.joinbase.ai/v1/weights/latest` — the same shape as SN53's, on subnets nobody has looked at.

The gaps block now carries `moving_target_surfaces`, so it's queryable through `list_gaps` the day it lands rather than being a note somebody reads.

## It asserts nothing about a sibling's existence

It reports that a tracked surface answers *"what is happening now"*, and that *"what happened then"* may be published beside it. Resolving that means reading the **operator's own documentation** — which is how SN53's was found.

Probing guessed paths would be unsolicited traffic against a third party, and a 200 on a guessed URL is not evidence that the operator *publishes* it as a surface, which is the standard every other entry in this registry is held to.

## Precision, measured rather than assumed

- **Anchored to the end of the path** — a `/latest/` in the middle is a directory name, and `/latest-releases` is a word.
- **Path only** — `?window=latest` is a parameter value, not a moving target.
- **Callable kinds only.** A `/latest` on a dashboard or website is a page a human opens, not an API with an indexed sibling; flagging one would send an enricher to read documentation that doesn't exist. Restricting to `subnet-api`/`openapi`/`sse`/`data-artifact` drops the one `dashboard` false positive and keeps all **24 subnet-api** and **3 data-artifact** hits.

Each of those is a test, including one that the helper survives junk (empty URL, unparseable URL, missing kind) without throwing.

## It also collapses buildGaps to one implementation

`scripts/build-artifacts.ts` and `scripts/validate.ts` each carried a **byte-for-byte copy**, and the reproducibility check compares one against the other — so adding this field to the build alone reported as `per-subnet detail artifact is not reproducible from registry inputs` on **every subnet**, rather than as the divergence it was.

That is the **third** time today a duplicated derivation produced a failure that named the wrong thing (after `subnetDisplayName` in #9767 and this). Two copies of a function that exists to be compared against itself is a bug generator, not a safety net.

## Validation run

```
npm run build                     # 7/7
npm run validate                  # 129 subnets, 3441 surfaces, 137 providers
npm run validate:api              # 201 routes
npm run validate:contract-drift   # passed
node scripts/validate-docs.ts     # passed
npm run lint && npm run format:check && npx tsc --noEmit
```

638 tests across formatting, the artifact gate, the gaps loader, the schemas and the lib helpers.

Closes #9746

### DIFF
--- a/packages/contract/index.d.ts
+++ b/packages/contract/index.d.ts
@@ -7912,6 +7912,7 @@ export interface components {
         Gaps: {
             gap_notes: string[];
             missing_kinds: components["schemas"]["SurfaceKind"][];
+            moving_target_surfaces?: string[];
             supported_kinds: components["schemas"]["SurfaceKind"][];
         };
         GapsArtifact: {
@@ -16956,6 +16957,9 @@ export interface operations {
                      *           ],
                      *           "missing_kinds": [
                      *             "archive"
+                     *           ],
+                     *           "moving_target_surfaces": [
+                     *             "example"
                      *           ],
                      *           "supported_kinds": [
                      *             "archive"
@@ -37114,6 +37118,9 @@ export interface operations {
                      *           "missing_kinds": [
                      *             "archive"
                      *           ],
+                     *           "moving_target_surfaces": [
+                     *             "example"
+                     *           ],
                      *           "supported_kinds": [
                      *             "archive"
                      *           ]
@@ -41289,6 +41296,9 @@ export interface operations {
                      *           "missing_kinds": [
                      *             "archive"
                      *           ],
+                     *           "moving_target_surfaces": [
+                     *             "example"
+                     *           ],
                      *           "supported_kinds": [
                      *             "archive"
                      *           ]
@@ -42024,6 +42034,9 @@ export interface operations {
                      *           ],
                      *           "missing_kinds": [
                      *             "archive"
+                     *           ],
+                     *           "moving_target_surfaces": [
+                     *             "example"
                      *           ],
                      *           "supported_kinds": [
                      *             "archive"
--- a/public/metagraph/openapi.json
+++ b/public/metagraph/openapi.json
@@ -8773,6 +8773,9 @@
               "missing_kinds": [
                 "archive"
               ],
+              "moving_target_surfaces": [
+                "example"
+              ],
               "supported_kinds": [
                 "archive"
               ]
@@ -9858,6 +9861,9 @@
               "missing_kinds": [
                 "archive"
               ],
+              "moving_target_surfaces": [
+                "example"
+              ],
               "supported_kinds": [
                 "archive"
               ]
@@ -10306,6 +10312,9 @@
               ],
               "missing_kinds": [
                 "archive"
+              ],
+              "moving_target_surfaces": [
+                "example"
               ],
               "supported_kinds": [
                 "archive"
@@ -31098,6 +31107,12 @@
           "missing_kinds": {
             "items": {
               "$ref": "#/components/schemas/SurfaceKind"
+            },
+            "type": "array"
+          },
+          "moving_target_surfaces": {
+            "items": {
+              "type": "string"
             },
             "type": "array"
           },
--- a/public/metagraph/types.d.ts
+++ b/public/metagraph/types.d.ts
@@ -7912,6 +7912,7 @@ export interface components {
         Gaps: {
             gap_notes: string[];
             missing_kinds: components["schemas"]["SurfaceKind"][];
+            moving_target_surfaces?: string[];
             supported_kinds: components["schemas"]["SurfaceKind"][];
         };
         GapsArtifact: {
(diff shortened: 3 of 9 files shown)


JSONbored:
Superseded by a branch rebuilt cleanly on current main. #9767 merged the shared-derivation extraction this depends on, producing four both-added conflicts across scripts/lib.ts, scripts/lib/formatting.ts, scripts/validate.ts and tests/formatting.test.ts — each one a hunk that cuts through a describe block or an import list. Rebuilt rather than resolved; same change, clean history.

PR #9805, opened 2026-08-07
fix(neon): make the SQL portable, and gate it, so the cutover holds

### LINKED ISSUE #9791: Interface-compatible runners are NOT dialect-compatible: SQLite date() silently emptied two routes on Neon
#9784 moved the `neuron_daily` analytics routes onto Neon. Two of them returned **zero rows**.

Measured against a baseline captured before the cutover:

```
/api/v1/subnets/1/history          28 rows -> 0    (4,226 bytes -> 83)
/api/v1/subnets/movers?window=30d   5 rows -> 0    (2,889 bytes -> 524)
```

Rolled back by removing `neuron_daily` from `NEON_READ_LANES`.

## The cause

Both queries anchor their window with SQLite date arithmetic:

```sql
WHERE snapshot_date >= (SELECT date(MAX(snapshot_date), '-30 days') FROM neuron_daily)
```

**Postgres has no `date(x, '-30 days')`.** The subquery yields nothing, `>=` matches nothing, and the route returns a schema-stable empty result — a 200 with no rows, which is exactly the shape that does not look like a failure.

## The wrong assumption, stated plainly

`createD1Sql` and `createPgSql` are interface-compatible, and the code says so:

> createPgSql is interface-compatible with createD1Sql -- so the store moves by choosing a runner, not by rewriting a query.

That is true of the **binding mechanism** and false of the **SQL**. `toPositionalPlaceholders` translates `?` to `$n`; nothing translates SQLite's function library. Every query moved to Neon has to be portable, and portability is a property of the query text that nothing currently checks.

Other SQLite-isms that would fail the same way and are worth grepping for before the next cutover: `date()`/`datetime()`/`julianday()`/`strftime()` with modifier strings, `IFNULL`, `INSTR`, `GROUP_CONCAT` (Postgres is `STRING_AGG`), `||` on non-text, `LIMIT -1`, `AUTOINCREMENT`, and `json_extract` (Postgres uses `->`/`->>`).

## Two fixes, both needed

**1. Port the two queries.** Compute the window boundary in TypeScript and pass it as a bound parameter, rather than asking the database to do date arithmetic. That is portable by construction and removes the dialect question entirely for these routes.

**2. Make the gate refuse un-portable SQL.** `NEON_READ_ROUTE_TABLES` says which routes MAY move; nothing says which routes CAN. A test that scans each mapped route's SQL for a deny-list of SQLite-only constructs would have failed #9784 at CI instead of in production. The map already extracts per-route SQL blocks for the table assertion, so the scan is nearly free.

## What worked

The pre-cutover baseline caught this within minutes of the deploy, and the rollback is one word. Both are worth keeping as the standard for every remaining cutover in #9787 — a cutover without a content baseline would have shipped two empty routes silently.

JSONbored:
Closes #9791

Two rollbacks preceded this (#9792, #9802) and both had **one cause**: `createD1Sql` and `createPgSql` are interface-compatible, so a query moves stores by choosing a runner. That is true of the **binding** and false of the **SQL**.

## Three classes, all fixed

| was | now |
|---|---|
| `date(MAX(snapshot_date), '-30 days')` | boundary computed in TypeScript and bound |
| `validator_permit = 1` | `validator_permit = TRUE` |
| `SUM(validator_permit)` | `SUM(CASE WHEN validator_permit THEN 1 ELSE 0 END)` |

**The second one is what emptied `/api/v1/validators`, and I missed it twice.** D1 stores those columns as `0`/`1` integers; Neon declares them `BOOLEAN`, because that is what the mirror writes. Postgres rejects `boolean = integer`, so the query errored, `tryPostgresTier` returned null, and the handler's `??` fallback built an **empty leaderboard** — a 200 with a plausible shape and no error anywhere. `SUM(boolean)` fails identically.

I spent two rollbacks blaming SQL *functions* when the real blocker was column *typing* — a difference the mirror itself introduced this morning.

## Verified equivalent before changing anything

```
int_form      1580   (validator_permit = 1)
bool_form     1580   (validator_permit = TRUE)
sum_raw       1580   (SUM(validator_permit))
sum_portable  1580   (SUM(CASE WHEN validator_permit THEN 1 ELSE 0 END))
```

Run against live D1. SQLite has supported `TRUE`/`FALSE` since 3.23, so both forms are correct in both stores.

## The gate that did not exist

`tests/neon-sql-portability.test.ts` scans each movable route's own SQL for SQLite-only constructs and **fails CI**, instead of letting Postgres return an empty result at runtime. It covers `date()`/`datetime()` with modifiers, `strftime`, `julianday`, `unixepoch`, `IFNULL`, `GROUP_CONCAT`, `json_extract`, `json_each`, `INSTR`, `LIMIT -1`, `AUTOINCREMENT`.

Two properties keep it honest:
- **it strips comments first** — a comment *naming* a construct is not a use of it, and it was flagging `/chain/turnover` for the note recording what it was ported away from
- **it proves the deny-list matches the exact SQL that caused #9791**, so it cannot pass on nothing (my first `date(` pattern couldn't cross the nested paren in `MAX(snapshot_date)` and silently matched nothing)

## Re-enables

`NEON_READ_LANES` goes back to `account_position_daily,neurons,neuron_daily`.

## Verification

- **719 files, 17,452 tests, all passing**; typecheck, eslint, prettier clean
- I'll diff all 15 routes against the pre-cutover D1 baseline within minutes of deploy, same as before — that baseline is what caught both regressions.

## Two false positives I caught in my own rewrite

A blind regex over six files hit `let active = 0` (TypeScript, not SQL) and a test helper's default parameter. Both reverted. Worth noting because it's the same lesson as the SQL itself: a pattern that looks like SQL isn't necessarily SQL.

### DIFF
--- a/src/chain-turnover.ts
+++ b/src/chain-turnover.ts
@@ -198,7 +198,7 @@ const EMPTY_SET = new Set<string>();
 // Null-safe: no data or an unresolvable boundary yields the schema-stable empty block
 // (comparable:false, empty leaderboard), never throws. `limit` caps the per-subnet leaderboard;
 // the network rollup, subnet_count, and distribution cover every subnet that had a validator set
-// at either boundary (the loader reads only validator_permit=1 rows, so subnets with no
+// at either boundary (the loader reads only validator_permit = TRUE rows, so subnets with no
 // validators in the window are absent) — subnet_count/distribution count all of those, not just
 // the returned page.
 export function buildChainTurnover(
@@ -311,7 +311,7 @@ export function buildChainTurnover(
 // Network-wide validator turnover, computed live: anchor the window to the newest STORED
 // snapshot (date() relative to MAX(snapshot_date)) so a lagging/restored store still compares
 // real boundary snapshots, read every subnet's validator rows at those two days (bounded by
-// validator_permit = 1; the date-first idx_neuron_daily_date_netuid_agg covers the boundary
+// validator_permit = TRUE; the date-first idx_neuron_daily_date_netuid_agg covers the boundary
 // scan), shape with buildChainTurnover. Cold/absent or single-snapshot D1 → comparable:false.
 export async function loadChainTurnover(
   d1: (
@@ -341,7 +341,7 @@ export async function loadChainTurnover(
   let rows: Array<Record<string, unknown>> = [];
   if (startDate != null && endDate != null && startDate !== endDate) {
     rows = await d1(
-      `SELECT ${CHAIN_TURNOVER_READ_COLUMNS} FROM neuron_daily WHERE validator_permit = 1 AND snapshot_date IN (?, ?)`,
+      `SELECT ${CHAIN_TURNOVER_READ_COLUMNS} FROM neuron_daily WHERE validator_permit = TRUE AND snapshot_date IN (?, ?)`,
       [startDate, endDate],
     );
   }
--- a/src/live-economics-refresh.ts
+++ b/src/live-economics-refresh.ts
@@ -273,7 +273,7 @@ export interface NeuronAggregate {
  */
 export const NEURON_AGGREGATE_QUERY =
   "SELECT netuid, COUNT(*) AS uid_count, " +
-  "SUM(CASE WHEN validator_permit = 1 THEN 1 ELSE 0 END) AS validator_count, " +
+  "SUM(CASE WHEN validator_permit THEN 1 ELSE 0 END) AS validator_count, " +
   "SUM(stake_tao) AS total_stake_alpha, MAX(stake_tao) AS max_stake_alpha " +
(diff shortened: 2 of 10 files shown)


JSONbored:
Superseded by #9810.

Splitting this was the right call once #9807 landed. This PR carried two fixes: the SQLite `date(x, 'modifier')` boundary (#9791) and the boolean column typing (#9802). #9807 shipped the first one independently — `src/iso-date-window.ts` plus `neuronDailyWindowBounds`, a cleaner split than my `portable-date-window.ts` because it keeps the anchor query and the bound query separate at the call site.

That closed #9791, which this PR was linked to. #9810 carries the remaining half — the boolean typing and the CI gate — against a fresh issue (#9809), rebased onto `28be07fb2`.

PR #9838, opened 2026-08-07
fix(ci): unred main — five failures, none visible to the PR that caused them

### LINKED ISSUE #9837: main is red: tests/neon-migrate.test.ts is not prettier-formatted
format:check fails on main as of #9829; every open PR inherits a red Lint step.

### LINKED ISSUE #9839: main is red: BuildSummaryArtifactSchema's coverage fixture is a two-key stub
`tests/zod-schemas.test.ts` fails on `main`:

```
build-summary: BuildSummaryArtifactSchema.parse(...) succeeds using real budget helpers
  coverage.root_subnet_count  expected number, received undefined
  coverage.source             expected object, received undefined
```

#9827 widened `coverage` from a loose object to the full `CoverageArtifactSchema`, which requires ~17 fields. The fixture passes `{ candidate_count: 300, surface_count: 900 }` — a two-key stub that validated under the old loose type and does not under the new one.

**The schema is right.** `GET /api/v1/coverage` returns every one of those fields, so the build summary really does embed a full coverage artifact; the fixture was the thing that had drifted from reality.

It went red on `main` rather than in #9827 because the fixture lives in a different file from the schema, so nothing in that PR's diff touched it — and `Validate` on the merge queue is where it surfaced.

Fix: replace the stub with a real coverage object captured from the live artifact.

JSONbored:
Closes #9837, closes #9839.

`main` was red on **five independent things**. Every open PR inherited all of them.

| failure | cause |
|---|---|
| `format:check` | `tests/neon-migrate.test.ts` unformatted — my miss in #9829 |
| `BuildSummaryArtifactSchema` | #9827 widened `coverage` to the full artifact; fixture was a two-key stub |
| `SAMPLE_CHANGELOG` | fixture predated `schema_version`, `contract_version`, `generated_at` and most of `summary` |
| `root-is-not-a-subnet` | #9816 wrote "21 of 129 **subnets**" into `data-api.ts` — 129 is the netuid count, and that prose is what the test forbids |

## The pattern

In every case the assertion and the thing it asserts about live in **different files**, so a PR-scoped diff couldn't see the break. #9827 changed a schema in `schemas-src/`; its fixture is in `tests/`. #9816 edited a comment in `workers/`; the rule lives in `tests/root-is-not-a-subnet.test.ts`. Each was green on its own PR and red on merge.

## Fixtures checked against reality, not loosened

The tempting repair — relax the schema until the fixture passes — would have been wrong three times. I checked the live endpoints first:

- `GET /api/v1/coverage` returns **every** field `CoverageArtifactSchema` requires
- `GET /api/v1/changelog` returns `schema_version`, `contract_version`, `generated_at` and the full `summary` including `coverage_delta`

So the schemas were right and the fixtures had drifted from what the API has been serving all along. Fixtures updated from live data.

## Verification

`typecheck`, `eslint .` and `prettier --check .` clean repo-wide; all four previously-failing test files pass locally.

### DIFF
--- a/src/table-freshness-watchdog.ts
+++ b/src/table-freshness-watchdog.ts
@@ -183,6 +183,18 @@ export const TABLE_FRESHNESS: Readonly<Record<string, FreshnessExpectation>> = {
   // test doing its job: a new table arrived and was unwatched until named.
   // Pass ledgers are written with their parent table, so they share its
   // cadence.
+  // 0030, added while this map already existed -- the coverage test caught it
+  // the same way it caught 0029's pair. `neurons` is the 15-minute producer,
+  // so its pass ledger is the freshest of the four, but the threshold stays at
+  // the family's 36h: a pass table records COMPLETED passes, and a producer
+  // that stops mid-pass leaves the last completed one behind. Tightening this
+  // would alarm on the producer's cadence rather than on its failure.
+  neurons_passes: {
+    column: "captured_at",
+    kind: "ms",
+    maxAgeMs: 36 * HOUR,
+    reason: "written with neurons",
+  },
   nominator_positions_passes: {
     column: "captured_at",
     kind: "ms",
--- a/tests/changelog-mcp.test.ts
+++ b/tests/changelog-mcp.test.ts
@@ -16,11 +16,30 @@ import { mockEnv, type Row } from "./row-type.ts";
 type ReadArtifact = (env: Env, path: string) => Promise<StorageReadResult>;
 
 const SAMPLE_CHANGELOG = {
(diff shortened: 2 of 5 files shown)


JSONbored:
Superseded by #9843, which fixed the same four failures (format, the BuildSummary coverage fixture, SAMPLE_CHANGELOG, and "129 subnets") plus the `neurons_passes` classification.

Two of us converged on the same repairs from opposite ends — worth noting the shared cause rather than the duplication: every one of these failures had its assertion and its subject in **different files**, so each originating PR was green on its own diff and red on merge. That is the thing to fix, not the individual fixtures.

Closing without merging; `main` already carries all of it.

Issue #9796, opened 2026-08-07
mcp: derive tool output schemas from the route ArtifactSchemas they duplicate


JSONbored:
Part of #9793.

**109 of 114** files in `schemas-src/mcp-tools/` hand-model a response shape. Only **5** derive
from `schemas-src/routes/`. Meanwhile **108 of 112** route files already export a reusable
`ArtifactSchema` describing the same bytes, more precisely (96.9% vs 70.1% declared precision).

The duplication is the direct cause of both the drift and the opacity tracked in this epic. The
copies were written when no route schema existed; route schemas arrived later and the copies
never followed, and nothing detects the divergence.

## Fix

Derive each MCP output schema from the route's `ArtifactSchema`:

- identical shape -> reuse the route schema directly;
- deliberately narrowed for payload size -> express the narrowing as `.pick()` / `.omit()` /
  `.extend()` against the route schema.

`.pick()` is typed against the source shape, so a route field rename becomes a **compile error
in the MCP schema file** instead of silent production drift — which is the durable property
this issue is really buying.

Four route files export no `ArtifactSchema` and need one first: `event-stream-honesty.ts`,
`governance-config-changes.ts`, `subnet-profile.ts`, `sudo.ts`.

Ship in reviewable batches by tool family, not as one diff. Each batch must hold the published
JSON Schema wire-compatible except where it is a documented correction.

JSONbored:
Done. Every tool schema now derives from the route `ArtifactSchema` it mirrors, or declares in writing why it cannot.

Verified on `main` at 5b2601055:

```
validate:single-schema-source  298 published components, all declared in schemas-src
validate:mcp-route-map         225 tools all classified (212 mirror a route,
                               13 declared route-less with a reason);
                               267 published routes all accounted for
validate:mcp                   225 tools, lifecycle + all tools/call
```

The route↔tool map (#9880, merged as #9970) is what makes this durable: the direction this issue cared about — a tool re-declaring a shape its route already models — is now a compile-time relationship, and the gate fails in **both** directions, so neither a new tool without a route nor a new route without a tool can land silently.

Landed across #9818, #9823, #9827, #9884, #9928, #9932, #9946, #9953 and #9970.

PR #9840, opened 2026-08-07
fix: turn main green — four Validate failures, two of them mine

JSONbored:
`main` has been red since at least `20dcf1e4f` with **four** `Validate` failures. Two were mine, two were not, and **no branch cut from main can prove itself green until all four land** — which is why this is one PR rather than three.

| failure | cause |
|---|---|
| `root-is-not-a-subnet` › no source file claims "129 subnets" | **mine (#9816)** |
| `table-freshness-watchdog` › every table in migrations/d1 is classified | **mine (#9816)** |
| `changelog-mcp` › SAMPLE_CHANGELOG validates against GET_CHANGELOG_OUTPUT_SCHEMA | stale fixture |
| `zod-schemas` › build-summary: BuildSummaryArtifactSchema.parse(...) | stale fixture |
| `tests/neon-migrate.test.ts` not prettier-formatted | same as #9838 |

## Mine

**"21 of 129 subnets"** in a comment. 129 is the *netuid* count — root is not a subnet, which is exactly what that test enforces. Fixed in `workers/data-api.ts` and in the `0030` migration header, which repeats the sentence.

**`neurons_passes` unclassified.** `0030` added it and nothing named it in `TABLE_FRESHNESS`. The map's own comment, three entries above where the new one goes, records `0029`'s pair arriving the same way — *"a new table arrived and was unwatched until named."* I read that while writing `0029`'s entries and still missed `0030`'s. It gets `neurons`' 2h threshold because the tally is written in the **same batch** as the rows it describes.

## Not mine — and the schemas are right, the fixtures were stale

Both verified against the live artifacts rather than assumed:

**`changelog-mcp`.** #9796 derived `GetChangelogOutputSchema` from `ChangelogArtifactSchema` instead of copying it, making `schema_version`, `generated_at` and `summary.coverage_delta` required. `GET /api/v1/changelog` carries all three today. `provider_count: null` is kept deliberately — the live artifact reports null for a metric it could not diff, and a fixture that only shows the populated shape never exercises that arm.

**`zod-schemas`.** `meta-contracts.ts` now types build-summary's `coverage` as `CoverageArtifactSchema` rather than an open record, because the build summary embeds the whole coverage card. The fixture was a two-field stub. Replaced with the real card from `GET /api/v1/build` — which is what *"real builder output"* in that test's name is supposed to mean.

**`neon-migrate`.** Prettier only, no behaviour change. Same fix as #9838, included because `format:check` runs in the same job as the tests above, so neither PR could go green without the other. **#9838 can be closed as redundant.**

## A note on how I found these

I misdiagnosed the changelog failure twice before getting it right. First I imported `SAMPLE_CHANGELOG` from `src/` — it is defined in the test file — and validated `undefined`, which reported a meaningless root-level `must be object`. The assertion itself is `assert.ok(validate(...))`, which reports *false* and nothing else; extracting `validate.errors` is what turned it into a one-line answer both times.

Worth flagging separately: an `assert.ok(validate(x))` that discards `validate.errors` makes every future failure of these schema tests cost the same detour.

## Validation

```
vitest changelog-mcp zod-schemas table-freshness-watchdog root-is-not-a-subnet   # 365 passed
npm run typecheck     # clean
npm run lint          # clean
npm run format:check  # clean
```

Refs #9837

### DIFF
--- a/migrations/d1/0030_neurons_passes.sql
+++ b/migrations/d1/0030_neurons_passes.sql
@@ -24,7 +24,7 @@
 -- 30 hours old."
 --
 -- MAX(captured_at) CANNOT SUBSTITUTE, and this is the specific trap: it
--- reflects only the netuids that DID land, so a pass covering 21 of 129 subnets
+-- reflects only the netuids that DID land, so a pass covering 21 of 129 netuids
 -- leaves a perfectly fresh-looking stamp behind it. That is the same shape as
 -- the 147,000-row account_balances incident this whole mechanism came from.
 --
--- a/src/table-freshness-watchdog.ts
+++ b/src/table-freshness-watchdog.ts
@@ -123,6 +123,15 @@ export const TABLE_FRESHNESS: Readonly<Record<string, FreshnessExpectation>> = {
     maxAgeMs: 2 * HOUR,
     reason: "metagraph sync every 15 min",
   },
+  // Added by 0030, the same way 0029's pair was: the coverage test caught it
+  // unwatched. Written in the same batch as `neurons` -- deliberately, so a
+  // tally cannot outlive the rows it describes -- so it shares that cadence.
+  neurons_passes: {
+    column: "captured_at",
+    kind: "ms",
+    maxAgeMs: 2 * HOUR,
+    reason: "written with neurons",
+  },
   neuron_daily: {
     column: "captured_at",
     kind: "ms",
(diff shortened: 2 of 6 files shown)


JSONbored:
Superseded by #9843, which landed the same four fixes while this was in CI.

What #9843 covered: the `neurons_passes` classification, the changelog fixture, the build-summary coverage fixture, the `neon-migrate` formatting, and the `129 subnets` claim in `workers/data-api.ts`.

One thing it did not: the identical `129 subnets` sentence in the `0030` migration header, which is outside `root-is-not-a-subnet`'s scanned file list. Carried forward alone in #9851.

Both of my regressions from #9816 are now fixed on main either way.

PR #9948, opened 2026-08-07
feat(neon): serve the two history routes from Neon

### LINKED ISSUE #9947: The history tables' surrogate id is exposed in a pagination cursor, and Neon's ids are NOT D1's
Found while scoping the read cutover for `account_identity_history`. Blocks moving that route, and any other paginating on a surrogate id.

## The finding

#9896 declared `id BIGINT GENERATED ALWAYS AS IDENTITY` on the four AUTOINCREMENT history tables, on the reasoning that the id is *"a tiebreaker that can never tie"* — the natural key `(account, observed_at)` is unique, so D1's `ORDER BY account, observed_at DESC, id DESC` never reaches it.

That reasoning is correct for **ordering** and wrong about **exposure**. `GET /api/v1/accounts/:ss58/identity-history` paginates with:

```sql
WHERE account = ? AND (observed_at, id) < (?, ?)
ORDER BY observed_at DESC, id DESC LIMIT ?
```

and encodes `(observed_at, id)` into the returned cursor. The id is part of the **public API surface**.

## The ids genuinely differ

Both stores hold 531 rows numbered 1–531, which looks reassuring and is a coincidence of count. The first row by id in each:

```
D1     id=1  account=5HdThE9WvusfeFpk7R3Lnc9wfP4zsDje3GeRew6mrQHNNFiY
Neon   id=1  account=5C5LVkCfXk4LyUGozdk9FC65gauUVuAoRqAPpXdrhHYQeF4y
```

D1 assigned ids in **insertion** order; the backfill inserted into Neon in **keyset** order `(account, observed_at)`. Different orders, so the numbering does not correspond row for row.

## Impact, bounded but real

Within one account, `(account, observed_at)` is unique, so `observed_at` alone determines both the ordering and the page boundary. The id only decides whether the **cursor row itself** is re-included. So a cursor minted against D1 and replayed against Neon can duplicate exactly one row at a page boundary — not lose rows, not reorder.

Transient, and confined to cursors that straddle a cutover. But it means:

1. the route cannot move to Neon while D1-minted cursors are in flight, without accepting that
2. **the id is an observable value**, so "Neon assigns its own" is an API-visible change, not purely internal

## Options

- **paginate on the natural key** — `(observed_at)` alone is unique per account here, so the id can leave the cursor entirely. Cleanest, and makes the cursor store-independent.
- **version the cursor** so a D1-era cursor is rejected rather than silently misread
- **copy D1's ids** — needs `GENERATED BY DEFAULT` plus a sequence fixup, and reintroduces the collision hazard #9896 deliberately avoided

The first is almost certainly right, and it is a change to the route rather than the schema.

Related: #9773 (the read-site port), #9896 (where the ids were chosen).

JSONbored:
Closes #9947. **The first read cutover of this family** — `/subnets/{netuid}/hyperparameters/history` and `/accounts/{ss58}/identity-history` now read Neon. Both tables are at exact parity (137/137, 531/531) and each route reads one table, so there's no side loader to drag along.

## The cursor had to stop carrying a surrogate id

#9896 gave these tables `id GENERATED ALWAYS`, reasoning the id is "a tiebreaker that can never tie". Right about **ordering**, wrong about **exposure** — both routes encoded `(observed_at, id)` into their pagination cursor, putting a store-local surrogate into the public API.

The ids genuinely differ. Both stores hold 531 rows numbered 1–531, which is a coincidence of count:

```
D1    id=1  5HdThE9WvusfeFpk7R3Lnc9wfP4zsDje3GeRew6mrQHNNFiY
Neon  id=1  5C5LVkCfXk4LyUGozdk9FC65gauUVuAoRqAPpXdrhHYQeF4y
```

D1 numbered by insertion order; the backfill inserted in keyset order.

The cursor now carries `observed_at` alone — which was always sufficient, since within the `account = ?` / `netuid = ?` each query already pins, the natural key is unique. An old two-part cursor fails `decodeCursor`'s arity check and falls through to the offset path: restarts a page rather than seeking to a row that isn't there.

## Both scanners were blind to these routes

They read `matchNeuronsD1Route` only; these routes live in `matchHyperparamsIdentityD1Route`. Moving them would have shipped SQL **no portability or dialect check had ever seen** — and the gate caught exactly that.

Fixed by **discovering matchers by name** rather than naming one, so the next matcher can't silently fall outside the scan.

That widening then exposed a second blind spot: `matchHealthStatusD1Route` guards by early return (`if (url.pathname !== "...") return null`), which the block splitter didn't recognise — so its `surface_status` SQL was folding into a neighbour's block and being attributed to `/subnets/movers`. Same misattribution class as #9826, in a form that didn't exist then.

201 tests pass across the 7 affected suites.

### DIFF
--- a/tests/neon-read-routes.test.ts
+++ b/tests/neon-read-routes.test.ts
@@ -126,7 +126,9 @@ function loaderTables(name: string, seen = new Set<string>()): string[] {
  * shows up as a drop in `checked` rather than as silent success.
  */
 function routePath(block: string): string | null {
-  const literal = /pathname === "([^"]+)"/.exec(block)?.[1];
+  const literal =
+    /pathname === "([^"]+)"/.exec(block)?.[1] ??
+    /pathname !== "([^"]+)"/.exec(block)?.[1];
   if (literal) return literal;
   const re = /pathname\.match\(\s*\/\^([^\n]*?)\$\//.exec(block)?.[1];
   if (!re) return null;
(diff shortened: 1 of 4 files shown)


JSONbored:
Superseded by #9956. This branch dropped the surrogate id from the pagination cursor, which the D1 suite correctly rejected: two revisions written in the same millisecond share `observed_at` and are ordered by the id alone. The real problem was deeper — the id is also how the write path picks the latest revision via `MAX(id)` — so #9956 makes Neon's ids match D1's instead, which is the prerequisite for moving these reads at all.

PR #9955, opened 2026-08-07
fix(prober): one host per worker, so our own fan-out stops marking subnets degraded

### LINKED ISSUE #9904: health-prober: 12 surfaces across 8 subnets are published as degraded because our own probe fan-out trips their rate limiters
We publish 12 surfaces across 8 subnets as **degraded** because our own prober triggers their rate limiters. All 12 return `HTTP 200` to a single sequential request.

## The distribution is the proof

Every `rate-limited` verdict in production sits on one of two hosts. There are no others anywhere in the 626 published verdicts:

| host | rate-limited | surfaces we probe there |
|---|---:|---:|
| `api.taomarketcap.com` | **7** | 30 |
| `agent-upload.ridges.ai` | **5** | 13 |
| *every other host* | **0** | — |

If these were the operators rate-limiting the public, the verdicts would be scattered across hosts and correlated with popularity. They are perfectly correlated with **our own fan-out per host**, and with nothing else.

## Independently verified

I re-probed all 1,846 probe-enabled HTTP surfaces myself, one pass, and diffed against what we publish. Every one of the 12 answered `200`:

```
ours=degraded rate-limited   mine=live http=200   sn-62-ridges-problem-statistics      agent-upload.ridges.ai
ours=degraded rate-limited   mine=live http=200   sn-62-ridges-scoring-screener-info   agent-upload.ridges.ai
ours=degraded rate-limited   mine=live http=200   sn-62-ridges-scoring-weights         agent-upload.ridges.ai
ours=degraded rate-limited   mine=live http=200   sn-62-ridges-subnet-api              agent-upload.ridges.ai
ours=degraded rate-limited   mine=live http=200   sn-62-ridges-top-scores-over-time    agent-upload.ridges.ai
ours=degraded rate-limited   mine=live http=200   sn-89-taomarketcap-subnet-api        api.taomarketcap.com
ours=degraded rate-limited   mine=live http=200   sn-92-taomarketcap-subnet-api        api.taomarketcap.com
ours=degraded rate-limited   mine=live http=200   sn-104-taomarketcap-subnet-snapshot  api.taomarketcap.com
ours=degraded rate-limited   mine=live http=200   sn-109-taomarketcap-subnet-api       api.taomarketcap.com
ours=degraded rate-limited   mine=live http=200   sn-115-taomarketcap-subnet-api       api.taomarketcap.com
ours=degraded rate-limited   mine=live http=200   sn-116-taomarketcap-subnet-api       api.taomarketcap.com
ours=degraded rate-limited   mine=live http=200   sn-118-taomarketcap-subnet-api       api.taomarketcap.com
```

## Mechanism

`src/health-prober.ts:536` schedules with a single global limit and no notion of host:

```ts
const PROBE_CONCURRENCY = 8;                       // line 71
const probed = await mapLimit(surfaces, concurrency, async (surface) => { … });
```

`operational-surfaces.json` is sorted `(netuid, surface_id)` (`scripts/build-artifacts.ts:2774`), so **all five `sn-62-ridges-*` surfaces share netuid 62 and are consecutive** — they land in one concurrency-8 window and hit `agent-upload.ridges.ai` simultaneously. taomarketcap's 30 surfaces are spread across netuids 89–118, so adjacency is weaker there, but 30 probes at concurrency 8 is enough sustained pressure on one host.

From there it is mechanical: `classifyProbe` maps `429 → "rate-limited"`, `statusForClassification` maps that to `degraded`, and the subnet's rollup goes with it.

## Why this matters more than a wrong flag

These are **public statements about other people's services**. Subnets 62, 89, 92, 104, 109, 115, 116 and 118 are shown as degraded on a registry whose purpose is telling people whether a subnet's surfaces work. Two teams are being marked down for answering our requests correctly.

It also inflates the degraded count that everything downstream reads — `/api/v1/health` currently reports 91 degraded of 626, and **12 of those are ours**.

## Fix

A per-host budget in the prober. The global `PROBE_CONCURRENCY = 8` can stay; what is missing is that no two in-flight probes should share a hostname, or at most N should. Options in rough order of effort:

1. **Interleave by host before scheduling.** Group surfaces by hostname, then round-robin the groups into the list `mapLimit` walks. Zero new concurrency machinery — it just stops same-host entries being adjacent, which is what the `(netuid, surface_id)` sort creates. Fixes the ridges case outright and thins taomarketcap.
2. **A per-host semaphore** (1–2 concurrent, small delay between same-host probes). Correct in general, more code.
3. **Do not let `rate-limited` mean `degraded`.** A 429 says nothing about whether the surface works — it says we asked too often. Consider mapping it to a distinct status, or holding the prior verdict, so a self-inflicted 429 cannot publish as a health claim about a third party. Worth doing regardless of 1/2, as defence in depth.

## Also seen, weaker signal

5 surfaces are published `timeout` and answered `200` to me, clustered 2/2/1 on three hosts (`api.graphite-ai.net`, `api.alpharidge.ai`, `api.investing88.ai`). Same shape, smaller sample — plausibly the same contention, plausibly genuine slowness. Worth re-checking after the per-host fix rather than filing separately.

## What was verified as correct

For completeness, since this came out of a full diff: the other 25 disagreements were **my** errors, not the prober's. `contentMismatch` correctly whitelists `raw.githubusercontent.com` / `.json` served as `text/plain` (22 cases), and `probeHttpSurface` correctly retries `GET` when a `HEAD` probe returns 400/403/405 (3 cases). My reimplementation had neither. Agreement was 565/614 before those, 590/614 after — the 12 above and the 5 timeouts are what remain.

JSONbored:
## Summary

Every `rate-limited` verdict in production sat on exactly **two hosts** — 7 on `api.taomarketcap.com`, 5 on `agent-upload.ridges.ai` — and **zero** on the other 120. All 12 answered `200` to a single sequential request. The hosts were fine; we were the load.

12 of the 91 degraded surfaces on `/api/v1/health` were ours, and two teams were publicly marked down for answering our requests correctly.

## Cause

`mapLimit` is a flat worker pool over one cursor, so which surfaces are in flight together is decided by **list order alone**. `operational-surfaces.json` is sorted `(netuid, surface_id)` (`scripts/build-artifacts.ts`), which puts a subnet's surfaces **adjacent** — all five `sn-62-ridges-*` share netuid 62, so all five hit one host inside a single concurrency-8 window.

From there it is mechanical:

```
429 → classifyProbe "rate-limited" → statusForClassification "degraded" → the subnet rollup
```

## Fix

The pool walks host **groups**; each worker drains its group serially. Exactly one request per host is ever in flight — by construction, no lock, no semaphore.

```ts
await mapLimit(groupSurfacesByHost(surfaces), concurrency, async (group) => {
  for (const surface of group) {
    probed[slotOf.get(surface) as number] = await probeOne(surface);
  }
});
```

Groups run **longest-first**, because the sweep is now bounded by its largest group rather than by total/concurrency — starting the 61-surface group last would strand it after the pool had nothing else to run.

## Not a wall-time tradeoff

Measured over the 626 monitored surfaces:

| | |
|---|---:|
| distinct hosts | **122** |
| largest group (`raw.githubusercontent.com`) | 61 |
| flat pool at concurrency 8 | ~78 sequential slots |
| **one host per worker** | **~61 sequential slots** |

Fewer, not more — the flat pool's tail contention was costing more than serializing the big groups does. And 122 hosts comfortably exceeds the 8 workers, so none idles.

## Ordering is preserved deliberately

Rows are written back into their **original slots** rather than flattened in group order. `persistToKv` ranks the RPC pool off these rows, and reordering them would quietly reshuffle tie-breaks between equally-healthy endpoints. A scheduling change must not move which endpoint the proxy picks.

## Tests

Added to `tests/health-prober.test.ts` — 114 pass in that file, 1,695 across the five suites that import the prober.

**Pure function:** `probeHostKey` (normalization, and that an unparseable URL groups with its twins rather than collapsing every bad entry into one serial bucket); `groupSurfacesByHost` (grouping, longest-first, original order within a group, every surface exactly once).

**The invariant, over the real `mapLimit`:** no two concurrent requests share a host; the pool is still fully used (peak concurrency 8 with more hosts than workers); a single host is probed serially and nothing is dropped.

**Positive control:** the flat pool it replaces is asserted to *actually* double up on that host — without it, "never two concurrent" could pass simply because the harness never overlapped anything.

**End-to-end through `runHealthProber`:** five same-host surfaces never overlap, and interleaved input hosts come back interleaved in `KV_HEALTH_CURRENT`. The pure-function tests prove the grouping; this proves the *sweep* uses it — reverting the call site would otherwise leave every other test green.

**Mutation-tested.** Reverting the call site to a flat per-surface pool fails with:

```
× five surfaces on one host never overlap
  AssertionError: peaked at 5 concurrent probes on one host
```

and passes again on restore.

## Validation run

```
npm run lint          clean
npm run format:check  All matched files use Prettier code style!
npm run typecheck     clean
vitest health-prober, health-probe-core, analytics, mcp-server, neon-prune
                      1695 passed
```

**Patch coverage:** 72 changed lines in `src/health-prober.ts`, **0 uncovered statements and 0 uncovered branches** — measured by intersecting the v8 report with `git diff -U0 origin/main`, not by reading a file-level percentage.

Rebased onto `336e5511c` and re-verified the change survived.

Closes #9904

### DIFF
--- a/src/health-prober.ts
+++ b/src/health-prober.ts
@@ -69,6 +69,60 @@ export const OPERATIONAL_SURFACES_PATH = "/metagraph/operational-surfaces.json";
 type Row = Record<string, unknown>;
 
 const PROBE_CONCURRENCY = 8;
+
+/**
+ * A surface's scheduling host — the thing that rate-limits us (#9904).
+ *
+ * Falls back to the raw url when it will not parse, so an unparseable entry
+ * groups with its identical twins rather than joining one giant "" bucket that
+ * would then be probed serially.
+ */
+export function probeHostKey(url: unknown): string {
+  try {
+    return normalizedHostname(new URL(String(url)).hostname);
+  } catch {
+    return String(url ?? "");
+  }
+}
+
+/**
+ * Surfaces grouped by host, longest group first, original order within a group.
+ *
+ * WHY THIS EXISTS. `mapLimit` is a flat worker pool over one cursor, so which
+ * surfaces are in flight together is decided purely by list order — and
+ * operational-surfaces.json is sorted `(netuid, surface_id)`
+ * (scripts/build-artifacts.ts), which puts a subnet's surfaces ADJACENT. All
+ * five `sn-62-ridges-*` share netuid 62, so all five hit
+ * agent-upload.ridges.ai inside one concurrency-8 window. The host answered
+ * 429, `classifyProbe` mapped that to `rate-limited`, `statusForClassification`
+ * to `degraded`, and we published a subnet as degraded for answering us.
+ *
+ * Measured 2026-08-07: every `rate-limited` verdict in production — 12 of them
+ * — sat on exactly two hosts (7 on api.taomarketcap.com, 5 on
+ * agent-upload.ridges.ai) and zero on the other 120. All 12 answered 200 to a
+ * single sequential request.
+ *
+ * The pool now walks GROUPS, not surfaces, and each worker drains its group
+ * serially — so exactly one request per host is ever in flight, by
+ * construction, without a lock.
+ *
+ * LONGEST FIRST because the sweep is now bounded by its largest group, not by
+ * total/concurrency: starting the 61-surface group last would strand it after
+ * the workers had nothing else to do. Measured over the 626 monitored surfaces:
+ * 122 distinct hosts, largest group 61, so this bounds the sweep at ~61
+ * sequential slots against the flat pool's ~78 — fewer, not more. 122 hosts
+ * also comfortably exceeds the 8 workers, so none idles.
+ */
+export function groupSurfacesByHost(surfaces: readonly Row[]): Row[][] {
+  const byHost = new Map<string, Row[]>();
+  for (const surface of surfaces) {
+    const host = probeHostKey(surface.url);
+    const group = byHost.get(host);
+    if (group) group.push(surface);
+    else byHost.set(host, [surface]);
+  }
+  return [...byHost.values()].sort((a, b) => b.length - a.length);
+}
 // Warn when a sweep nears the 15-minute Cron Trigger ceiling (~8 min = generous
 // headroom). Early signal to raise concurrency or shard before runs overlap.
 const PROBE_WALLTIME_WARN_MS = 8 * 60 * 1000;
@@ -537,7 +591,7 @@ export async function runHealthProber(
     priorStatus.set(row.surface_key || row.surface_id, row);
   }
 
-  const probed = await mapLimit(surfaces, concurrency, async (surface) => {
+  const probeOne = async (surface: Row): Promise<Row> => {
     const input: ProbeSurface = {
       id: surface.surface_id,
       netuid: surface.netuid,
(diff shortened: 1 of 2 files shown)


JSONbored:
Closing — the implementation hand-rolled host scheduling in the caller (a grouping helper, plus manual result slotting through an identity Map and a cast) instead of teaching the concurrency primitive about it. Reworking so `mapLimit` owns the invariant and the call site is one argument. Same issue, better shape.

Issue #9789, opened 2026-08-07
mcp: six tools exceed 5 s, five of them account-scoped


JSONbored:
Swept **219 of 225** tools against production (6 mutating tools excluded: credential/webhook/alert/watch). Every one called with real arguments. Two findings, one systemic.

## 1. The payload is serialized twice, on every call

`dispatchTool` (`src/mcp-server.ts`) returns both on success:

```ts
content: [{ type: "text", text: JSON.stringify(payload) }],
structuredContent: payload as Row,
```

Measured on `get_health_trends`:

```
content:            487,172 bytes   (text is 436,526 chars)
structuredContent:  478,945 bytes
identical payload:  true
```

REST serves the same body in **436,783 bytes**; MCP serves **923,778**. **2.1×, on every tool.**

The MCP spec makes the text block a **SHOULD**, not a MUST — it exists so clients that predate `structuredContent` still get something readable. So this is a negotiable compatibility cost, not a requirement: a client on protocol ≥ 2025-06-18 understands `structuredContent` and does not need the duplicate.

Halving every response is the single largest optimisation available to this server, and it is one function.

## 2. Seventeen tools still exceed 100 KB

#9730 fixed the tools that page through `applyQueryFilters`. None of these do:

| Tool | Bytes | ms |
|---|---:|---:|
| `get_health_trends` | **923,778** | 219 |
| `get_subnet_ohlc` | **760,130** | 4,372 |
| `list_validator_economics` | **691,668** | 472 |
| `get_coverage_depth` | **562,952** | 246 |
| `list_fixtures` | **493,286** | 246 |
| `get_agent_catalog` | 346,594 | 139 |
| `get_contracts` | 320,880 | 82 |
| `get_subnet_trajectory` | 257,952 | 79 |
| `get_subnet_metagraph` | 205,886 | 105 |
| `list_profiles` | 201,880 | 132 |
| `list_schemas` | 163,842 | 47 |
| `get_subnet_detail` | 149,080 | 336 |
| `list_extrinsics` | 141,140 | 8,027 |
| `get_subnet_profile` | 132,258 | 233 |
| `get_account_extrinsics` | 131,474 | 6,827 |
| `get_block_chain_events` | 107,060 | 771 |
| `get_validator_detail` | 103,010 | 181 |

`get_health_trends` takes **no required arguments** and returns 2 windows × 123 subnets × a daily `points` array — ~230K tokens, larger than a context window.

[...]

PR #9978, opened 2026-08-07
perf(mcp): stop serializing every tool payload twice

### LINKED ISSUE #9977: mcp: every successful tool result serializes its payload twice
Split out of #9789, which bundled three independent findings. This is finding 1 — the systemic one, and the only one that is a single function.

## The measurement

`dispatchTool` (`src/mcp-server.ts`) returns both on every success:

```ts
content: [{ type: "text", text: JSON.stringify(payload) }],
structuredContent: payload as Row,
```

Measured on `get_health_trends`:

```
content:            487,172 bytes   (text is 436,526 chars)
structuredContent:  478,945 bytes
identical payload:  true
```

REST serves the same body in **436,783 bytes**; MCP serves **923,778**. **2.1×, on every tool, on every call.**

## Why it is safe to fix

The MCP spec makes the text block a **SHOULD**, not a MUST, and states the reason inline: *backwards compatibility*. It exists so clients that predate `structuredContent` (introduced in revision **2025-06-18**, alongside `outputSchema`) still receive something renderable.

So the duplicate is not a requirement — it is a compatibility cost, and it should be spent only on the clients that need it.

Two conditions decide that, and both are load-bearing:

1. **The client negotiated 2025-06-18 or later.** Per spec such a client MUST send `MCP-Protocol-Version` on every request after `initialize`, and a server seeing no header MUST assume `2025-03-26`. That default is what makes this safe: a client too old to send the header is also too old to read `structuredContent`, and lands on the compatibility path either way.
2. **The tool publishes an `outputSchema`.** A client that sees no declared output schema has no reason to look at `structuredContent` at all — it renders `content`, and an empty one would be a silent break rather than an optimisation. (All 225 tools publish one today; the condition stays because it is *why* the rule is correct.)

Error results are unaffected — their text is a short `code: message`, not a copy of the payload, and it is what a human reads when a call fails.

## Scope

`dispatchTool`'s success path, plus threading the request's negotiated protocol version onto the MCP context. Findings 2 (17 tools over 100 KB) and 3 (6 tools over 5 s) stay on #9789.

Part of #9793.

JSONbored:
## Summary

Every successful MCP tool result carried its payload **twice** — once as a JSON text block, once as `structuredContent`. On `get_health_trends` that is 924 KB for a body REST serves in 437 KB. **2.1×, on every tool, on every call.**

The MCP spec makes the text block a `SHOULD`, and states the reason inline: *backwards compatibility*. It exists so clients that predate `structuredContent` — introduced in revision **2025-06-18** alongside `outputSchema` — still receive something renderable. That makes it a compatibility cost, not a requirement, and it should be spent only on the clients that actually need it.

## The rule

`toolResultContent` drops the duplicate when **both** hold:

| condition | why it is load-bearing |
|---|---|
| client negotiated **≥ 2025-06-18** | that is the revision that introduced `structuredContent`. Per spec such a client MUST send `MCP-Protocol-Version` on every request after `initialize`, and a server that sees no header MUST assume `2025-03-26`. |
| the tool publishes an **`outputSchema`** | a client that sees no declared output schema has no reason to look at `structuredContent` — it renders `content`, and an empty one would be a silent break, not an optimisation. |

The spec's assumed-`2025-03-26` default is what makes this safe rather than risky: **a client too old to send the header is also too old to read the structured channel, and lands on the compatibility path either way.** Nothing has to opt in, and nothing that works today stops working.

All 225 tools publish an `outputSchema` (now asserted by a test), so the second condition is always true in production. It stays because it is *why* the rule is correct, not a prediction about the future.

**Error results are untouched.** Their text is a short `code: message` — not a copy of the payload — and it is what a human reads when a call fails.

## On the shape of the change

The rule lives in one exported function rather than inline in `dispatchTool`. That is deliberate: the `outputSchema` half of the condition cannot be reached through any real tool, so inline it would have been an untestable branch. As a function it is pinned by direct unit tests across all four combinations, with no coverage annotation.

## Validation

```
npm run lint && npm run format:check && npm run typecheck && npm run build
npm run validate:mcp
npm run validate:contract-drift
npm run validate:schemas
npm run validate:schema-opacity
npm run validate:single-schema-source
npm run validate:mcp-route-map
npx vitest run tests/mcp-server.test.ts tests/mcp-server-branch-coverage.test.ts \
  tests/call-subnet-surface-mcp.test.ts
```

All green — **1,536 tests**, of which **1,528 are the pre-existing MCP suite passing unchanged**. That is the most useful number here: none of those 337 assertions on `content[0].text` sends a protocol header, so every one of them exercises the compatibility path and confirms it still behaves exactly as before.

The 8 new tests cover the other side: each supported version, the absent-header default, an error result on the newest version, and both halves of the condition in isolation.

Patch coverage by diff intersection against `coverage/lcov.info`: **6/6 lines, 8/8 branches, 100%**.

Closes #9977

### DIFF
--- a/src/mcp-server.ts
+++ b/src/mcp-server.ts
@@ -1631,6 +1631,11 @@ interface McpCtx {
   env: Env;
   domain?: string;
   sessionId?: string | null;
+  // #9789: the protocol revision THIS request declared, which decides whether
+  // a tool result still needs the compatibility text block. Optional because
+  // every direct-call test builds a context without a request to read it from
+  // -- and absent means the spec's assumed 2025-03-26, i.e. keep the block.
+  protocolVersion?: string | null;
   clientIp?: string | null;
   // #8994: the session id a single `initialize` will hand back, generated
   // before dispatch so the $mcp_initialize event can carry the session it is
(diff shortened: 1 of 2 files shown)


JSONbored:
Re-swept all 225 tools against production to size this properly, rather than extrapolating from the one tool #9789 measured.

**219 of 225 tools carry a byte-identical duplicate. It is 47.5% of everything this server sends.**

```
duplicate text blocks: 219/225 tools
                       5,185,540 of 10,917,479 bytes = 47.5%
```

The 6 that do not are the mutating tools (credential/webhook/alert), which the sweep does not call.

Effect on the tools over 100 KB — this is what a compliant client sees after the change:

| tool | now | after | |
|---|---:|---:|---|
| `get_subnet_ohlc` | 940,643 | 490,383 | |
| `get_health_trends` | 924,193 | 487,460 | |
| `list_validator_economics` | 691,543 | 362,440 | |
| `get_subnet_detail` | 567,696 | 297,432 | |
| `get_coverage_depth` | 560,734 | 293,023 | |
| `list_fixtures` | 497,250 | 262,579 | |
| `get_subnet_profile` | 424,977 | 223,272 | |
| `get_agent_catalog` | 347,648 | 182,653 | |
| `get_contracts` | 320,882 | 164,535 | |
| `get_subnet_uptime` | 283,745 | 151,731 | |
| `list_subnet_apis` | 271,989 | 141,650 | |
| `get_subnet_trajectory` | 258,229 | 134,797 | |
| `get_subnet_metagraph` | 201,829 | 105,847 | |
| `list_profiles` | 201,985 | 105,739 | |
| `get_subnet_endpoints` | 192,514 | 102,030 | |
| `list_schemas` | 163,844 | 85,867 | **under** |
| `get_subnet_surfaces` | 159,230 | 82,995 | **under** |
| `get_block_chain_events` | 152,048 | 79,910 | **under** |
| `list_extrinsics` | 116,198 | 60,129 | **under** |

**19 tools over 100 KB → 15**, with no change to what any tool actually returns.

The remaining 15 are the genuine fat-payload problem, which stays on #9789 — this PR does not touch what a tool returns, only how many times it is written.

Issue #10003, opened 2026-08-07
validate: a subnet renaming itself on-chain hard-fails CI — 'registry validates' is red on main right now


JSONbored:
`tests/artifacts.test.ts > registry validates` fails on a clean `origin/main` (verified at `fa711aaf2`, in a throwaway worktree with no local changes):

```
sn-99: curated name "Leoma" (netuid 99) matches the on-chain identity of netuid 36,
not its own — possible mis-keyed overlay.
```

## It is not a mis-keyed overlay

The heuristic's own hedge is the actual explanation. From `registry/native/finney-subnets.json`:

| netuid | on-chain `subnet_name` | our stored `name` | manifest |
|---|---|---|---|
| 36 | **`Leoma`** | `Eirel` | `registry/subnets/eirel.json` |
| 99 | `Thirty Spokes` | **`Leoma`** | `registry/subnets/leoma.json` |

Subnet **36 renamed itself on-chain from Eirel to Leoma**. Our stored label for 99 is still `Leoma` (its own on-chain name is now `Thirty Spokes`). Both appear in the same run's "26 manifest label(s) now stale" list, which is reported as *informational* — the served name already follows the chain since #9748.

So the two names collide, but nothing is mis-keyed: `leoma.json` really is netuid 99, and `eirel.json` really is netuid 36.

## Why this matters beyond the red test

`scripts/validate.ts:1993` turns this into a **hard assertion failure**, and the input is entirely operator-controlled — the assertion text says so itself:

> on-chain identity names are operator-controlled and may be duplicated maliciously or accidentally

Any subnet operator can therefore red-line our CI by renaming their subnet on-chain to a string we happen to store as another subnet's stale label. Today that happened by accident. The same guard reads as a deliberate availability hazard.

Note `node scripts/validate.ts` at the repo root exits **0** — the failure only reproduces through the test's sandboxed tree, which is why it can sit red without being obvious locally.

## What to decide

1. **Demote to a warning** and let the existing stale-label report carry it — consistent with #9748 having already made the *served* name follow the chain, so no output is wrong.
2. **Keep it fatal but scope it to a genuine mis-key** — only fire when the colliding manifest's `netuid` does not match its own on-chain row, which is the condition the message actually claims.
3. **Refresh the stale labels** (`leoma.json` → `Thirty Spokes`, `eirel.json` → `Leoma`) so the collision disappears. Fixes today's red; the guard stays weaponizable tomorrow.

[...]

PR #10004, opened 2026-08-07
feat(registry): re-check the URLs the registry asserts, after the PR merges

### LINKED ISSUE #9870: adapters: 'unsafe URL' is what we report for dead DNS, and the two SSRF guards disagree on that exact case
Nine of 65 adapter snapshots are `degraded` in `GET /api/v1/freshness`, all 5.4h old against a 24h limit — so their **fetch** is failing, not their freshness. The causes are not the same, and two of them are ours.

| adapter | recorded error | content-type | whose |
|---|---|---|---|
| `sn-120` | `unsafe URL` | — | **ours** |
| `sn-94` | `unsafe URL` | — | **ours** |
| `sn-10` | `response was not JSON` | `text/yaml; charset=utf-8` | **ours** |
| `sn-100`, `sn-37`, `sn-58` | HTTP 404 | — | upstream |
| `sn-102` | HTTP 403 | `text/html` | upstream |
| `sn-36` | HTTP 521 | — | upstream |
| `sn-70` | HTTP 502 | — | upstream |

## 1. "unsafe URL" is what we say when DNS does not resolve

`https://api.affine.io/openapi.json` and `https://autoresearch.bitsota.com/openapi.json` are both reported as `unsafe URL`. Neither is unsafe — **neither resolves**:

```
api.affine.io              -> NO A RECORD (NXDOMAIN)
www.affine.io              -> 104.21.77.43, 172.67.204.119     (same domain, resolves fine)
autoresearch.bitsota.com   -> NO A RECORD, CNAME und2zuh8bf.eu-central-1.awsapprunner.com  (dead App Runner)
```

The path is `scripts/lib.ts`:

```ts
const records = await resolver(host, { all: true, verbatim: true });
if (records.length === 0 || records.some((r) => isUnsafeIpAddress(r.address))) {
  return [];          // -> safeFetch: { ok: false, unsafe: true } -> "unsafe URL"
}
```

Failing closed on an unresolvable host is a defensible **security** posture. Reporting it as *unsafe* is a **diagnostics** bug: it sends whoever reads it into an SSRF investigation when the actual problem is a registry entry pointing at a hostname that no longer exists. Both cases here are dead DNS, and both looked like our guard misfiring until resolved by hand.

`"unresolvable host"` and `"blocked: resolves to a private address"` are different facts and should read differently.

## 2. The two guards disagree on that exact condition

The Worker-side guard reaches the opposite verdict on the same input:

```ts
// src/health-prober.ts — workerResolvedUrlSafetyGuard
// No confirmed private answer / DoH failure → fail open.
return answers.some(isUnsafeIpAddress);   // [] → false → SAFE
```

```ts
// scripts/lib.ts — resolvePublicUrlAddresses
if (records.length === 0 || ...) return [];   // [] → UNSAFE
```

Same URL, same NXDOMAIN, two answers depending on which path evaluates it — the build-time snapshot blocks it, the runtime prober would fetch it. Neither is obviously wrong on its own; having both, undocumented, is. Confirmed directly: `isUnsafePublicUrl` (the literal guard both share) returns `false` for all four URLs tested, so the divergence is entirely in the DNS layer.

## 3. YAML OpenAPI specs are not parsed

`sn-10` publishes its spec as `text/yaml; charset=utf-8` and we record `response was not JSON`, `captured_count: 0`. YAML is at least as common as JSON for OpenAPI — the spec's own examples are YAML — so this is a capability gap rather than a broken subnet. One adapter today; it will recur.

The snapshot fetch already sends `accept: application/json`, which is honest about what it can handle, but the surface is still registered as an `openapi` surface and still reports degraded.

## 4. Six adapters point at dead upstreams

404 ×3, 403, 502, 521. Not our bugs, but they are registry data-quality facts: those subnets' `openapi` surfaces do not currently serve. Worth a pass against the enrichment queue, since the registry's premise is that a listed surface is a live one.

## Suggested split

- Rename the failure mode so dead DNS stops reading as a security block, and reconcile (or document) the fail-open/fail-closed divergence between the two guards. That is the part with the highest confusion cost.
- YAML parsing for OpenAPI capture is its own small change.
- The six dead upstreams belong in the registry-curation lane, not here.

### LINKED ISSUE #9907: surfaces: 1,233 surfaces declare probe.enabled and are never probed — 28 are broken, including three subnets whose website 404s
1,233 surfaces declare `probe: {enabled: true, …}` and have **no published health verdict**. I probed them; **28 are broken** and nothing in the system knows.

## What is actually down

```
dead (404)     netuid 14   docs         https://cacheon.ai/docs/miners/api-contract
dead (404)     netuid 36   website      https://eirel.ai/
dead (404)     netuid 36   docs         https://eirel.ai/docs
dead (404)     netuid 36   docs         https://eirel.ai/whitepaper.pdf
dead (404)     netuid 36   dashboard    https://leaderboard.eirel.ai
dead (404)     netuid 37   docs         https://new-collector-api-production.up.railway.app/docs
dead (404)     netuid 37   openapi      https://new-collector-api-production.up.railway.app/openapi.json
dead (404)     netuid 58   website      https://handshake58.com
dead (404)     netuid 58   docs         https://handshake58.com/skill.md
dead (404)     netuid 58   openapi      https://handshake58.com/openapi.json
dead (404)     netuid 92   source-repo  https://github.com/tensorclaw/tensorclaw
dead (404)     netuid 100  openapi      https://chain.joinbase.ai/openapi.json
dead (404)     netuid 103  source-repo  https://github.com/Djinn-Inc/djinn
dead (404)     netuid 120  source-repo  https://github.com/AffineFoundation/affine-cortex
timeout        netuid 94   openapi      https://autoresearch.bitsota.com/openapi.json
timeout        netuid 94   docs         https://autoresearch.bitsota.com/api/v1/disclaimer.md
timeout        netuid 94   website      https://bitsota.ai/
timeout        netuid 120  openapi      https://api.affine.io/openapi.json
transient(521) netuid 36   openapi      https://api.eirel.ai/openapi.json
transient(502) netuid 66   website      https://ninja66.ai/
transient(502) netuid 70   openapi      https://api.nexisgen.ai/openapi.json
transient(521) netuid 92   website      https://www.tensorclaw.ai/
transient(500) netuid 97   docs         https://us-east-1.hippius.com/docs
transient(500) netuid 116  website      https://taolend.io/
transient(502) netuid 125  website      https://8ball125.com/
unsupported(402) netuid 10 website      https://www.taofi.com/pool
unsupported(405) netuid 80 dashboard    https://www.f2pool.com/mining-user-ltc/…
unsupported(405) netuid 123 website     https://mantis123.com/
```

**Three subnets have a hard-404 website**: 36 (Eirel), 58 (Handshake58) — whose entire domain 404s including its `openapi.json` — and effectively 92 (repo gone, site 521). We publish those URLs as the subnet's website.

Three `source-repo` links point at repositories that no longer exist. Seven `openapi` surfaces are unreachable, which is the specific input `get_api_schema` needs.

## Why nothing catches it

`/api/v1/health` scopes itself to operational kinds (`scope: "operational"`), and the split is exactly along kind:

| | with a verdict | without |
|---|---:|---:|
| `subnet-api`, `data-artifact`, `sse`, `subtensor-rpc/wss` | 626 | 0 |
| `dashboard` | 0 | 436 |
| `docs` | 0 | 394 |
| `source-repo` | 0 | 168 |
| `website` | 0 | 146 |
| `openapi` | 0 | 70 |
| `sdk` / `example` / `repo-registry` | 0 | 19 |

Scoping the *health rollup* to callable surfaces is defensible — a docs page being down is not a subnet outage. The problem is that these surfaces still **declare `probe.enabled: true`**, so the registry asserts they are meant to be checked, and then nothing checks them. Of the 1,233, 919 have `last_verified_at: null` and the other 314 were last verified **2026-07-16 at the newest — 22 days ago**.

The result is a registry that confidently serves a subnet's website URL with no signal that it has been 404ing.

## Two things to decide

**1. Who probes these?** Either bring the non-operational kinds into the prober (they are cheap — mostly `HEAD`, and `openapi` at least should be probed since it feeds schema capture), or set `probe.enabled: false` on kinds nothing will ever check, so the registry stops claiming they are monitored.

**2. Keep them out of the health rollup either way.** A 404 docs page should not make a subnet read as degraded. This wants a separate signal — a link-rot report, or a `link_status` distinct from `health.status` — not a change to what `/api/v1/health` means.

`openapi` is the one I would not leave out: 7 of 70 are unreachable right now, and #9893 is a separate case of an `openapi` surface being misreported. Whatever answers "can we still fetch this subnet's schema" should be automatic.

Found while auditing subnet registry accuracy: I independently probed all 1,846 probe-enabled HTTP surfaces (see #9904 for the other half of that sweep) and diffed against what we publish. Related: #9906 (`stale: false` on never-verified surfaces — the same 1,233 are the bulk of it).

### LINKED ISSUE #9914: registry: 355 surfaces cite a dead source_url — the exact condition that auto-closes a contributor PR, never re-checked after merge
A dead `source_url` is a **deterministic auto-close** for a contributor PR. In the merged registry, **355 surfaces across 25 subnets cite a `source_url` that is dead right now**, and nothing ever re-checks one after merge.

## The numbers

I probed every `source_url` in `registry/subnets/*.json` — 777 distinct URLs, 3,666 references:

| | |
|---|---:|
| distinct `source_url`s | 777 |
| **broken** | **55** |
| surface references to a broken one | **355** (9.7%) |
| subnets affected | **25** |

Breakdown: 28 × `404`, 14 × DNS-fail/timeout, 4 × `502`, 3 × `521`, 3 × `503`, 2 × `530`, 1 × `500`.

Affected netuids: 4, 12, 14, 16, 36, 37, 43, 58, 66, 70, 74, 92, 94, 98, 99, 100, 102, 103, 106, 111, 116, 120, 121, 125, 126.

## Four dead OpenAPI documents underpin 237 surfaces

The damage is concentrated, and the shape explains why:

```
121 surfaces   HTTP 521   https://api.eirel.ai/openapi.json                                (netuid 36)
 58 surfaces   HTTP 404   https://new-collector-api-production.up.railway.app/openapi.json (netuid 37)
 32 surfaces   timeout    https://autoresearch.bitsota.com/openapi.json                    (netuid 94)
 26 surfaces   HTTP 404   https://chain.joinbase.ai/openapi.json                           (netuid 100)
```

These are specs whose individual endpoints were expanded into one `subnet-api` surface each — which is why a single URL backs 121 rows. **The spec is gone, so not one of those 121 surfaces can be re-verified, and none of them can be re-derived if the expansion ever needs redoing.** Netuid 37's is a `railway.app` preview deployment, which is a URL class that was always going to expire.

Also gone: three GitHub repos cited as evidence (`Djinn-Inc/djinn`, `AffineFoundation/affine-cortex`, `tensorclaw/tensorclaw` — confirmed 404 against `api.github.com`, see #9912), the docs pages at `cacheon.ai/docs/miners/api-contract` and `connito.ai/docs/reference/sn-owner-api`, and `learnbittensor.org/subnets/12`.

## The asymmetry is the point

`CLAUDE.md` states the rule plainly — a contributor PR is auto-**CLOSED** on a "deterministic fail (duplicate / dead `source_url` / private URL / secret)". The gate fetches the `source_url` and rejects the PR if it 404s.

We apply that standard exactly once, at merge, and never again. A contributor submitting `https://api.eirel.ai/openapi.json` today would have their PR closed automatically; the 121 surfaces already merged on that evidence keep their `authority` and `curation_level` untouched.

This is not an argument for leniency at the gate — it is that **evidence decays and nothing measures the decay.** `last_verified_at` tracks the *surface*, not the evidence behind it, and 2,731 surfaces do not have even that (#9906).

## What this changes about the surfaces themselves

Nothing automatic, and that is deliberate — a dead `source_url` does not mean the surface is dead. Netuid 36's `api.eirel.ai/openapi.json` returns 521 while several of its `subnet-api` surfaces still answer. The claim that decays is *"we can still show this subnet publishes this"*, which is the claim the registry exists to make.

So the fix is a signal, not a purge.

## Fix

1. **Probe `source_url`s on a cadence** and record the result per surface — a `source_status` / `source_last_checked` beside `last_verified_at`. Cheap: 777 distinct URLs, and it is the same fetch the gate already performs for a PR.
2. **Surface it in curation.** A surface whose only evidence 404s should not hold the same `curation_level` as one whose evidence resolves. The demotion machinery already exists (`resolveSurfaceCurationLevel` demotes on stale verification) — this is another input to it.
3. **Prioritise the four OpenAPI docs.** 237 of the 355 references hang off them. Either the subnet republished the spec somewhere (find it, update the evidence) or those surfaces are now unverifiable and should be marked as such.
4. Consider rejecting ephemeral hosts as evidence at intake — `*.up.railway.app`, preview deployments, and similar are guaranteed link-rot. Netuid 37's 58 surfaces are the demonstration.

## Adjacent, found in the same pass

`sn-101-eni-website` (netuid 101) is registered as `http://tag101.ai/` — plain HTTP, the only non-`https` HTTP surface in the registry (the four `wss://` RPC endpoints are legitimately not HTTPS). It is worse than a scheme problem: `https://tag101.ai/` fails outright, and `http://tag101.ai/` **redirects to `https://github.com/tag101-ai/tag101`** — so a surface registered as this subnet's `website` is in fact a redirect to a source repo, over an insecure scheme.

Otherwise the registry's structural integrity is clean, for the record: **0** duplicate URLs within any subnet file, **0** surfaces citing a provider with no `registry/providers/` file (137 provider files, all referenced ones present), and only 23 of 2,740 surfaces carry no `source_url` at all (14 on netuid 7, 5 on 74, 4 on 0 — all first-party/maintainer-curated).

Found while auditing every tracked subnet for registry accuracy.

### LINKED ISSUE #9917: registry: 21 provider URLs are dead and nothing probes them — including our own gittensory entry, which 5 surfaces still cite
21 of 407 URLs in `registry/providers/*.json` are dead. Nothing probes them — the health prober covers subnet surfaces, not provider identity.

## Broken

```
http=522  alphacore.json          website_url  https://alpha-core.ai/
http=0    alveus-labs.json        website_url  https://bitsota.ai/
http=404  backprop-finance.json   docs_url     https://backprop.finance/dtao/subnets/
http=404  bitads.json             github_url   https://github.com/FirstTensorLabs/BitAds
http=0    bitads.json             docs_url     https://bitads.ai/docs
http=0    bitads.json             website_url  https://bitads.ai/
http=404  djinn.json              github_url   https://github.com/Djinn-Inc/djinn
http=530  dogelayer.json          website_url  https://dogelayer.ai/
http=502  eightball.json          website_url  https://8ball125.com/
http=404  eirel.json              website_url  https://eirel.ai/
http=0    gittensory.json         docs_url     https://gittensory.aethereal.dev
http=404  handshake58.json        docs_url     https://handshake58.com/skill.md
http=404  handshake58.json        website_url  https://handshake58.com
http=404  herald.json             social.x     https://x.com/heraldmedia_ai
http=0    nodexo.json             docs_url     https://docs.nodexo.ai/
http=530  oneoneone.json          website_url  https://oneoneone.io/
http=0    shiftlayer.json         website_url  https://play.shiftlayer.ai/
http=500  taolend.json            website_url  https://taolend.io/
http=521  tensorclaw.json         website_url  https://www.tensorclaw.ai/
http=502  unarbos.json            website_url  https://ninja66.ai/
http=0    vectorchat.json         website_url  https://subnet.chunking.com/
```

By field: 14 `website_url`, 5 `docs_url`, 2 `github_url`, 1 `social.x`. Every provider file carries a `website_url`, so **roughly 1 in 10 provider websites is down** — `bitads.json` has all three of its links dead.

## Our own entry is one of them

`registry/providers/gittensory.json` is first-party and stale in three ways:

```json
{
  "id": "gittensory",
  "name": "Gittensory",
  "website_url": "https://gittensory.aethereal.dev",   ← DNS fails (000)
  "docs_url":    "https://gittensory.aethereal.dev",   ← same, dead
  "github_url":  "https://github.com/JSONbored/gittensory"
}
```

`gittensory` was renamed to **loopover** (cutover verified 2026-07-22). The GitHub URL survives only because GitHub auto-redirects a renamed repo — it resolves to `https://github.com/JSONbored/loopover` — so it passes a probe while pointing at a name we no longer use. `loopover.aethereal.dev` does not resolve either, so the site appears to be gone rather than moved.

**5 surfaces on netuid 74 (`registry/subnets/gittensor.json`) cite `provider: "gittensory"`**, so this is not an orphan record.

The logo (`metagraph.sh/logos/gittensory.png`) is first-party and fine — #8288 cached provider logos locally, which is why it is the one link that cannot rot.

## Why nothing catches it

`probe.enabled` lives on *subnet surfaces*. Provider files carry `website_url` / `docs_url` / `github_url` / `logo_url` / `social.*` with no probe config at all, so they were never in scope for the prober and have no `last_verified_at` equivalent. `/api/v1/providers` serves them as-is.

This is the third population in the registry with the same gap — subnet surfaces the prober skips (#9907), `source_url` evidence (#9914), and now provider identity. The pattern is that only one of the four URL populations in the registry is checked.

## Fix

1. Probe provider URLs on a cadence — 407 distinct URLs, the same fetch the surface prober already does. A `link_status` per field is enough; provider links do not need the full health model.
2. Fix `gittensory.json` ourselves: it is our own record, the rename is ours, and either the site moved or the entry should say so. Renaming the provider `id` would touch the 5 netuid-74 surfaces that reference it, so that is a decision, not a typo fix.
3. `bitads.json` (all three links dead) and the four subnets whose provider *and* surfaces are both down — 36, 58, 92, 125 — look like genuinely departed teams rather than transient outages, and are worth a curation decision rather than a link fix. Note netuid 16's provider is `bitads` while the subnet itself was renamed to **TC** (#9909); same team, two stale records.

Found while auditing every tracked subnet for registry accuracy.

### LINKED ISSUE #9982: registry: one daily lane to re-check the 1,067 reference URLs nothing verifies after merge
Umbrella for #9914, #9917 and #9907, which are three symptoms of one gap: **the registry asserts URLs are real, and only checks them on the way in.** A dead `source_url` is a deterministic auto-CLOSE for a contributor PR; after merge, nothing looks again.

## The measured populations

| population | refs | distinct URLs | broken today |
|---|---:|---:|---:|
| surface `source_urls` (#9914) | 777 | | 355 refs / 25 subnets |
| provider URLs (#9917) | 384 | | 21 |
| non-operational surface URLs (#9907) | 494 | | 28 |
| **union** | 1,655 | **1,067** | |

They overlap heavily — a subnet's repo is often also its provider's `github_url` and a `source_url` on three of its surfaces — so the unit of work is the **URL**, with citations fanning back out.

## Why not just widen the health prober

`OPERATIONAL_SURFACE_KINDS` limits the prober to `subtensor-rpc/wss`, `archive`, `subnet-api`, `sse`, `data-artifact` — things with uptime. The 1,233 surfaces in #9907 are `docs`/`website`/`source-repo`/`openapi`/`dashboard`/`sdk`: references, not services. They have no uptime, they must never enter incidents or the pool breaker, and checking them every 15 minutes would be 1,067 pointless requests an hour. Separate concern, separate cadence.

## Constraints that shaped it

- **1000 subrequests per invocation.** 102 URLs are our own `metagraph.sh/logos/*` — verified against files on disk by a build gate (free, deterministic, blames the commit that dropped the file). That leaves 965 network checks, inside one daily invocation, with a per-run cap rotating the remainder by least-recently-checked.
- **`unsafe URL` was what we reported for dead DNS (#9870).** Both `dead` and `unsafe` are in `DEMOTING_CLASSIFICATIONS`, so the wrong reason already reached production verdicts. Had to be fixed first or the lane would inherit it for every NXDOMAIN link.
- **False positives are the real risk.** 25 of the audit's apparent breakages were the harness, not the target — 403 is Cloudflare, 429 is us. Reuse `probeUrl`/`classifyProbe`; never hand-roll status handling.
- **Reusing github-signals for the 165 repo roots was considered and rejected**: it DROPS a gone repo (#9969) rather than flagging it, so absence there conflates "gone" with "never tracked". Deriving death from absence is the exact error #9969 fixed.

## Decisions taken

1. **A dead link demotes after 3 consecutive daily failures**, through the existing `DEMOTING_CLASSIFICATIONS` machinery. Only verdicts meaning "did not serve the document" accrue a strike; a surface loses verification when EVERY `source_url` is dead, since those are alternative proofs of one claim.
2. **GitHub URLs go through the authed contents API**, not the HTML page — github.com serves a soft-404 that reads as 200, and the token lifts the limit from 60/hr to 5000/hr.
3. **One PR**, since the SSRF-guard fix and the lane are the same subject.

JSONbored:
## Summary

A dead `source_url` is a deterministic **auto-CLOSE** for a contributor PR. Nothing re-checked it after merge — so the bar only applied on the way in. Today **355 surfaces across 25 subnets** cite a dead `source_url`, and **21 provider URLs** are dead including our own `gittensory` entry, which five surfaces still cite.

This adds a daily link-rot lane over the three URL populations the health prober deliberately does not cover, and fixes the SSRF-guard bug that had to land first.

## Why this is not the health prober

`OPERATIONAL_SURFACE_KINDS` limits the prober to `subtensor-rpc/wss`, `archive`, `subnet-api`, `sse`, `data-artifact` — things with uptime. The 494 surfaces in #9907 are `docs`/`website`/`source-repo`/`dashboard`/`sdk`: **references, not services**. They have no uptime, they must never reach incidents or the pool breaker, and probing them every 15 minutes would be 1,067 pointless requests an hour. Separate concern, separate cadence — a widening of the prober would have corrupted the numbers that describe live services.

## The budget, and what it bought

The three populations overlap heavily — 777 + 384 + 494 references are only **1,067 distinct URLs**, so the unit of work is the URL and citations fan back out to whoever asserted it.

| | URLs | how |
|---|---:|---|
| `metagraph.sh/logos/*` | 102 | **filesystem gate at build** — free, deterministic, blames the commit that dropped the file |
| everything else | **965** | network, inside one invocation's **1000-subrequest** ceiling |

`LINK_STATUS_MAX_CHECKS_PER_RUN` caps a tick at 900 and orders by least-recently-checked, so the 65-URL remainder rotates in next tick rather than starving — every URL covered within ~2 days, no cross-day split needed.

Reusing github-signals for the 165 repo roots **was considered and rejected**: that lane *drops* a gone repo (#9969) rather than flagging it, so absence there conflates "gone" with "never tracked" and "rate-limited". Deriving death from absence is the exact error #9969 existed to fix. All 429 github.com URLs go through the GitHub API instead, which answers directly — and catches the rot github-signals structurally cannot see, where the repo is alive and the file moved.

## Following today's patterns

- **Lane shape** — `src/link-status-sync.ts` mirrors `github-signals-sync.ts` (#233): daily cron → R2 store → build projects. Freshness never depends on a bot PR.
- **Same loader both sides** — build and `validate.ts` read the store through one loader, exactly like `loadSurfaceProbeEvidence`, so they cannot disagree and fail artifact parity.
- **Reuse, don't reimplement** — `probeUrl` + `classifyProbe`, not a hand-rolled checker. 25 of the audit's apparent breakages were the harness, because it didn't know 403 means Cloudflare and 429 means us.
- **`mapLimit` + `groupKey`** (#9959) for per-host serialization. `api.github.com` is exempted per-URL: that invariant exists to protect a subnet's single-box API, not GitHub, and serializing its 429 checks would be the entire runtime.
- **Outcome gates in `validate.ts`** (#9963, #9969) rather than call-site checks.

## What a dead link costs

Only after **3 consecutive daily failures**, and only for verdicts meaning "did not serve the document". 401/403/429 never accrue — they prove the host is alive. 5xx and timeouts *do*, precisely because the streak rule makes them safe: the largest rot cluster is `api.eirel.ai`, serving Cloudflare **521 behind live DNS for 121 surfaces**; excluding 5xx would have made the lane blind to its biggest case.

A surface loses verification when **every** `source_url` is dead — alternative proofs of one claim, not a checklist, so one moved README cannot mass-demote a subnet that documented itself twice.

## Fixes #9870, which this depended on

`unsafe URL` was what we reported for a host that no longer resolves. Since **both** `dead` and `unsafe` are in `DEMOTING_CLASSIFICATIONS`, the wrong reason was already reaching production verdicts — and a link lane built on that guard would have mislabelled every NXDOMAIN link.

`resolveUrlSafety` now separates three cases, matching the vocabulary `resolvedWebhookUrlStatus` already uses: `unsafe` (guard or private address), `unresolvable` (authoritative NXDOMAIN), `resolve-error` (a transient failure — **ours**, never scored as the target being dead). Verified against the two hosts in the issue:

```
unresolvable — "host does not resolve"     api.affine.io          (was: unsafe URL)
unresolvable — "host does not resolve"     autoresearch.bitsota.com
```

`isUnsafeResolvedUrl`/`resolvePublicUrlAddresses` keep their exact signatures, so all 8 callers are unchanged by construction.

## Also in this diff

- **Logo referential-integrity gate** — every served logo URL must resolve to a shipped file (75 hashed cache entries + 101 curated). This is what makes checking our own assets from disk *sound* rather than merely cheaper; without it the self-hosted subset would simply be unchecked. Currently 0 missing.
- **Double-encoding fix**, caught by its own test: `URL.pathname` is already percent-encoded, so `encodeURIComponent` produced `my%2520file.md` — a false 404 for every path with a space.
- **Two dead branches removed** rather than annotated away (`pathSegments`' unreachable catch, a `githubApiUrl(url) || url` that could never be taken).

## Validation

```
npm run typecheck                      clean
npm run lint / format:check            clean
npm run build                          passed
npm run validate                       Validated 129 native, 129 overlays, 3441 surfaces, 137 providers, 2037 candidates
vitest (link-status, public-safety,    123 passed
       adapter-snapshot, surface-verification)
```

**Patch coverage** (diff ∩ v8, `src/**` + `workers/**`): statements **100%**, branches **100%** — `link-status-core` 64/64, `link-status-sync` 48/48, `surface-verification` 8/8, `workers/api.ts` 4/4 (the cron-chain false arms are covered once a sibling cron test runs).

**Proven able to fail**, not just passing:
- `surfaceEvidenceIsDead` `every`→`some` → the ALL-not-ANY test fails
- `LINK_DEAD_STRIKES` 3→1 → 4 tests fail
- dropping `transient` from the strike set → the api.eirel.ai test fails
- hiding one logo file → `served logo URLs must resolve to a shipped file (1)`

## Noted, not fixed here

`tests/artifacts.test.ts > registry validates` is **red on clean `origin/main`** (reproduced in a throwaway worktree at `fa711aaf2`), unrelated to this change: subnet 36 renamed itself on-chain to `Leoma`, colliding with our stale stored label for netuid 99. Filed as #10003 with the evidence — it needs a registry-data judgement, and a reviewer would be surprised to find it in a link-rot PR.

Closes #9907
Closes #9914
Closes #9917
Closes #9870
Closes #9982

### DIFF
--- a/scripts/build-artifacts.ts
+++ b/scripts/build-artifacts.ts
@@ -69,6 +69,7 @@ import {
   writeJson,
   loadSurfaceProbeEvidence,
 } from "./lib.ts";
+import { deadLinkUrls, loadLinkStatusRecords } from "./link-status.ts";
 import {
   buildAgentReadiness,
   buildCoverageDepthArtifact,
@@ -221,8 +222,20 @@ const nativeByNetuid = new Map(
 // editing a registry file. Loaded through lib's shared loader so this and
 // validate.ts cannot disagree about the evidence and fail artifact parity.
 const surfaceProbeEvidence = await loadSurfaceProbeEvidence();
+// #9914: the link lane's confirmed-dead URLs. A surface whose source_urls have
+// ALL died has lost the evidence that it exists, and loses its verification the
+// same way a dead probe takes it — see surfaceEvidenceIsDead. Loaded here (and
+// re-loaded identically by validate.ts) so the two cannot disagree and fail
+// artifact parity, exactly like surfaceProbeEvidence above.
+const linkStatusRecords = await loadLinkStatusRecords();
+const deadLinks = deadLinkUrls(linkStatusRecords);
 const surfaces: Row[] = withSurfaceFreshness(
-  flattenSurfaces(activeOverlays, surfaceProbeEvidence, nativeByNetuid),
+  flattenSurfaces(
+    activeOverlays,
+    surfaceProbeEvidence,
+    nativeByNetuid,
+    deadLinks,
+  ),
   Date.parse(nativeSnapshot.captured_at),
 );
 // #1002: dedup candidate ↔ curated surface. A candidate that shares a curated
--- a/scripts/lib.ts
+++ b/scripts/lib.ts
@@ -45,6 +45,38 @@ type Row = Record<string, unknown>;
 // Deliberately not a general-purpose knob: unset (the normal case, including
 // all of CI's own build/publish steps) it resolves exactly as before.
 const defaultRepoRoot = fileURLToPath(new URL("..", import.meta.url));
+/**
+ * Join `relative` onto `root`, or return null if it would escape.
+ *
+ * Registry files are CONTRIBUTOR-SUBMITTED, so any path derived from one is
+ * attacker-influenced: a `logo_url` of
+ * `https://metagraph.sh/logos/../../../../etc/passwd` survives a naive
+ * `path.join`, which normalizes `..` away rather than rejecting it. Decoding
+ * happens before the check because `%2e%2e` is also `..`.
+ *
+ * Containment is asserted on the RESOLVED path with a trailing separator, so a
+ * sibling directory sharing a name prefix (`/public-evil` vs `/public`) cannot
+ * pass a bare startsWith.
+ */
+export function resolveWithinRoot(
+  root: string,
+  relative: string,
+): string | null {
+  let decoded: string;
+  try {
+    decoded = decodeURIComponent(relative);
+  } catch {
+    // A malformed percent-escape is not a path we will look up.
+    return null;
+  }
+  const resolvedRoot = path.resolve(root);
+  const resolved = path.resolve(resolvedRoot, `.${path.sep}${decoded}`);
+  return resolved === resolvedRoot ||
+    resolved.startsWith(resolvedRoot + path.sep)
+    ? resolved
+    : null;
+}
+
 export const repoRoot = process.env.METAGRAPH_REPO_ROOT
   ? // Trailing separator to match fileURLToPath's directory form, so the
     // handful of `path.relative(repoRoot, …)` / string-prefix callers behave
--- a/scripts/lib.ts
+++ b/scripts/lib.ts
@@ -16,6 +16,7 @@ import {
 import { entityLabelsIndex } from "../src/entity-labels.ts";
 import {
   isProbeDemoted,
+  surfaceEvidenceIsDead,
   verifyFromProbeEvidence,
   type SurfaceProbeRecord,
 } from "../src/surface-verification.ts";
@@ -1007,6 +1008,11 @@ export function flattenSurfaces(
   subnets: Row[],
   surfaceProbeEvidence: Record<string, SurfaceProbeRecord> = {},
   nativeByNetuid?: Map<unknown, Row>,
+  // Optional like nativeByNetuid (#9963), so every read-only caller --
+  // capture-fixtures, probes-smoke, snapshot-openapi, tests -- is unchanged by
+  // construction. Absent means "the link lane has no verdict", which is not the
+  // same as "nothing is dead".
+  deadLinkUrls: ReadonlySet<string> = new Set(),
 ): Row[] {
   return subnets
     .flatMap((subnet) =>
@@ -1050,7 +1056,14 @@ export function flattenSurfaces(
         // advertise that as verified for the rest of the TTL is exactly the
         // complaint in #8658's title. Live evidence outranks a historical
         // hand-edit when the two disagree about whether something still exists.
-        const confirmedDead = isProbeDemoted(probeRecord);
+        // Two independent ways to lose a verification, both meaning "we looked
+        // and it is gone": the prober confirmed the surface itself dead, or the
+        // link lane confirmed every source_url proving it exists is dead
+        // (#9914). The second is what makes the gate's dead-source_url rule
+        // apply after merge and not only on the way in.
+        const confirmedDead =
+          isProbeDemoted(probeRecord) ||
+          surfaceEvidenceIsDead(surface.source_urls, deadLinkUrls);
         // #1006: the as-of timestamp every served surface should carry. A
         // per-surface verification wins; otherwise only official surfaces may
         // inherit subnet curation verified_at (when a maintainer last vetted the
@@ -1414,42 +1427,132 @@ interface ResolvedAddress {
   family: number;
 }
 
+/**
+ * Why a URL yielded no address to connect to (#9870).
+ *
+ * All three refuse the connection, but they are NOT the same claim about the
+ * world, and collapsing them is what made us report `unsafe URL` for a host
+ * that simply no longer exists:
+ *
+ *   - `unsafe`        — we looked, and connecting would be a safety problem: the
+ *                       literal guard rejected it, or DNS answered with a
+ *                       private/loopback/link-local address (rebinding).
+ *   - `unresolvable`  — we looked, and the name authoritatively does not exist
+ *                       (NXDOMAIN / no records). Nothing unsafe about it; it is
+ *                       DEAD, and that is a fact about the target worth
+ *                       publishing.
+ *   - `resolve-error` — we could not look. A transient resolver error, a
+ *                       timeout, our own network. This is a fact about US, and
+ *                       must never be scored as the target being dead —
+ *                       "absence of evidence is not death"
+ *                       (src/surface-verification.ts).
+ *
+ * The `unsafe` / `resolve-error` split (and those exact names) already exist in
+ * `resolvedWebhookUrlStatus` (src/webhooks.ts), for the same reason: a DNS blip
+ * must be retryable, not a terminal verdict. This adds the third case webhooks
+ * has no use for — an authoritative "no such name".
+ */
+export type UrlResolutionFailure = "unsafe" | "unresolvable" | "resolve-error";
+
+export interface UrlResolution {
+  addresses: ResolvedAddress[];
+  /** null exactly when `addresses` is non-empty. */
+  failure: UrlResolutionFailure | null;
+}
+
+/**
+ * The one place each failure is put into words. Every caller that reports a
+ * reason to a human or an artifact reads this map, so `api.affine.io` cannot be
+ * described as "unsafe URL" in one lane and "dead" in another.
+ */
+export const URL_RESOLUTION_FAILURE_MESSAGE: Record<
+  UrlResolutionFailure,
+  string
+> = {
+  unsafe: "unsafe URL",
+  unresolvable: "host does not resolve",
+  "resolve-error": "DNS lookup failed",
+};
+
+/**
+ * Node error codes that mean the resolver got an authoritative "no such name"
+ * rather than failing to ask. EAI_AGAIN (temporary failure) is deliberately
+ * absent — it is a `resolve-error`, not a death certificate.
+ */
+const AUTHORITATIVE_DNS_MISS = new Set(["ENOTFOUND", "ENODATA", "EAI_NODATA"]);
+
+/**
+ * Resolve a URL to the public addresses it is safe to connect to, and say WHY
+ * when there are none. `resolvePublicUrlAddresses`/`isUnsafeResolvedUrl` are
+ * thin wrappers so every existing caller keeps its exact signature — only
+ * callers that report a reason to a human need the richer result.
+ */
+export async function resolveUrlSafety(
+  value: unknown,
+  resolver: typeof lookup = lookup,
+): Promise<UrlResolution> {
+  let host: string;
+  try {
+    const url = new URL(value as string);
+    if (isUnsafeUrl(url.toString())) {
+      return { addresses: [], failure: "unsafe" };
+    }
+    host = normalizeHostname(url.hostname);
+  } catch {
+    // Not a URL at all. A malformed input is a safety refusal, not a claim
+    // that some host is dead.
+    return { addresses: [], failure: "unsafe" };
+  }
+
+  if (isIP(host)) {
+    return {
+      addresses: [{ address: host, family: isIP(host) }],
+      failure: null,
+    };
+  }
+
+  try {
+    const records = await resolver(host, { all: true, verbatim: true });
+    if (records.length === 0) {
+      return { addresses: [], failure: "unresolvable" };
+    }
+    if (records.some((record) => isUnsafeIpAddress(record.address))) {
+      return { addresses: [], failure: "unsafe" };
+    }
+    return {
+      addresses: records.map((record) => ({
+        address: record.address,
+        family: record.family,
+      })),
+      failure: null,
+    };
+  } catch (error) {
+    // The resolver threw. ENOTFOUND/ENODATA is how Node reports NXDOMAIN — an
+    // authoritative "no such name", i.e. dead. Anything else (EAI_AGAIN,
+    // SERVFAIL, timeout) means we failed to ask, which is our problem and must
+    // not be recorded against the target.
+    const code = (error as NodeJS.ErrnoException)?.code;
+    return {
+      addresses: [],
+      failure: AUTHORITATIVE_DNS_MISS.has(String(code))
+        ? "unresolvable"
+        : "resolve-error",
+    };
+  }
+}
+
 export async function resolvePublicUrlAddresses(
   value: unknown,
   resolver: typeof lookup = lookup,
 ): Promise<ResolvedAddress[]> {
-  try {
-    const url = new URL(value as string);
-    if (isUnsafeUrl(url.toString())) {
-      return [];
-    }
-
-    const host = normalizeHostname(url.hostname);
-    if (isIP(host)) {
-      return [{ address: host, family: isIP(host) }];
-    }
-
-    const records = await resolver(host, { all: true, verbatim: true });
-    if (
-      records.length === 0 ||
-      records.some((record) => isUnsafeIpAddress(record.address))
-    ) {
-      return [];
-    }
-    return records.map((record) => ({
-      address: record.address,
-      family: record.family,
-    }));
-  } catch {
-    return [];
-  }
+  return (await resolveUrlSafety(value, resolver)).addresses;
 }
 
 export async function isUnsafeResolvedUrl(
   value: unknown,
   resolver: typeof lookup = lookup,
 ): Promise<boolean> {
-  return (await resolvePublicUrlAddresses(value, resolver)).length === 0;
+  return (await resolveUrlSafety(value, resolver)).failure !== null;
 }
 
 const SAFE_FETCH_REDIRECT_CODES = new Set([301, 302, 303, 307, 308]);
--- a/scripts/link-status.ts
+++ b/scripts/link-status.ts
@@ -1,0 +1,62 @@
+// Build-side reader for the link-rot store the daily cron writes
+// (src/link-status-sync.ts). Mirrors scripts/github-signals.ts exactly: R2
+// store first (the cron's fresh copy), committed seed second, empty last.
+//
+// The seed materialization is not an optimization. The publish build holds the
+// R2 credentials and reads the store; the publish job's later `npm run
+// validate` step runs CREDENTIAL-LESS against the restored registry/ tree. Both
+// must see the same snapshot or the validation gates would judge artifacts
+// built from data they cannot read. Writing the store back to the seed path is
+// what carries it across — the same travel-with-the-artifacts design the native
+// snapshot, candidate inputs and github-signals already use.
+
+import path from "node:path";
+import {
+  isConfirmedDeadLink,
+  LINK_STATUS_R2_KEY,
+  type LinkStatusRecord,
+} from "../src/link-status-core.ts";
+import { readGeneratedStoreJson } from "./r2-rest.ts";
+import { readJson, repoRoot, stableStringify, writeJson } from "./lib.ts";
+
+type Row = Record<string, unknown>;
+
+export const linkStatusPath = path.join(
+  repoRoot,
+  "registry/generated/link-status.json",
+);
+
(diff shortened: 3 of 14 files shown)


JSONbored:
## Correction: the logo gate was wrong, and it was the CI failure

The first CI run was red on **my own gate**, not on anything pre-existing — 17,678 tests passed, 1 failed:

```
served logo URLs must resolve to a shipped file (235)
```

`tests/helpers/repo-sandbox.ts` copies only the **data** directories (`registry/`, `public/`) and deliberately not `apps/`, so `apps/ui/public/logos` does not exist in a sandboxed run. My gate read "the tree ships no logo directory" as "all 235 logos were deleted".

Fixed by skipping the gate when that directory is absent: a data-only tree asserts nothing about UI assets. CI's own `npm run validate` runs against the full checkout, where the directory exists and the gate stays live.

**Re-proven able to fail after the fix**, on both populations independently:

| scenario | result |
|---|---|
| hide one curated `/logos/*.png` | gate fires |
| hide one cached `/logos/cache/<sha>.png` | gate fires |
| clean tree | `validate` exits 0 |
| sandboxed `registry validates` | passes |

## Also correcting the note in the PR description

I claimed above that `registry validates` was red on clean `origin/main` and filed #10003 for it. **That was my misdiagnosis** — now closed. The comparison worktree had never been built, so validate failed there with `ENOENT: public/metagraph/providers.json`. The `sn-99 / Leoma` text I attributed it to is a **warning**; validate exits 0 with it present, and `scripts/validate.ts:1985` makes that deliberate:

> On-chain identity names are operator-controlled, though, so this must not hard-fail validation ... Emit an actionable warning instead.

So the hazard I "found" was already handled, correctly, before I got there. Nothing outstanding from that thread.

PR #10032, opened 2026-08-08
fix(test): let an e2e navigation wait out the dev-server restart it already expects

### LINKED ISSUE #10013: ci: the sticky-table-header e2e turned main red on a commit that could not have caused it
`main` went red on `673f567c` (#10007's merge) and green again on the very next commit, `133c3522`, with no fix in between.

```
  133c3522  success
  673f567c  failure   <- #10007
  a3c66c7f  success
  6c9d3efe  success
```

## The failing check

`ui` → **Responsive overflow check (e2e)**:

```
[interaction] › tests/e2e/sticky-table-header.spec.ts:147:7
  › /chain/extrinsics › table header pins to its own scrollport at desktop (1280px)
```

## It cannot have been that commit

#10007 changed `src/mcp-server.ts`, three `schemas-src/mcp-tools/*` files and one test file. It touched:

- no `apps/ui` file,
- no `openapi.json`,
- no generated types,
- no `packages/*`.

Confirmed by `git show --stat` filtered on those paths — empty. There is no path by which an MCP tool's input schema changes how a table header pins at 1280px, and the next commit proved it by passing without any change to that test or the code under it.

## Why it is worth an issue rather than a shrug

A flaky check on `main`'s required set is not cosmetic:

- it makes "is main green?" unanswerable at a glance, which is the question every merge decision depends on;
- it trains people to merge past a red `ui` job, which is exactly how a *real* UI regression gets through;
- it cost a real investigation here — the only way to rule it out was to diff the commit against the paths the test could possibly depend on.

## Suggested direction

`apps/ui`'s e2e harness already has known instability (ProxyController crashes, spawn races in CI). A sticky-position assertion at a fixed viewport is timing-sensitive by nature — it depends on layout settling and on rAF, which is paused when the pane is not visible. Either await a settled layout signal rather than a timeout, or quarantine this spec out of the required set until it is deterministic.

Whichever it is, **a required check that fails on unchanged code is worse than no check**: it spends the signal it exists to provide.

JSONbored:
## The diagnosis in #10013 was right that the commit was innocent, and wrong about why

`main` went red on `673f567c` and green on the next commit with no fix. The issue reasoned it must be a timing-sensitive sticky-position assertion and suggested awaiting a settled layout signal or quarantining the spec.

The CI log says otherwise — the assertion never ran:

```
23:19:22  [e2e-server] wrangler exited (code=1 signal=null) after 226648ms
          -- restarting (1/10). Expected cause: ProxyController
          "Network connection lost". Tests hitting the gap retry.

23:19:25  Error: page.goto: net::ERR_CONNECTION_REFUSED
          at http://localhost:8080/chain/extrinsics
```

The dev server was **down**. `tests/e2e/serve-e2e.ts` already anticipates this crash and restarts it — the failure is three seconds into that restart, and Playwright's retry fired in the same second and hit the same window.

So the supervisor's own closing promise, *"Tests hitting the gap retry"*, was the bug: nothing made a navigation wait for the server to come back. Quarantining the spec would have removed a check that was never wrong.

## The fix

`gotoThroughRestart` waits out the restart, and **only** for the error class meaning "nothing is listening yet". A failure the server *answered* — a 500, a redirect loop, a genuine timeout — still propagates on the first attempt, so a real breakage is never converted into a slow pass. That safety property is the one I care most about and it is mutation-tested.

Three details worth calling out:

- **`offline.spec.ts` is deliberately excluded at one site.** Its third navigation runs after `context.setOffline(true)`, where a refused connection *is* the assertion. Retrying it would be retrying the thing under test. The two pre-offline navigations do use the helper.
- **A per-worker latch.** Once a full budget elapses with the server still refusing, it is gone rather than restarting, so the rest of that worker fails fast. This came from measurement, not theory: when I killed the port holder mid-run, every remaining test paid the full 45s. The latch caps that at one budget per worker.
- **Per-test timeout 30s → 90s.** The default was already marginal — `sticky-table-header` spends up to 5s on `networkidle` plus up to 20s waiting for the `<thead>` (`/chain/extrinsics` needs ~10s to paint), leaving under 5s of slack *before* any restart. A cap is not a cost; a passing test still finishes when it finishes.

`.unit.ts`, not `.test.ts`: `playwright.config.ts` sets `testDir` to `./tests/e2e`, and Playwright's default `testMatch` claims both `*.spec.ts` and `*.test.ts` — a `.test.ts` there would be collected by Playwright as an e2e file with no `test()` calls. `vitest.config.ts` is widened to pick up `tests/e2e/**/*.unit.ts`, with that reason recorded. Verified: `playwright test --list` does not collect it.

## Validation

```
apps/ui typecheck            clean
apps/ui vitest               235 files, 1869 tests passed
apps/ui lint                 0 errors (9 pre-existing warnings elsewhere)
prettier --check             clean
playwright --project=interaction   104 passed, 2 skipped (2.1m)
```

**Mutation-tested** — each of these makes a test fail, so the tests are load-bearing:

| mutation | caught by |
|---|---|
| drop the `isServerUnavailable` guard (retry everything) | `rethrows a non-restart error on the FIRST attempt` |
| remove the latch short-circuit | `stops re-waiting the full budget once the server is confirmed gone` |
| never set the latch | same |

One thing I removed rather than tested: an earlier `attempts > 1` guard on the latch. Mutating it to `attempts > 0` changed nothing any test could see — with a real 45s budget the two are behaviorally identical, so it encoded a distinction that does not exist. Simplified it away instead of writing a test that pretended to cover it.

## Not claimed

I could not reproduce the ProxyController crash on demand. Killing the port holder externally does **not** trigger the supervisor (it watches the `wrangler` child, not the port), so that run exercised the "server never came back" path instead — which is how the latch got measured. The restart path itself is covered by the unit tests against a fake page, plus a full clean e2e run.

Closes #10013

### DIFF
--- a/apps/ui/playwright.config.ts
+++ b/apps/ui/playwright.config.ts
@@ -10,6 +10,18 @@ export default defineConfig({
   fullyParallel: true,
   forbidOnly: !!process.env.CI,
   retries: process.env.CI ? 1 : 0,
+  // #10013: Playwright's default per-test budget is 30s, and these specs were
+  // already spending most of it before any restart: sticky-table-header waits
+  // up to 5s for networkidle and then up to 20s for the <thead> (measured --
+  // /chain/extrinsics needs ~10s to paint its table), leaving under 5s of
+  // slack on a route that was expected to be slow. That left no room for
+  // gotoThroughRestart to wait out a supervised `wrangler dev` restart, so a
+  // test landing in the restart window failed on the timeout instead.
+  //
+  // A cap is not a cost: a passing test finishes when it finishes, and this
+  // only changes how long a genuinely stuck one is allowed to hang. 90s is the
+  // 45s restart budget plus the ~25s these specs already use, rounded up.
+  timeout: 90_000,
   reporter: process.env.CI ? "github" : "list",
   // #8928: pinned, because Playwright's default is os.cpus().length / 2 -- 2
   // workers on the 4-core ubuntu-latest runner, leaving half the machine idle
--- a/apps/ui/tests/e2e/evidence-deep-link.spec.ts
+++ b/apps/ui/tests/e2e/evidence-deep-link.spec.ts
@@ -1,6 +1,7 @@
 import { existsSync } from "node:fs";
 import { test, expect } from "@playwright/test";
 import { harPathForRoute, DATED_ENDPOINT_PATTERNS, findHarFixture } from "./har-path.ts";
+import { gotoThroughRestart } from "./server-restart.ts";
 
 // #6434 (historical): /subnets/:netuid used to render EvidencePanel twice -- a
 // preview embedded in the Overview tab and the full section under a dedicated
@@ -47,7 +48,7 @@ async function openWithHar(page: import("@playwright/test").Page, url: string) {
       await page.route(pattern, (route) => route.fulfill(fixture));
     }
   }
-  await page.goto(url);
+  await gotoThroughRestart(page, url);
   // HAR responses resolve instantly, which starves "networkidle" of the quiet
   // window it needs on a route with recurring refetches -- fall back to a fixed
   // settle rather than hanging (same rationale as responsive-overflow.spec.ts).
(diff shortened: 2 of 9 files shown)


JSONbored:
Rebased onto `2f4167f67`. The `test` job's failure on the first run was inherited from main, not from this branch — `AssertionError: blocks_head is written to Neon but not in PARITY_TABLES`, 731 passed / 1 failed, in the Neon parity watchdog. This diff is apps/ui-only. Main fixed it in #10030 (`fix(neon): the parity sweep watches the two lanes #10020 added`), which this branch now includes.

The `ui` job — the one this change actually affects — passed on the first run.

Issue #9881, opened 2026-08-07
neon-parity watchdog has reported `unknown` on every tick since it shipped: D1 caps compound SELECT at 5 terms


JSONbored:
`src/neon-parity-watchdog.ts` sweeps `PARITY_TABLES` (10 tables) as one `UNION ALL`. D1 will not parse it.

```
lane        verdict   detail
neon-parity unknown   counts unreadable: D1_ERROR: too many terms in compound SELECT: SQLITE_ERROR
```

Every row in `lane_health` for this lane since #9850 deployed is that same line, hourly.

## The limit is 5, measured

Probed against the production D1 database 2026-08-07 with synthetic `SELECT n AS t, 1 AS n` terms:

| terms | result |
|---|---|
| 3 | ok |
| 5 | ok |
| 6 | `too many terms in compound SELECT` |
| 8 | same |
| 10 | same |

Upstream SQLite defaults `SQLITE_MAX_COMPOUND_SELECT` to 500. D1 builds it at 5. Ten tables is five past it, so the D1 half of the comparison throws before reading a row and the lane correctly records that it cannot tell — which is indistinguishable from healthy to anything that only checks for `fail`.

## Why the tests did not catch it

The test double accepted any statement width, so the suite asserted the code's own assumption. The fix has to include teaching the fake to reject >5 terms like the real store does; with that in place the pre-fix code fails 7 tests.

## Fix

Batch the sweep into statements of at most 5 terms and merge the counts. This matters more as the migration proceeds — every table added to `PARITY_TABLES` adds a term, and the failure is total rather than partial.

PR #10035, opened 2026-08-08
fix(neon): watch the two backfilled tables nothing was comparing

### LINKED ISSUE #10034: ci: main is red — blocks_head and raw_capture_state are backfilled into Neon but unwatched
`main` is failing `tests/neon-parity-watchdog.test.ts`:

```
FAIL the deployed wiring > every mirrored or backfilled table is watched
AssertionError: blocks_head is written to Neon but not in PARITY_TABLES
```

Reproduced on `main` directly, so it is not any one PR's diff — it is currently blocking every open PR.

## What it means

`NEON_BACKFILL_LANES` in `wrangler.data.jsonc` names 20 tables that are reconciled D1 → Neon on a cron. `PARITY_TABLES` in `src/neon-parity-watchdog.ts` lists 22 tables whose row counts are compared between the two stores.

**`blocks_head` and `raw_capture_state` are in the first list and not the second.** Both were added to Neon by `migrations/neon/0005_remaining_d1_tables.sql` and wired into the backfill lane; neither was added to the watchdog.

So both are being written into Neon on a schedule with **nothing comparing the two sides** — which is, in the test's own words, "the gap this lane exists to close: a table written to both stores with nothing comparing them".

The test is doing exactly its job. It is not a flake and should not be relaxed.

## Fix

Add both to `PARITY_TABLES`. The list is deliberately hand-maintained rather than discovered — its comment explains why:

> Hand-listed rather than discovered, because "what is in Neon" is the thing being checked — deriving the list from Neon would make a table that silently vanished there look like a table nobody asked about.

That reasoning is right, and it is also why the list can fall behind a lane: nothing forces the two to be edited together except this test. Which is the point of the test.

No batching concern: `parityCountBatches` already chunks the sweep because D1 parses at most five `UNION ALL` terms (#9881), so the list going from 22 to 24 is free.

JSONbored:
**`main` is red**, and has been — reproduced on `main` directly, so this is blocking every open PR rather than belonging to any one of them.

```
FAIL tests/neon-parity-watchdog.test.ts > the deployed wiring > every mirrored or backfilled table is watched
AssertionError: blocks_head is written to Neon but not in PARITY_TABLES
```

## What it means

`NEON_BACKFILL_LANES` names 20 tables reconciled D1 → Neon on a cron. `PARITY_TABLES` lists the tables whose row counts are compared between the two stores.

**`blocks_head` and `raw_capture_state` are in the first and not the second.** Both were added to Neon by `migrations/neon/0005_remaining_d1_tables.sql` and wired into the backfill lane; neither reached the watchdog. So both have been written into Neon on a schedule with **nothing comparing the two sides** — in the test's own words, *"the gap this lane exists to close: a table written to both stores with nothing comparing them"*.

The test is doing exactly its job. It should not be relaxed; the list is what changes.

## Why the list can fall behind

`PARITY_TABLES` is deliberately hand-maintained, and its comment explains why:

> Hand-listed rather than discovered, because "what is in Neon" is the thing being checked — deriving the list from Neon would make a table that silently vanished there look like a table nobody asked about.

That reasoning is right. It is also why the list can drift from a lane: nothing forces the two to be edited together **except this test** — which is the point of the test, and why it caught this.

No batching concern: `parityCountBatches` already chunks the sweep because D1 parses at most five `UNION ALL` terms (#9881), so 22 → 24 is free.

## Validation

```
npm run lint && npm run format:check && npm run typecheck
npx vitest run tests/neon-parity-watchdog.test.ts tests/ledger-neon-write.test.ts \
  tests/neon-backfill.test.ts tests/table-freshness-watchdog.test.ts
```

All green — 160 tests across every file naming `PARITY_TABLES`, `neon-parity`, or either table (grepped, not guessed).

Closes #10034

### DIFF
--- a/src/neon-parity-watchdog.ts
+++ b/src/neon-parity-watchdog.ts
@@ -87,6 +87,12 @@ export const PARITY_TABLES = [
   "surface_failure_daily",
   "subnet_burn_history",
   "tao_usd_index",
+  // Added by migrations/neon/0005 and named in NEON_BACKFILL_LANES, but never
+  // added here -- so both were reconciled into Neon with nothing comparing the
+  // two sides, which is the exact gap this lane exists to close. The deployed
+  // -wiring test caught it and had been failing on main.
+  "blocks_head",
+  "raw_capture_state",
 ] as const;
 
 /**


JSONbored:
Superseded by #10030, which landed the same fix while this was in CI and annotates the per-table thresholds better than mine did (blocks_head unpruned at the default; raw_capture_state two rows, where any drift is meaningful).

Verified on main: `tests/neon-parity-watchdog.test.ts` passes, 41 tests. Closing rather than rebasing — the change is already there.

PR #10042, opened 2026-08-08
fix(registry): stale:false meant "never checked" for 78% of served surfaces

### LINKED ISSUE #9906: surfaces: stale:false is published for 2,731 surfaces that have never been verified, making it indistinguishable from freshly verified
`GET /api/v1/surfaces` publishes `stale: false` for **2,731 of 3,493 surfaces (78%)** that have never been verified even once.

```
total surfaces                     3,493
never verified (last_verified_at null)  2,731   → stale: false  (100% of them)
ever verified                            762   → stale: false ×753, true ×9
```

A consumer filtering on `stale === false` to mean "this verification is current" gets a set in which four out of five entries have no verification behind them at all.

## Cause

`isSurfaceStale` (`scripts/lib.ts`) opens with a null guard that answers the wrong question:

```ts
export function isSurfaceStale(lastVerifiedAt, kind, nowMs): boolean {
  if (!lastVerifiedAt) return false;
  const verifiedMs = Date.parse(lastVerifiedAt as string);
  if (!Number.isFinite(verifiedMs) || !Number.isFinite(nowMs)) return false;
  return nowMs - verifiedMs > surfaceFreshnessTtlDays(kind) * 86_400_000;
}
```

Read narrowly — *"is this verification stale?"* — `false` for "there is no verification" is defensible. Read as the field name reads to an API consumer — *"is this surface's status stale?"* — it is exactly backwards: a surface nobody has ever checked is the **most** unknown, not the least.

`false` also collapses two genuinely different states into one value, which is the same distinction this codebase is careful about elsewhere: `exists` is null rather than false on RPC failure in `/crowdloans/{id}`, `leased` is null rather than false in `/subnets/{netuid}/lease`, and `verdict: "unknown"` is deliberately kept distinct from `"ok"` in `lane_health` ("a watchdog that could not evaluate has not observed health, and collapsing the two would report an unmeasured lane as a healthy one" — `migrations/d1/0014`).

## Scope check — this is a labelling bug, not a trust bug

I checked whether the false `stale` leaks into the trust tier, since `withSurfaceFreshness` uses it to decide whether to demote `curation_level`. **It does not:**

```
curation_level, never verified:            candidate-discovered × 2,731
curation_level, verified but stale=true:   candidate-discovered × 9
```

Never-verified surfaces already floor at the lowest tier through `resolveSurfaceCurationLevel`, so they land where a demoted stale surface lands. The trust ordering is intact — the damage is confined to the published `stale` field itself.

Worth stating because it bounds the fix: this does not require rethinking the freshness model, only making one field honest.

## Fix

Make `stale` nullable and return `null` when `last_verified_at` is null — "not applicable / never measured", distinct from a measured-and-fresh `false`. That is a schema change (`schemas-src/`, plus regenerated `openapi.json` and types) and a breaking-ish change for anyone doing `stale === false`, so it may be worth pairing with the freshness TTL docs.

If nullable is unattractive, the alternative is to leave the boolean alone and stop implying currency from it — but then `last_verified_at: null` is the only honest signal, and the field is decoration.

## Related context, not part of this fix

Of 1,859 surfaces declaring `probe.enabled: true`, only 626 have a published health verdict. The 1,233 without are `dashboard` (436), `docs` (394), `source-repo` (168), `website` (146), `openapi` (70), `sdk` (11), `example` (5), `repo-registry` (3) — i.e. the non-operational kinds, which `/api/v1/health` scopes itself out of by design (`scope: "operational"`). Among those 1,233, 314 do carry a `last_verified_at` and the **newest is 2026-07-16**, 22 days ago. Whether those kinds should be verified on a cadence is a separate question from whether `stale` should lie about them; filing only the latter here.

Found while auditing subnet registry accuracy.

JSONbored:
## The measurement

Paging the whole live collection (`GET /api/v1/surfaces`, 7 pages, 3,493 rows):

| `last_verified_at` | `stale` | count |
|---|---|---:|
| never | `false` | **2,731** |
| ever | `false` | 753 |
| ever | `true` | 9 |

A consumer filtering on `stale === false` to mean *"this verification is current"* got a set in which **four out of five entries had no verification behind them at all**.

## Why `false` was the wrong answer

Read narrowly — *"is this verification stale?"* — `false` for "there is no verification" is defensible. Read as the field name reads to an API consumer — *"is this surface's status stale?"* — it is backwards: a surface nobody has ever checked is the **most** unknown, not the least.

`null` is the distinction this codebase already keeps everywhere else: `exists` is null rather than false on RPC failure in `/crowdloans/{id}`, `leased` is null rather than false in `/subnets/{netuid}/lease`, and `verdict: "unknown"` stays distinct from `"ok"` in `lane_health` — *"a watchdog that could not evaluate has not observed health, and collapsing the two would report an unmeasured lane as a healthy one"* (`migrations/d1/0014`).

The two unmeasurable cases go the same way: an unparseable `last_verified_at` and a non-finite `nowMs` are both "could not evaluate", not "evaluated and fine".

## One change covers all three contract surfaces

`SurfaceSchema` backs REST `/surfaces`, the MCP `list_surfaces` output (through `SurfacesArtifactSchema` → `z.array(SurfaceSchema)`), and the committed artifact — so the fix lands in one place. I confirmed the live MCP tool reproduced the identical bug first:

```json
{"last_verified_at":null, ... ,"stale":false}
```

The GraphQL SDL already declares `stale: Boolean`, which permits null, so it needed no change. I checked the other `stale` occurrences and they are a **different field** — `health.stale`, nested under `health` in `health-serving.ts`, `mcp-server.ts:13294` and `schemas-src/mcp-tools/ai-integration.ts` — deliberately untouched. `apps/ui` has no consumer of the surface-level flag.

## This is a labelling fix, not a trust change

Never-verified surfaces already floor at `candidate-discovered` through `resolveSurfaceCurationLevel`, so the tier ordering comes out byte-identical. A new test asserts that **outcome** for all three states.

The demotion branch now compares `stale === true` rather than relying on truthiness. Being straight about this: `null` is falsy, so the two agree today and **no test can separate them** — I verified that by mutation. The explicit compare is there because the value is no longer a boolean, and any future truthy representation of "unmeasured" would silently start demoting all 2,731. I corrected a comment that had overclaimed the test pinned this.

## The artifact invariant got stronger, not weaker

`tests/artifacts.test.ts` asserted `typeof surface.stale === "boolean"`. Rather than relax that to accept null, it now asserts the **correlation**:

> `stale` is null **exactly when** `last_verified_at` is null

so the flag and the evidence it describes cannot drift apart again. Mutation-tested — reverting `isSurfaceStale` to return `false` fails it with that message.

## Validation

```
npm run typecheck              clean
npm run lint / format:check    clean
npm run build                  passed (openapi.json + client/contract types regenerated)
npm run validate               129 subnets, 3441 surfaces, 137 providers, 2037 candidates
npm run validate:contract-drift  passed
vitest (9 files: mcp-server, script-utils, artifacts, graphql,
        surfaces-mcp, zod-schemas, lib-helpers, …)   3,176 passed
```

Built artifact after the change:

```
last_verified=never  stale=None   -> 2615
last_verified=ever   stale=False  ->  820
last_verified=ever   stale=True   ->    6
```

`stale === false` now selects the 820 surfaces that actually have a current verification.

The published schema is `anyOf: [boolean, null]` with the three states described, so the contract states the distinction rather than leaving consumers to infer it.

Closes #9906

### DIFF
--- a/packages/contract/index.d.ts
+++ b/packages/contract/index.d.ts
@@ -11704,7 +11704,8 @@ export interface components {
             schema_status?: "machine-readable" | "ui-only" | "not-captured";
             schema_url?: string;
             source_urls?: string[];
-            stale?: boolean;
+            /** @description Whether this surface's verification has aged past its per-kind freshness TTL. NULL when there is no verification to judge (`last_verified_at` is null, or unparseable) — distinct from a measured `false`, which means we checked and it is current. Filtering on `stale === false` therefore selects surfaces with a CURRENT verification, not merely surfaces that are not known to be stale. */
+            stale?: boolean | null;
             status?: components["schemas"]["HealthStatus"];
             subnet_name?: string;
             subnet_slug?: string;
--- a/public/metagraph/openapi.json
+++ b/public/metagraph/openapi.json
@@ -52454,7 +52454,15 @@
             "type": "array"
           },
           "stale": {
-            "type": "boolean"
+            "anyOf": [
+              {
+                "type": "boolean"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "description": "Whether this surface's verification has aged past its per-kind freshness TTL. NULL when there is no verification to judge (`last_verified_at` is null, or unparseable) — distinct from a measured `false`, which means we checked and it is current. Filtering on `stale === false` therefore selects surfaces with a CURRENT verification, not merely surfaces that are not known to be stale."
           },
           "status": {
             "$ref": "#/components/schemas/HealthStatus"
--- a/public/metagraph/types.d.ts
+++ b/public/metagraph/types.d.ts
@@ -11704,7 +11704,8 @@ export interface components {
             schema_status?: "machine-readable" | "ui-only" | "not-captured";
             schema_url?: string;
             source_urls?: string[];
-            stale?: boolean;
+            /** @description Whether this surface's verification has aged past its per-kind freshness TTL. NULL when there is no verification to judge (`last_verified_at` is null, or unparseable) — distinct from a measured `false`, which means we checked and it is current. Filtering on `stale === false` therefore selects surfaces with a CURRENT verification, not merely surfaces that are not known to be stale. */
+            stale?: boolean | null;
             status?: components["schemas"]["HealthStatus"];
             subnet_name?: string;
             subnet_slug?: string;
--- a/schemas-src/routes/subnet-detail.ts
+++ b/schemas-src/routes/subnet-detail.ts
@@ -550,7 +550,17 @@ export const SurfaceSchema = z
       .optional(),
     schema_url: HttpOrWssUrlSchema.optional(),
     source_urls: z.array(z.url()).optional(),
-    stale: z.boolean().optional(),
+    stale: z
+      .boolean()
+      .nullable()
+      .optional()
+      .describe(
+        "Whether this surface's verification has aged past its per-kind freshness TTL. " +
+          "NULL when there is no verification to judge (`last_verified_at` is null, or " +
+          "unparseable) — distinct from a measured `false`, which means we checked and it " +
+          "is current. Filtering on `stale === false` therefore selects surfaces with a " +
+          "CURRENT verification, not merely surfaces that are not known to be stale.",
+      ),
     status: HealthStatusSchema.optional(),
     subnet_name: z.string().optional(),
     subnet_slug: z.string().optional(),
--- a/scripts/lib.ts
+++ b/scripts/lib.ts
@@ -1111,14 +1111,38 @@ export function surfaceFreshnessTtlDays(kind: string): number {
 // deterministic reference — never wall-clock — so the flag is reproducible). A
 // surface with no last_verified_at is NOT stale: that's "unverified", a distinct
 // state the null timestamp already signals.
+/**
+ * Is this surface's verification stale? `null` when there is nothing to judge.
+ *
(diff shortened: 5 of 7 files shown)


JSONbored:
Superseded by #10050, which merged first and fixes #9906 the same way. Closing rather than resolving the conflicts — the work is duplicate, not divergent.

I checked before starting (no open PR referenced #9906, and the API chat's queue was all MCP tools), so this was a concurrent start rather than a missed collision.

Comparing the two, the merged fix is equivalent on everything that matters:

- `isSurfaceStale` returns `boolean | null`, null for a missing `last_verified_at` **and** for both unmeasurable cases (unparseable date, non-finite `nowMs`) — same reasoning, same three states.
- `SurfaceSchema.stale` nullable, with the regenerated `openapi.json` + client/contract types.
- `tests/artifacts.test.ts` asserts the **correlation** (`stale === null` exactly when `last_verified_at === null`), not merely the type — which was the part I most wanted to see survive, since that is what stops the two collapsing back together.

One difference, which I do **not** think warrants a follow-up: `withSurfaceFreshness` still branches on `curation_level: stale` (truthiness) rather than an explicit `stale === true`. I mutation-tested that on my branch — `null` is falsy, so the two are behaviourally identical and no test can distinguish them. The explicit compare only guards against a hypothetical future where "unmeasured" is represented by something truthy (a `"unknown"` sentinel). Worth a line if anyone touches that function again; not worth a PR of its own.

Nothing from this branch is lost.

Issue #5597, opened 2026-07-15
Backfill subnet_hyperparams_history via archive-node state_call (blocked on #2111)


JSONbored:
## Problem

`subnet_hyperparams_history` (migration `0037`, `src/subnet-hyperparams-history.mjs`, shipped in
#4477 closing #4309 — deliverable 1.6 of the subnet-hyperparameters epic #4303) only captures
**forward-only diffs from the day it shipped**. Any subnet whose hyperparameters changed before
that date has no history — `GET /api/v1/subnets/{netuid}/hyperparameters/history` returns nothing
prior to first capture, even though the on-chain history exists.

PR #4477 explicitly deferred this: "a full backfill needs archive-node `state_call` at past block
heights (#2111, not yet at tip), explicitly deferred to a follow-up PR."

## Area

Backend / chain-data backfill — `src/subnet-hyperparams-history.mjs`, a new one-shot backfill
script (likely `scripts/backfill-subnet-hyperparams-history.mjs` or similar, mirroring how other
one-time backfills in this repo are scripted rather than added to the live cron path).

## Proposal

Once #2111 (bare-metal archive node) is synced to tip and its `state_call` RPC surface is
reachable, walk each subnet's hyperparameter-affecting extrinsics/blocks (or a coarser periodic
sample, if a full per-block walk is prohibitively expensive — needs a design call) via archive
`state_call`, diff against the previous known state the same way
`recordSubnetHyperparamsChanges` already does for live writes, and backfill
`subnet_hyperparams_history` rows with their real historical `observed_at`/block height instead of
the capture-date they'd otherwise get.

## Deliverables

- A design note on the backfill strategy (full per-block walk vs. periodic/change-detection
  sampling against archive state) — cost/time tradeoff depends on real archive `state_call`
  latency, unknown until #2111 lands.
- The backfill script itself, reusing `subnet-hyperparams-history.mjs`'s existing diff/hash logic
  rather than re-deriving it.
- Idempotency: safe to re-run without duplicating rows already captured by the live forward-only
  writer.
- Test coverage matching this repo's 99%+ patch gate.

## Resources

- Deferred by: PR #4477 (https://github.com/JSONbored/metagraphed/pull/4477)
- Precedent pattern: `src/subnet-hyperparams-history.mjs`, `migrations/0037_subnet_hyperparams_history.sql`
- Sibling history pattern this was itself cloned from: `src/subnet-identity-history.mjs`,
[...]

PR #10299, opened 2026-08-09
fix(neon): weights_version is a u64 on chain, not an int32

### LINKED ISSUE #10298: fix(neon): weights_version is a u64 on chain and an int32 in the table — two subnets have already exceeded it
## `weights_version` is a u64 on chain and an int32 in Neon

`SubtensorModule.WeightsVersionKey` is a **u64**. Both columns that store it are `integer` (int32, ceiling 2,147,483,647):

```
subnet_hyperparams          weights_version   integer
subnet_hyperparams_history  weights_version   integer
```

Declared as `INTEGER` in `migrations/neon/0001_side_tables.sql:75` and `migrations/neon/0003_append_only_histories.sql:151`.

## It works today by luck

Live max across all 129 subnets is **1,778,428,951** — 83% of the way to the ceiling, and no nulls. Every write succeeds right now.

## It has already been exceeded

Found while flushing the #5597 historical backfill: 15 replayed rows fail with

```
error: value "3000000030" is out of range for type integer
  where: "unnamed portal parameter $9 = '...'"     ($9 = weights_version)
```

Three subnets have carried an out-of-range value:

| netuid | weights_version |
|---|---|
| 16 | 3,000,000,030 |
| 18 | 9,223,372,036,854,775,807 |
| 28 | 9,223,372,036,854,775,807 |

9,223,372,036,854,775,807 is `i64::MAX` — the sentinel a subnet sets to disable version gating outright. Two subnets have used it. This is not a hypothetical value the chain could theoretically produce; it is a value the chain has produced, on mainnet, more than once.

## The failure mode is a whole-lane stall, not a dropped field

The live hourly hyperparameters lane writes this column through the same shared column list and the same statement. The insert is one statement per row, so a single subnet setting `weights_version_key` above 2^31−1 fails **that subnet's row**, and the value is one sudo call away at any time. Nothing in the write path clamps, validates, or reports it — the guard in `src/subnet-hyperparams-backfill.ts`'s `rejectRow` checks `netuid`, `block_number` and the epoch-millis floor, not field ranges, so this surfaces as a Postgres error rather than a counted rejection.

## Fix

Widen both columns to `BIGINT`. Non-destructive, matches the chain's own type, and the tables are small (129 and ~2,000 rows), so the rewrite is negligible.

`BIGINT` is int64 and the chain type is u64, so the top half of the range is still unrepresentable in principle — but `i64::MAX` is the value actually used as the "never" sentinel, and it fits. Worth a note in the migration rather than reaching for `NUMERIC`.

Blocks the remaining #5597 backfill flush.

JSONbored:
## The column could not hold the value the chain publishes

`SubtensorModule.WeightsVersionKey` is a **u64**. Both columns storing it were `INTEGER` — int32, ceiling 2,147,483,647.

## Not theoretical — already exceeded, twice

Found flushing the #5597 historical backfill: 15 replayed rows failed with

```
error: value "3000000030" is out of range for type integer
  where: "unnamed portal parameter $9 = '...'"      ($9 = weights_version)
```

| netuid | weights_version |
|---|---|
| 16 | 3,000,000,030 |
| 18 | 9,223,372,036,854,775,807 |
| 28 | 9,223,372,036,854,775,807 |

That second value is `i64::MAX`, the sentinel a subnet sets to disable version gating outright. **Two subnets have used it on mainnet.**

## It works today by luck, and the margin is 17%

Live maximum across all 129 subnets, measured 2026-08-09: **1,778,428,951** — 83% of the way to the ceiling, no nulls. Every current write succeeds.

The live hourly lane writes this column through the same shared column list and the same one-statement-per-row insert, and nothing in the write path clamps, range-checks or reports it — `rejectRow` guards `netuid`, `block_number` and the epoch-millis floor, not field ranges. So a subnet raising its key past 2^31−1 fails that row as a Postgres error, not as a counted rejection. This is one sudo call from being a live fault instead of a backfill one.

## Why BIGINT and not NUMERIC

`BIGINT` is int64 and the chain type is u64, so the top half of the chain's range is still unrepresentable. That is deliberate: `i64::MAX` is the value actually used as the "never" sentinel and it fits, while `NUMERIC` would cost every reader an arbitrary-precision decode for a range the chain does not use. The migration says so, so the next person does not rediscover it.

## Applied

Both columns are `bigint` in production as of this change — verified against `information_schema`. Widening only; every existing value is valid unchanged, and the tables are small (129 and ~3,150 rows), so the rewrite was negligible.

The remaining backfill then flushed clean: **2,329 rows attempted, 0 rejected**, `subnet_hyperparams_history` 2,066 → 3,150.

`0001_side_tables.sql` and `0003_append_only_histories.sql` are left alone deliberately — an applied migration is a record of what was applied, and 0014 is the record of this change. Nothing in the repo derives column types from those files.

Closes #10298

### DIFF
--- a/migrations/neon/0014_weights_version_bigint.sql
+++ b/migrations/neon/0014_weights_version_bigint.sql
@@ -1,0 +1,37 @@
+-- weights_version is a u64 on chain and was an int32 here (#10298).
+--
+-- SubtensorModule.WeightsVersionKey is a u64. Both columns storing it were
+-- declared INTEGER -- migrations/neon/0001_side_tables.sql and
+-- migrations/neon/0003_append_only_histories.sql -- which caps at 2,147,483,647.
+--
+-- THIS IS NOT A THEORETICAL RANGE. Measured 2026-08-09, the live maximum across
+-- all 129 subnets was 1,778,428,951: working, and 83% of the way to the ceiling.
+-- But three subnets have historically carried a value above it --
+--
+--   netuid 16  3,000,000,030
+--   netuid 18  9,223,372,036,854,775,807
+--   netuid 28  9,223,372,036,854,775,807
+--
+-- -- and that second value is i64::MAX, the sentinel a subnet sets to disable
+-- version gating outright. Two subnets have used it. Found when 15 rows of the
+-- #5597 historical replay failed with "value 3000000030 is out of range for
+-- type integer".
+--
+-- The live hourly lane writes this column through the same shared column list,
+-- one statement per row, so a subnet setting the key above 2^31-1 fails that
+-- row -- and nothing in the write path clamps or reports it. This is one sudo
+-- call away from being a live fault rather than a backfill one.
+--
+-- BIGINT is int64 and the chain type is u64, so the top half of the chain's
+-- range is still not representable. That is deliberate rather than overlooked:
+-- i64::MAX is the value actually used as the "never" sentinel and it fits,
+-- while NUMERIC would cost every reader an arbitrary-precision decode for a
+-- range the chain does not use.
+--
+-- Widening only -- every existing value is valid unchanged, and both tables are
+-- small (129 and ~2,000 rows), so the rewrite is negligible.
+ALTER TABLE subnet_hyperparams
+  ALTER COLUMN weights_version TYPE BIGINT;
+
+ALTER TABLE subnet_hyperparams_history
+  ALTER COLUMN weights_version TYPE BIGINT;


JSONbored:
## Swept the rest of the family — `weights_version` was the only one

`subnet_hyperparams` has **14** remaining `integer` columns. None carries a live maximum above **1% of the int32 ceiling** (21.4M), so nothing else is near overflow.

The stronger check is empirical rather than a `max()` over current values: the full historical replay — **2,329 rows spanning 2023-12-14 to now, across every subnet** — flushed with **0 rejections** once this column was widened. Every other column held every value the chain has published over that period.

So this is a single mis-typed column, not the first instance of a class.

Issue #10228, opened 2026-08-08
44 D1-named identifiers outlive the database — three of them are contract, not cosmetics


JSONbored:
D1 is deleted (#10179) and its files are renamed (#10225), but the identifiers are still everywhere. An AST-free sweep of `src/` and `workers/` finds 44 distinct symbols carrying `D1`/`d1`, ~200 occurrences. They are not one job — they split three ways, and the split is the point of this issue.

**1. Safe mechanical renames.** Internal symbols with no external referent. `currentD1ReadFailureGeneration`, `coerceD1Number`, `toD1Flag`, `matchNeuronsD1Route`, `NeuronsD1RouteHandler`, `neuronsD1Handler`, `matchHyperparamsIdentityD1Route`, `matchHealthStatusD1Route`, `hyperparamsIdentityD1Handler`, `healthStatusD1Handler`, `loadAlphaPricesByNetuidD1`, `loadD1AlphaPricesByNetuid`, `loadAccountPositionsD1`, `loadSubnetTemposD1`, `loadLeaderboardD1Rows`, `loadRealizedStakeBaselinesD1`, `loadNominatorCountsD1`, `loadNominatorCountD1`, `loadAccountRegistrationsD1`, `loadNeuronStakeByHotkeysD1`, `writeAccountBalancesToD1`, `writeNominatorPositionsToD1`, `applyRegistrySyncToD1`, `rollupUptimeDailyToD1`, `persistProbesToD1`, `pruneD1Checks`, `neuronsServedFromD1`, `markD1FallbackRows`, `hasD1FallbackRows`, and the `*_D1_READ_COLUMNS` constant families.

Two deserve a look rather than a `sed` while renaming: `toD1Flag` and `coerceD1Number` encode a boolean→`1`/`0` and a numeric coercion that D1 needed. Postgres columns typed `int` still need them, so the behaviour stays — but confirm which columns before assuming, because a boolean column does not (#10088's typing trap).

**2. `DATA_API_D1_FLAGS` is a DEPLOYED VAR NAME.** It is set per-config in the wrangler files, it is per-flag rather than global, and widening or renaming it breaks every path reading it. Renaming means a coordinated var rename plus a deploy of every config that declares it, and `vars` do not inherit between configs — that has produced four separate defects already. Do it on its own, or leave it.

[...]

PR #10305, opened 2026-08-09
fix(neon): bind() must return a new statement, and the lane must count what landed

### LINKED ISSUE #10304: fix(neon): pg bind() mutates a shared statement — subnet_burn_history has written 1 row per tick for 34h while reporting 129
## `bind()` mutates a shared statement, and production has been losing writes for 34 hours

`src/pg-statement-client.ts`'s `prepare()` returns one statement object and one `bound` wrapper. `bind()` writes into that shared object and returns it:

```ts
function prepare(text: string) {
  const stmt: PendingStatement = { text, values: [] };   // ONE object
  const bound = { …, __stmt: stmt };                      // ONE wrapper
  return {
    bind(...values) {
      stmt.values = values;   // mutates it
      return bound;           // returns the same object every call
    },
    ...bound,
  };
}
```

**D1's `bind()` is immutable** — it yields a fresh statement — and every lane here is written against that contract. The idiom this module exists to preserve is quoted in its own header:

```ts
const insert = db.prepare(SQL);
await db.batch(rows.map((r) => insert.bind(r.a, r.b)));
```

With a mutating `bind()`, all N elements of that array are **the same object** carrying only the **last** row's values. `batch()` runs the statement N times with identical parameters, and an `ON CONFLICT … DO UPDATE` folds them into **one row**.

## Measured in production

`subnet_burn_history`, 2026-08-09:

| | |
|---|---|
| rows per 15-minute tick | **1** (should be 129) |
| netuids frozen | **126 of 129**, all at `2026-08-08T15:31:01` |
| still accruing | 107, 120, 122 — whichever row happened to be last that tick |
| `lane_health` verdict | `ok` — `"captured 129, pruned"`, every pass |

126 netuids have exactly 352 rows each and stop dead at the same instant. That instant is when the table went sole-store on Neon and this shim took over from D1.

It is not append-on-change: at least one of those netuids has 352 rows all carrying the **same** `burn_tao`, so the writer appends every tick regardless of change. The rows are simply not landing.

`/api/v1/subnets/{netuid}/burn/history` therefore serves a series whose newest point is 34 hours old for 98% of subnets, while `/subnets/{netuid}/burn` returns a live value — a caller comparing them sees the series as dead.

## Blast radius

Only `src/subnet-burn-history.ts` uses prepare-once/bind-many today. The other `db.batch` call site (`workers/api.ts`, the emission-gate change log) calls `db.prepare(SQL).bind(…)` fresh inside its `.map()`, so each statement gets its own object and it is unaffected.

But the broken idiom is the **documented** one, and nothing catches it: the call site looks correct, the batch reports success, the lane reports its verdict, and the row count is the only evidence.

## The secondary defect

`captured: rows.length` counts what was **read from the chain**, not what was **written**. A lane that writes 1 of 129 rows reports `captured 129, pruned` and stays green forever. Whatever counts a pass should count rows that landed — otherwise this exact class of failure is invisible by construction, which is how it survived 34 hours.

## Fix

`bind()` returns a new statement, chainable, so re-binding an already-bound statement is independent too. Three regression tests, each verified to go **red** against the mutating implementation and green against the fix — asserting the bound *values*, not the statement count, because a mutating `bind()` still produces N statements and only the values expose it.

The lost ticks cannot be recovered; the lane resumes full coverage on the next tick after deploy.

JSONbored:
## Production has been writing 1 row per tick instead of 129, for 34 hours, while reporting success

`prepare()` returned one statement object and one `bound` wrapper, and `bind()` wrote into that shared object and returned it. **D1's `bind()` is immutable** — it yields a fresh statement — and every lane here is written against that contract, including the idiom quoted in this module's own header:

```ts
const insert = db.prepare(SQL);
await db.batch(rows.map((r) => insert.bind(r.a, r.b)));
```

With a mutating `bind()`, all N elements of that array are **the same object** carrying only the **last** row's values. `batch()` runs the statement N times with identical parameters, and `ON CONFLICT … DO UPDATE` folds them into one row.

## What that did

`subnet_burn_history`, measured 2026-08-09:

| | |
|---|---|
| rows per 15-minute tick | **1** (should be 129) |
| netuids frozen | **126 of 129**, all at `2026-08-08T15:31:01` |
| still accruing | 107, 120, 122 — whichever row was last that tick |
| `lane_health` | `ok` — `"captured 129, pruned"`, every pass |

That instant is when the table went sole-store on Neon and this shim took over from D1.

Ruled out append-on-change: at least one frozen netuid has **352 rows all carrying the same `burn_tao`**, so the writer appends every tick regardless of change. The rows were simply not landing.

The user-visible symptom: `/subnets/{netuid}/burn/history` serves a series whose newest point is 34 hours old for 98% of subnets, beside `/subnets/{netuid}/burn` returning a live value.

## Blast radius

Only `src/subnet-burn-history.ts` uses prepare-once/bind-many today. The other `db.batch` call site — the emission-gate change log in `workers/api.ts` — calls `db.prepare(SQL).bind(…)` fresh inside its `.map()`, so each statement owns its object and it is unaffected. I checked every `.bind(` and `batch(` call site rather than assuming.

The broken idiom is the documented one, though, so this was a landmine for the next lane as much as a bug in this one.

## The tests were verified to fail

Three regressions, each run against the mutating implementation first:

```
× prepare once, bind per row -- each statement keeps its OWN values
× binding does not disturb the statement it was bound from
× re-binding an already-bound statement is independent too
```

and green against the fix. They assert the bound **values**, not the statement count — a mutating `bind()` still produces N statements, and only the values expose it. The third covers the chainable case (`stmt.bind(a).bind(b)` must leave `bind(a)` alone).

## Not fixed here

`captured: rows.length` counts what was read from the chain, not what was written, so the lane reported `captured 129` throughout. A pass counter that counts intent cannot see this class of failure — noted in #10304 as a separate defect rather than widened into this fix.

The lost ticks are unrecoverable; the lane resumes full coverage on the first tick after deploy. I will verify row counts against production once it lands.

Closes #10304

### DIFF
--- a/src/pg-statement-client.ts
+++ b/src/pg-statement-client.ts
@@ -162,7 +162,16 @@ export function createPgStatementClient(
             toPositionalPlaceholders(stmt.text),
             stmt.values,
           );
-          out.push({ results: res.rows ?? [] });
+          // D1's per-statement envelope is `{ results, meta: { changes } }`,
+          // and `meta.changes` was the missing half. Without it a batching
+          // lane has NO WAY to learn how many rows it actually wrote, so the
+          // only number it can report is the one it intended -- which is how
+          // subnet-burn-history reported `captured 129` on ticks that wrote 1.
+          // A write path that cannot count its own effect cannot be watched.
+          out.push({
+            results: res.rows ?? [],
+            meta: { changes: res.rowCount ?? 0 },
+          });
         }
         await c.query("COMMIT");
         return out;
--- a/src/pg-statement-client.ts
+++ b/src/pg-statement-client.ts
@@ -86,32 +86,57 @@ export function createPgStatementClient(
     return c.query(toPositionalPlaceholders(stmt.text), stmt.values);
   };
 
+  /**
+   * BIND RETURNS A NEW STATEMENT. It must, and this is not a style choice.
+   *
+   * D1's `bind()` is immutable -- it yields a fresh statement and leaves the
+   * prepared one alone -- and the lanes are written against that contract. The
+   * idiom this whole module exists to keep working is:
+   *
+   *     const insert = db.prepare(SQL);
+   *     await db.batch(rows.map((r) => insert.bind(r.a, r.b)));
+   *
+   * An earlier version mutated one shared `stmt` and returned one shared
+   * `bound` object, so every element of that array was THE SAME OBJECT holding
+   * only the LAST row's values. `batch()` then ran the statement N times with
+   * identical parameters, and an `ON CONFLICT ... DO UPDATE` collapsed the lot
+   * into a single row.
+   *
+   * It failed silently and it failed in production: measured 2026-08-09,
+   * `subnet-burn-history` had written exactly ONE row per 15-minute tick since
+   * 2026-08-08T15:31 -- the instant the table went sole-store on Neon and this
+   * shim took over from D1 -- while reporting `captured 129` on every pass.
+   * 126 of 129 subnets stopped accruing history and every lane card stayed
+   * green. Nothing about the shape of the call looks wrong at the call site,
+   * which is exactly why the invariant belongs here.
+   */
   function prepare(text: string) {
-    const stmt: PendingStatement = { text, values: [] };
-    const bound = {
-      // `results` mirrors D1's envelope so a caller reading either shape works.
-      async all() {
-        const res = await run(stmt);
-        return { results: (res.rows ?? []) as unknown[] };
-      },
-      async run() {
-        const res = await run(stmt);
-        return { meta: { changes: res.rowCount ?? 0 } };
-      },
-      async first() {
-        const res = await run(stmt);
-        return ((res.rows ?? [])[0] ?? null) as unknown;
-      },
-      /** Read by batch() -- the pending statement, not a promise. */
-      __stmt: stmt,
+    const statement = (values: unknown[]) => {
+      const stmt: PendingStatement = { text, values };
+      return {
+        // `results` mirrors D1's envelope so a caller reading either shape works.
+        async all() {
+          const res = await run(stmt);
+          return { results: (res.rows ?? []) as unknown[] };
+        },
+        async run() {
+          const res = await run(stmt);
+          return { meta: { changes: res.rowCount ?? 0 } };
+        },
+        async first() {
+          const res = await run(stmt);
+          return ((res.rows ?? [])[0] ?? null) as unknown;
+        },
+        /** Read by batch() -- the pending statement, not a promise. */
+        __stmt: stmt,
+        /** A FRESH statement, never `this` -- see the note above. Chainable, so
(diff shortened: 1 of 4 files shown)


JSONbored:
## Second commit: the lane now counts what landed

The original PR left `captured: rows.length` alone and noted it as a separate defect. On reflection that was the wrong call — it is the half that made the bug *invisible*, and shipping the fix without it leaves the next occurrence just as silent.

Two changes:

- **the runner reports `meta.changes` per statement.** It returned only `{ results }`, so a batching lane had no way to learn how many rows it wrote. A write path that cannot count its own effect cannot be watched.
- **`subnet-burn-history` reports the sum of those counts** instead of the row count it read.

The guard that matters is the fallback: it falls back to the intended count only when **no** statement reports a numeric `meta.changes`. An empty or meta-less result means "this store cannot count its own writes", not "wrote nothing" — reading it as zero would invert the bug into a permanent false alarm on a working lane. Two tests cover both directions: two statements reporting one write ⇒ `captured: 1`, and an empty batch result ⇒ falls back rather than reporting 0.

## On the framing

The original description justified the invariant with "the way D1's does". D1 is deleted, so citing it as the authority is wrong — the invariant holds because **the callers** require it, not because a database we no longer run happened to behave that way. The code comments say that now.

That prompted a wider look: **~4,300 D1 references remain** across the repo (2,104 in `tests/`, 1,183 in `src/`, 456 in `workers/`), including live identifiers — `D1Runner` 49, `D1Like` 48, `d1All` 29, `toD1Flag` 22, `coerceD1Number` 18. #10228 scoped that as "44 identifiers"; I have corrected it there.

Filed **#10309** for the architectural half: the shim exists to emulate an API whose database we deleted, and this bug is what that emulation costs — an adapter diverging from an interface with no live reference implementation to check against. Not folded into this PR; this one stays the fix.

JSONbored:
## Correction: the freeze instant is 08:31Z, not 15:31Z

The timestamp I quoted for when the 126 netuids stopped — `2026-08-08T15:31:01` — is **+7 hours off**. The correct instant is **2026-08-08T08:31:01Z**.

Cause: I read it as `to_timestamp(max(observed_at)/1000) AT TIME ZONE 'UTC'`, which yields a naive timestamp that the pg driver then serialized as if it were local. The same query via `to_char(…)` returns the right value, and the API independently reported netuid 1's newest point as `2026-08-08T08:31:01.855Z` — which agrees with the corrected figure and disagreed with mine.

**Nothing else changes.** The duration (~34h), the counts (126 of 129 frozen, 1 row per tick, 3 survivors), the 1:1 `serve_axon`/`AxonServed` match, and the diagnosis were all computed with `extract(epoch from (now() - …))` or as plain counts, none of which are affected by the display artifact.

It matters only for correlation: anyone lining this up against a deploy or a Workers Build timeline should look at **08:31Z on 2026-08-08**, not 15:31Z.

Issue #10190, opened 2026-08-08
refactor(tiers): sweep the tryPostgresTier calls that can never forward — 8 flags read "retired", 0 read "postgres"


JSONbored:
`tryPostgresTier` forwards when the flag is `"postgres"`, **or** `"d1"` and the flag name is in `DATA_API_D1_FLAGS`. Measured across all three wrangler configs on 2026-08-08:

- **8 flags read `"d1"`** — `METAGRAPH_NEURONS_SOURCE`, `_SUBNET_HYPERPARAMS_`, `_ACCOUNT_IDENTITY_`, `_SUBNET_SNAPSHOTS_`, `_HEALTH_` (main), plus the three re-declared in `wrangler.data.jsonc`
- **8 flags read `"retired"`** — `_BLOCKS_`, `_EXTRINSICS_`, `_ACCOUNT_EVENTS_`, `_TOP_HOLDERS_`, `_SELF_HEALTH_`, `_SUBNET_IDENTITY_`, `_RPC_USAGE_`, `_SUBNET_OWNERSHIP_`
- **0 flags read `"postgres"`**

Two consequences follow.

**1. The `value === "postgres"` disjunct is itself inert.** No deployed configuration reaches it. The only live path through `tryPostgresTier` is `"d1"` + `DATA_API_D1_FLAGS`, which holds exactly three flags (`NEURONS`, `SUBNET_HYPERPARAMS`, `ACCOUNT_IDENTITY`).

**2. Every `tryPostgresTier` call on a `"retired"` flag is dead, and invisible.** The flag arrives as a `keyof Env` **string**, so the comparison happens inside the helper and `tsc` cannot see it. #10186 regenerated the Env types and `tsc` flagged six comparisons — but those were only the handful written *inline* as `env.X === "postgres"`. Rough counts of the string-arg form: ~120 in `src/mcp-server.ts`, ~90 in `src/graphql.ts`, ~60 in `workers/request-handlers/entities.ts`, ~25 in `workers/request-handlers/analytics.ts`. A large fraction sit on `"retired"` flags and can never forward.

**Do not read "tsc is happy" as "this tier is reachable."** That is precisely the inference the string argument defeats.

## Why this matters beyond tidiness

[...]

Issue #9369, opened 2026-08-04
weights/serving/stake-moves/stake-transfers/registrations: four more per-subnet cards serve a confident zero


JSONbored:
Follow-up to #9367. Having fixed one instance, I probed the rest of the family the same way and found four more.

## Measured live, 2026-08-04

Chain-wide sibling versus the per-subnet card, subnet 64:

| family | chain-wide | `/subnets/64/…` | verdict |
|---|---|---|---|
| `serving` | 3,036 servers over 20 subnets | **0** | broken |
| `stake-moves` | 674 movers over 128 subnets | **0** | broken |
| `stake-transfers` | 430 senders / 12,168 over 126 subnets | **0** | broken |
| `registrations` | 6,317 registrants / 8,055 over 98 subnets | **0** | broken |
| `prometheus` | 0 over 0 subnets | 0 | **correct** |
| `axon-removals` | 0 over 0 subnets | 0 | **correct** |
| `weights` | 1,259 / 70,216 | 14 / 2,745 | fixed by #9368 |

`stake-moves` is independently confirmed: `/api/v1/subnets/64/events` returns 3 `StakeMoved` events in the same window the card reports 0 movers for.

`prometheus` and `axon-removals` report 0 per-subnet **and** 0 chain-wide, because `PrometheusServed` and `AxonInfoRemoved` do not occur in the window at all. Their zero is the right answer and they are deliberately left alone — "fix everything that looks like the bug" would have made two correct routes wrong.

## Root cause, same as #9367

Each of the four had a two-tier fallback:

```ts
tryPostgresTier(...) ?? buildSubnetX(null, netuid, ...)
```

`METAGRAPH_ACCOUNT_EVENTS_SOURCE` is `"retired"` in `wrangler.jsonc`, so `tryPostgresTier` returns null **unconditionally** and the zeroed card is the only thing left. 200, no degraded marker, and a number that reads as "this subnet has no activity of this kind".

## The wider finding

Auditing every source flag against `DATA_API_D1_FLAGS`:

```
METAGRAPH_ACCOUNT_EVENTS_SOURCE       retired    forwards=False
METAGRAPH_BLOCKS_SOURCE               retired    forwards=False
METAGRAPH_EXTRINSICS_SOURCE           retired    forwards=False
METAGRAPH_TOP_HOLDERS_SOURCE          retired    forwards=False
METAGRAPH_SELF_HEALTH_SOURCE          retired    forwards=False
[...]

PR #10317, opened 2026-08-09
fix(api): give chain/prometheus the cold-tier rung its axon twin has

### LINKED ISSUE #10248: fix(prometheus): the PrometheusServed stream is neither captured nor loadable — chain/prometheus answers a hard zero beside a live axon twin
`/api/v1/chain/prometheus`, `get_chain_prometheus` and the GraphQL `chain_prometheus` field all answer a hard zero with `degraded: { reason: "prometheus_stream_not_curated" }`, while the axon twin beside them is fully live:

```
/api/v1/chain/serving?window=7d     subnet_count=20  distinct_servers=2655  announcements=6638  degraded=null
/api/v1/chain/prometheus?window=7d  subnet_count=0   distinct_exporters=0   announcements=0     degraded=prometheus_stream_not_curated
```

`PrometheusServed` should be live from Neon like everything else. It is not, and there are **two independent reasons** — fixing either alone changes nothing.

## 1. The event is not captured (infra side)

Neon's raw chain-event stream carries 25+ SubtensorModule methods, including obscure ones (`BasketDeposited`, `IncentiveAlphaEmittedToMiners`, `RAORecycledForRegistrationSet`, `AlphaBurned`) — so it is not a curated shortlist. `AxonServed` is present with 158 rows in the hot window, newest seconds old.

`PrometheusServed` has **zero rows** in both `chain_detail_chain_events` and `chain_detail_account_events`.

```sql
SELECT pallet, method, count(*) FROM chain_detail_chain_events
WHERE pallet='SubtensorModule' GROUP BY 1,2 ORDER BY 3 DESC;
-- AxonServed 158 ... no PrometheusServed row at any count
```

The producer is metagraphed-infra's live-follow poller (see `handleChainDetailSync`'s own header), and the lakehouse `account_events` curation is the Rust extractor. Both exclude it. `src/chain-prometheus.ts` records the historical scale: the chain emitted **18,041** PrometheusServed events and our curation carries 0.

Nothing in THIS repo filters it — `src/chain-detail-sync-payload.ts` accepts any bare variant name, with no kind allowlist. So metagraphed will store the event the moment infra emits it.

## 2. There is no loader to serve it (this repo)

Even once the rows land, all three surfaces stay at zero: `buildChainPrometheus`/`buildAccountPrometheus`/`buildSubnetPrometheus` are called with a hardcoded empty list. There is no `loadChainPrometheusColdTier`, and no `CHAIN_PROMETHEUS_ROLLUP` spec.

The axon twin is the exact template:

```ts
// src/chain-event-rollup-cold-tier.ts
export const CHAIN_SERVING_ROLLUP: ChainEventRollupSpec = {
  eventKind: "AxonServed",
  countField: "announcements",
  distinctField: "distinct_servers",
  distinctColumn: "hotkey",
};
```

A `CHAIN_PROMETHEUS_ROLLUP` (`eventKind: "PrometheusServed"`, `distinctField: "distinct_exporters"`) plus a loader mirroring `src/chain-serving-loader.ts`, wired into REST + MCP + GraphQL together, is the whole of the metagraphed half.

## Sequencing

The metagraphed half is safe to land first and is independently testable with a stubbed rollup — it simply keeps returning the degraded zero until the events arrive, then starts answering with no further change. The infra half is the one that needs the cross-repo request.

Note `/api/v1/chain/axon-removals` is a *different* case and is NOT a bug: it reports `axon_removals_never_emitted` because the runtime has never emitted an `AxonRemoved` event at all.

JSONbored:
## The route had two rungs where its twin has three

`/chain/prometheus` and `/chain/serving` are twins — the pallet emits `PrometheusServed` and `AxonServed` from the same `serving.rs`, as the same `(netuid, hotkey)` tuple. They resolved data differently:

```
serving:     tryPostgresTier(…) ?? loadChainServingColdTier(…) ?? empty stub
prometheus:  tryPostgresTier(…) ??                                empty stub
```

`METAGRAPH_ACCOUNT_EVENTS_SOURCE` reads `"retired"` (one of the eight #10190 counts), so that first rung **can never forward**. Serving survives on its second rung, added in #9216. Prometheus had no second rung at all — it fell straight from a tier that always misses to a zeroed card.

So the `prometheus_stream_not_curated` marker was pointing at the wrong system. Curating the stream perfectly would have changed nothing: **nothing was wired to read it.** Verified live — `/chain/prometheus?window=7d` answers `subnet_count: 0` beside `/chain/serving?window=7d` answering 38.

## Wired on all three surfaces, deliberately

REST, MCP and GraphQL each had the identical two-rung shape. `src/chain-serving-loader.ts` exists *because* #9216 wired only REST and left `get_chain_serving` and GraphQL's `chain_serving` returning zeroed cards — "a caller could not tell which surface was lying". Repeating that here would have been repeating a documented mistake, so all three now call one `loadChainPrometheusColdTier`.

## A drift the twins had already accumulated

`buildChainServing` takes a `subnetCount`; `buildChainPrometheus` did not, and derived `subnet_count` from `subnets.length` — the **returned page**. With the default limit of 20 over a window covering 128 subnets, that publishes `subnet_count: 20` as a measurement. Added the parameter to match, with a test that a `limit: 1` over two subnets still reports `subnet_count: 2`.

## The gate caught its own stale exemption

`tests/graphql-tier-cascade.test.ts` carries `chain_prometheus` in `NO_TIER_ANYWHERE` with the note *"no lane exists for it"*. That stopped being true in this diff, and the gate failed **because the exemption was now stale** — it checks both directions. Removed.

## Data side, already done

`chain.account_events` carried 0 `PrometheusServed` against 18,041 in `chain.chain_events`. Backfilled today by projecting `chain_events` (its `args` already carry `[netuid, [[32 bytes]]]`) — **18,041 rows, blocks 109 → 8,655,704**, verified against the `serve_prometheus` extrinsic signer on 40/40 sampled events and committed through `iceberg_load.py`'s exactly-once ledger.

Not a re-decode: those events reach back to block 109, far below the raw R2 capture floor of 8,759,336, so the bytes to re-decode do not exist.

**Expect the card to stay quiet, and that is correct.** The last `serve_prometheus` on mainnet was block 8,655,704 / 2026-07-19, so a 7d or 30d window legitimately covers zero announcements. What changes is that the zero is now *measured* rather than structural, the marker no longer blames a curation gap that is closed, and the historical data is reachable.

## Validation

```
npx vitest run       774 files, 18,126 passed, 11 skipped
new tests            6 (right event kind, response shape, window-vs-page
                     subnet_count, window/label resolved together, limit
                     resolved once, declining lakehouse -> null)
validate, validate:api, validate:contract-drift, graphql-component-parity,
validate-docs, validate-mcp-route-map                                  clean
build / typecheck / lint / format                                      clean
```

Closes #10248

### DIFF
--- a/src/chain-event-rollup-cold-tier.ts
+++ b/src/chain-event-rollup-cold-tier.ts
@@ -61,6 +61,17 @@ export const CHAIN_SERVING_ROLLUP: ChainEventRollupSpec = {
   distinctColumn: "hotkey",
 };
 
+// PrometheusServed is AxonServed's twin -- the pallet emits both as
+// (netuid, hotkey) from the same serving.rs -- so it rolls up identically.
+// It had no spec at all, which is why /chain/prometheus had no cold-tier rung
+// to fall to and answered a permanent zero beside a live axon card (#10248).
+export const CHAIN_PROMETHEUS_ROLLUP: ChainEventRollupSpec = {
+  eventKind: "PrometheusServed",
+  countField: "announcements",
+  distinctField: "distinct_exporters",
+  distinctColumn: "hotkey",
+};
+
 export const CHAIN_WEIGHTS_ROLLUP: ChainEventRollupSpec = {
   eventKind: "WeightsSet",
   countField: "weight_sets",
--- a/src/chain-prometheus-loader.ts
+++ b/src/chain-prometheus-loader.ts
@@ -1,0 +1,70 @@
+// Shared chain-prometheus loader for REST + MCP + GraphQL parity.
+//
+// THE RUNG THAT WAS MISSING (#10248). Both prometheus and its axon twin resolve
+// their data the same way -- `tryPostgresTier(METAGRAPH_ACCOUNT_EVENTS_SOURCE)`
+// first -- and that flag reads "retired", so the call can never forward (#10190
+// counts eight such flags). `/chain/serving` survives that because it has a
+// SECOND rung, `loadChainServingColdTier`, which reads the lakehouse rollup.
+// Prometheus had no second rung at all: it fell straight from a tier that
+// always misses to the empty stub, and published a confident zero.
+//
+// So the route was never "waiting for curation" in the sense its degraded
+// marker implied. Even with a fully curated stream it had nothing to read from.
+//
+// Mirrors src/chain-serving-loader.ts deliberately, down to resolving BOTH the
+// window and the limit once here: the rollup reader caps at 200 and the builder
+// at 20, so an omitted limit otherwise scans ten times the rows the response
+// can carry -- the exact defect #9239 fixed on the serving side.
+//
+// Returns null rather than a zeroed card so each caller keeps its own fallback
+// and error contract.
+import {
+  ANALYTICS_WINDOW_DAYS,
+  DEFAULT_ANALYTICS_WINDOW,
+} from "../workers/config.ts";
+import {
+  buildChainPrometheus,
+  CHAIN_PROMETHEUS_LIMIT_DEFAULT,
+} from "./chain-prometheus.ts";
+import {
+  CHAIN_PROMETHEUS_ROLLUP,
+  loadChainEventRollup,
+} from "./chain-event-rollup-cold-tier.ts";
+import type { r2SqlQuery } from "./r2-sql.ts";
+
+/**
+ * One window of prometheus-serving activity from the lakehouse, already built
+ * into the response shape -- or null when the lakehouse cannot answer.
+ */
+export async function loadChainPrometheusColdTier(
+  env: Parameters<typeof r2SqlQuery>[0],
+  {
+    window,
+    limit,
+    query,
+  }: {
+    window: string;
+    /** Optional so callers that leave it to the builder's own default do not
+     * have to restate it here. */
+    limit?: number;
+    /** Injectable for tests; forwarded to the rollup reader. */
+    query?: typeof r2SqlQuery;
+  },
+): Promise<ReturnType<typeof buildChainPrometheus> | null> {
+  const label = Object.hasOwn(ANALYTICS_WINDOW_DAYS, window)
+    ? window
+    : DEFAULT_ANALYTICS_WINDOW;
+  const rowLimit = limit ?? CHAIN_PROMETHEUS_LIMIT_DEFAULT;
+  const rollup = await loadChainEventRollup(env, CHAIN_PROMETHEUS_ROLLUP, {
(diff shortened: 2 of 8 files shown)


JSONbored:
## Extended: all three prometheus cards, not just the chain-wide one

After wiring `/chain/prometheus` I swept every `tryPostgresTier` ladder on a retired flag across REST and MCP, reusing the detector `tests/graphql-tier-cascade.test.ts` already uses. 126 ladders sit on a retired flag; 45 looked rung-less.

**Most of that 45 was my detector being crude** — it stops at the first `;`, so a ladder split across statements reads as missing a rung it has. Probing six live: `/chain/calls`, `/chain/stake-moves`, `/chain/stake-transfers` and `/chain/signers` all return real data (false positives), and `/chain/axon-removals` returns a **correct** zero — `AxonInfoRemoved` has never been emitted, which `uncurated-event-streams.ts` already documents.

The real finding was that prometheus is broken as a **family**:

| route | before | its serving twin |
|---|---|---|
| `/chain/prometheus` | 0, degraded | 38 subnets |
| `/subnets/{netuid}/prometheus` | 0, degraded | 3 announcements, 1 server |
| `/accounts/{ss58}/prometheus` | 0, degraded | works |

All three had the identical shape, and in each case the fix was the rung the twin already had:

- **per-subnet** — its serving twin got `loadSubnetEventCardColdTier` in #9369; the prometheus card was left out. Same generic helper, `CHAIN_PROMETHEUS_ROLLUP` instead of `CHAIN_SERVING_ROLLUP`.
- **per-account** — `handleAccountServing` declares a `coldTier:` property; `handleAccountPrometheus` simply did not have the line. Added `loadAccountPrometheusColdTier`, mirroring the serving loader query for query.

So this is one defect in three places, each introduced by wiring a card's twin and not the card. That is the same failure `chain-serving-loader.ts` was extracted to prevent, one level down.

## Still green

```
npx vitest run    774 files, 18,126 passed, 11 skipped
build / typecheck / lint / format                     clean
validate, validate:api, validate:contract-drift,
graphql-component-parity, validate-docs               clean
```

## Worth a follow-up, not folded in here

The GraphQL gate that caught the stale `chain_prometheus` exemption only scans `src/graphql.ts`. Its header assumes "REST and MCP served real rows for every one" — prometheus was the counterexample on all three surfaces. Widening that gate to REST and MCP would have caught this family on the first commit; my ad-hoc scan needed hand-verification against production to separate signal from noise. Filing separately.

Issue #8922, opened 2026-07-31
perf(test): isolate:false cuts the coverage pass CPU by 68% — segregate the ~19 vi.mock files


JSONbored:
## Where CI time is now

#8909 landed and did what it promised — measured on main's first post-merge run:

| | before | after |
| --- | --- | --- |
| **`test` job total** | 905–1011s | **669s** |
| Test (artifact writers, serial) | 433–477s | **115s** |
| Test (parallel, with coverage) | 282–316s | 327s |

The coverage pass is now **49% of the job** and the only item worth attacking:

| Item | Time | Share |
| --- | --- | --- |
| **Test (parallel, with coverage)** | **327s** | **49%** |
| Test (artifact writers, serial) | 115s | 17% |
| Lint + format | ~60s | 9% |
| 39 validator steps | 64s | 10% |
| Install / Typecheck / Checkout / Build | ~90s | 13% |

(The 39 validator steps total only 64s — parallelising them is not worth it.)

## Measured options

Local `test:ci`, same exclusions CI uses. **User CPU is the number that matters**: CI runners are 4-core and compute-bound, so wall clock there tracks CPU, unlike a core-rich laptop where everything looks flat.

| variant | wall | **user CPU** | failures |
| --- | --- | --- | --- |
| baseline (forks + isolate) | 63s | **443s** | 0 |
| `--pool=threads` | 98s | 404s | 1 file / 4 tests |
| **`--isolate=false`** | 65s | **140s** | 2 files / 8 tests |
| `--pool=threads --isolate=false` | 85s | — | 5 files / 8 tests |

- **`pool: threads` is dead.** Slower wall clock, only ~9% less CPU, and it breaks `tests/lib-helpers.test.ts` (4 tests, `process.chdir() is not supported in workers`). Not worth fixing for that payoff.
- **`isolate: false` cuts CPU 443s → 140s, a 68% reduction (3.2x).** That directly targets the ~234s of repeated module imports: 540 files each re-import `mcp-server.ts` (~12.5k lines) and `graphql.ts` (~10k) into a fresh registry.

Projected: coverage pass **327s → ~120–150s**, i.e. **~180s off every PR**, no extra runners. Combined with #8909 that is ~1000s → ~470s for the `test` job.

## Why it currently fails, and the fix

[...]

PR #10417, opened 2026-08-10
ci: split the phases that share nothing, and shard the pass that dominates

### LINKED ISSUE #10414: ci: 752s of test job, 65% of it one step — split the phases that share nothing
## The measurement

Green `main` run 31363433811:

| job | wall |
|---|---|
| `test` | **752s** |
| `ui` | 494s (parallel) |
| `changes` / `python` | 17s / 11s |

Inside `test`, one step dominates:

| step | wall |
|---|---|
| **50. Test (split passes, with coverage gate)** | **486s (65%)** |
| 7. Lint + format | 85s |
| 10. Validate committed contract (drift) | 23s |
| 5. Install | 23s |
| the other ~45 steps | ~135s |

And the test step's two passes are lopsided:

| pass | files | wall | of which import |
|---|---|---|---|
| shared registry | 722 | 320s | 63s |
| module-mocking (isolated) | **71** | **158s** | **129s** |

**Pass 2 runs 9% of the suite for 33% of the time, and 82% of that is module import.** Those files each need their own registry — a `vi.mock` requirement established in #8922 — so each re-imports `src/mcp-server.ts` (~12.5k lines) and `src/graphql.ts` (~10k) from scratch: ~1.8s of import per file before a single assertion runs.

## What is actually wrong

Nothing here is slow because the tests are slow. It is slow because **~266s of validators, lint, typecheck and build run sequentially in the same job as a 486s test step**, and because the two vitest passes run one after the other when they share nothing.

## Proposed work, in impact order

1. **Split the `test` job in two** — validators/lint/typecheck/build as one job, the test passes as another. Wall becomes `max(266, 486)` instead of `266 + 486`. One workflow edit, no test changes. **~266s.**
2. **Run the two vitest passes as parallel jobs** — `max(320, 158)` instead of `478`. **~158s.**
3. **Shard the 722-file shared pass across 3 runners** — ~107s + ~60s setup. **~150s.**
4. **Attack pass 2's import cost at the root** — the same class of fix as #10376, where breaking two constant imports cut the data-api bundle 8328→3280 KiB. If those 71 files reach the heavy modules transitively, a leaf-module extraction removes the 129s rather than parallelising it.

Steps 1–3 take the critical path from **752s to roughly 270s (~65%)**, after which the bottleneck is the validator job and `Lint + format` (85s) is the next target.

## Constraints any implementation must respect

- **The two-pass partition stays computed, never hand-listed.** A hand-maintained list rots into exactly the cross-file mock contamination #8922 removed.
- **Coverage must not silently narrow.** `Verify coverage reports exist` currently assumes both lcov files land in one job. Codecov merges multi-upload commits per commit SHA, but the check needs restructuring rather than deleting — a partial denominator understates coverage silently, which is the failure it was added to prevent.
- **`codecov/patch` at 99% branch-counted** must still measure the same diff.

## Acceptance

- [ ] Critical-path wall time for a green PR run at least halves, measured on a real run and quoted
- [ ] The mock-isolation partition is still computed at run time
- [ ] Coverage denominator provably unchanged — same total, not merely "green"

JSONbored:
## The measurement

Green `main` run 31363433811 — the `test` job is 752s, and **65% of it is one step**:

| step | wall |
|---|---|
| **50. Test (split passes, with coverage gate)** | **486s** |
| 7. Lint + format | 85s |
| the other ~45 steps | ~180s |

| pass | files | wall | of which import |
|---|---|---|---|
| shared registry | 722 | 320s | 63s |
| module-mocking | **71** | **158s** | **129s** |

**Nothing here is slow because the tests are slow.** It's slow because ~266s of validators, lint, typecheck and build run *sequentially in the same job* as a 486s test step, and because the two vitest passes run one after the other when they share nothing at all.

## What changes

- **`validate`** keeps the ~45 gates, lint, typecheck and build.
- **`test`** becomes a 4-leg matrix: the shared pass in three file shards, the mocking pass on its own.

Wall becomes `max(validate, slowest leg)` instead of the sum. Expected: **752s → ~270s**, after which the bottleneck is `validate` and `Lint + format` (85s) is the next target.

`run-ci-tests.ts` grows `--pass` and `--shard`. **The default is unchanged** — both passes, sequentially — because that's what a developer runs locally and what `npm run test:ci` has always meant.

## Things I deliberately kept honest

- **The partition is still computed** from source on every run. Sharding changes how many runners share the work, never *which* files are mocking files — a hand-listed partition is the contamination #8922 removed.
- **`--shard` is refused on the mocking pass** rather than silently ignored. 71 files don't need slicing, and asking for it means the caller believes something untrue about what's running.
- **Coverage is not narrowed.** Each leg uploads its own report; Codecov merges by commit SHA exactly as it already merged the two passes' separate uploads. `Verify coverage reports exist` still refuses an empty report *per leg*, so a partial denominator can't slip through — that check is the reason it exists.
- **`fail-fast: false`**, so a shard-specific failure doesn't cancel the legs whose results you need to interpret it.

## One gate had to move with it

`validate-workflows` matched the literal `./coverage/lcov.info`. The uploads are matrix-driven now (`./${{ matrix.cov }}/lcov.info`), so that literal finds **nothing** and the gate reports the token/tokenless split as *missing* — the opposite of what it checks. It now matches `lcov.info` and `vitest*.xml`: the rule is "a coverage upload exists on both the token and tokenless paths", not "the path is spelled this way". Commented so nobody tightens it back.

## Verification

- `--pass=mocked` → 71 files, **5,184 tests green, 49s** locally.
- `--pass=shared --shard=1/3` → 242 of 724 files, **4,119 tests green, 47s**.
- Both new guards proven to fail: `--pass=bogus` and `--shard` without `--pass=shared` exit 1 with a stated reason.
- All 45 validators, `tsc`, `eslint`, `prettier` clean; workflow YAML parses and `validate:workflows` passes.

**The real proof is this PR's own run** — the acceptance criterion is a measured halving on a real run, and I'll quote the numbers from it here once it completes.

Closes #10414

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -728,18 +728,22 @@ jobs:
       matrix:
         include:
           - name: shared 1/3
+            id: shared-1
             args: --pass=shared --shard=1/3
             cov: coverage
             junit: ""
           - name: shared 2/3
+            id: shared-2
             args: --pass=shared --shard=2/3
             cov: coverage
             junit: ""
           - name: shared 3/3
+            id: shared-3
             args: --pass=shared --shard=3/3
             cov: coverage
             junit: ""
           - name: mocked
+            id: mocked
             args: --pass=mocked
             cov: coverage-mocked
             junit: "-mocked"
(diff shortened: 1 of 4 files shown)


JSONbored:
## The split works; the coverage merge does not

Its own run proves the wall-clock claim:

| job | wall |
|---|---|
| `test (mocked)` | 239s |
| `validate` | 233s |
| `test (shared 1/3)` | 208s |
| `test (shared 3/3)` | 192s |
| `test (shared 2/3)` | 149s |

Critical path **752s → 239s, a 68% cut** — better than the ~270s estimate, and the legs are well balanced (149–239s).

**But `codecov/project` fails at 95.77%, down 2.17 points.** `codecov/patch` passes, so the diff itself is fully covered; what moved is the *denominator*. That is exactly the constraint this issue named — "coverage must not silently narrow" — and a 2.17-point drop is not a rounding artefact of four uploads instead of two.

So this **must not merge** until that is understood. Two candidates worth checking in order:

1. **Per-leg Codecov flags.** All four legs upload unflagged for the same SHA. Codecov merges multi-upload commits, but without flags there is nothing distinguishing a shard's partial report from a replacement of the previous one.
2. **What each shard instruments.** If coverage is not collected with `all: true`, each shard reports only the files its own tests touched, and the union of four partial reports is not the same denominator as one whole run — untested files that appeared in the single run's report would simply be absent.

I would not guess between them: the right next step is to compare the file *count* in one leg's `lcov.info` against the pre-split run's, which distinguishes (2) from (1) directly.

Holding this PR rather than lowering the project threshold to accommodate it — a floor moved to fit the measurement is not a floor, which is the same reasoning `run-ci-tests.ts` already applies to the per-pass vitest thresholds.

JSONbored:
**Candidate (2) is ruled out by measurement.** Both a shard's report and the mocked pass's report carry **670 `SF:` entries** — the v8 provider with an explicit `include:` instruments every matching file regardless of what actually ran, so each leg already reports the same denominator. Nothing is missing from the individual reports.

So the loss is in the **merge**, not the instrumentation, and `-2.17%` is consistent with Codecov finalising the check before all four uploads land rather than with any leg being wrong. Four concurrent unflagged uploads for one SHA is the race; two sequential ones from a single job never exposed it.

That makes **per-leg flags** the fix to try (Codecov's documented pattern for matrix jobs), not a threshold change. I have not pushed it — it should be one deliberate change measured against a real run, not stacked on top of a failing one.

Issue #10238, opened 2026-08-08
ci: the metagraphed-data-api Workers Build fails intermittently, and a failure on merge means production silently misses the deploy


JSONbored:
`Workers Builds: metagraphed-data-api` failed twice on 2026-08-08 and passed on retry both times, with **no change to the diff** — the second retry was a `--amend --no-edit` force-push, i.e. byte-identical content.

| PR | first attempt | retry (same content) |
|---|---|---|
| #10230 | fail | pass |
| #10237 | fail | pass |

Ruled out as causes:

- **Not the diff.** #10230 touched `wrangler.jsonc` (the *main* Worker's config), `workers/api.ts` and a test. `workers/data-api.ts` does not import `workers/api.ts`, and the data-api bundle was byte-comparable across branches: `5131.05 KiB / gzip 1024.82` on the branch vs `5131.13 / 1024.85` on a main-based worktree — the 0.08 KiB is an unrelated branch's renames.
- **Not local bundling.** `npx wrangler deploy --dry-run --config wrangler.data.jsonc` is clean on both branches.
- **Not the cron ceiling.** 250 per account, 13 Workers.

The reason is only visible in the Cloudflare dashboard; the GitHub check carries a link and no text. The standing suspicion (from #10194) is the bundle size — **1024.8 KiB gzip**, by far the largest of the fleet (`registry-sync-api` is 141) — against the 1-second startup limit, which is timing-sensitive and would explain an identical build passing on a second run.

## Why this is worth fixing rather than retrying

Workers Builds is the deploy path. A build that fails **on the merge commit** leaves `main` merged and production running the previous version, with nothing that reads as an incident — the PR is green-and-merged, and the only symptom is that the fix does not take effect. That is the same shape as #10190/#9779: a silent gap between "landed" and "running".

Suggested, in order of value:

1. **Detect it.** A check that the deployed SHA for each Worker is an ancestor of `main`, on a schedule, writing a `lane_health` verdict. `npx wrangler deployments list --config <file>` already gives the deployed SHA; comparing it to `main` is one `git merge-base --is-ancestor`. This makes the failure visible regardless of cause, and is worth having even if the build is fixed.
2. **Reduce the bundle.** 1024 KiB gzip on a Worker whose job is store access suggests the artifact/serving code is being pulled in. Worth a `--metafile` breakdown before assuming.
3. Capture the dashboard reason once, so this stops being a guess.

[...]

PR #10425, opened 2026-08-10
ci: run lint concurrently, and shard the pass I was wrong to refuse

### LINKED ISSUE #10424: ci: the bottleneck moved — lint runs serially, and the pass I refused to shard is now the longest job
#10414 took the critical path from 752s to ~281s. The bottlenecks moved; these are the new ones, measured on the merged run.

```
249s  validate        — of which Lint + format is 80s
251s  test (mocked)   — the single longest job
222s  test (shared 3/3)
204s  test (shared 1/3)
149s  test (shared 2/3)
 30s  coverage
```

## Three things, all measured

**1. `Lint + format` runs eslint and prettier sequentially.** Timed separately: eslint 38s, prettier 29s. They are independent and share nothing — run concurrently the step is bounded by the slower one instead of their sum. No extra job, no extra setup cost.

**2. The mocked pass is not sharded, and I was wrong to refuse it.** #10414 added a guard rejecting `--shard` on `--pass=mocked`, reasoning that "71 files do not need slicing". That sized the work by FILE COUNT when the cost is per-file IMPORT: 129s of the mocked pass's 158s is module import, ~1.8s per file, because each file needs its own registry (#8922) and re-imports `mcp-server.ts` (~12.5k lines) and `graphql.ts` (~10k). Import cost shards perfectly. It is now the longest job in the run, which is the evidence the reasoning was wrong.

**3. The shared shards are unbalanced** — 149s / 204s / 222s. vitest shards by file count, not by duration, so a shard that draws the slow files runs 50% longer than its sibling. More shards narrows the spread as well as the total.

## Not in scope here

Attacking the import cost at its ROOT — breaking the transitive reach from those 71 files into `mcp-server.ts`/`graphql.ts`, the same class of fix as #10376 — remains the real win, because it removes work rather than distributing it. Sharding is the cheap version and should not be mistaken for the fix.

## Acceptance

- [ ] Critical path measurably below the current ~281s, quoted from a real run
- [ ] The mocked pass shards, and the guard that refused it is gone rather than worked around
- [ ] `after_n_builds` stays 1 and the consolidation guard still holds at the new leg count

JSONbored:
#10414 moved the bottleneck rather than removing it. Measured on its merged run:

```
249s  validate        — of which Lint + format is 80s
251s  test (mocked)   — the single longest job
222s / 204s / 149s    — the shared shards, badly unbalanced
 30s  coverage
```

Three things, each measured before changing anything.

## 1. Lint ran serially

eslint and prettier read the same tree and write nothing, so they share nothing but CPU. Timed separately: **38s and 29s**. In series that's the 80s dominating the job.

Both are waited on and the exit code carried out explicitly, because a bare `cmd & cmd & wait` reports the **shell's** status — it would turn a red lint into a green step, which is worse than a slow one. Verified against every combination: `(0,0)→0`, `(1,0)→1`, `(0,1)→1`, `(3,5)→5`. Both still run when the first fails, so one push shows you both lists.

## 2. I was wrong to refuse sharding the mocking pass

#10414 added a guard rejecting `--shard` on `--pass=mocked`, reasoning that *"71 files do not need slicing"*. That sized the work by **file count** when the cost is per-file **import**: 129s of that pass's 158s is module import, ~1.8s per file, because each file needs its own registry (#8922) and re-imports `mcp-server.ts` and `graphql.ts` from scratch.

Import cost shards exactly as well as test time does — and the unsharded pass then became the longest job in the run, which is the evidence the reasoning was wrong. Locally: `--pass=mocked --shard=1/2` runs 36 files in **37s** against 49s for all 71.

The guard is **removed**, not worked around. What remains is that `--shard` still requires `--pass`, or it would slice one pass and silently run the other whole.

## 3. The shared shards were unbalanced

149s to 222s — a 50% spread, because vitest shards by file count and not by duration. Five shards instead of three narrows the spread as well as the total.

## Coverage is unaffected

Seven legs still publish artifacts merged by the single upload from #10414, `after_n_builds` stays 1, and both consolidation guards still hold at the new leg count.

## What this does *not* fix

It distributes the import cost rather than removing it. Breaking the transitive reach from those 71 files into `mcp-server.ts`/`graphql.ts` — the same class of fix as #10376, which cut the data-api bundle 8328→3280 KiB — is the real win, because it deletes the work instead of buying more runners to absorb it. Worth its own issue rather than being quietly folded in here.

## Verification

All 45 validators, `tsc`, `eslint`, `prettier` clean; workflow YAML parses and `validate:workflows` passes at the new 7-leg count. Both new guard behaviours exercised locally. **The measured critical path from this PR's own run will be quoted here before it merges.**

Closes #10424

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -343,10 +343,25 @@ jobs:
       # private-boundary leak, broken doc-contract reference, or malformed
       # intake template can land in a "docs-only" PR just as easily as in a
       # code PR).
+      # CONCURRENTLY, not sequentially (#10424). eslint and prettier read the
+      # same tree and write nothing, so they share nothing but CPU: timed
+      # separately they are 38s and 29s, and in series that is the 80s that
+      # dominated this job.
+      #
+      # Both are waited on and the exit code is carried out explicitly, because
+      # a bare `cmd & cmd & wait` reports the SHELL's status, not the jobs' --
+      # it would turn a red lint into a green step, which is worse than a slow
+      # one. Verified against every combination: (0,0)->0, (1,0)->1, (0,1)->1,
+      # (3,5)->5. Both still run even when the first fails, so one push shows
+      # you both lists rather than making you fix them one at a time.
       - name: Lint + format
         run: |
-          npm run lint
-          npm run format:check
+          npm run lint & lint=$!
+          npm run format:check & fmt=$!
+          rc=0
+          wait $lint || rc=$?
+          wait $fmt || rc=$?
+          exit $rc
 
       # Same class of check as Lint + format above: cheap, no build, no network
       # — runs unconditionally rather than gating on DOCS_ONLY since a real
@@ -727,24 +742,39 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - name: shared 1/3
+          - name: shared 1/5
             id: shared-1
-            args: --pass=shared --shard=1/3
+            args: --pass=shared --shard=1/5
             cov: coverage
             junit: ""
-          - name: shared 2/3
+          - name: shared 2/5
             id: shared-2
-            args: --pass=shared --shard=2/3
+            args: --pass=shared --shard=2/5
             cov: coverage
             junit: ""
-          - name: shared 3/3
+          - name: shared 3/5
             id: shared-3
-            args: --pass=shared --shard=3/3
+            args: --pass=shared --shard=3/5
             cov: coverage
             junit: ""
-          - name: mocked
-            id: mocked
-            args: --pass=mocked
+          - name: shared 4/5
+            id: shared-4
+            args: --pass=shared --shard=4/5
+            cov: coverage
+            junit: ""
+          - name: shared 5/5
+            id: shared-5
+            args: --pass=shared --shard=5/5
+            cov: coverage
+            junit: ""
+          - name: mocked 1/2
+            id: mocked-1
+            args: --pass=mocked --shard=1/2
+            cov: coverage-mocked
+            junit: "-mocked"
+          - name: mocked 2/2
+            id: mocked-2
+            args: --pass=mocked --shard=2/2
             cov: coverage-mocked
             junit: "-mocked"
     steps:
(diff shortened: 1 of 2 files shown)


JSONbored:
## Updated — and one thing in the first push measured worse

The first push's own run said five shared shards is **worse than three**, so it is reverted:

| shards | per-leg | max | sum |
|---|---|---|---|
| 3 shared | 149 / 204 / 222s | **222s** | 575s |
| 5 shared | 98 / 130 / 149 / 153 / **263s** | **263s** | 793s |

Sharding is not free — each shard is its own process with its own module registry, so splitting further re-imports the shared modules more times. Five shards added **218s of total work** and made the *slowest* leg worse, which is the number that sets the critical path. vitest shards by file count, not duration, so a finer split also concentrates the slow files: 263s against a 98s sibling.

What the same run confirmed as real wins, both kept:

- **Lint concurrency: 80s → 63s.** Less than the theoretical max because a 4-core runner has eslint and prettier contending, but real.
- **Mocked sharding: 251s → 105s / 163s.** The pass I had refused to shard.

## And the root fix is now in

Sharding distributes the import cost; this removes it. Crawling the static import graph from the 71 mocking tests, 39 reached a heavy module — through **three edges, not seventy-one**:

```
23  workers/api.ts                        -> src/mcp-server.ts
19  workers/request-handlers/discovery.ts -> src/mcp-server.ts
20  workers/chain-firehose-hub.ts         -> src/graphql.ts
```

The first two are accidents of module scope, not real dependencies. `api.ts` uses exactly **one** binding from `mcp-server.ts` and **one** from `graphql.ts`; `discovery.ts` uses eight, six of them plain constants. All sit inside async functions, so they are imported there instead. Reach drops **39 → 32**, and both modules come off the Worker's startup path — where #10238 already established that what runs at import is what exceeds the 400ms ceiling.

**The third edge stays.** `chain-firehose-hub.ts` validates subscription documents against the real schema in a *synchronous* exported function. That is a genuine dependency, and making it async to save import time would change a tested contract for a smaller win.

## A guard, because I broke this twice while writing it

Every matrix leg must now appear in the merge job's existence check **and both** Codecov `files:` lists. Changing the shard count and missing one of those four places is silent — a leg left out of `files:` simply is not counted, which reads as a coverage drop, and a leg left out of the existence check removes the only thing that would have said so. Proven to fail on a dropped leg.

## On measurement

Local timing was unusable here: the same command took **170s and 69s** minutes apart on a contended machine. Every number above is from CI, and I will quote this push's critical path before merging.

PR #10515, opened 2026-08-10
feat(ui): render every published route — 30 unrendered surfaces down to 0

### LINKED ISSUE #10300: ui: 30 of 205 published routes are rendered nowhere — and the TAO price we publish is not the one we display
## 30 of 205 published GET routes are rendered nowhere

Measured 2026-08-09 by resolving every `API_ROUTES` GET path against every `.ts/.tsx/.mdx/.json` file under `apps/ui/src` and `apps/ui/content`, with `{netuid}`-style tokens matched as path segments so template-literal construction (`` `/api/v1/subnets/${netuid}/profile` ``, the UI's actual idiom) counts as a reference. The generated `content/docs/api-reference/` tree is excluded — it mentions every route by construction and would make the sweep report 100%.

```
published GET routes:  205
referenced by apps/ui: 175
NOT referenced:         30
```

Spot-checked 9 of the 30 by substring; **0 false positives**. One near-miss worth recording: `concentration/history` returns 6 hits, but all six are the per-subnet `/subnets/{netuid}/concentration/history` — the network-wide `/chain/concentration/history` is genuinely unrendered. A naive substring sweep would have cleared it.

## The one that is not a gap but a duplication

**`/api/v1/network/tao-usd` is unused, and the UI gets its TAO price from a third party instead.**

`apps/ui/src/lib/metagraphed/market.functions.ts` fetches `https://api.coinpaprika.com/v1/tickers/tao-bittensor` (server-side, via `createServerFn` — not from the browser). Meanwhile we publish our own price with materially better provenance:

| | ours (`/network/tao-usd`) | coinpaprika (what the UI shows) |
|---|---|---|
| price | 207.0295 | 207.1395 |
| basis | `wrapped_onchain_median` across 2 Uniswap pools, with addresses + liquidity | centralized ticker |
| pinned to | Ethereum block 25,719,199 | — |
| history | 1,440 points over 24h | none |

They agree to **0.05%**, so this is not a correctness bug today. It is two sources of truth for one number, where the one we do not use is the one we can explain.

**It is not a drop-in swap.** `registry-ticker.tsx` also renders `market_cap` and `volume_24h` from that payload, and `/network/tao-usd` serves neither. So the honest options are: take the price from our own endpoint and keep coinpaprika for the two market aggregates, or add those aggregates to our surface. Worth deciding rather than silently swapping and dropping the ticker.

## The rest, classified

Not everything unrendered is a gap — some of these are API-only by design. Splitting them so the list is actionable:

### A — user-facing data with no home (17)

```
/api/v1/accounts/top-holders                      network-wide holder leaderboard
/api/v1/chain/holders                             cross-subnet alpha holders
/api/v1/subnets/{netuid}/burn                     live registration cost
/api/v1/subnets/{netuid}/burn/history             …and its series
/api/v1/subnets/{netuid}/validator-economics      validator entry economics
/api/v1/subnets/{netuid}/validator-economics/history
/api/v1/validators/economics                      the cross-subnet ranking
/api/v1/chain/concentration/history               network concentration drift
/api/v1/chain/concentration/subnets               per-subnet concentration ranking
/api/v1/subnets/{netuid}/emission-pipeline/history
/api/v1/chain/governance/emission-changes         the emission-gate change log
/api/v1/health/failure-reasons                    why probes fail, and whether the mix is moving
/api/v1/chain/indexer-lag                         how current our own capture is
/api/v1/crowdloans, /api/v1/crowdloans/{id}
/api/v1/domains/{tag}/summary
/api/v1/subnets/{netuid}/surface-history          a subnet's surface audit trail
/api/v1/accounts/{ss58}/root-claim
```

`/subnets/{netuid}/burn` is the one I would fix first: "what does it cost to register on this subnet right now" is a question every prospective miner asks, we serve both the value and its history, and neither appears on the subnet page.

### B — landed this session, nothing renders them yet (3)

```
/api/v1/chain/subnet-lifecycle          (#10263)
/api/v1/subnets/{netuid}/lifecycle      (#10263)
/api/v1/chain/deregistration-ranking    (#10285, in review)
```

Expected — the surfaces are new. Listed so they do not quietly join category A. The deregistration ranking in particular is the kind of thing a subnet page should carry a badge for.

### C — plausibly API-only, listed for a decision rather than a fix (7)

```
/api/v1/profiles              /api/v1/candidates
/api/v1/curation              /api/v1/source-snapshots
/api/v1/fixtures/{surface_id} /api/v1/alerts/triggers/{id}
/api/v1/network/randomness
```

Registry-operations and integration surfaces. If they are deliberately API-only that is a fine answer — but it should be a recorded decision, because right now "no page exists" and "no page should exist" are indistinguishable from the outside, which is the same ambiguity #10222 fixed for lane health.

## Why this matters beyond tidiness

Every one of these routes costs a capture lane, a schema, a contract entry, tests, and a share of the 99% patch-coverage gate. Category A is data we pay to collect, validate and serve, that no human ever sees. The recurring question in this repo has been "is this surface real or is it a confident zero" — this is the mirror image: surfaces that are real and unlooked-at.

Method note: the sweep is ~40 lines and could be a CI gate that fails when a NEW route lands with no UI reference and no explicit opt-out — the same shape as `NO_QUERY_PARAMETERS` and `AGENT_UNREACHABLE`. That would keep category B from silently becoming category A. Worth doing only if the opt-out list is honest, not as a rubber stamp.

JSONbored:
Closes #10300.

`/api/v1/network/tao-usd` and 24 other published routes were rendered nowhere. All of them render now, and the ratchet's ceiling drops from 25 to **0**: every published non-feed GET route is referenced by `apps/ui`.

```
UI route coverage: all 205 published route(s) are rendered
```

## The principle

Each panel renders the **distinction its route went to the trouble of publishing**, not just the headline number. That is the whole reason these were worth building rather than stubbing:

| surface | the distinction it would have lost |
|---|---|
| emission-pipeline/history | `point_count` vs `distinct_observations` — a flat stretch is either a steady pipeline or a lane that stopped, and only `repeats_previous_observation` tells them apart |
| chain/burn, holders, concentration ×2 | every one publishes a **read count beside a total**; an aggregate over a partial read describes the subset that was read |
| concentration/history | `builder_versions` — a trend across a version change compares two definitions, not a movement |
| chain/indexer-lag | **fast is not current**: a lane that stopped an hour ago still reports excellent write latency |
| health/failure-reasons | `share` (of all checks) and `failure_share` (of the failures) have different denominators |
| network/randomness | `stored_round_span`, not the newest round — a reveal timelocked to a round that aged out cannot be verified |
| validators/economics | `excluded` is **rendered, not filtered**; a leaderboard that hides its exclusions reports a subset as the whole |
| accounts/top-holders | free vs delegated are different positions; 7d/30d/90d flows can disagree |
| accounts/{ss58}/root-claim | `field_sources` says the claim type is *measured* and the hotkey list *reconstructed* |
| subnet-lifecycle, emission-changes | a NULL block means "before we were watching", never block 0 |
| crowdloans | met the cap and settled are **independent** — four states, not two |
| candidates/curation/profiles/source-snapshots | each stage says whether it was **reachable**; unreadable renders as unknown, never as 0 |
| fixtures/{surface_id} | an absent fixture is an answer, not an error |
| alerts/triggers/{id} | the same 404 for a wrong token and a missing id, on purpose — the panel refuses to guess which |

## Two judgement calls I want visible

**`/domains/{tag}/summary` adds no information today.** The rollup list already carries every field the summary returns. Rather than bolt on a second fetch of identical data to move a number, I wired it as the *authoritative* record behind the expandable row — lazily (`enabled: open`), with the list values as fallback. That is the list/detail split done properly; it is not new data, and the code says so.

**Category C got rendered, not exempted.** #10300 listed six routes as "plausibly API-only... but it should be a recorded decision". Rendering them *is* the decision, and it means the ceiling can be 0 — which is only enforceable because there is no backlog left to hide behind an allowance.

## The ceiling at zero

`MAX_UNRENDERED_ROUTES` 25 → 0 turns the ratchet into the gate #10300 proposed: **a new route with no UI reference fails CI**. The zero is asserted in `tests/validate-ui-route-coverage.test.ts` as well as the script, so raising it is a visible test change rather than a one-character edit. If a route genuinely should not render, the fix is a named exemption with a reason — raising the ceiling would hide every other gap behind one number, which is the shape that let this reach 35 unnoticed.

## Verification

- `validate:ui-route-coverage` — 205/205, exit 0
- `tsc --noEmit` clean at repo root **and** in `apps/ui`
- eslint clean in both workspaces (0 errors; 9 pre-existing `react-refresh` warnings, none in new files)
- prettier clean under each workspace's own config
- `apps/ui` component suite: **53 files, 324 tests passing**
- `tests/validate-ui-route-coverage.test.ts` + `tests/pipeline-validator-parity.test.ts`: 18 passing
- Response shapes taken from **live production**, not inferred from the spec

New unit tests cover the reasoning that is easy to get wrong: a null flow window is not a disagreement, zero is its own direction, an unmeasured subnet is dropped rather than ranked, an exclusion with no stated reason still gets counted, and an unknown coverage count resolves to "cannot claim coverage" rather than "covered everything".

Screenshots waived per the maintainer convention on this repo.

---

## Changes since this was opened

Self-review after opening found four defects, three of them in my own new code. Recorded here rather than left in scattered comments.

**1. The candidate total was the exact bug this PR argues against.** `/api/v1/candidates` is 3.6 MB unlimited, and the panel fetched it whole to count an array. It accepts `?limit=` and **no total survives the trim** — so the obvious later optimisation would have silently turned the total into the limit. The total now comes from `source-snapshots`'s server-computed `summary.candidate_count`; list routes are fetched bounded as display samples that are never counted. Curation stays a whole fetch because `gap_total` sums across every curated subnet, and a sum over a truncated list is wrong, not smaller. **4,413,759 → 254,999 bytes (-94%).**

**2. `completeness_score` is 0–100, not a 0..1 ratio** like every neighbouring share on this API (measured 25..97 live). The panel multiplied by 100 and would have shipped "9000% complete".

**3. Crowdloans were mounted inside `{showAnalytics && …}`, which defaults to closed.** The sweep counts a source reference, so this would have passed the gate at 0 unrendered while the surface stayed effectively unseen — the rubber stamp #10300 warned against. Moved to the always-rendered body. I audited every other mount added here: the rest are unconditional or on a default view. The emission-gate log stays under `?view=admin` deliberately, because that is a real linkable tab beside the config-change table it belongs with, not a disclosure that starts closed.

**4. The header ticker credited CoinPaprika for a price we compute.** This is #10300's *second half*. The fetch had already moved to `/api/v1/network/tao-usd`, but the tooltip still read "Current TAO price · CoinPaprika", attributing our own liquidity-weighted median across qualifying wTAO/WETH pools to a venue that no longer supplies it. A tooltip is the only provenance a reader gets for that number. Fixed; the remaining `coinpaprika` mentions in `market.functions.ts` are the historical note explaining why the swap happened and how `market_cap`/`volume_24h` are now different quantities — those are correct and stay.

Both leaderboards also now say **"cheapest 15 of 123"** and **"largest 15 of 3,843"** rather than quoting a server total beside a silently truncated list. Every total quoted that way was checked against production at `limit=2/15/25/200` first — `total`, `excluded` and `account_count` are all server-computed and limit-independent.

## One caveat on this gate, recorded as #10517

`validate:ui-route-coverage` proves a route is **mentioned** in `apps/ui`, not rendered. Verified against the real exported function: a bare comment, an `.mdx` line, and an `ApiSourceFooter` `paths` string all count as rendered. That last one is a plain string array someone will reasonably add while touching a page.

Nothing is gamed today and all 205 routes genuinely render — but at a ceiling of 0 a green tick invites a reading the evidence cannot support, so I corrected **the check's own success message** here ("are referenced by apps/ui", not "are rendered"). Narrowing the evidence itself is left to #10517 for a decision.

Re-verified after all of the above: `tsc` clean in both workspaces, eslint clean in both, **198 test files / 1,517 tests passing**, sweep 205/205.

### DIFF
--- a/apps/ui/src/components/metagraphed/account-root-claim.tsx
+++ b/apps/ui/src/components/metagraphed/account-root-claim.tsx
@@ -1,0 +1,69 @@
+import { useQuery } from "@tanstack/react-query";
+import { accountRootClaimQuery } from "@/lib/metagraphed/queries";
+import { Panel } from "@/components/metagraphed/primitives";
+import { Skeleton, ErrorState } from "@/components/metagraphed/states";
+import { formatNumber, formatRelative } from "@/lib/metagraphed/format";
+
+/**
+ * `/api/v1/accounts/{ss58}/root-claim` (#10300), published and rendered
+ * nowhere.
+ *
+ * What this account's root stake would do in a swap, and which hotkeys it
+ * reaches. The panel exists because of one thing the payload says about
+ * itself:
+ *
+ * `field_sources` MARKS THE HOTKEY LIST AS RECONSTRUCTED, not measured. The
+ * claim type is read from chain storage; the hotkey list is inferred. Rendering
+ * them the same way would present an inference with the authority of a reading,
+ * which is the whole failure mode `field_sources` exists to prevent -- so the
+ * provenance is shown next to the thing it qualifies, not buried in a footer.
+ */
+export function AccountRootClaim({ ss58 }: { ss58: string }) {
+  const { data, isLoading, isError, error, refetch } = useQuery(accountRootClaimQuery(ss58));
+
+  if (isLoading) return <Skeleton className="h-[90px] w-full" />;
+  if (isError) return <ErrorState error={error} onRetry={() => void refetch()} />;
+
+  const c = data?.data;
+  // No root claim is a real answer for most accounts, not a gap — rendered as
+  // a plain statement rather than an error or an empty-state alarm.
+  if (!c || !c.claim_kind) {
+    return (
+      <Panel as="section" dense>
+        <p className="mg-type-data-sm text-ink-muted">
+          This account has no root-claim state recorded.
+        </p>
+      </Panel>
+    );
+  }
+
+  return (
+    <Panel as="section" dense>
+      <div className="grid grid-cols-2 gap-x-6 gap-y-3">
+        <div>
+          <div className="mg-type-label text-ink-muted">claim type</div>
+          <div className="mg-type-data text-ink" title="Read from SubtensorModule.RootClaimType.">
+            {c.claim_kind}
+          </div>
+        </div>
+        <div>
+          <div className="mg-type-label text-ink-muted">hotkeys reached</div>
+          <div className="mg-type-data tabular-nums text-ink">{formatNumber(c.hotkeys.length)}</div>
+        </div>
+      </div>
+
+      {/* The provenance sits beside the figure it qualifies. A reconstructed
+          list is inferred from other state, not read from it. */}
+      {c.hotkeys_source && c.hotkeys_source !== "measured" ? (
+        <p className="mt-3 mg-type-label text-ink-muted">
+          The hotkey list is <strong>{c.hotkeys_source}</strong> — inferred from other state rather
+          than read from chain storage, unlike the claim type above.
+        </p>
+      ) : null}
+
+      {c.queried_at ? (
+        <p className="mt-2 mg-type-label text-ink-muted">Read {formatRelative(c.queried_at)}.</p>
+      ) : null}
+    </Panel>
+  );
+}
--- a/apps/ui/src/components/metagraphed/alert-trigger-lookup.tsx
+++ b/apps/ui/src/components/metagraphed/alert-trigger-lookup.tsx
@@ -1,0 +1,134 @@
+import { useState, type FormEvent } from "react";
+import { useMutation } from "@tanstack/react-query";
+import { apiFetch } from "@/lib/metagraphed/client";
+import { Panel } from "@/components/metagraphed/primitives";
+import { formatNumber, formatRelative } from "@/lib/metagraphed/format";
+
+/** Mirrors src/alert-triggers.ts. */
+const OWNER_TOKEN_HEADER = "x-alert-trigger-owner-token";
+
+interface TriggerView {
+  id: string;
+  channel: string;
+  destination: string;
+  active: boolean;
+  netuid: number | null;
+  event_kind: string | null;
+  last_matched_at: string | null;
+  match_count: number;
+}
+
+/**
+ * `/api/v1/alerts/triggers/{id}` (#10300), published and rendered nowhere.
+ *
+ * The UI has always been able to CREATE a trigger — that is what the watch
+ * forms do — and never to read one back. So a caller who created an alert had
+ * no way to answer "is it still active, and has it ever fired", which is the
+ * only question you have after setting one up.
+ *
+ * It needs the `owner_token` issued once at creation, and the route returns the
+ * SAME 404 for a wrong token and a nonexistent id so it cannot be used to
+ * enumerate other callers' triggers. The panel says that rather than guessing
+ * which of the two happened — reporting "no such trigger" for a mistyped token
+ * would send someone hunting for a trigger that exists.
+ *
+ * `last_matched_at` being null is DISTINCT from `match_count` of 0: one means
+ * it has never fired, the other is the count. They agree today, and the panel
+ * shows the count with its own "never fired" wording rather than deriving one
+ * from the other.
+ */
+export function AlertTriggerLookup() {
+  const [id, setId] = useState("");
+  const [token, setToken] = useState("");
+
+  const lookup = useMutation({
+    mutationFn: async (vars: { id: string; token: string }): Promise<TriggerView> => {
+      const res = await apiFetch<TriggerView>(
+        `/api/v1/alerts/triggers/${encodeURIComponent(vars.id)}`,
+        { init: { headers: { [OWNER_TOKEN_HEADER]: vars.token } } },
+      );
+      return res.data;
+    },
+  });
+
+  function onSubmit(e: FormEvent<HTMLFormElement>) {
+    e.preventDefault();
+    lookup.mutate({ id: id.trim(), token: token.trim() });
+  }
+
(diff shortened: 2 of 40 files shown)


JSONbored:
Follow-up on my own panel — two defects found by checking the routes against production rather than trusting the shapes.

**The candidate total was a trap.** `/api/v1/candidates` is **3.6 MB** unlimited, and the panel fetched it whole to count an array. It *accepts* `?limit=`, and no total survives the trim — the response is `{candidates, generated_at, notes, schema_version}`. So anyone later adding a limit to shrink that payload would have silently turned the total into the limit: a confident wrong number, in the panel whose whole argument is that those are the ones worth catching.

The total now comes from `source-snapshots`'s server-computed `summary.candidate_count` (2,184 — verified equal to the full array), and the list routes are fetched **bounded, as display samples that are never counted**. Curation stays a whole fetch, because `gap_total` sums over every curated subnet and a sum over a truncated list is not a smaller sum, it is a wrong one. Sample lists are headed "latest N", never "N of M".

```
page load: 4,413,759 → 254,999 bytes  (-94%)
```

**`completeness_score` is already a percentage.** 0–100, measured 25..97 across a live sample — not the 0..1 ratio every neighbouring share on this API uses. The panel multiplied by 100 and would have shipped "9000% complete".

Re-verified after both: `tsc` clean, eslint clean, 324 `apps/ui` tests passing, sweep still 205/205.

PR #10598, opened 2026-08-11
feat(graphql): mirror the coverage ratio, with its evidence attached

### LINKED ISSUE #10476: graphql(revenue): coverage fields on the SDL
Mirror the revenue routes into GraphQL.

## Deliverable

Revenue + coverage fields on the subnet type and a network-level coverage query.

## Watch

- Parity must be checked **in both directions** — a one-directional test hides the reverse gap.
- `Int` is 32-bit in GraphQL. Revenue in cents, or cumulative TAO in rao, overflows it. Use the appropriate scalar.
- Nullability must match the Zod source exactly: `coverage_ratio` is genuinely nullable and the SDL must say so, or consumers will treat absent as zero.

---
Part of #10439 — Phase 4 · surfacing. Maintainer-only.

JSONbored:
## Summary

`revenue` was **explicitly dropped** from the GraphQL `Surface` projection, with a comment deferring it to this issue — so the coverage ratio existed over REST and MCP and nowhere here. Adds two Query fields and the eight types behind them.

## The evidence travels with the number

That is the whole point. `provenance` is non-null on every figure, and `sources[]` reports every declared surface with `contributes` and `excluded_reason` — so a caller can see **why** a subnet with visible figures reports a null headline rather than inferring it. There is no selection set that yields a revenue figure while making it impossible to ask how it was obtained.

## Nullability mirrors the Zod source exactly

`revenue_usd`, `coverage_ratio` and `subsidy_multiple` are nullable because **absent revenue is `null`, never `0`** — and that is the *normal* case, true of 127 of 129 subnets. A non-null `Float` here would have forced every consumer to read absent as zero, which is precisely the false-claim-at-scale this epic exists to avoid.

## Every amount is Float, not Int

GraphQL's `Int` is 32-bit signed. A cumulative revenue figure in cents, or a TAO amount in rao, overflows it — and a non-null `Int` carrying one raises on every request and **nulls its whole surrounding object**, which is how a well-formed, confident, wrong answer reaches a client that renders `data` without inspecting `errors`.

## One shape, not two

Both artifacts nest the same `SubnetRevenue`, so the per-subnet card and the network table resolve to **one** published type rather than two structurally identical ones — declared in `PUBLISHED_NAMES` for both traversal paths.

The resolvers are a thin projection over the same `src/revenue-load.ts` that REST and MCP compose from. Nothing is recomputed, so the three surfaces cannot disagree about a subnet's ratio. The network field reads observations **once** for all 129 subnets rather than per subnet.

Neither field errors on a subnet with no revenue data: 127 of 129 are in that state, so an error would make the normal case look like a broken endpoint and a caller sweeping the network would see 127 failures instead of 127 answers.

## The gates drove this, in both directions

`graphql-component-parity` compares each SDL type field-by-field against the component it claims to mirror, so it named every divergence as I introduced it — the auto-generated nested component names, then `contract_version`/`notes` from the artifact envelope. I first tried declaring those as dropped and the gate correctly kept failing; they are real published fields, so the SDL declares them.

`published-names` additionally requires the SDL description and the registry description to be **character-identical**, which caught two that had drifted while I was writing them.

## Verification

- **All 40 CI `validate:*` scripts pass**, including all six `graphql-*` gates: `types-drift` · `route-parity` · `component-parity` · `query-arguments` · `tier-parity` · `hand-written-checks` · `published-names`
- **Full suite: 823 files, 18,845 tests.** Two hardcoded count assertions move with the two new bindings — `QUERY_BINDINGS` 196→198 and exact argument pairs 187→189. Found by running the full suite rather than the files I touched, after the same class of assertion failed CI on #10578
- `build` · `typecheck` · `lint` · `format:check` clean; `generated/graphql/types.ts` regenerated and committed
- Rebased onto `c8bf08431`; rebuild produced no diff beyond an unrelated `operational-surfaces.json` re-drift from #10596's registry edit, which is deliberately not gated and is not included here

Closes #10476

### DIFF
--- a/generated/graphql/types.ts
+++ b/generated/graphql/types.ts
@@ -1422,6 +1422,21 @@ export type ChainRegistrationsSubnet = {
   registrations_per_registrant?: Maybe<Scalars['Float']['output']>;
 };
 
+/** Every subnet's coverage in one response. Subnets with NO observed revenue are INCLUDED with null ratios rather than dropped -- omitting them would make the covered set look like the whole network. Mirrors GET /api/v1/chain/revenue-coverage. */
+export type ChainRevenueCoverage = {
+  __typename?: 'ChainRevenueCoverage';
+  contract_version?: Maybe<Scalars['String']['output']>;
+  generated_at: Scalars['String']['output'];
+  /** Public-safe notes; a string or a string list depending on the adapter. */
+  notes?: Maybe<Scalars['JSON']['output']>;
+  /** How many subnets have a readable revenue figure. Against subnet_count this is the honest headline, stated rather than left to be inferred from nulls. */
+  observed_count: Scalars['Int']['output'];
+  schema_version: Scalars['Int']['output'];
+  subnet_count: Scalars['Int']['output'];
+  subnets: Array<SubnetRevenue>;
+  window_days: Scalars['Int']['output'];
+};
+
 /** Network-wide axon-serving announcement leaderboard (#5873). The network-wide counterpart of subnet_serving. Mirrors GET /api/v1/chain/serving's data envelope. */
 export type ChainServing = {
   __typename?: 'ChainServing';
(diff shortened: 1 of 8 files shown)


JSONbored:
## Follow-up: patch coverage, and a stale catalog picked up on the way

`codecov/patch` read **0.00% of diff hit**. The two resolvers and their three helpers are runtime code in `src/graphql.ts`, and nothing exercised them — I ran the validators and the full suite on this PR and skipped the diff-intersection check, which is the one that measures what the gate measures.

Seven tests now drive both fields through the real `handleGraphQLRequest`, covering what the field is *for* rather than that it returns:

- the normal case (no observations) serves a real emission side with every ratio `null`
- a superseded surface is reported with its `excluded_reason` and never contributes
- a surface with no `revenue` block is not listed among `sources` at all
- an unknown netuid answers rather than erroring
- the network table includes the uncovered with null ratios
- malformed artifacts degrade rather than throw — a non-array `surfaces` key declares nothing, `usd: 0` prices the emission at 0 and yields null ratios rather than dividing by it, and a junk netuid row is skipped instead of counted into `subnet_count`

Diff-intersection now shows **0 uncovered changed lines across all 120**.

## The unrelated file in this diff

`public/metagraph/operational-surfaces.json` is regenerated here. #10596 added `sn-53-engy-gateway-meta` to the registry without rebuilding the derived catalog:

```
eligible in surfaces.json    : 620
in operational-surfaces.json : 619
  - sn-53-engy-gateway-meta
```

The surface is probe-enabled, public-safe and a callable kind, so as it stands it will never be health-probed and `call_subnet_surface` rejects its id. **`validate:operational-surface-parity` is failing on main right now** and would fail every PR until someone rebuilds. I reverted this file once as out-of-scope before realising it was the blocker.

Issue #10659, opened 2026-08-11
perf(neon): buffer capture writes through a Durable Object so the compute can autosuspend


JSONbored:
## Problem

The Neon compute **can never autosuspend**, so we pay for 24/7 allocated capacity to do a few minutes
of real work. Measured 2026-08-11 on project `green-dawn-75468244`:

```
lane_health inter-write gaps, 6h window:
  8,953 writes | avg gap 2.4s | MAX gap 16.7s
  gaps over 60s: 0    gaps over 300s: 0

cpu_used_sec / active_time = 208,931 / 343,284 = 0.61 CU while active
active_time / elapsed      = 72%
utilization                = ~2.8%
```

`suspend_timeout_seconds` is 300, and Neon's minimum is 60 — **neither is reachable** at a 16.7s
maximum inter-write gap. The compute is therefore always on, and ~97% of what we pay for is idle
allocation rather than work.

**The crons are a red herring.** `TAO_USD_INDEX_CRON = "* * * * *"` (`workers/config.ts:450`) looks
like the culprit and is ~4% of the volume. The real stream is the capture lanes writing continuously.

## Why a Durable Object, and not a queue

**Cloudflare Queues cannot do this.** `max_batch_timeout` caps at **60 seconds**
(https://developers.cloudflare.com/queues/platform/limits/), so a consumer still writes every minute
and the compute still never suspends. The primitive has to be able to hold state and flush on an
arbitrary interval:

- **Durable Object alarm** — flushes on any interval, storage is durable so nothing is lost on
  eviction. Existing precedents in-repo: `workers/mcp-session-hub.ts:136`, `workers/alerter-hub.ts:428`.
- **Workflow `step.sleep`** — also works, but bills per step and is an awkward fit for a metronome.
- **Isolate buffering** (the `USAGE_ROLLUP_FLUSH_COUNT` / `_AGE_MS` pattern in `workers/data-api.ts:5129`)
  is lossy on isolate eviction — acceptable for usage telemetry, **not** for capture data.

## Proposed change

A Durable Object that buffers Neon writes in its own storage and flushes them on a ~10-minute alarm,
with the `*-neon-write.ts` lanes routed through it: `src/capture-state-neon-write.ts`,
`src/chain-detail-neon-write.ts`, `src/neurons-neon-write.ts`,
`src/nominator-positions-neon-write.ts`, `src/ledger-neon-write.ts`,
[...]

PR #10679, opened 2026-08-11
refactor(neon): route every write through one uniform, bufferable runner

### LINKED ISSUE #10677: perf(neon): buffer tao_usd_index, the last writer keeping the compute awake
## Problem

#10659 routed the six `src/*-neon-write.ts` capture lanes through the write-behind buffer so the Neon
compute can autosuspend. **`tao_usd_index` is the remaining writer that keeps it awake**, and it is the
most frequent one by interval: `TAO_USD_INDEX_CRON = "* * * * *"` (`workers/config.ts:450`) writes
**exactly 60 rows/hour**, verified against production:

```
hour   writes  first      last
05:00      60  05:00:47   05:59:47
04:00      60  04:00:47   04:59:35
03:00      60  03:00:47   03:59:47
```

`suspend_timeout_seconds` is 300 and Neon's floor is 60, so a write every 60s means the compute can
never suspend **no matter how many other lanes buffer**. Partial buffering saves exactly nothing.

## Why it could not just be wired like the others

The six capture lanes were drop-in because every statement is a write whose result is discarded — no
`RETURNING`, every `sql.unsafe(...)` awaited bare. `ingestTaoUsdIndex` is neither
(`workers/data-api.ts:7664`):

```ts
const written = await sql<Record<string, unknown>>`
  INSERT INTO tao_usd_index (...) VALUES (...)
  ON CONFLICT (block_number, observed_at) DO NOTHING
  RETURNING block_number`;
return { inserted: written.length > 0 };
```

Two blockers:

1. **It uses the tagged-template call form**, not `sql.unsafe(text, values)`. `createBufferedPgSql`
   implements the one-method `PgUnsafe` surface only.
2. **It consumes the result.** `RETURNING block_number` + `written.length > 0` is how the caller learns
   whether the row was new. A buffered enqueue cannot know that, and reporting `inserted: true`
   optimistically would be a confident lie about a row that has not been written yet.

## Proposed approach

Report what is actually true rather than flattening it. The insert is already idempotent
(`ON CONFLICT DO NOTHING` on `(block_number, observed_at)`), so the honest shapes are `inserted` for a
direct write and a distinct `buffered` for an enqueue — never `inserted: true` for a row still sitting
in the buffer. Confirm first what actually consumes `inserted`: `runTaoUsdIndexWatchdog` reads the
**table**, not this return value, so the blast radius may be only the cron's diagnostic result.

The buffered path also needs `unsafe(text, values)`, so the statement moves from the tagged template
to the parameterised form the other lanes already use.

## Acceptance criteria

- [ ] `tao_usd_index` writes go through the buffer when its lane is flagged.
- [ ] No caller is told a buffered row was `inserted`.
- [ ] Measured after enabling every lane: inter-write gaps exceed `suspend_timeout_seconds`, and the
      compute is observed suspending (`pg_postmaster_start_time()` moving, `active_time` growing
      slower than wall clock).

Blocks the payoff from #10659. Until this lands, `NEON_WRITE_BUFFER_LANES` cannot deliver a saving
however many other lanes are named in it.

### LINKED ISSUE #10678: refactor(neon): route every write lane through one shared runner and buffer the remaining five
## Problem

#10659 shipped the write-behind buffer and wired `blocks-head` / `raw-capture-state`. The other five
`src/*-neon-write.ts` lanes still open their own connection, and **partial buffering saves exactly
nothing** — one lane writing every 12s pins the Neon compute as hard as ten do.

Measured verdict rates over 6h (2026-08-11), the lanes still unwired:

| lane | writes/hour |
|---|---:|
| `neon:chain-detail` | 295 |
| `neon:account-balances` (+`-pass`) | 161 each |
| `neon:nominator-positions` (+`-pass`/`-prune`) | 79 each |
| `neon:validator-nominator-counts` | 42 |
| `neon:neurons` / `neuron_daily` / `account_position_daily` | 22 each |

## The duplication that made this worth doing properly

All six runners built their connection with the **identical** line:

```ts
const sql =
  deps.sql ??
  (hyperdrive?.connectionString && ctx ? createPgSql(hyperdrive, ctx) : null);
```

Adding the buffer branch to each would have made six copies of a decision with one right answer, and
a lane that forgets the branch still works, still passes its own tests, and silently denies the buffer
its saving.

## Proposed change

Extract one `neonWriteRunner(env, ctx, lane, hyperdrive)` into `src/neon-write-buffer.ts` and collapse
all six runners onto it, including the one #10659 wired inline. Add a test asserting no
`src/*-neon-write.ts` calls `createPgSql` directly, so a seventh lane cannot quietly bypass it.

## Acceptance criteria

- [ ] All six lanes route through one shared runner.
- [ ] A guard test fails if any `-neon-write.ts` builds its own connection.
- [ ] Behaviour unchanged with the flag empty (the default).

Does not on its own let the compute suspend — `tao_usd_index` still writes every 60s.

JSONbored:
## Summary

Every Neon write now goes through **one** runner, and every lane can be buffered — including the one
that could not be before.

### One runner, not six

All six `src/*-neon-write.ts` lanes built their connection with the identical line:

```ts
const sql =
  deps.sql ??
  (hyperdrive?.connectionString && ctx ? createPgSql(hyperdrive, ctx) : null);
```

Adding the buffer branch to each would have made six copies of a decision with one right answer, and
the failure mode is silent: a lane that forgets the branch still works, still passes its own tests,
and simply denies the buffer its saving. They now share `neonWriteRunner`, and a test asserts that no
`src/*-neon-write.ts` calls `createPgSql` directly — so a seventh lane cannot quietly bypass it.

### tao_usd_index is no longer the odd one out

It was the **most frequent single writer** — `TAO_USD_INDEX_CRON = "* * * * *"`, verified at exactly
60 rows/hour against production — so on its own it kept the compute from ever reaching the 300s
suspend timeout. It could not be deferred because it used the tagged-template form with
`RETURNING block_number` and reported `inserted: written.length > 0`.

It now issues the same parameterised, result-discarding statement every other lane does. **Nothing
real is lost:** the Workers runtime discards a `scheduled()` return value, so the only consumer of
`inserted` was this repo's own tests; `ON CONFLICT DO NOTHING` already makes a repeated height a
genuine no-op; and whether heights are landing is measured where it belongs, by
`src/tao-usd-index-watchdog.ts` reading the table. The return is now `{ written: boolean }`, which
says what actually happened.

### A latent trap, closed

`createBufferedPgSql` previously returned `[]` for **any** statement, including one with `RETURNING`.
That is the one way buffering could produce a *confidently wrong* answer rather than a slow one:
`RETURNING id` + `rows.length > 0` is how a caller learns its row was new, and an empty result reads
as "already present" for a row that has not been written yet. The buffered runner now refuses such a
statement outright, so the contract — fire-and-forget writes only — is enforced rather than
documented. The check is word-bounded and case-insensitive, with a test proving a column named
`returning_at` does not trip it.

### zod instead of a hand-rolled guard

`decodeStatement` parsed untrusted JSON with three hand-written `typeof` checks. It now parses with a
`.strict()` zod schema, and `BufferedStatement` is inferred from it. The shape crosses two trust
boundaries — a Durable Object request body, and durable storage written by a possibly older deploy —
and a hand-rolled check drifts from the type it is supposed to enforce the moment a field is added.

## Validation

```
npm run typecheck                              # clean
npm run lint · npm run format:check            # clean
npm run validate                               # exit 0
npm run validate:worker-types-parity           # 3 workers match
npx vitest run <10 suites>                      # 193 passed
```

**Patch coverage measured by diff intersection — zero uncovered statements and zero uncovered
branches across every changed file:**

```
src/neon-write-buffer.ts                stmt: NONE | branch: NONE
src/capture-state-neon-write.ts         stmt: NONE | branch: NONE
src/chain-detail-neon-write.ts          stmt: NONE | branch: NONE
src/hyperparams-identity-neon-write.ts  stmt: NONE | branch: NONE
src/ledger-neon-write.ts                stmt: NONE | branch: NONE
src/neurons-neon-write.ts               stmt: NONE | branch: NONE
src/nominator-positions-neon-write.ts   stmt: NONE | branch: NONE
workers/data-api.ts                     stmt: NONE | branch: NONE
```

## What this does and does not change

`NEON_WRITE_BUFFER_LANES` still **defaults to empty**, so this deploy changes no write path. What it
changes is that every lane is now *able* to buffer — which is the precondition for the saving, since
partial buffering yields exactly nothing.

Turning the flag on for all lanes is a separate, reversible config change, and the accepted trade-off
from #10659 applies when it happens: `blocks_head` buffered at a ten-minute flush means `/blocks`
routes serve a head up to ~10 minutes behind. Every affected watchdog bound is `2 * HOUR`, so this
sits twelve ticks inside a contract that already exists.

Closes #10677
Closes #10678

### DIFF
--- a/scripts/validate-untyped-db-reads.ts
+++ b/scripts/validate-untyped-db-reads.ts
@@ -53,7 +53,7 @@ import { repoRoot } from "./lib.ts";
  */
 // Annotated `number` rather than left as the literal so the "none left"
 // message below stays reachable code as the ceiling falls to zero.
-export const MAX_UNTYPED_READS: number = 105;
+export const MAX_UNTYPED_READS: number = 104;
 
 /** Where a read can live. `scripts/` is excluded: it does not serve traffic. */
 const SOURCE_DIRS = ["src", "workers"];
--- a/src/capture-state-neon-write.ts
+++ b/src/capture-state-neon-write.ts
@@ -20,21 +20,13 @@
 // upsert would silently drop that and replace a known value with NULL -- the
 // same shape as #9634's last_ok, one table over.
 import { laneHealthStore } from "./lane-health-store.ts";
-import {
-  createBufferedPgSql,
-  neonWriteBufferEnabled,
-  type NeonWriteBufferNamespace,
-} from "./neon-write-buffer.ts";
+import { neonWriteRunner } from "./neon-write-buffer.ts";
 import {
   neonDualWriteEnabled,
   recordNeonWriteVerdict,
   type NeonWriteResult,
 } from "./neon-write.ts";
-import {
-  createPgSql,
-  type HyperdriveLike,
-  type WaitUntilLike,
-} from "./pg-sql.ts";
+import { type HyperdriveLike, type WaitUntilLike } from "./pg-sql.ts";
 import type { LaneHealthDb } from "./lane-health.ts";
 
 export const BLOCKS_HEAD_NEON_LANE = "blocks-head";
@@ -68,19 +60,9 @@ async function runner(
     return { sql: null, laneDb, now, attempted: false };
   }
   const hyperdrive = env?.HYPERDRIVE as HyperdriveLike | undefined;
-  // #10659: a flagged lane enqueues into the write-behind buffer instead of
-  // opening its own connection. Drop-in because PgUnsafe is one method, these
-  // statements carry no RETURNING, and the caller awaits the result bare --
-  // see src/neon-write-buffer.ts for how that was established. Defaults OFF
-  // (empty lane list), so this deploy changes nothing until a lane is named.
-  const buffer = env?.NEON_WRITE_BUFFER as NeonWriteBufferNamespace | undefined;
-  const sql =
-    deps.sql ??
-    (buffer && neonWriteBufferEnabled(env, lane)
-      ? createBufferedPgSql(buffer, lane)
-      : hyperdrive?.connectionString && ctx
-        ? createPgSql(hyperdrive, ctx)
-        : null);
+  // #10659: buffered when the lane is flagged, direct otherwise. Defaults OFF
+  // (empty lane list), so this changes nothing until a lane is named.
+  const sql = deps.sql ?? neonWriteRunner(env, ctx, lane, hyperdrive);
   if (!sql) {
     // Enabled but unbound is a MISCONFIGURATION, not a quiet no-op.
     await recordNeonWriteVerdict(
--- a/src/chain-detail-neon-write.ts
+++ b/src/chain-detail-neon-write.ts
@@ -19,11 +19,8 @@ import {
   writeRowsToNeon,
   type NeonWriteResult,
 } from "./neon-write.ts";
-import {
-  createPgSql,
-  type HyperdriveLike,
-  type WaitUntilLike,
-} from "./pg-sql.ts";
+import { type HyperdriveLike, type WaitUntilLike } from "./pg-sql.ts";
+import { neonWriteRunner } from "./neon-write-buffer.ts";
 import type { LaneHealthDb } from "./lane-health.ts";
 
 // ---------------------------------------------------------------------------
@@ -199,9 +196,10 @@ export async function mirrorChainDetailToNeon(
   }
 
   const hyperdrive = env?.HYPERDRIVE as HyperdriveLike | undefined;
+  // #10659: buffered when the lane is flagged, direct otherwise. Defaults OFF
+  // (empty lane list), so this changes nothing until a lane is named.
   const sql =
-    deps.sql ??
-    (hyperdrive?.connectionString && ctx ? createPgSql(hyperdrive, ctx) : null);
+    deps.sql ?? neonWriteRunner(env, ctx, CHAIN_DETAIL_NEON_LANE, hyperdrive);
   const laneDb = laneHealthStore(env, deps.laneHealthDb);
   const now = deps.now ?? Date.now;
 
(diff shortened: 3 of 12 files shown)


JSONbored:
## Updated: `blocks-head` is now excluded from buffering

Investigating the freshness requirement turned up something that changes this PR's ceiling, so recording it here.

`blocks_head` is not merely a write — it is the **block explorer's live read path**. `src/blocks-cold-tier.ts` routes `block_number > seam` at `blocks_head b LEFT JOIN chain_detail_blocks c`, which is precisely the window "between a block being seen and being decoded". Deferring that write defers the visible chain tip by the whole flush interval.

So `NEVER_BUFFER_LANES` now makes naming it in `NEON_WRITE_BUFFER_LANES` a **no-op**, enforced rather than documented — the tempting move when the compute still will not suspend is to add the one remaining high-frequency lane and see what happens.

### The honest consequence

`blocks-head` writes every ~12s, and the compute only suspends after **300s** of quiet. Those two facts are irreconcilable:

> **Neon suspension and a live block explorer are mutually exclusive while the head is served from Neon.**

So this PR no longer promises a sleeping compute. What it buys is a lower CU *while active* — far fewer statements per hour, so autoscaling sits nearer the 0.25 floor — on top of the ~40x write-rate cut already measured from #10664.

Making suspension possible would mean moving the above-seam read off Neon entirely. That is a much larger change than it looks, because the `LEFT JOIN` above cannot span two stores: `src/blocks-cold-tier.ts` already warns that one side in the other store returns rows with every `c.` column null — "a wrong answer with a valid shape". I have not attempted it here.

Also lowers `MAX_UNTYPED_READS` 105 → 104: dropping tao_usd_index's `sql<Record<string, unknown>>` escape hatch improved the read ratchet, and that validator fails until the gain is locked in.

Issue #10680, opened 2026-08-11
api(owner-cut): the share is read from an artifact that never carried it, so every subnet reports "not read"


JSONbored:
`/api/v1/subnets/{netuid}/owner-cut` and the `get_subnet_owner_cut` MCP tool report `"owner cut share not read"` for **all 129 subnets**, so the accrual, its TAO/USD pricing and the whole disposition are `null` everywhere.

## Cause

Both read the share from the served artifact:

```ts
const parameters = await readArtifact(env, "/metagraph/network/parameters.json");
const ownerCut = parameters.data?.subnet_owner_cut_effective;
```

That artifact publishes **no owner-cut field at all**. Verified in production:

```
$ curl -s https://api.metagraph.sh/metagraph/network/parameters.json | jq '[to_entries[]|select(.key|test("owner"))]'
[]

$ curl -s https://api.metagraph.sh/api/v1/network/parameters | jq '.data|{subnet_owner_cut, subnet_owner_cut_effective}'
{ "subnet_owner_cut": null, "subnet_owner_cut_effective": 0.179995422293431 }
```

The **live route** computes it. The **artifact** does not carry it. The read has therefore never returned anything.

## Why it went unnoticed

This is #10566's shape exactly: a correct-looking decline standing in for a read that never happened. `"owner cut share not read"` is a true sentence and the honest output for an unresolvable share — it just was not supposed to be the answer for every subnet on the network, forever. Nothing failed, because refusing to assume 18% is the designed behaviour.

## The distinction that makes the live read necessary

`subnet_owner_cut_effective` is a **tri-state** reconstruction (#10484):

- storage **has a value** → that value;
- storage is **unset** → the runtime default, `11796/65535` ≈ 18%;
- the read **failed** → `null`.

So a local constant is not a substitute: "unset on chain, so use the default" and "we could not ask" are different answers, and only a real read tells them apart.

## Consequence

Blocks item 2 of every per-subnet attribution issue (#10489–#10500) — the owner-cut disposition cannot be recorded when the accrual is null — and makes the #10511 money-map panel show `Not read` in every bucket for reasons unrelated to the flow streams.

## Fix

Read through `loadNetworkParameters` (KV-cached) on both surfaces, injected so tests do not reach a live node. `get_subnet_owner_cut` becomes genuinely open-world and its annotation must say so.

## Second defect, found while fixing the first

[...]

Issue #10566, opened 2026-08-10
worker(revenue): the probe lane has no scheduled caller


JSONbored:
#10444 shipped the probe lane as a library. Nothing calls it.

## Observed

```
$ grep -rn "revenue-probe\|revenue-observation\|revenue-coverage\|revenue-serving\|revenue-chain-inflow" src/ workers/ scripts/
tests/revenue-observation.test.ts:9:  import { extractRevenue } from "../src/revenue-observation.ts";
tests/revenue-probe.test.ts:12:  } from "../src/revenue-probe.ts";
tests/revenue-coverage.test.ts:12: } from "../src/revenue-coverage.ts";
tests/revenue-chain-inflow.test.ts:4: import { aggregateChainInflow } from "../src/revenue-chain-inflow.ts";
tests/revenue-serving.test.ts:5:  import { buildSubnetRevenue } from "../src/revenue-serving.ts";
```

Only `tests/` imports them. No `workers/` module calls `runRevenueProbe`, and no `wrangler*.jsonc` carries a revenue cron. PR #10560 shipped exactly three files — `migrations/neon/0016_revenue_observations.sql`, `src/revenue-probe.ts`, `tests/revenue-probe.test.ts`. The scheduled invocation was never part of it.

`revenue_observations` has never been written to. Every route the epic serves reports `revenue_usd: null` for all 129 subnets — including SN64 and SN51, the two the epic exists to measure.

## Why nothing failed

The epic's own rule is that absent revenue serialises as `null`, never `0`. A dead producer is therefore indistinguishable from the correct, honest answer for 127 of 129 subnets. There is no error, no empty-artifact alarm, and no lane verdict — because there is no lane.

## Deliverable

- A `REVENUE_PROBE_CRON` in `workers/config.ts` and a matching `triggers.crons` entry, dispatched in `workers/api.ts` alongside the other capture lanes
- A store module writing `revenue_observations` (UPSERT on `(surface_id, period)`) and `revenue_probe_failures`, via `producerStore` — the `SUBNET_BURN_CAPTURE_CRON` lane is the closest analogue
- A lane verdict through `recordLaneVerdict`, so `lane-alarm` can see it. A capture lane with no verdict is how #10172's frozen series went unnoticed for five hours
- A staleness watchdog, since a lane that stops is otherwise invisible for exactly the reason above

## Cadence

[...]

Issue #10548, opened 2026-08-10
registry(sn37): nine surfaces on a Railway host that no longer exists, invisible because probe.enabled is false


JSONbored:
All nine of SN37 Aurelius's registered stats surfaces answer the same thing:

```
GET https://new-collector-api-production.up.railway.app/stats/deposits
404 {"status":"error","code":404,"message":"Application not found","request_id":"..."}
```

`Application not found` is **Railway reporting the host itself no longer exists** — not a 404 from the app for a route that moved. Every surface on that origin is dead, not just these.

## The surfaces

`/stats/deposits` · `/stats/submissions` · `/stats/validators` · `/stats/pipeline` · `/stats/transparency-report` · `/stats/consensus-deviation` · `/stats/version-distribution` · `/stats/archetypes` · `/stats/influence-distribution`

All nine are registered `auth_required: true`, `probe.enabled: false` — so **the prober never touches them and their health has never been measured.** That is why nine dead surfaces sat in the registry without a single incident being raised.

## Why this is worth its own issue

Found while running the #10464 revenue sweep, which is the only reason anything fetched them. The revenue classification is recorded there (`provenance: none`, dated), but that PR is about revenue and this is a coverage problem:

**a surface registered `probe.enabled: false` is invisible to health forever.** SN37 is the demonstration — an entire origin can disappear and the registry keeps advertising nine endpoints on it.

## Worth checking

- [ ] How many other surfaces are `probe.enabled: false` on an origin with no live surface at all? That is the population where this can recur silently.
- [ ] Should an auth-gated surface still be probed for *liveness* (does the host resolve, does it answer anything) even when the body cannot be read? A 401 is a healthy signal; `Application not found` is not, and today we distinguish neither.
- [ ] Whether SN37 has a live replacement origin, in which case these nine want re-pointing rather than removing.

Issue #354, opened 2026-06-12
Local-dev + testnet→mainnet quickstart (real SKILL + fixtures)


JSONbored:
`/api/v1/local` hands a ws:// URL and walks away. Ship a real 'stand up a local Bittensor chain' guide + a `metagraphed localnet` helper (docker-compose, chain-spec, register a test subnet) — the develop-before-mainnet path a new builder needs.

PR #10716, opened 2026-08-11
feat(queues): sweep one subnet per message, not eight per invocation (#10715)

### LINKED ISSUE #10715: arch(queues): the sweep's slice size is the invocation budget, not a cadence
`SWEEP_BATCH_SIZE = 8` in `src/attribution-sweep.ts` exists for one reason: 128 subnets × up to 8 sources is ~1000 outbound requests, and one scheduled invocation cannot do that. So the cron sweeps eight subnets and a full pass takes sixteen hours.

**That is not a cadence. It is the invocation budget, dressed as one.**

## What a queue changes

One message per subnet is one invocation per subnet:

- the whole registry is swept from a single tick, instead of over sixteen hours;
- the slice size stops being a number anyone has to justify;
- a subnet whose fetch fails is **retried on its own**, instead of being dropped until the next hour comes round to it again;
- a subnet that keeps failing lands in a **dead-letter queue**, which `src/dead-letter.ts` already records as a lane verdict. Today it produces an `unreachable` row forever and nothing says the lane itself is struggling.

The last two are the real argument. A cron branch has no retry and no dead letter — a failure inside it is simply gone.

## The cron does not disappear, and should not

Queues do not self-trigger. Something has to be the clock, and the honest options are:

1. **Cron as producer** — the cron enqueues and does no work. One cron per lane, but it is a clock rather than the lane.
2. **Self-perpetuating queue** — the consumer re-enqueues itself with `delaySeconds`. No cron at all, but **no recovery**: if the chain breaks once, that item stops forever and nothing restarts it. That is exactly the silent-stoppage class #10566, #10680 and #10548 all turned out to be.
3. **One heartbeat** — a single cron whose only job is "ensure everything that should be running, is", feeding queues and workflows.

(3) is the end state, and (1) is the step that gets there: once every lane is a producer, the producers collapse into one heartbeat trivially. (2) is the tempting one and the wrong one — it trades a supervisor for a silent failure mode.

## Deliverable

Migrate the attribution sweep as the proving ground — it is the newest lane, has no legacy behaviour, and its slice size is the clearest example of the problem.

- `attribution-sweeps` + `attribution-sweeps-dlq`, batch 3, 3 retries, concurrency 5.
- The cron becomes a producer, enqueuing **every** due subnet.
- The consumer sweeps one subnet per message, reading the registry record at **delivery** so a queued message cannot sweep a stale copy of its surfaces.
- The DLQ registered in `DEAD_LETTER_LANES` so an exhausted message becomes a lane verdict rather than vanishing.

Then the same shape for the origin-reachability and revenue-probe lanes, and finally the heartbeat that replaces their three crons with one.

JSONbored:
## The problem, stated precisely

`SWEEP_BATCH_SIZE = 8` exists because 128 subnets × up to 8 sources is ~1000 outbound requests, and one scheduled invocation cannot do that. So the cron swept eight subnets and a full pass took **sixteen hours**.

**That is not a cadence. It is the invocation budget, dressed as one.**

## What changes

One message per subnet is one invocation per subnet. The whole registry is swept from a single tick, and the slice size stops being a number anyone has to justify.

| | cron branch | queue |
| --- | --- | --- |
| Full pass | ~16 hours | one tick |
| A subnet whose fetch fails | dropped until next hour | **retried on its own** |
| A subnet that always fails | `unreachable` row forever, lane looks fine | **dead letter → lane verdict** |
| Budget | 8 subnets share one invocation | one each |

The middle two are the real argument. A cron branch has no retry and no dead letter — a failure inside it is simply gone.

## The cron does not disappear, and should not

Queues do not self-trigger; something has to be the clock. This demotes the cron from *being* the lane to being a **producer that enqueues and does no work**.

The tempting next step — the consumer re-enqueueing itself with `delaySeconds`, no cron at all — has **no recovery**. If the chain breaks once, that subnet stops forever and nothing restarts it. That is exactly the silent-stoppage class #10566, #10680 and #10548 all turned out to be.

The end state is **one heartbeat** feeding every lane, and every lane becoming a producer is the step that gets there. Once they all are, the producers collapse into one trivially.

## Details worth flagging

**The registry record is read at delivery**, not carried on the message — a message that waited in the queue must not sweep a stale copy of a subnet's surfaces. Same reason the webhook lane reads its subscription at delivery. A subnet deregistered while its message waited resolves to `no-sources` rather than throwing; asserted.

**A malformed body is acked, not retried.** A body that is not a netuid never will be, and retrying spends the whole budget to reach the dead letter with a message nobody can act on.

**A failed write retries**, because an unwritten sweep is a sweep that did not happen as far as anything reading the store is concerned.

**The DLQ is registered in `DEAD_LETTER_LANES`** — a dead letter nobody consumes is a second log (#354/#363), and the consumer branch on `batch.queue` comes first so a sweep message is never handed to the webhook deliverer.

## Operational

Both queues created out of band (`wrangler queues create`), as the four existing ones were:

```
attribution-sweeps · attribution-sweeps-dlq
```

Config validated with `wrangler deploy --dry-run` — both bindings resolve.

## Validation

```
npx tsc --noEmit                clean
npm run lint / format:check     clean
npm run validate                pass
npm run build                   no drift
npx vitest run                  19464 passed, 0 failed
npm run scan:public-safety      pass
```

Patch coverage by diff intersection: **199 changed lines across `src/**` and `workers/**`, 0 uncovered**. 66 tests, including the consumer driven through the Worker's own `queue()` export.

Closes #10715

### DIFF
--- a/src/attribution-sweep.ts
+++ b/src/attribution-sweep.ts
@@ -297,123 +297,158 @@ export async function loadSweepRecord(
   }
 }
 
-// ── the lane ────────────────────────────────────────────────────────────────
+// ── the queue lane ──────────────────────────────────────────────────────────
+//
+// WHY THIS IS A QUEUE AND NOT A BIGGER CRON SLICE. The cron lane this replaced
+// swept a slice of eight subnets per tick, because 128 subnets x up to 8 sources
+// is ~1000 outbound requests and one scheduled invocation cannot do that. A full
+// pass took sixteen hours. That was never a cadence -- it was the invocation
+// budget, dressed as one.
+//
+// A queue removes the constraint rather than working around it. One message per
+// subnet is one invocation per subnet, so the whole registry is swept from a
+// single tick.
+//
+// AND IT REMOVES THE SCHEDULING MACHINERY ENTIRELY. The old lane ordered subnets
+// by staleness and took the eight oldest, which meant a store read on every tick
+// purely to decide what to skip. Enqueuing everything skips nothing, so the
+// staleness ordering, the slice size, and the producer's database handle all go
+// with it -- `swept_at` is now only a fact the API serves, never an input to
+// scheduling.
+//
+// It also turns two silent failures into visible ones:
+//
+//   - a subnet whose fetch fails is RETRIED on its own, instead of being
+//     dropped until the next hour came round to it again;
+//   - a subnet that keeps failing lands in the dead-letter queue, which
+//     src/dead-letter.ts records as a lane verdict. Under the cron it simply
+//     produced a `unreachable` row forever and nothing said the lane itself was
+//     struggling.
 
-/** Subnets swept per tick.
- *
- * A full pass is ~129 subnets x up to 8 sources = a thousand outbound requests,
- * which is not a tick, it is an incident. The lane sweeps a SLICE and picks it
- * by staleness, so the pass completes over a day without ever bursting. */
-export const SWEEP_BATCH_SIZE = 8;
-
-/**
- * The subnets to sweep next: least recently swept first, never-swept before
- * that.
- *
- * Self-balancing on purpose. A fixed rotation drifts out of step the moment a
- * tick is skipped or a subnet is added; ordering by staleness means the lane
- * converges on whatever is oldest no matter what happened before it.
- */
-export function nextToSweep(
-  netuids: number[],
-  sweptAt: Map<number, number>,
-  batch = SWEEP_BATCH_SIZE,
-): number[] {
-  return [...netuids]
-    .sort((a, b) => {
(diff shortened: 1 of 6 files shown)


JSONbored:
Superseded — this content landed in main via #10727, which was stacked on it and squash-merged with both commits.

Verified before closing: `attribution-sweeps` is declared in main's `wrangler.jsonc` (4 references) and `enqueueSweeps` is in `src/attribution-sweep.ts`. The branch itself is now stale — diffed against current main it shows 36 files and ~1980 deletions, because it would revert work merged since. Rebasing it would be reverting other people's changes, not resolving a conflict.

Nothing outstanding: #10715 is closed by #10727, and the follow-up collapse is tracked in #10735.

Issue #10696, opened 2026-08-11
observability(neon): page on write-buffer failures instead of only recording them


JSONbored:
## Problem

The write-behind buffer (#10659) writes lane verdicts, which are the durable **record** — but it
emits **no PostHog `$exception`**, which is the **notification** path. `src/lane-health.ts`'s header
draws that distinction deliberately: PostHog can be dropped by quota, so it is not the record; but a
row in `lane_health` pages nobody.

So today, every failure mode of the buffer is silent until someone reads the table or a freshness
watchdog trips up to **two hours** later:

| Failure | Consequence today |
|---|---|
| Flush throws mid-drain | Backlog held, retried next alarm — invisible |
| Hyperdrive unbound at flush | Backlog grows unboundedly — invisible |
| A statement is undecodable | **Discarded. This is data loss** — invisible |
| Buffer full, enqueue refused | Capture rows turned away — invisible |
| `alarm()` itself throws | **The buffer stops draining entirely** — invisible, and no alarm re-armed |

The last two are the ones that matter most. A full buffer means producers are outrunning the flush;
an alarm that throws means nothing drains ever again.

## Proposed change

Emit `recordExceptionEvent` from each path, all under one `route` so PostHog groups them as a single
issue rather than splitting one incident across several that each look half as urgent:

- `flush_failed` — a statement failed mid-drain
- `flush_hyperdrive_unbound` — nowhere to flush to, N statements held
- `flush_undecodable_statement` — a statement was discarded (data loss)
- `buffer_full` — an enqueue was refused (backpressure)
- `flush_alarm_threw` — the alarm's own catch, which also has to keep re-arming

A clean flush must page **nothing** — an inbox is only useful if the healthy path is silent.

## Not adding tracing spans

Deliberate. `POSTHOG_TRACES_SAMPLE_RATE` is left unset repo-wide and only MCP opts in at 0.25, because
tracing volume is what produced the PostHog bill this project already had to unwind. The alarm fires
6×/hour; spans would add recurring cost for a lane whose failures the `$exception` path already
covers. Revisit only if a flush failure ever needs per-step attribution.

## Acceptance criteria

- [ ] Every failure path above emits an `$exception` with a distinct `errorCode`.
- [ ] All captures share one `route`.
- [ ] A successful flush emits nothing.
- [ ] `alarm()` catches its own throw, captures it, and still re-arms.
- [ ] The capture seam is injectable so tests assert the codes without a live PostHog.

PR #10717, opened 2026-08-11
revert(neon): disable the write buffer -- it stalled and stopped draining

JSONbored:
## Incident

Buffered lanes stopped landing in Neon after **08:47**, and `neon:buffer-flush` last fired at
**08:59:52**. At 09:47 that was ~47 minutes of `tao_usd_index` minute-ticks, plus `account_balances`
and `nominator_positions`, sitting undrained.

**The block explorer was never affected.** `blocks-head` and `chain-detail` are in
`NEVER_BUFFER_LANES`, wrote throughout, and `blocks_head` stayed current — the exclusion did its job.

## Rollback

Emptying `NEON_WRITE_BUFFER_LANES` makes `neonWriteRunner` fall through to `createPgSql`, restoring
direct writes on every lane immediately. The buffer code stays in place and inert — **the flag is the
whole rollback, which is why it was built as a flag.**

## What is known, and what is not

Known: the Durable Object was reset repeatedly by deploys — three in twenty minutes — and enqueues
failed against a resetting object. `Durable Object reset because its code was updated` surfaced as a
stale verdict on `neon:validator-nominator-counts` at 09:33:50, and the new PostHog capture (#10696)
paged on it correctly.

**Not** known: whether that alone left the alarm unarmed, or whether the re-arm change in #10693 did.
`enqueueStatement` only arms when `getAlarm() === null`, and #10693 removed the post-flush re-arm on
the reasoning that the next enqueue would arm it — if an enqueue fails, nothing arms it, and the
buffer has no self-healing path. That is the leading hypothesis and it is mine.

This PR does not claim to fix it. It restores the data path so diagnosis is not happening while chain
price data goes unwritten.

## Follow-up

Root cause and a re-enable plan go in a separate issue. The buffer should not go back on until it has
a path that recovers from a failed enqueue without a human.

### DIFF
--- a/wrangler.data.jsonc
+++ b/wrangler.data.jsonc
@@ -139,7 +139,7 @@
     //
     // Every lane below is held to `2 * HOUR` by src/table-freshness-watchdog.ts,
     // so a ten-minute flush is twelve ticks inside a bound that already exists.
-    "NEON_WRITE_BUFFER_LANES": "raw-capture-state,neurons,subnet-hyperparams,account-identity,nominator-positions,account-balances,hotkey-alpha,validator-nominator-counts,tao-usd-index",
+    "NEON_WRITE_BUFFER_LANES": "",
     "UNKEY_API_ID": "api_2Dv8cVjrTSiD",
     // Tier-source flags. These are READ by this Worker -- neuronsServedFromD1()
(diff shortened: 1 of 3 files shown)


JSONbored:
Not rolling back — fixing forward. Root cause identified: see the fix PR.

Issue #10760, opened 2026-08-11
graphql: the 42 differences between the generated schema and the published one that are not nullability


JSONbored:
`report:graphql-schema-diff` builds the schema from the declared sources and compares it to the hand-written `src/graphql-sdl.ts`. Before this issue it reported:

```
416 generated object types against the schema's 420; 301 reproduce field for field.
  285 field(s) the generator would TIGHTEN to non-null
    4 type(s) the schema has that it does not build
   38 other difference(s)
```

The 285 are one question — whether a producer can answer null — and they need production evidence, so they belong in their own issue. **The other 42 are not that question.** They are a mix of generator gaps, contract gaps, and places the hand-written SDL is simply wrong, and every one of them has a determinate right answer available from the code.

## What the 42 turned out to be

**Four types the generator could not reach, because a name belongs to a REFERENCE and the generator resolved it per COMPONENT.** `SubnetAlphaVolumeArtifact` is `SubnetVolume` from its own route and `ChainAlphaVolumeSubnet` from inside the chain rollup; `SubnetEconomics` is `OpportunityEntry` when a leaderboard ranks it. `publishedName()` answers per component, so whichever name won was the only one built and the other was registered but never referenced. The remaining two — `EmissionGateChange`, `SubnetTrajectoryDelta` — are the opposite shape of the same gap: REST genuinely serves an un-flattened union and a window-keyed record, so the Zod says `JSON`, and the concrete component GraphQL publishes had nothing pointing at it.

**Eight fields the generator did not build.** Not one cause:

- `SubnetUptime.observed_at` — **the contract under-declares a field production always serves.** `formatUptime` emits it on every card, both transports feed it, REST reads it back into the response meta, and `UptimeArtifactSchema` never declared it. `.passthrough()` is why nothing noticed.
- `AlphaUsdFieldSource.storage` — **the emitter dropped it silently.** A `z.null()` property was dropped and reported into `nullOnly`, which exactly one assertion read, naming exactly one entry. When `field_sources_usd.storage` became the second and third entries they left the schema with no gate and no failure — and the SDL had already drifted into publishing one of the two and not the other.
- `Changelog.coverage_delta` — a genuine GraphQL-only flatten (#9892); the artifact nests it under `summary`.
[...]

Issue #10215, opened 2026-08-08
graphql: never exercised against production — 194 Query fields, zero runtime verification


JSONbored:
MCP has `conformance:mcp` (221 tools called against production, responses validated against their own published schemas, scheduled daily). REST has `check-response-conformance.ts`. **GraphQL has neither.**

Everything asserted about it is static: the SDL is compared to `openapi.json` offline. Nothing has ever *executed* a query against production and checked the answer.

## What that misses, concretely

The static gate is blind to every class the MCP sweep was built to catch:

- **A field that resolves to null.** `DomainSummary.emission_concentration` was declared `Float` against a 12-key object, so GraphQL raised a coercion error and nulled it on all 14 domains — and the SDL's own comment defined that null as "the domain has no members." A confident, wrong answer. That was found by reading, not by running.
- **A field that answers empty.** #10200's three routes serve `surfaces: []` with no degraded marker; the GraphQL mirrors of those fields have never been checked at all.
- **A resolver that throws.** Nothing dispatches these 194 fields.
- **The tier cascade.** `tests/graphql-tier-cascade.test.ts` covers it hermetically; production is where a dead tier actually shows.

## Shape

Mirror `scripts/check-mcp-conformance.ts`, which already solves the hard parts (serial with spacing against a rate-limited endpoint, real arguments per field, a violation is an error rather than a flag):

1. build a query per Query field from its arguments, using the same path fixtures the parity gates use
2. execute against `https://api.metagraph.sh/api/v1/graphql`
3. fail on `errors[]`, on a null non-null field, and on a field whose value contradicts the REST route it says it mirrors
4. sweep the `fields`-capable fields under projection too, the way the MCP sweep does — that is how #10175 found four tools whose projected rows failed their own schema

Scheduled out of band like its MCP sibling, not in CI: it needs production data.

## Done when

Every one of the 194 Query fields has been executed against production and its answer checked against the route it mirrors, on a schedule.

Part of #10213.

PR #10770, opened 2026-08-11
feat(graphql): join the nullability evidence to the tightenings it licenses

### LINKED ISSUE #10769: graphql: the 300 nullability tightenings need evidence per field, not one aggregate
With #10760 closed, the generated schema reproduces the published one field for field — 420 types against 420, **0 other differences**. One class is left:

```
300 field(s) the generator would TIGHTEN to non-null
```

## The count is not the decision

Each of the 300 is a claim about a **producer** ("it cannot answer null"). What the generator knows is a claim about a **schema** ("the Zod has no `.nullable()`"). Those are different claims, and where they disagree the served schema has been the one that was right: graphql-js enforces non-null at execution, so a single null nulls the whole surrounding object and attaches an error — which is what `SelfHealthLane.detail` did on every `self_health` request until #10215.

`conformance:graphql-nullability` already asks production the producer question, over both transports. What is missing is the **join**. Its answer covers every non-null leaf the generator promises, and only *some* of those are ones the cutover would change — so "3176 of 3403 observed, 0 answered NULL" reads as though it licenses all 300, when the 300 are a different set and the overlap is the entire question.

## What the join finds

Once the two reports are intersected, the 300 split three ways — and the interesting buckets are not the big one:

- **5 answered NULL.** All on `OpportunityEntry`: `max_uids`, `max_validators`, `miner_count`, `registration_allowed`, `validator_count`. `formatLeaderboards` ranks each board off the fields that board sorts by and leaves the rest unset, and the boards disagree with each other about which — `open_slots[]` answers null for `max_validators`/`miner_count`/`validator_count`; `validator_headroom[]` fills two of those and still answers null for `miner_count`. Verified on every row of both against api.metagraph.sh.

  This **cannot** be fixed in the Zod: `/api/v1/economics` fills all five on all 129 rows, so relaxing `SubnetEconomics` would weaken a contract the full card honours in order to describe a view that does not. The relaxation belongs to the projection.

- **1 that is a real contract bug**, and one #8055 already found once. `AccountIdentityHistoryEntry.identity_hash` is `z.string()`, non-null — but `formatIdentityHistoryEntry` builds it as `row.identity_hash ?? null` (`src/account-identity-history.ts`). #8055 hit exactly this on the *subnet* sibling and corrected `SubnetIdentityHistoryEntry.identity_hash` to nullable, with the reasoning written into the file. The account route was never corrected, and nothing noticed because the key is always **present** — required and nullable are orthogonal, and only the first was true.

- **The rest**, which production has answered repeatedly with values and never once with null.

## A blind spot the join also exposes

Four `Adapter.*` fields came back "no evidence either way" — and they are not unanswerable, they are unasked. The probe holds one value per argument NAME, and `slug` is two vocabularies: `provider(slug:)` takes a provider slug, `adapter(slug:)` takes a subnet slug. So `adapter(slug: "apex")` resolved null, every `Adapter` field went unobserved, and the report printed that silence as an absence of evidence. That is the failure this script exists to prevent, one level up.

## Done when

- the join exists and runs as a script, so the evidence for a tightening and the tightening itself are in one place
- it reports **0 answered NULL**
- the 5 partial-row fields are declared nullable on the view, not on the component
- `AccountIdentityHistoryEntry.identity_hash` matches its producer
- no tightening is left unproved because the probe never asked

Part of #10214.

JSONbored:
With #10760 closed the generated schema reproduces the published one field for field — 420 types against 420, **0 other differences**. One class was left: **300 fields the generator would tighten to non-null.**

## The count was not the decision

Each of the 300 is a claim about a **producer** ("it cannot answer null"). What the generator knows is a claim about a **schema** ("the Zod has no `.nullable()`"). Where those disagree the served schema has been the one that was right — graphql-js enforces non-null at execution, so a single null nulls the whole surrounding object and attaches an error (#10215).

`conformance:graphql-nullability` already asks production the producer question. What was missing is the **join**: its answer covers every non-null leaf the generator promises, and only *some* of those are ones the cutover would change. So "3176 of 3403 observed, 0 answered NULL" read like a licence for all 300, when the 300 are a different set and the overlap is the whole question. `report:graphql-tightening-evidence` intersects them and exits 1 on the one bucket that is not a judgement call.

## What the join found

**5 answered NULL** — all on `OpportunityEntry`. `formatLeaderboards` ranks each board off the fields that board sorts by and leaves the rest unset, and the boards disagree about which:

| board | `max_validators` | `validator_count` | `miner_count` |
|---|---|---|---|
| `open_slots[]` | null | null | null |
| `validator_headroom[]` | set | set | **null** |

On every row of both, against api.metagraph.sh. This **cannot** be fixed in the Zod — `/api/v1/economics` fills all five on all 129 rows, so relaxing `SubnetEconomics` would weaken a contract the full card honours in order to describe a view that does not. `ProjectedType.nullable` puts the relaxation on the view, and only one way: a projection may publish a component's non-null field as nullable, never the reverse.

**1 real contract bug**, and one #8055 already found once. `AccountIdentityHistoryEntry.identity_hash` is declared non-null while `formatIdentityHistoryEntry` builds it as `row.identity_hash ?? null`. #8055 hit exactly this on the *subnet* sibling, corrected it, and wrote the reasoning into the file — the account route was never corrected. Nothing noticed because the key is always **present**: required and nullable are orthogonal, and only the first was true. This fixes REST, MCP and GraphQL at once.

## A blind spot one level up

Four `Adapter.*` fields came back "no evidence either way" — not unanswerable, **unasked**. The probe holds one value per argument NAME, and `slug` is two vocabularies: `provider(slug:)` takes a provider slug, `adapter(slug:)` takes a subnet slug. So `adapter(slug: "apex")` resolved null, every `Adapter` field went unobserved, and the report printed that silence as an absence of evidence — which is precisely the failure the script exists to prevent. `ARGUMENTS_BY_FIELD` gives the field its own namespace; verified against production, `adapter(slug: "sn-1")` returns netuid 1.

## Where it lands

```
294 tightening(s); 288 proved, 0 answered NULL, 6 unproved
```

Of the six the probe never saw a value for: four are the `Adapter.*` fields the override now reaches, and two — `AccountCounterpartyTransfer.from`/`.to` — are non-null by **control flow** rather than by sampling. A row only reaches `transfers.push` when `from`/`to` equal a validated ss58 string (`src/counterparties.ts`), which null cannot.

Verified: `report:graphql-schema-diff` 0 other differences · `report:graphql-sdl-equivalence` 418/418 types, 198/198 return types, 0 JSON under-typings · all six graphql validators OK · `validate:contract-drift` OK · `npm run validate` exit 0 · lint/format/tsc clean · 390 tests green across the four graphql suites · `schemas-src/graphql/build-schema.ts` at **100% statements and 100% branches** (the enum-inside-a-list arm that `codecov/patch` flagged on #10761 is now covered by a test that drives it).

Closes #10769

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,6 +111,7 @@
     "validate:graphql-query-arguments": "node scripts/validate-graphql-query-arguments.ts",
     "report:graphql-sdl-equivalence": "node scripts/report-graphql-sdl-equivalence.ts",
     "report:graphql-schema-diff": "node scripts/report-graphql-schema-diff.ts",
+    "report:graphql-tightening-evidence": "node scripts/report-graphql-tightening-evidence.ts",
     "validate:published-names": "node scripts/validate-published-names.ts",
     "validate:graphql-tier-parity": "node scripts/validate-graphql-tier-parity.ts",
     "build:openapi-zod": "node scripts/generate-openapi-zod-components.ts",
--- a/packages/contract/index.d.ts
+++ b/packages/contract/index.d.ts
@@ -5312,8 +5312,8 @@ export interface components {
                 description?: string | null;
                 discord?: string | null;
                 github?: string | null;
-                /** @description Stable hash of this entry's tracked identity fields -- unchanged across entries where nothing actually differs. */
-                identity_hash: string;
+                /** @description Stable hash of this entry's tracked identity fields -- unchanged across entries where nothing actually differs. Null on a row with no computed hash. */
+                identity_hash: string | null;
                 image?: string | null;
                 name?: string | null;
                 observed_at: string | null;
--- a/public/metagraph/openapi.json
+++ b/public/metagraph/openapi.json
@@ -14741,8 +14741,15 @@
                   ]
                 },
                 "identity_hash": {
-                  "description": "Stable hash of this entry's tracked identity fields -- unchanged across entries where nothing actually differs.",
-                  "type": "string"
+                  "anyOf": [
+                    {
+                      "type": "string"
+                    },
+                    {
+                      "type": "null"
+                    }
+                  ],
+                  "description": "Stable hash of this entry's tracked identity fields -- unchanged across entries where nothing actually differs. Null on a row with no computed hash."
                 },
                 "image": {
                   "anyOf": [
--- a/public/metagraph/types.d.ts
+++ b/public/metagraph/types.d.ts
@@ -5312,8 +5312,8 @@ export interface components {
                 description?: string | null;
                 discord?: string | null;
                 github?: string | null;
-                /** @description Stable hash of this entry's tracked identity fields -- unchanged across entries where nothing actually differs. */
-                identity_hash: string;
+                /** @description Stable hash of this entry's tracked identity fields -- unchanged across entries where nothing actually differs. Null on a row with no computed hash. */
+                identity_hash: string | null;
                 image?: string | null;
                 name?: string | null;
                 observed_at: string | null;
--- a/schemas-src/graphql/build-schema.ts
+++ b/schemas-src/graphql/build-schema.ts
@@ -225,13 +225,22 @@ export function buildGeneratedSchema(
           // nulled parent. So the union publishes it, nullable -- which is
           // exactly what the hand-written SDL already did by hand for
           // `featured`, `uid_count` and `stake_dominance`.
+          // Two ways a promise the component makes cannot be kept here. A
+          // field only SOME of the producers carry: ask the other one and it
+          // answers null, which graphql-js turns into a nulled parent -- what
+          // the SDL already did by hand for `featured`, `uid_count` and
+          // `stake_dominance`. And a field a PROJECTION declares nullable,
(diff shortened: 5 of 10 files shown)


JSONbored:
Post-merge-check confirmation of the number the PR body reports conservatively.

The body says `288 proved, 0 answered NULL, 6 unproved`, noting that four of the six were the `Adapter.*` fields `ARGUMENTS_BY_FIELD` now reaches. That run predated the override. Re-run with it in place:

```
tightening-evidence: 294 tightening(s); 292 proved, 0 answered NULL, 2 unproved.

UNPROVED -- no evidence either way:
  - AccountCounterpartyTransfer.from
  - AccountCounterpartyTransfer.to
```

The override moved exactly the four it was expected to and nothing else. The two left are the ones proved by control flow rather than by sampling — a row only reaches `transfers.push` when `from`/`to` equal a validated ss58 string (`src/counterparties.ts`), which null cannot.

Issue #10710, opened 2026-08-11
fix(identity): subnet_identity_history has no writer and no Neon table — restore it as a Neon lane


JSONbored:
**`subnet_identity_history` has no writer and no Neon table. It exists only as a frozen lakehouse export, and it is the root cause of #10700 and #10706.**

This is the same shape as #9779 (the registry cluster whose only sync path was a retired GitHub Actions lane) — a table whose producer was removed, discovered from the consumers rather than from the producer.

## The evidence

**Nothing writes it.** `grep -rn 'INTO subnet_identity_history' src workers scripts` returns nothing.

**D1 was the primary writer.** `src/subnet-identity-history.ts`'s own docstring, still describing the pre-cutover world:

> #4832 gap-closure: mirror recordSubnetIdentityChanges' **D1 write** into Postgres via the DATA_API service binding … Best-effort: never throws, and a failure here must never block **the D1 write above (the primary contract)**

So the design was D1-primary with a best-effort Postgres mirror. D1 was eliminated 2026-07-17, and the mirror went through `METAGRAPH_SUBNET_IDENTITY_SOURCE` — retired in every config and absent from `DATA_API_FORWARD_FLAGS`. Both legs are gone.

**There is no Neon table.** No migration under `migrations/neon/`, and no entry in `generated/db/types.ts`. It appears only in `generated/lakehouse/types.ts:399` as `chain.subnet_identity_history`.

**The contract already says so**, in `src/contracts.ts`:

> served from the frozen lakehouse export of subnet_identity_history … **The table is append-only and stops at the export — no D1 table backs it and identity changes since are recorded nowhere** — so read observed_at on each entry as how current the answer is.

That sentence is accurate and easy to read past. "Recorded nowhere" is the whole bug.

## What is broken downstream

| consumer | symptom | issue |
| --- | --- | --- |
| `recordSubnetIdentityChanges` | diffs against an always-empty baseline, so reports ~129 changes every tick, and writes nothing | #10700 |
| `previously_known_as` on 3 routes | published, `.optional()`, always absent | #10706 |
| `/api/v1/subnets/{netuid}/identity-history` | serves the frozen export; no change since it was taken | — |
| `/api/v1/chain/identity-history` | same | — |

[...]

PR #10771, opened 2026-08-11
fix(identity): point the subnet-identity read tier at Neon

JSONbored:
The last switch on #10710, and the one that makes the rest visible.

#10740 built the writer, and the poller lane's first pass wrote **124 rows across 129 netuids** — verified in production. The reads did not move: `/api/v1/chain/identity-history` still served `2026-07-31`.

`METAGRAPH_SUBNET_IDENTITY_SOURCE` was `"retired"`, and `tryPostgresTier` forwards only on `"postgres"` — so both serving routes kept falling through to the frozen lakehouse export while the new table filled up beside them.

**The flag's history is this whole bug in miniature.** Flipped to `"postgres"` in 2026-07 on a live parity pass; set to `"retired"` when D1 went away, on the assumption that `syncSubnetIdentityToPostgres` would take over. That function is named as "the sole writer now" in four places and was never written. So the table had **no writer and no reader at once**, and nothing said so — a Postgres miss degrades to a schema-stable empty feed rather than a 404.

Without this, #10740 would have been filling a table nothing selects from: the same failure wearing the other shoe.

Types regenerated, parity gates green.


JSONbored:
**This flag flip cannot reach the reads, and I think it is a no-op. Two independent reasons, both checkable on this commit.**

### 1. No identity handler reads the flag any more

`#10190` did not just observe that the tier was unreachable — it **removed the `tryPostgresTier` call**. All five sites now pass `null` and carry the same comment:

```
workers/request-handlers/entities.ts:1743   // NO TIER READ (#10190). METAGRAPH_SUBNET_IDENTITY_SOURCE reads "retired" …
workers/request-handlers/entities.ts:2022   // NO TIER READ (#10190). …
src/graphql.ts:3024                          // NO TIER READ (#10190). …
src/graphql.ts:3068                          // NO TIER READ (#10190). …
src/mcp-server.ts:3509                       // NO TIER READ (#10190). …
```

`grep -c "METAGRAPH_SUBNET_IDENTITY_SOURCE"` finds it in comments on those paths and in no live expression. Setting it to `"postgres"` changes the value of something nothing evaluates.

### 2. DATA_API has no route to answer even if it did

```
$ grep -c "FROM subnet_identity_history" workers/data-api.ts
0
$ grep -oE '"/api/v1/[a-z0-9/{}_-]*identity[a-z0-9/{}_-]*"' workers/data-api.ts | sort -u
"/api/v1/internal/account-identity-sync"
"/api/v1/internal/subnet-identity-sync"
```

Both are the **write** endpoints the poller POSTs to. There is no read route, so a forward would fall to the gone-tier 503, `tryPostgresTier` would degrade to the schema-stable empty response, and the handler would land on the frozen export exactly as it does today — one wasted subrequest later.

### It is also worth undoing rather than leaving

Since #10767 a 5xx from the binding is retried once before degrading. If a `tryPostgresTier` call is ever restored on these paths while the flag reads `"postgres"`, every identity request pays **two** DATA_API subrequests to reach the same frozen answer.

The flag is also only flipped in `wrangler.jsonc`, not `wrangler.data.jsonc` — moot given the above, but it would matter if the forward were real.

### The reads still need repointing

Verified against production after this merged (it had not deployed yet, so this is the code argument, not a live one) — the feed is still pinned:

```
GET /api/v1/chain/identity-history  →  newest change 2026-07-31T11:00:20.942Z
```

while `lane neon:subnet-identity` reports `248 row(s) in 2 statement(s)`.

#10774 does the repoint in code: a live leg in `src/identity-history-answer.ts`, so all five call sites get it and none can diverge — cascade `tier → live → frozen → empty`, with a live miss falling through so the frozen export still covers everything before the writer existed. Filed as #10773.

Issue #10817, opened 2026-08-11
watchdog: TABLE_FRESHNESS names 49 of 60 Neon tables, and absence reads as healthy


JSONbored:
Part of #10806.

`TABLE_FRESHNESS` (`src/table-freshness-watchdog.ts:127-533`) is the estate's per-table staleness
map, swept hourly at `46 * * * *`. It names **49** tables. Neon has **60**.

Ten are absent entirely — neither watched nor explicitly exempted:

| table | rows | last write (2026-08-11) | producer | why it should be watched |
|---|---|---|---|---|
| `subnet_deregistration_daily` | 258 | 05:17 | cron `17 5 * * *` | daily lane, and also has no reader (#10818) |
| `subnet_identity` | 124 | 16:18 | poller subnet-identity | landed 2026-08-11 in migration 0023 |
| `subnet_lifecycle` | 169 | 2026-08-10 17:51 | subnet-lifecycle lane | 22 h old at audit time |
| `self_health_checks` | 2,622 | 16:28 | registry-sync-api cron | our own uptime record |
| `self_health_daily` | 12 | — | registry-sync-api cron | rollup of the above |
| `revenue_observations` | 91 | 16:27 | queue `revenue-probes` | |
| `revenue_probe_failures` | 66 | 16:27 | queue `revenue-probes` | |
| `attribution_sweeps` | 128 | 16:27 | queue `attribution-sweeps` | |
| `attribution_candidates` | 0 | — | queue `attribution-sweeps` | **empty**, and unread (#10818) |
| `origin_reachability` | 326 | 16:28 | queue `origin-reachability` | |

Several have their own dedicated watchdogs, so this is not "ten unwatched lanes" — but
`TABLE_FRESHNESS` is the one place that reads as a census of Neon, and it is not one. Being absent
from it is indistinguishable from being healthy.

The existing exemptions are the model to follow: `null` with a stated reason
(`coveredBy: "registry-sync"`, "a watch list, edited by hand", "append-on-change").

## Acceptance

- [ ] Every table in `db/schema.sql` appears in `TABLE_FRESHNESS`, with a threshold or a declared exemption
- [ ] A new Neon table that is added without an entry fails a gate
- [ ] The gate is proven able to fail

Issue #10812, opened 2026-08-11
api(ownership-history): every subnet returns a confident zero over 135 rows we hold


JSONbored:
Part of #10806. The store half is #10811; this is the contract half, and it is worth fixing
**independently** — the surface should never have answered `[]` without saying why.

## Measured 2026-08-11

Every subnet returns `data: []` with `meta.degraded` **absent**. A caller cannot distinguish:

- "this subnet has never changed hands" (true for most subnets), from
- "the only source is frozen at 2026-08-01 and the Neon table does not exist" (true for all of them)

`loadSubnetOwnershipHistoryColdTier` returns `null` when the lakehouse cannot answer, and the caller
turns that into an empty list rather than a declined one.

The cold-tier module already knows the honest ceiling — it computes `observed_through` precisely so
that *"a caller is never told 'no transfers' when the truth is 'not watched since'"* — and the route
drops it.

## Scope

Where the tier cannot answer, declare it: `degraded` with a reason, and publish `observed_through`
so an empty list is bounded by a date rather than implying "never".

This is the same rule the estate already applies elsewhere — an empty answer is only correct when the
filter is honest.

## Acceptance

- [ ] A tier that cannot answer produces `degraded`, never a bare `[]`
- [ ] `observed_through` is published on the payload
- [ ] A test pins the difference between "no changes in range" and "source unavailable"
- [ ] Applies to the GraphQL mirror and the MCP tool, not just REST

JSONbored:
Closing — **the premise is wrong, and it was my measurement error.** The surface is behaving
correctly.

I probed `/api/v1/subnets/{netuid}/ownership-history` reading the wrong field, concluded "empty with
no degraded flag", and filed on that. The actual payload for netuid 18, one of the subnets that did
change hands:

```json
{ "count": 1,
  "ownership_changes": [{ "netuid": 18,
      "old_coldkey": "5DHwWLjtpwnZQUQKKXE2N5Gdy2N8PpqhgjLUuzgSB7yuGZkF",
      "new_coldkey": "5GgvCi6h7dNsC489T8UnUMv912SoEXpEUDVt71VJU1Td7WKh",
      "block_number": 8724813, "observed_at": "2026-07-29T03:01:36.000Z",
      "source": "chain-event" }],
  "observed_through": "2026-07-29T03:05:51.987Z",
  "meta": { "source": "lakehouse-cold-tier" } }
```

Three things I got wrong:

[...]

PR #10821, opened 2026-08-11
feat(neon): give the subnet-ownership lane the two tables it writes to

### LINKED ISSUE #10811: data(neon): subnet_ownership and subnet_ownership_history do not exist, so the lane has nowhere to write
Part of #10806 (2026-08-11 audit). The worst finding: a surface that answers confidently and wrongly
over data we already hold.

## Measured 2026-08-11

`/api/v1/subnets/{netuid}/ownership-history` returns `data: []` for **every** subnet, with **no
`degraded` flag**:

| netuid | rows returned | `meta.degraded` |
|---|---|---|
| 1 | 0 | null |
| 2 | 0 | null |
| 3 | 0 | null |
| 64 | 0 | null |

Meanwhile `chain.subnet_ownership_history` holds **135 rows covering 129 subnets**, and
`chain.subnet_ownership` holds 128.

## Three breaks, stacked

1. **The Neon tables do not exist.** `to_regclass('public.subnet_ownership')` → `null`;
   `to_regclass('public.subnet_ownership_history')` → `null`. No table anywhere matches `%ownership%`.
   The infra lane writes them with a direct `INSERT INTO subnet_ownership`, so it has had nowhere to
   land since the move off the old box.
2. **The lakehouse copies are frozen** at 2026-08-01 / 2026-08-02.
   `src/subnet-ownership-cold-tier.ts` says so in its own header: *"FROZEN, LIKE EVERYTHING ELSE THE
   BOX WROTE."*
3. **The lane was unscheduled entirely** until metagraphed-infra#450, and had no Workflow instance
   until one was created by hand on 2026-08-11 16:45 UTC (metagraphed-infra#454).

## Scope (this issue = the Neon half)

Create both tables so the lane has a destination, and repoint the read path at Neon with the frozen
lakehouse as the cold fallback it already is.

Per the existing checklist, declaring a Neon table touches six places — migration, `db/schema.sql`,
generated db types, `NEON_SOLE_STORE_TABLES` in all three wrangler configs, `TABLE_FRESHNESS`, and
the read-store table group.

Columns should match what the lakehouse already carries, so the seed is loadable as history:
`subnet_ownership(netuid, owner_hotkey, owner_coldkey, captured_at)` and
`subnet_ownership_history(id, netuid, owner_hotkey, owner_coldkey, captured_at)`.

The serving contract is #10812; this issue is the store.

## Acceptance

- [ ] Both tables exist with a migration, in `db/schema.sql`, and in generated types
- [ ] Both are in `NEON_SOLE_STORE_TABLES` in **all three** wrangler configs (see #10814)
- [ ] Both are in `TABLE_FRESHNESS` with a threshold sized to the lane's 300 s cadence
- [ ] The 135 lakehouse rows are loaded as history, once, idempotently
- [ ] `/subnets/{netuid}/ownership-history` returns rows for a subnet that has changed hands

JSONbored:
Closes #10811

## What was wrong

The `subnet-ownership` lane has run in metagraphed-infra since 2026-08-03 and writes Postgres **directly** (`DATABASE_URL`, not a sync route), so unlike every other poller lane there is no handler in this repo that would have failed loudly when its destination went missing.

Measured 2026-08-11:

| check | result |
|---|---|
| `to_regclass('public.subnet_ownership')` | `null` |
| `to_regclass('public.subnet_ownership_history')` | `null` |
| tables matching `%ownership%` in `information_schema` | none |
| `chain.subnet_ownership_history` (lakehouse) | **135 rows across 129 subnets** |
| `/api/v1/subnets/{netuid}/ownership-history` | `data: []` for netuid 1, 2, 3, 64 — with **no `degraded` flag** |

They existed on the old indexer box, whose Postgres seeded the lakehouse on 2026-08-02. The move to Neon never brought them.

## What this does

`migrations/neon/0024_subnet_ownership.sql` creates both, with columns taken from the lane's own INSERTs rather than redesigned — so the first pass after its container image ships doesn't need a migration chasing it:

- `netuid` is the card's PRIMARY KEY, because the upsert's conflict target is `ON CONFLICT (netuid)`
- the history takes no `id` from the producer, hence `BIGSERIAL`, matching 0021
- both key columns are `NOT NULL`, because the Rust side binds `String`, not `Option<String>` — a netuid whose owner resolves to the zero account is pruned rather than written null

Applied to Neon, then `db/schema.sql` + `generated/db/{schema.json,types.ts}` regenerated from the live database (62 tables, 546 columns). Registered in `NEON_SOLE_STORE_TABLES` across **all three** wrangler configs, verified byte-identical afterwards.

## Seeded, and idempotent

135 history rows and 128 card rows loaded from the lakehouse, guarded by `NOT EXISTS` / `ON CONFLICT DO NOTHING`. Re-running the whole seed leaves the count at 135 — verified, not assumed.

Seven subnets carry two observations — **14, 18, 27, 53, 86, 102, 104** — genuine owner changes the history now records. (The epic's 2026-08-03 measurement said three; it has grown since.)

## What is deliberately not here

**No `TABLE_FRESHNESS` entry.** I wrote one, then removed it: that map's own header says it covers "every table in `tests/fixtures/sqlite-schema`", and its coverage test derives the table set from that directory — the frozen **D1-era** migration set, 49 tables, exactly the 49 the map names. No Neon-era table added since is classifiable without writing SQLite DDL for a Postgres-only table, which would entrench the mechanism rather than fix it.

That mechanism is the actual cause of #10817 ("names 49 of 60"), and `subnet_ownership` should be classified there alongside the other eleven it currently hides.

What *is* here on that axis: `subnet_ownership` in `PRODUCER_CADENCE_SECS` at its 300s interval, and `"subnet-ownership"` in `LANE_PRODUCER`, which gives `lane-alarm` a silence bound for the lane immediately.

## Still blocked downstream

The lane cannot write until its container image ships #450's lane list (metagraphed-infra#461) — both bootstrapped Workflow chains currently fail `404 no such lane`. Until then the card carries the 2026-08-02 seed's `captured_at` and is honestly stale. Serving the rows, and declining honestly when the tier cannot answer, is #10812.

## Verification

- `npx tsc --noEmit` clean
- `validate:db-types-drift` — current at 62 tables / 546 columns
- `validate:schemas`, `validate:schema-enums`, `validate:single-schema-source` pass
- `tests/{table-freshness-watchdog,table-freshness-cadence,producer-cadence,neon-sole-store,neon-sole-store-tables-exist}.test.ts` — 86 passed


JSONbored:
Correction to this PR's description, measured after it merged.

I wrote that **seven** subnets carry a genuine owner change — 14, 18, 27, 53, 86, 102, 104. Seven carry two *observations*; only **three** changed hands. The other four are **hotkey rotations under an unchanged coldkey**, and the contract is explicitly coldkey-to-coldkey — `subnet-ownership-cold-tier.ts` states it: *"a hotkey rotation under an unchanged coldkey is not a change of ownership."*

Distinct coldkeys per netuid in the seeded history:

| netuid | coldkeys | genuine transfer |
|---|---|---|
| 18, 86, 102 | 2 | **yes** |
| 14, 27, 53, 104 | 1 | no — hotkey rotation only |

Nothing in the diff changes: the seed is the producer's own rows and the count is right. Only my characterisation of them was.

I also filed #10812 off a bad probe of this endpoint and have closed it — the surface was answering correctly the whole time, including `observed_through` as the bound on an empty list.

Issue #10814, opened 2026-08-11
crons(data-api): three declared crons have no handler, and no test can catch that direction


JSONbored:
Part of #10806.

## Three crons fire into nothing, ~23 wasted invocations/hour

`wrangler.data.jsonc` declares six crons on `metagraphed-data-api`. Three have no branch in
`workers/data-api.ts:8335-8371`:

| cron | line | was | producer |
|---|---|---|---|
| `*/3 * * * *` | 476 | `NEON_BACKFILL_CRON` | `src/neon-backfill.ts` **deleted** |
| `26 * * * *` | 477 | mirror-lag watchdog | `src/neon-mirror-lag.ts` **deleted** |
| `38 * * * *` | 478 | D1↔Neon row-count parity | `src/neon-parity.ts` **deleted** |

All three fall through to `{ ok: false, skipped: true, reason: "unknown cron" }` at
`workers/data-api.ts:8367`. The constants they referenced no longer exist anywhere; the lanes are
already in `RETIRED_LANES` / `RETIRED_LANE_PREFIXES` (`src/lane-health.ts:84,102`).

The live `26 * * * *` handler is on the **api** Worker, which is why this reads as harmless in a
grep.

## Why no test caught it

Every cron gate in `tests/` runs **constant → declared-in-wrangler**. The reverse — every declared
cron resolves to a handler — has never been asserted, in either direction, on either Worker.
`tests/neon-prune.test.ts:131` and `tests/table-freshness-watchdog.test.ts:928` are both the forward
direction only.

## Acceptance

- [ ] The three crons are removed from `wrangler.data.jsonc`
- [ ] A test asserts every cron in **every** wrangler config resolves to a dispatch branch
- [ ] The gate is proven able to fail by adding a bogus cron

PR #10827, opened 2026-08-11
fix(crons): make the api Worker's */15 branch explicit, so an unknown cron stops running a producer

### LINKED ISSUE #10815: crons(api): the */15 branch is the unconditional default, so an unknown cron runs the health prober
Part of #10806. Same class as #10814, on the other Worker — found while checking it.

`wrangler.jsonc:1028` declares `*/15 * * * *`, and `workers/api.ts:3287` runs the health prober as
the **unconditional default** of `dispatchScheduled` rather than as a branch matched on that string.

So on the `metagraphed` Worker, *any* cron string that does not match an earlier branch silently runs
`runHealthProber` — a lane that writes `surface_checks`, `surface_status`, `surface_uptime_daily`,
`subnet_snapshots` and `lane_health`, and POSTs `/poll-start` to `CHAIN_FIREHOSE_HUB`.

A typo'd or retired cron here does not go quiet like the three in #10814 — it runs a real producer at
the wrong cadence, which is worse and harder to see.

## Acceptance

- [ ] `*/15 * * * *` is matched explicitly
- [ ] An unrecognised cron returns a skipped/unknown result rather than running a producer
- [ ] Covered by the same reverse gate as #10814

JSONbored:
Closes #10815

## The mirror of #10814, and the worse half

`dispatchScheduled` matched 36 cron constants and then, for anything unmatched, fell through to the firehose-hub bootstrap and `runHealthProber`. `*/15 * * * *` was the only one of `wrangler.jsonc`'s **37** expressions with no constant and no branch — it **was** the fall-through.

| | data-api (#10814) | api Worker (this) |
|---|---|---|
| unmatched cron does | `{ skipped: true }` | **runs `runHealthProber`** |
| cost | ~23 no-op invocations/hour | a real producer on a cadence nobody chose |
| writes | nothing | `surface_checks`, `surface_status`, `surface_uptime_daily`, `subnet_snapshots`, `lane_health` — plus a `/poll-start` to `CHAIN_FIREHOSE_HUB` |

A wrong lane at a wrong cadence is worse than a no-op, and much harder to see: nothing looks broken.

## The fix

- **`HEALTH_PROBER_CRON`** names that expression, so it is a declared schedule like the other 36 rather than the absence of one.
- **`API_HANDLED_CRONS`** declares all 37, built from the constants themselves.
- The dispatcher **declines against that set before its branches** — so the set is load-bearing rather than documentation, and a cron missing from it is refused even where a branch would have matched.
- The bottom guard is kept as the *other* direction's net (an entry added to the set whose branch nobody wrote), with reason `"declared but unhandled"` so the two cases are distinguishable in a log.

## Tests

`tests/api-crons-have-handlers.test.ts` — 7 assertions, covering the bookkeeping *and* the behaviour:

- every declared cron resolves to a handled lane, and every handled cron is declared
- the prober's cron is a **named member**, not the default — the assertion that stops it becoming the fall-through again
- the set has no duplicates (two entries for one expression means two branches believe they own it)
- through the **real `scheduled` export**: an unknown cron and an empty cron are both declined, so the guard itself is what's tested rather than a reimplementation

Verified the config direction fails by declaring `"7 7 7 7 7"`:

```
× every DECLARED cron resolves to a handled lane
  + '7 7 7 7 7'
```

## Verification

- `npx tsc --noEmit` clean · `format:check` clean · `lint` clean
- **All 21 test suites that exercise `handleScheduled` pass unchanged** — the top guard does not reject anything the dispatcher used to accept
- `tests/api-crons-have-handlers.test.ts` — 7 passed

### DIFF
--- a/tests/api-coverage.test.ts
+++ b/tests/api-coverage.test.ts
@@ -3102,10 +3102,27 @@ describe("handleScheduled", () => {
     assert.ok(result === undefined || typeof result === "object");
   });
 
-  test("any other cron runs the health prober", async () => {
-    // No bindings → runHealthProber should not throw with an empty env.
-    const result = await handleScheduled(
+  test("an undeclared cron is DECLINED, and does not run the health prober", async () => {
+    // INVERTED BY #10815, and the old name said the bug out loud: "any other
+    // cron runs the health prober". It did -- dispatchScheduled ran the prober
+    // as its unconditional fall-through, so a typo'd or retired expression
+    // wrote surface_checks, surface_status, surface_uptime_daily,
+    // subnet_snapshots and lane_health on whatever cadence it happened to have.
+    const result = (await handleScheduled(
       { cron: "*/2 * * * *" } as unknown as ScheduledController,
+      {} as unknown as Env,
+      { waitUntil() {} } as unknown as ExecutionContext,
+    )) as Record<string, unknown>;
+    assert.equal(result.skipped, true);
+    assert.equal(result.reason, "unknown cron");
+  });
+
+  test("the health prober's own cron still reaches it", async () => {
+    // The other half: declining unknowns must not decline the real one.
+    const result = await handleScheduled(
+      {
+        cron: workerConfig.HEALTH_PROBER_CRON,
+      } as unknown as ScheduledController,
       {} as unknown as Env,
       { waitUntil() {} } as unknown as ExecutionContext,
     );
--- a/tests/api-crons-have-handlers.test.ts
+++ b/tests/api-crons-have-handlers.test.ts
@@ -1,0 +1,117 @@
+// Both directions between wrangler.jsonc's 37 crons and the api Worker that
+// dispatches them (#10815) -- the twin of tests/data-api-crons-have-handlers,
+// and the worse of the two failures it guards.
+//
+// On data-api an unrecognised cron fell through to `{ skipped: true }`: a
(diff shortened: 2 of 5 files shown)


JSONbored:
Restructured after `codecov/patch`, which was right about the first cut.

That version had `dispatchScheduled` decline against `API_HANDLED_CRONS` **before** its branches — which made the guard at the bottom **unreachable**, since nothing could get past the first check without also matching a branch. An unreachable guard is a branch no test can cover and no reader can trust, and reaching for a coverage-ignore would only have hidden that.

The dispatcher now reads plainly:

- `HEALTH_PROBER_CRON` is an `if` like the other 36, guarding the firehose bootstrap and the prober together (the bootstrap is piggybacked on that tick by design)
- the function ends with one `return { skipped: true, reason: "unknown cron" }` that anything unmatched reaches — both an undeclared expression and a declared-but-unbranched one, which are the same fault from a caller's side

`API_HANDLED_CRONS` is now read by the **test** rather than the dispatcher, and `tests/api-crons-have-handlers.test.ts` ties them together: every entry must have an `if (cron === ...)` branch. Verified it fails by breaking one branch and watching it name `ABUSE_SCAN_CRON`.

All 21 suites exercising `handleScheduled` pass — 1151 tests — plus the full 41-validator set and format.

Issue #10051, opened 2026-08-08
Delete the migration scaffolding — mirrors, reconciler and parity sweep are transitional, and the reconciler is dangerous if it outlives the inversion


JSONbored:
The end state is **one store**, not two kept in step. Every mechanism below exists only to carry tables across, and each becomes dead weight the moment the last table lands. Filing now so the migration has a defined finish rather than trailing off with the scaffolding still deployed.

## What has to be deleted, and when it becomes safe

| scaffolding | safe to delete once |
|---|---|
| `NEON_DUAL_WRITE_LANES` + every `mirror*ToNeon` call | every table is in `NEON_SOLE_STORE_TABLES` |
| `NEON_BACKFILL_LANES`, `NEON_BACKFILL_PLANS`, `src/neon-backfill.ts` | nothing writes D1, so there is nothing to reconcile from |
| `neon-parity` + `PARITY_TABLES` | same — a sweep comparing one store to an empty one reports the whole estate as a gap |
| `neon:` mirror lane verdicts | the writes they describe are no longer mirrors |
| `NEON_READ_LANES` / `NEON_READ_ROUTE_TABLES` / `routeStore` | there is no second runner to choose between |
| `NEON_SOLE_STORE_TABLES` itself | it answers "is D1 behind this?" and D1 is gone |
| `createD1Sql`, `METAGRAPH_HEALTH_DB` binding, `src/neon-prune.ts`'s D1 half | last |

## The order is forced, and getting it wrong is silent

`neon-parity` and the reconciler must go **after** the write inversion, not before: while D1 is still written they are the only things that would catch a Neon write path that quietly drops rows. Delete them early and the estate loses its evidence exactly when it needs it most.

Equally, they must not survive the inversion. A reconciler pointed at a D1 that nothing writes will copy a frozen snapshot over live Neon rows — it does not know the direction has reversed. **That is a data-loss bug, not stale scaffolding**, and it is the single most dangerous artifact in this list.

## `lane_health` is the deliberate exception, until it isn't

It moves last and alone. Every watchdog writes to it, including the ones watching this migration, so until the estate is verified it is the only thing that can report a lane that broke while being moved. It is also the reason `d1_migrations` and the D1 binding cannot be removed before it.

## Acceptance

- `grep -r "NEON_DUAL_WRITE\|NEON_BACKFILL\|PARITY_TABLES\|createD1Sql" src/ workers/` returns nothing
[...]

Issue #10831, opened 2026-08-11
test(flake): the SafeMode wiring test proves the cron by opening a socket, and fails on loaded runners


JSONbored:
`tests/safe-mode-watchdog.test.ts` → *"the cron is wired: the schedule reaches the watchdog"* fails
intermittently in CI and passes locally. It has cost two unrelated PRs a re-run today (#10826, then
#10827), on branches that touch neither the watchdog nor its cron.

## Why it flakes

It proves the wiring by making a **real TCP connection** and asserting the failure it produces:

```ts
const result = await worker.scheduled(
  { cron: SAFE_MODE_WATCHDOG_CRON, scheduledTime: Date.now() } as never,
  { SAFE_MODE_RPC_URL: "http://127.0.0.1:1/unreachable" } as never,
  { waitUntil: () => {} } as never,
);
assert.equal(result.reason, "unreachable");
```

So the assertion is not "the cron reached the watchdog" — it is "the cron reached the watchdog **and
the socket failed in the specific way that yields `reason: "unreachable"`**". On a loaded runner that
second clause can resolve differently (a timeout rather than a refusal), and the test reports the
first clause as broken. Locally it takes ~2.3s, which is most of the suite's runtime.

## It no longer needs the socket

The property it wants is now assertable statically. #10827 makes the api Worker's handled set data
(`API_HANDLED_CRONS`) and declines anything outside it, and
`tests/api-crons-have-handlers.test.ts` already asserts both directions between that set and
`wrangler.jsonc` — including, through the real `scheduled` export, that an unknown cron is declined
and a declared one is not.

So this test can assert membership rather than a socket outcome, and keep its intent without its
flakiness. Its sibling below it (*"the cron constant matches a wrangler schedule"*) is already
static and does not flake.

## Acceptance

- [ ] The wiring assertion does not depend on network behaviour
- [ ] The watchdog's `unreachable` path keeps a test — it is real behaviour, just not evidence of wiring
- [ ] Suite time for this file drops (the ~2.3s is the connect timeout)

PR #10839, opened 2026-08-11
feat(sync): route subnet-ownership through Hyperdrive, and stop the identity history rewriting its own timestamps

### LINKED ISSUE #10836: feat(sync): give subnet-ownership a sync route, so the last direct-Postgres lanes go through Hyperdrive
The last two lanes that write Postgres from inside the container should go through Hyperdrive like
the other eight. This is the route half; the producer half is metagraphed-infra#461.

## Why

The poller container **cannot use Hyperdrive**. It is a plain Linux process, not a Worker isolate,
and Hyperdrive's connection string is *"only accessible from your Worker"*. That is precisely why
eight of the eleven poller lanes POST to a sync route here and let `metagraphed-data-api` — which
holds the `HYPERDRIVE` binding — do the write:

| pattern | lanes |
|---|---|
| POST → data-api → **Hyperdrive** → Neon | `account_balances`, `hotkey_alpha`, `account_identity`, `chain_detail`, `metagraph`, `subnet_hyperparams`, `subnet_identity`, `validator_nominators` |
| direct `DATABASE_URL` from the container | `subnet_ownership`, `self_stake`, `self_health` |

The direct pattern is a leftover from the self-hosted box, where a raw URI made sense. On Cloudflare
it means shipping a Neon credential into a container image and bypassing the connection pool every
other lane uses — while Neon's compute is exactly what the pool exists to protect.

**`self_stake` needs nothing here.** It writes `nominator_positions`, and
`/api/v1/internal/nominator-positions-sync` already exists with `NOMINATOR_POSITIONS_SYNC_SECRET`
already provisioned on the poller. It is a producer-side change only.

**`subnet_ownership` needs this route.** Its two Neon tables landed in #10811.

## The shape is already in the codebase

`subnet_ownership` is a card plus an append-on-change history — structurally identical to
`subnet_identity`, which `FAMILY_MIRROR_PLANS` already models. The new plan entry is declarative:

```ts
export const SUBNET_OWNERSHIP_NEON_LANE = "subnet-ownership";

[SUBNET_OWNERSHIP_NEON_LANE]: {
  lane: SUBNET_OWNERSHIP_NEON_LANE,
  latest: {
    table: "subnet_ownership",
    columns: SUBNET_OWNERSHIP_INSERT_COLUMNS,     // netuid, owner_hotkey, owner_coldkey, captured_at
    conflict: ["netuid"],
  },
  history: {
    table: "subnet_ownership_history",
    columns: SUBNET_OWNERSHIP_HISTORY_COLUMNS,
    conflict: ["netuid", "owner_hotkey", "owner_coldkey"],
  },
},
```

### The history conflict key, and why not `observed_at`

The producer re-reads every subnet's owner each pass, so the same ownership arrives repeatedly at
different timestamps. Conflicting on the timestamp — what `subnet_hyperparams_history` and
`account_identity_history` do — would append a duplicate row every 300 s and bury the provenance
the table exists for. `subnet_identity_history` already solved this by conflicting on a content
hash instead; the owner pair **is** the content here, so it is the direct analogue and needs no hash.

**Verified against the seeded data**: 135 rows, **135 distinct `(netuid, owner_hotkey, owner_coldkey)`
triples**, and **no subnet has ever returned to a previous owner**. So the unique index holds today,
and the A→B→A collapse this key implies is not a case that has occurred. It is the same trade-off
`subnet_identity_history` already accepted, and worth stating rather than discovering later.

## Work

- [ ] `migrations/neon/0025_subnet_ownership_history_revision.sql` — `CREATE UNIQUE INDEX ... ON
      subnet_ownership_history (netuid, owner_hotkey, owner_coldkey)`. Required for `ON CONFLICT`;
      #10811 created the tables without it because nothing wrote them yet.
- [ ] `SUBNET_OWNERSHIP_INSERT_COLUMNS` / `_HISTORY_COLUMNS` + the `FAMILY_MIRROR_PLANS` entry
- [ ] A row schema alongside `subnetIdentitySyncRowSchema` — bounded ss58 strings, `netuid` ceiling,
      `SYNC_MIN_CAPTURED_AT_MS`
- [ ] `handleSubnetOwnershipSync` in `workers/data-api.ts`, mirroring `handleSubnetIdentitySync`:
      secret gate → body/row caps → `validateSyncRows` → `mirrorFamilyToNeon` → **authoritative**
      failure (Neon is the only store, so its failure is the request's)
- [ ] Route registration + `SUBNET_OWNERSHIP_SYNC_SECRET` on `wrangler.data.jsonc`
- [ ] **The prune.** The producer currently ends with
      `DELETE FROM subnet_ownership WHERE netuid <> ALL($1)`, so a deregistered subnet's card is
      removed. That has to move here or be dropped deliberately — a card table that only ever grows
      would keep dead netuids forever.
- [ ] Tests to 99 % patch coverage, branch-counted

Once this lands, metagraphed-infra#461 converts the producer to POST here, both lanes take an
`on_demand.register(...)` receiver, and `POSTGRES_JOBS` shrinks to `self-health` alone — removing
the last direct-Postgres writer from the container.

### LINKED ISSUE #10838: fix(neon): subnet_identity_history rewrites its own first-seen stamp on every pass, so the identity feed reports a mass rename hourly
`subnet_identity_history` is an append-on-CHANGE table whose timestamps track **last seen**, not
first seen. Every unchanged identity has its `observed_at` rewritten on every hourly pass, so
`/api/v1/chain/identity-history` publishes a network-wide identity change once an hour over
subnets that have not changed at all.

## The mechanism

The history conflicts on `(netuid, identity_hash)` — content, not timestamp — which is right, and
is what stops repetition appending a row. But `buildPgUpsert` emits a `DO UPDATE SET` for every
column that is not part of the conflict key, and `observed_at` is one of those:

```sql
INSERT INTO subnet_identity_history (netuid, block_number, observed_at, subnet_name, ...)
VALUES (...)
ON CONFLICT (netuid, identity_hash) DO UPDATE SET
  block_number = EXCLUDED.block_number,
  observed_at  = EXCLUDED.observed_at,   -- <-- rewrites first-seen to last-seen
  ...
```

So repetition writes no new ROW, and silently moves the existing one's stamp.

0021's own comment describes the intended contract accurately — *"a pass whose hash matches the
newest row for that netuid writes nothing, so a lane running hourly against an identity that never
changes adds no rows at all"* — and it is true about rows. It is silent about the timestamp.

## Measured on production, 2026-08-11

```
SELECT to_timestamp(observed_at/1000), count(*) FROM subnet_identity_history GROUP BY 1;
 2026-08-11 15:17:41 |   1
 2026-08-11 18:19:44 | 124
```

Seven passes have landed (`lane_health`, `neon:subnet-identity`, `verdict='ok'`), from 13:15:49
onward. If first-seen were preserved, the bulk of those 125 rows would sit at **13:15:49**. Instead
124 of 125 sit at the single newest pass. The one row at 15:17:41 is netuid 108's superseded
revision — it stopped being sent when that subnet's identity changed, so nothing has bumped it
since, which is the control case.

## Why it matters

`observed_at` is the column both serving routes page on and order by, so the corruption is not
cosmetic:

- `/api/v1/chain/identity-history` — the network-wide feed reports ~124 identity changes an hour
- `/api/v1/subnets/{netuid}/identity-history` — a subnet's trail says it was renamed an hour ago
- `previously_known_as` provenance, which is the reason the table exists

## The fix

Guard the history's `DO UPDATE` so a LATER sighting of a revision cannot move its stamp:

```
subnet_identity_history.observed_at > EXCLUDED.observed_at
```

`>` rather than `DO NOTHING` deliberately: an EARLIER observation (a backfill) is a *better*
first-seen and should still win. Only a later one is refused. The card keeps its existing
`captured_at < EXCLUDED.captured_at` — the two guards point in opposite directions, which is the
distinction between a latest-only card and a first-seen history.

The two timestamp-KEYED histories (`subnet_hyperparams_history`, `account_identity_history`) need
no guard: their conflict key contains `observed_at`, so a conflict there is the identical row
arriving twice and the update is a no-op.

## Repair, which must come AFTER the deploy

The next pass re-corrupts until the guard ships. Once it has:

```sql
-- netuid 108's newer revision genuinely first appeared at the pass after its
-- predecessor stopped being sent.
UPDATE subnet_identity_history SET observed_at = 1786465102177
 WHERE netuid = 108 AND observed_at = 1786472384283;

-- everything else was first observed on the first landed pass, 13:15:49.
UPDATE subnet_identity_history SET observed_at = 1786454149023
 WHERE observed_at IN (1786472384283, 1786461461498);
```

Fixed in the same PR as #10836, which builds `subnet_ownership_history` on the same machinery and
would have inherited the identical bug.

JSONbored:
The last poller lane writing Postgres from inside the container gets a sync route, so its writes
go through Hyperdrive like the other eight — and a live bug in the sibling table it is modelled on
gets fixed, because this PR would otherwise have inherited it.

Closes #10836
Closes #10838

## Why the container cannot just use Hyperdrive

It is a plain Linux process, not a Worker isolate, and the pooled connection string is *"only
accessible from your Worker"*. That is why eight of the eleven lanes already POST to a sync route
here and let `metagraphed-data-api` — which holds the `HYPERDRIVE` binding — do the write.
`subnet-ownership` kept a raw `DATABASE_URL` from the self-hosted box: a Neon credential in a
container image, and every write bypassing the pool Neon's compute is protected by.

## Two things moved out of Rust, because neither survives the trip

**The diff.** The producer SELECTed the whole card (`fetch_current_owners`), compared each resolved
owner (`owner_changed`), and appended to the history only on a change. That read-then-write is a
race between overlapping passes, so `0025`'s unique index on
`(netuid, owner_hotkey, owner_coldkey)` decides what is new instead — the same move
`subnet_identity_history` made with its content hash. Checked against the seeded rows before
choosing it: **135 rows, 135 distinct triples, and no subnet has ever returned to a previous
owner**, so the index is creatable today and the A→B→A collapse it implies is not a case that has
occurred.

**The prune.** The producer ended each pass with `DELETE FROM subnet_ownership WHERE netuid <> ALL($1)`.
Dropped, a latest-only card keeps deregistered netuids forever. It is `plan.prune` now: after the
upsert and only if the upsert landed, keyed on what was **posted** rather than on what was written,
and refusing an empty key set — `<> ALL('{}')` is true for every row, which against production
would have deleted all 128.

## The bug this found in `subnet_identity_history`

`buildPgUpsert` emits `DO UPDATE SET` for every column outside the conflict key. For a history
keyed on **content** rather than on its timestamp, that includes `observed_at` — so every hourly
re-send rewrote first-seen to last-seen:

```
SELECT to_timestamp(observed_at/1000), count(*) FROM subnet_identity_history GROUP BY 1;
 2026-08-11 15:17:41 |   1
 2026-08-11 18:19:44 | 124
```

Seven passes have landed since 13:15:49; 124 of 125 rows sit at the single newest one. An
append-on-CHANGE table was publishing a network-wide identity change every hour —
`/api/v1/chain/identity-history`, the per-subnet trail, and `previously_known_as` provenance all
read from that column.

Both content-keyed histories now carry a guard that keeps the **earlier** observation. `>` rather
than `DO NOTHING` on purpose: a backfill supplying an older sighting is a *better* first-seen and
still wins. The two timestamp-keyed histories need no guard and are untouched — asserted as a
derived rule over `FAMILY_MIRROR_PLANS`, so a fifth plan inherits it.

The existing `an unchanged identity appends nothing on a second pass` test could not catch this: it
counted rows, and the row was already there. It now asserts the stamp, and I confirmed it fails
without the guard.

## Note on `NEON_DUAL_WRITE_LANES`

The lane has to be added there or `mirrorFamilyToNeon` returns `{attempted: false}` and the route
502s. The name is a fossil — there is no second store, and `neonDualWriteEnabled` is now just a
per-lane on/off switch. Renaming belongs to #10051, not here.

## Verification

- 41 CI validators, derived from the workflow files; `typecheck` + `lint` + `format:check`
- full suite unsharded: **19,789 passed**, 11 skipped
- **patch coverage 53/53 changed lines = 100%**, no uncovered branches
- both new behaviours mutation-checked: removing the history guard fails 1 test, removing the prune
  fails 2
- the prune's cast and the empty-set hazard were both run against production Neon before shipping

## Sequencing

The producer half is metagraphed-infra's; it needs `SUBNET_OWNERSHIP_SYNC_SECRET` provisioned on
this Worker after this merges. Until then the lane simply stays disabled, which is the same state
it is in today.


JSONbored:
Added a third fix, because it was blocking this PR and is fully closeable on its own.

`test (shared 3/3)` failed on `tests/safe-mode-watchdog.test.ts > the cron is wired`. That is **#10831**, open, and already the cause of re-runs on #10826 and #10827 — branches touching neither the watchdog nor its cron. It passes 3/3 locally; the shard passes locally at 5,098 tests. The `coverage` job failed only because it could not assemble artifacts from the failed shard, so it is the same root cause.

Rather than re-run and hope, the test now asserts what it actually means: `SAFE_MODE_WATCHDOG_CRON ∈ API_HANDLED_CRONS`. It was opening a real TCP connection to `127.0.0.1:1` and requiring the failure to arrive as `reason: "unreachable"` — two claims in one assertion, and the socket half is what moves under load. #10827 made the handled set data and made anything outside it an explicit decline, so membership is now provable statically, and `api-crons-have-handlers` separately asserts every named cron has a dispatch branch.

The `unreachable` path keeps its own test, per that issue's acceptance — driven through an injected rejecting fetch, so it tests the watchdog's error handling rather than socket semantics. Both mutation-checked.

JSONbored:
Rebased onto main; three things changed and are worth stating.

**The SafeMode flake fix is gone from this branch, deliberately.** #10842 landed the same fix on main while this was in review, and #10831 is closed. I dropped my commit rather than carry a duplicate — the PR closes #10836 and #10838 only, and the body is updated to match.

**The migration is now `0026`, not `0025`.** #10837 landed `0025_nominator_positions_shares.sql` in the meantime; `validate:migrations` requires prefixes to be unique and gap-free, and caught the collision. Renumbered to `0026_subnet_ownership_history_revision.sql` with every reference to it — code comments, the test that applies the DDL verbatim — updated with it.

**Two conflicts in `src/neon-write.ts` and its test**, both "each side added a different thing in the same place": #10837's `normalizeShareFractionsInNeon` and this PR's `pruneCardOutsideKeySet`. Resolved by keeping both.

Re-verified after the rebase rather than assuming the diff survived it: `typecheck` · `build` · 41 CI validators · `lint` · `format:check` · 187 tests across the eight affected suites.

JSONbored:
Rebased again and re-verified. Two separate causes for the last red run, one of them a real gap in the PR.

**A missing proxy — my omission, now fixed.** The producer is a container, and a container only knows the public host: it POSTs to `api.metagraph.sh`, never to data-api directly. Every internal sync route therefore needs an explicit proxy branch on the main Worker, and I had added the data-api route without one. The pass would have 404'd at the edge — which is precisely the state `subnet-identity` was in for its entire life (#10710), a lane faithfully posting to a URL nothing served.

Added `handleSubnetOwnershipSyncProxy` plus its dispatch, a functional test mirroring the identity proxy's, and — because nothing structural would have caught this — `tests/internal-sync-routes-are-proxied.test.ts`, which asserts both directions between data-api's 11 internal `*-sync` routes and the main Worker's proxies for them. The reverse direction had to be narrowed to handlers that actually call `proxyToDataApi`: `emission-gate-sync` is handled on the main Worker and `registry-sync` forwards to a different Worker entirely, so both are legitimately not data-api routes. Mutation-checked — deleting the ownership dispatch fails it.

**The rest was staleness, not breakage.** `validate:nullable-overpromises` did not exist on my branch because #10841 added it to main after my base; the branch was simply behind. Re-derived the validator list from the workflows after rebasing — it is **42** steps now, not 41.

Green locally on the current main: typecheck · build (no artifact drift) · 42 validators · lint · format:check · 118 tests across the eight affected suites.

PR #10848, opened 2026-08-11
feat(sync): give nominator_positions a prune domain, so self-stake can share it

### LINKED ISSUE #10845: feat(sync): give nominator_positions a prune domain, so self-stake can share it without losing rows
`nominator_positions` is about to have a second writer, and the prune it already has would delete
that writer's rows. This is the route half; the producer half is metagraphed-infra#473.

## Why a second writer

`validator-nominators` scans `SubtensorModule::Alpha`. `self-stake` exists precisely for the pairs
that scan **misses**: an owner's own stake on their own hotkey frequently has no Alpha entry at all
— `{bits: 0}` on ~91% of one hotkey's registered pairs, measured by the Python script the lane
replaced — even when the runtime-computed stake is large.

Those rows belong in `nominator_positions`, not beside it. A validator's own stake is often its
largest position, and eight surfaces read that table: `subnet-holders`, `top-holders-holdings`,
`validator-nominator-positions`, both serving tiers, the staleness watchdogs,
`hotkey-alpha-completeness`. A separate table means a UNION in every one of them, and the first
reader anybody forgot would under-report silently.

## Why it cannot reuse `/api/v1/internal/nominator-positions-sync`

That route **prunes per coldkey**: every row for a posted coldkey older than that coldkey's newest
`captured_at` in the request. `validator-nominators` may do that because `pack_coldkey_chunks`
never splits a coldkey across requests.

`self-stake` structurally cannot. Its rows are absent from the Alpha scan *by construction*, so from
the other lane's point of view they are always "older than this pass". Any owner who also nominates
elsewhere appears in a validator-nominators pass, and that pass deletes their self-stake row —
written weekly, deleted within a day, with survival depending on whether the owner happened to
nominate anywhere else.

## The fix: a prune domain, not a new table

`0027` adds `source TEXT NOT NULL DEFAULT 'alpha'`. All 123,057 existing rows become `'alpha'` —
which is what wrote them — so the scoped prune deletes exactly what the unscoped one did. The new
clause is strictly **narrower**, which is the direction a change to a `DELETE` should always run.

| producer | writes | prunes |
|---|---|---|
| `validator-nominators` | `source='alpha'` | `… AND source='alpha'` |
| `self-stake` | `source='self-stake'` | `… AND source='self-stake'` |

`source` is stamped by the **writer**, never accepted from the wire: a producer that could name its
own source could claim the other lane's domain and delete its rows.

## Work

- [ ] `0027_nominator_positions_source.sql` — the column plus a `(coldkey, source, captured_at)`
      index for the prune's predicate
- [ ] `pruneKeysInNeon` takes an optional source, bound as a parameter
- [ ] `mirrorNominatorPositionsToNeon` takes `source` and `lane`, stamps the column, scopes the prune
- [ ] `handleSelfStakeSync` + its proxy on the main Worker
- [ ] `self-stake` in `NEON_DUAL_WRITE_LANES`, its own lane verdict (not the sibling's — different
      cadences, and a shared verdict lets one lane's silence hide behind the other's success)
- [ ] **no pass tally** — `nominator_positions_passes` records completeness of a full keyspace scan,
      and self-stake is a targeted read; declaring it complete would corrupt the ledger the other
      lane's consumers gate on
- [ ] a test that measures a self-stake row surviving a validator-nominators pass, and its inverse

JSONbored:
`nominator_positions` is about to have a second writer, and the prune it already has would delete
that writer's rows.

Closes #10845

## Why a second writer, and why it belongs in this table

`validator-nominators` scans `SubtensorModule::Alpha`. `self-stake` exists precisely for the pairs
that scan **misses** — an owner's own stake on their own hotkey frequently has no Alpha entry at all
(`{bits: 0}` on ~91% of one hotkey's registered pairs) even when the runtime-computed stake is large.

A validator's own stake is often its largest position, and eight surfaces read this table:
`subnet-holders`, `top-holders-holdings`, `validator-nominator-positions`, both serving tiers, the
staleness watchdogs, `hotkey-alpha-completeness`. A separate table means a UNION in every one of
them, and the first reader anybody forgot would under-report silently.

## Why it cannot reuse the existing route

That route prunes per coldkey: every row for a posted coldkey older than that coldkey's newest
`captured_at` in the request. `validator-nominators` may do that because `pack_coldkey_chunks` never
splits a coldkey.

`self-stake` structurally cannot. Its rows are absent from the Alpha scan **by construction**, so
from that lane's point of view they are always "older than this pass". Any owner who also nominates
elsewhere would lose their self-stake row within a day of it being written — partial, unpredictable
survival, which is worse than either extreme.

## The fix: a prune domain

`0027` adds `source TEXT NOT NULL DEFAULT 'alpha'`. All 123,057 existing rows become `'alpha'` —
which is what wrote them — so the scoped prune deletes **exactly** what the unscoped one did. The new
clause is strictly *narrower*: it cannot delete anything the current prune keeps, which is the
direction a change to a `DELETE` should always run.

| producer | writes | prunes |
|---|---|---|
| `validator-nominators` | `source='alpha'` | `… AND source='alpha'` |
| `self-stake` | `source='self-stake'` | `… AND source='self-stake'` |

`source` is stamped by the **writer** and rejected on the wire — a producer that could name its own
source could claim the other lane's domain and delete its rows.

**No pass tally**, unlike its sibling. `nominator_positions_passes` records completeness of a full
keyspace scan; self-stake is a targeted read, and declaring it complete would corrupt the ledger the
other lane's consumers gate on.

## Measured, not argued

The headline test is `a self-stake row SURVIVES a validator-nominators pass over the same coldkey`,
with its inverse alongside it. Removing the scope fails **both** — i.e. the mutation reproduces the
exact data loss this route exists to prevent.

## Verification

`typecheck` · `build` (no artifact drift) · all **42** CI validators · `lint` · `format:check` ·
**137 tests** across the seven affected suites, incl. 11 new route tests against real Postgres with
`0027` applied verbatim.

## Sequencing

The producer half is metagraphed-infra's and needs `SELF_STAKE_SYNC_SECRET` provisioned after this
merges. Until then the lane stays disabled — which is its state today, since it has never once run.


JSONbored:
Pushed fixes for both failures.

**Public-safety scan.** My comments used bare `coldkey` in prose, which that scanner flags — an ss58 coldkey is public data, but the rule exists to catch *"your coldkey seed phrase"* wording, and it only exempts the forms where it reads as a code identifier or a field name. Backticks cleared three of the four; the fourth was a sentence where the word genuinely ran as prose (`… older than that coldkey's newest captured_at …`), so I reworded it rather than smuggling it past the rule.

**codecov/patch.** Measured it directly rather than guessing, by intersecting the diff with an lcov run:

```
src/neon-write.ts                     1/1
src/nominator-positions-neon-write.ts 12/12
workers/api.ts                        3/3
workers/data-api.ts                   35/35
TOTAL 51/51 = 100.00%
```

Two real gaps turned up doing that, both now closed:

- the `{rows:[...]}` envelope was never exercised — every test used a bare array, so that arm of the body parse was dead. It is the form the producer actually posts.
- the **false** side of the `self-stake-sync` path check. Covering it needed a request that falls *past* this branch to a route dispatched below it, which is a property worth asserting on its own: a loosened match here would silently swallow `validator-nominator-counts-sync` and the four routes under it, sending another lane's POST to this handler to fail its token check. That is now a named test rather than a coverage artifact.

Full gate green on current main: `typecheck` · `build` · **42** validators · `lint` · `format:check` · `scan:public-safety` · 177 tests across the affected suites.

Issue #10790, opened 2026-08-11
schemas: 340 .passthrough() means 37% of the contract ships fields it never declared


JSONbored:
**340 `.passthrough()` against 573 `.strict()`** — 37% of the contract accepts fields it does not declare.

## What it costs

`SubnetUptime.observed_at` is the worked example (#10761): `formatUptime` emits it on every card, both transports feed it, REST reads it back into the response meta — and `UptimeArtifactSchema` never declared it. `.passthrough()` is why nothing noticed, for months.

That is the benign version. The security version is the same mechanism: a producer that starts emitting an internal field ships it to clients, past a response tripwire that validates declared fields and waves the rest through. **`safeParse` over `.passthrough()` is a leak guard that does not guard** — which is why this issue and the response-validation work are one decision, not two.

## What to do

**Measure before enforcing.** Flipping 340 schemas blind will break things, and the breakage will be indistinguishable from the leak you are trying to find.

1. Add a report-only strict mode: parse with `.strict()`, log what the payload carried that the schema did not, serve anyway.
2. Run it across the served surface — production for shape, and the test fixtures for the degraded paths production will not show you (the lesson from #10786: production is not degraded when you ask it).
3. Triage each undeclared field: **declare it** (it belongs in the contract, like `observed_at`), or **stop emitting it** (it does not).
4. Then flip to enforcing, per schema, as each is cleared.

## What to avoid

- **Do not flip and fix forward.** An undeclared field is a fact about the producer; discovering it via a 500 in production is the worst way to learn it.
- **Do not declare a field just to clear the report.** Ask whether a client should see it at all — that judgement is the security value of this work.
- **Do not leave `.passthrough()` on anything a response validator covers.** The combination is strictly worse than either alone: it costs CPU and guarantees nothing about what leaves.
- **Do not treat `.strict()` on INPUT the same as on output.** Input strictness rejects a caller; output strictness protects one. Sequence them separately.

## Done when

Zero `.passthrough()` in `schemas-src/`, every previously-undeclared field either declared or no longer emitted, and a gate that fails a new one.

Part of #10780.

---

## Completion requirements (all must hold)

**Order within this issue — measuring before enforcing is the whole method:**
1. Add a report-only strict mode: parse with `.strict()`, log what the payload carried that the schema did not, **serve anyway**.
2. Run it across the served surface — production for shape, and the test fixtures for the degraded paths production will not show you.
[...]

PR #10905, opened 2026-08-12
fix(mcp): declare the fallback tier's five surface-row fields on get_subnet_health

### LINKED ISSUE #10904: fix(mcp): get_subnet_health's outputSchema refuses the live-postgres-fallback tier's rows -- netuid 0 drifts on every call
Alerted 2026-08-12 20:25Z (McpResponseSchemaDriftError). Reproduced deterministically: `get_subnet_health {"netuid": 0}` refuses with `response_schema_drift` on every call. The root network's response comes from the live-postgres-fallback tier, whose surface rows carry `provider`, `url`, `classification`, `status_code`, `observed_by` — all five declared by the ROUTE's own row schema (HealthSubnetSurfaceSchema, routes/health-surfaces.ts) and all five missing from the MCP tool's hand-copied row schema. The #10790 drift class, caught by the outbound tripwire doing its job. Fix: declare the five (optional — the tiers differ in what they attach), verified against the live payload.

JSONbored:
Closes #10904

`get_subnet_health {"netuid": 0}` refuses with `response_schema_drift` on every call (alerted 2026-08-12 20:25Z, reproduced deterministically against production). The root network's response is served by the live-postgres-fallback tier, whose surface rows carry `provider`, `url`, `classification`, `status_code`, and `observed_by` — exactly the five keys the ROUTE's own row schema (`HealthSubnetSurfaceSchema`, routes/health-surfaces.ts) has declared all along and the MCP tool's hand-copied row schema lacked. The #10790 drift class, caught live by the outbound tripwire doing precisely its job.

All five declared as optional (the tiers differ in what they attach; the route requires several, so optional is the safe superset of every tier). Verified by validating the live production payload against the schema: the five `unrecognized_keys` issues on every surfaces[] row are gone.

`tsc` clean, `validate:mcp` (235 tools against real responses) and `validate:contract-drift` pass.

### DIFF
--- a/schemas-src/mcp-tools/get-subnet-health.ts
+++ b/schemas-src/mcp-tools/get-subnet-health.ts
@@ -7,7 +7,10 @@ import { McpUnsortedPageFields, netuidSchema } from "./shared.ts";
 import { ListSubnetHealthInputSchema } from "./subnet-scoped-lists.ts";
 import { HealthSubnetSummarySchema } from "../routes/health.ts";
 import { LIVE_HEALTH_OVERLAY } from "../routes/subnet-detail.ts";
-import { ReliabilityScoreSchema } from "../routes/health-surfaces.ts";
+import {
+  HealthSubnetSurfaceSchema,
+  ReliabilityScoreSchema,
+} from "../routes/health-surfaces.ts";
 
 /**
  * DERIVED FROM THE NETWORK-WIDE SIBLING (#9998).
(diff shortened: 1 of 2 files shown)


JSONbored:
Revised per review: the widened hand-copy was the same two-declarations disease #10790 exists to kill — the copy would just lag the route again on the next field. The row schema is now the ROUTE's own `HealthSubnetSurfaceSchema`, exported from routes/health-surfaces.ts and aliased in the tool. Safe because the tool serves the same `overlaySubnetHealth()`/fallback-tier rows the route serves (which pass the route's tripwire across every tier in production), and the tool advertises no `fields`, so rows are never projected — the route's exact strict shape is the right one. Zero surfaces[] issues against the live netuid-0 payload; `validate:mcp` (235 tools, real responses) and `contract-drift` pass.

PR #11049, opened 2026-08-13
fix(lakehouse): a table read through R2 SQL must be snapshotted, and now is

JSONbored:
`chain.subnet_snapshots` is joined by `src/neuron-daily-cold-tier.ts` and was **not** in the catalog snapshot.

Not incidentally joined, either — the query prices stake and emission in TAO:

```sql
nd.stake_tao * CASE WHEN nd.netuid = 0 THEN 1
                    ELSE s.tao_in_pool_tao / s.alpha_in_pool END AS total_stake_tao
```

So three columns behind a **published figure** had no snapshot, no generated tuple, no Zod schema and no drift coverage.

## The rule existed and was a comment

`refresh-lakehouse-schema.ts` snapshots *"every `chain.*` table this repo READS"* and its list says: *"add one here the moment a reader does"* — while naming `subnet_snapshots` among the tables nothing reads. The rule was right. Nothing checked it.

`scripts/validate-lakehouse-readers.ts` checks it now, derived from the source both sides already have: tables named in SQL vs tables in the list. Both spellings, because both are in use — `chainTable("x")` and a literal `${NAMESPACE}.x` in a hand-written JOIN. **It was the literal one that escaped.**

## The gate's own first draft was wrong

Anchored on `FROM`/`JOIN` rather than a bare `chain.`, because the first version matched `https://archive.chain.opentensor.ai` and reported `chain.opentensor` as an unsnapshotted table. A hostname is not a table, and a gate that cries wolf on one gets muted.

Verified to fail on the real case by removing the entry again:

```
lakehouse-readers: 1 table(s) read but not snapshotted:
  chain.subnet_snapshots — read by src/neuron-daily-cold-tier.ts
```

## Also

`refresh-lakehouse-schema.ts` fetched the live catalog at **import time**, so reading its table list required a catalog token. Guarded behind `main()`.

## Follow-up needed

The columns still need a refresh run (`npm run refresh:lakehouse-schema`, which needs `R2_CATALOG_TOKEN`) before `subnet_snapshots` gets its generated types and Zod schema. The weekly workflow will open that PR on its next pass — this PR makes the omission fail loudly in the meantime.

16 tables read across 441 files, all snapshotted.


JSONbored:
Superseded by a clean rebuild on current main — #11008 squash-merged, so this branch was replaying commits already there and produced nine conflicts across five files. Reopened as a focused 12-file PR with the same content.

Issue #11034, opened 2026-08-13
e2e: the p50 contract rename broke 7 hermetic fixtures, and every UI PR with them


JSONbored:
Since #10989 (the breaking p50 rename) merged, the committed HAR fixtures serve pre-rename rows the pages no longer parse: /validators and /subnets render near-empty tables (sticky-header correctly refuses to measure them), the evidence deep-link redirect never runs, and the multisig error states time out — 7 failures on every UI PR's e2e job, reproduced on clean main. Fix: re-record the browser HARs against the production-based worker per the documented record flow, plus the SSR supplement entries the sweep recorded; verify 104/104 and the har-coverage ratchet.

Issue #11023, opened 2026-08-13
perf(api): /api/v1/health burns 300-660ms CPU per request parsing a 253 KB snapshot to serve a summary


JSONbored:
`/api/v1/health` is the most-polled route on the API and, per CPU, one of the most expensive. From a 3-day sample, `cpu_time_ms` on plain `GET /api/v1/health`:

```
294  318  343  411  457  483  489  591  660
```

and its per-subnet sibling `/api/v1/subnets/1/health`: `275, 296, 304, 477`.

For comparison, in the same window a full lakehouse-backed `/api/v1/blocks/{ref}` scan — a multi-second R2 SQL query — costs `cpu_time_ms: 1`. The scan is I/O; **health is CPU**, and CPU is the metered resource.

## Why it costs that

`/api/v1/health` is live-only by design (`LIVE_OVERLAY_ROUTE_IDS` holds `health` and `subnet-health`, so it is deliberately excluded from the static edge cache — correct, current health must not be pinned). It resolves through `loadGlobalOperationalHealth` → `resolveLiveHealth`, reading the KV key `health:current`.

Observed in the earlier trace sample, that value is **253,490 bytes**, read with `{ type: "json" }` — so every request pays a 253 KB fetch plus a full `JSON.parse` of it, then re-serialises a projection. That is the shape of a 300–660 ms CPU cost, and it scales with the size of the health snapshot rather than with the size of the answer.

Two consequences beyond the bill:

- **It is the route a prober hits hardest.** Every `Network connection lost.` in the sample is the same `HeadlessChrome/151` prober abandoning a response, and `/api/v1/health` is its most frequent target. A prober timing out against the health endpoint means the health signal itself is unreliable — see #11022.
- **CPU is per-isolate contention.** 660 ms of parse on a hot path shares an isolate with everything else in that colo, including the account routes that already OOMed one (#11019).

## Fixed looks like

The answer is not "cache it" — the exclusion from the static cache is right. It is that **serving current health should not require parsing the whole snapshot.**

Options, roughly in order of value:

1. **Don't parse what you don't serve.** If `/api/v1/health` returns a summary and `/api/v1/subnets/{n}/health` returns one subnet, neither needs all 253 KB deserialised. A pre-reduced `health:summary` key written by the same 2-minute cron that writes `health:current` turns both into a small read. The cron already owns this data; the projection is free at write time and quadratic-ish at read time.
[...]

Issue #11022, opened 2026-08-13
obs: ~80% of the error channel is cache misses and client disconnects, and it is hiding the OOM and the poller hang


JSONbored:
Across a 3-day error sample, **two fingerprints that are not errors account for roughly 80% of the volume**:

| fingerprint | what it actually is | share |
| --- | --- | --- |
| `a3351df5…` | `r2_get ERROR — The specified key does not exist.` on `icon-cache/…` | ~55% |
| `7ec1a015…` | `Network connection lost.` | ~27% |
| `7cc07c51…` | `Durable Object reset because its code was updated.` | ~12% |
| `5fe07c03…` | `Worker exceeded memory limit.` | ~7% |
| `9ef85f28…` | the Workers runtime killed a hung poller invocation | 1 event |

The bottom two rows are the only ones that mean something is wrong. They are also the two that are impossible to see.

## Each of the top three is a normal event wearing an error's clothes

- **A cache miss is not an error.** `icon-cache/{subject}/{size}` is a *cache* key; a miss is the cold path working. (Why the same four subjects miss forever is #11020 — a real defect, but the logging is separately wrong.)
- **A client disconnecting is not our failure.** Every `Network connection lost.` in the sample carries `outcome: "ok"` and comes from one `HeadlessChrome/151` prober (ASN 8075) against `/api/v1/health`, `/api/v1/subnets/1/health`, `/api/v1/subnets/1/endpoints`, `/api/v1/economics`, `/api/v1/validators/economics`, `/api/v1/endpoint-incidents`. The Worker did its job; the caller hung up.
- **A DO reset on deploy is the platform working as designed** (#11021 covers the part that *is* wrong: it reaches the client as a 500).

## Why this is worth an issue rather than a shrug

This is the same class as the memory I keep re-learning about this stack: **a channel that cannot distinguish a normal event from a fault is not a signal.** Concretely, in this window it hid:

- an isolate OOM that 503'd a request and took its neighbours with it (#11019)
- a poller invocation the runtime had to kill (metagraphed-infra#505)

Both are single-digit event counts under ~8,000 lines of routine noise. Nobody scrolls that far, and no threshold alarm survives a 95%+ false-positive base rate.

It also has a cost dimension: these are Workers Logs writes and Observability retention, paid for at ~2.3 events/minute sustained to record cache misses.

## Fixed looks like

[...]

Issue #11018, opened 2026-08-13
perf(api): the two block sub-resources #11010 left behind are still uncached, and hardcode 'short'


JSONbored:
#11010 wrapped `/api/v1/blocks/{ref}` and `/api/v1/extrinsics/{hash}` in the edge cache. The two block **sub-resources** dispatched immediately above them in the same function were left behind:

- `/api/v1/blocks/{ref}/extrinsics`
- `/api/v1/blocks/{ref}/events`

```ts
// workers/api.ts, dispatchChainHistoryRoute — all four are adjacent
blockExtrinsicsMatch → return handleBlockExtrinsics(...)              // uncached
blockEventsMatch     → return handleBlockEvents(...)                  // uncached
blockDetailMatch     → return withChainDetailEdgeCache(... handleBlock)      // #11010
extrinsicDetailMatch → return withChainDetailEdgeCache(... handleExtrinsic)  // #11010
```

Same properties as the two that were fixed: block-scoped, unbounded (~8.8M × 2), lakehouse-backed, and immutable once the block is settled. They *are* rate-limited (`chainDetailRateLimit`), so this is cost and latency rather than an abuse gap — the pure #11001 class.

## Both halves are needed, and the second is the one that bites

Wrapping them in `withChainDetailEdgeCache` alone **changes nothing**: both hardcode `"short"` in their `envelopeResponse` call, and the wrapper only stores a response whose `x-metagraph-cache-profile` is `static`. So each handler has to name its own settledness first, exactly as `handleExtrinsic` was made to in #11016.

And these two are the case where getting that wrong is *most* tempting. Both read through `answerBlockDetail`, whose contract is `cold ? { kind: "answer", tier: "cold" } : { kind: "miss" }` — and the cold loaders return a schema-stable payload for a confirmed absence rather than null. So `tier === "cold"` is true for "the lakehouse looked and this block is not there" here too.

Worse, "empty" is genuinely ambiguous on these two routes, and deliberately so: an empty extrinsics array and a block that genuinely had no extrinsics are the same bytes, which is the entire argument of #9208 and the reason `chainDetailGapResponse` exists. Both handlers are already handed an `isEmpty` predicate (`isEmptyExtrinsicPayload`) for precisely this distinction — so the storable condition is **cold AND not-empty**, and `isEmpty` is already in scope to express it.

Pinning an empty answer for an hour would convert "we cannot see it yet" into "this block had none", which is the one outcome #9208 was written to prevent.

## Also

[...]

PR #11050, opened 2026-08-13
chore: release main

### LINKED ISSUE #10600: api: the two composite subnet routes need a lever, and `fields` would mean two things
`/api/v1/subnets/{netuid}` (272,825 B) and `/api/v1/subnets/{netuid}/profile` (202,948 B) are the two
routes #9981 named that did **not** take the ordinary collection lever, and the reason is worth stating
rather than leaving as a gap.

## Why paging does not apply

Measured against production, their bulk is not one dominant list:

```
/subnets/64          endpoints=76, surfaces=76, verified_surfaces=76, candidate_surfaces=16
/subnets/64/profile  endpoints=76, surfaces=76, candidate_surfaces=16
```

Four parallel arrays of the same subject. A query collection pages **one** `data_key`, so declaring
one here would page a quarter of the payload and silently leave the rest — worse than not paging,
because the response would look narrowed while staying fat.

## The lever is `fields`, and that is the problem

`fields` already exists on 5 routes (`/subnets/{netuid}/metagraph`, `/neurons/{uid}`,
`/subnets/{netuid}/validators`, `/economics/*`). **Everywhere it exists it is a per-row column
projection** — pick columns out of the rows of a list.

What these two routes need is different: pick **top-level sections** out of a composite document
(`?fields=economics,health` returning those cards and not the four surface arrays). Same parameter
name, different unit of selection.

That is a contract decision, not an implementation detail:

- **One name, two meanings** is a real cost. A caller who learns `fields` on `/validators` and applies
  it to `/subnets/{netuid}` gets a different kind of answer, and no error tells them so.
- **#9884's rule** — a tool may serve a partial object only where it advertises `fields` — was written
  with the row-projection meaning in mind. Extending the name extends that rule to a shape it was not
  written for.
- The alternative is a **different parameter** (`sections=`, `include=`), which keeps `fields` honest
  at the cost of a second vocabulary for the same idea.

## What would settle it

A decision on which of those three the contract wants, and then the implementation is small: both
routes compose their response from named sections already.

## Done when

`/subnets/{netuid}` and `/subnets/{netuid}/profile` give a caller a way to ask for less, under a
parameter whose meaning is stated once and holds everywhere it appears.

Split out of #9981, whose other four routes took the ordinary `limit`/`offset` lever in #10599.

### LINKED ISSUE #10723: fix(watchdog): lane-alarm guesses a cadence from bursty writes, so four Neon lanes false-alarm every pass
`lane-alarm` calibrates its silence threshold from the **observed mean gap** between `lane_health` writes, and never consults the **declared** producer cadence — even for lanes that `LANE_PRODUCER` already maps to one. For bursty writers that collapses the threshold to its 90-minute floor, so a normal gap between passes fires a silence alarm every pass.

## Measured in production, 2026-08-11T02:28Z

Four lanes alarmed in a single tick:

```
lane neon:account-balances is silent: 1.8h
lane neon:account-balances-pass is silent: 1.8h
lane neon:hotkey-alpha is silent: 1.7h
lane neon:hotkey-alpha-pass is silent: 1.7h
```

Their producers' configured intervals:

| lane | producer | interval | correct silence bound (3x) | fired at |
| --- | --- | --- | --- | --- |
| `neon:account-balances`, `…-pass` | `account_balances` | **6h** (`ACCOUNT_BALANCES_POLL_SECS=21600`) | ~18h | **1.8h** |
| `neon:hotkey-alpha`, `…-pass` | `hotkey_alpha` | **24h** (`HOTKEY_ALPHA_POLL_SECS=86400`) | ~72h | **1.7h** |

Alarming a 24-hour lane after 1.7 hours is off by a factor of forty.

## Why the observed cadence is wrong for these lanes

`laneCadenceMs` is a mean gap (`(last - first) / (n - 1)`, `LANE_CADENCE_SQL`). These lanes write a verdict **per batch inside a pass** — `neon:account-balances` reports `775 row(s) in 1 statement(s)`, and a pass is many such batches seconds apart. Those tiny within-pass gaps dominate the mean, so the computed cadence collapses, `laneSilenceThresholdMs` floors it at 90 minutes, and any between-pass gap over 90 minutes trips an alarm.

For a 6-hour producer that is **every pass**. For a 24-hour producer, also every pass. Consistent with the data: each of the four fired exactly once, at the tick where the gap crossed 90 minutes.

## This is #10634's failure mode, one class over

`LANE_SILENCE_EXEMPT`'s own comment describes the same mechanism for `poller-build`:

> Five deploys inside the 7-day cadence window were enough to calibrate a "cadence", and the first quiet stretch longer than 3× it opened an alarm.

That was fixed by exempting the lane, which was right *there*: `poller-build`'s silence genuinely carries no information. It is **not** right here — a dead Neon writer matters, and these lanes are the only thing that would say so. So the fix is the threshold, not an exemption.

## The fix is already in the repo, unused

`src/producer-cadence.ts` exists for exactly this and already maps all four:

```ts
export const LANE_PRODUCER: Readonly<Record<string, ProducerLane>> = {
  "neon:account-balances": "account_balances",
  "neon:account-balances-pass": "account_balances",
  "neon:hotkey-alpha": "hotkey_alpha",
  "neon:hotkey-alpha-pass": "hotkey_alpha",
  …
};
```

with declared, *measured* cadences beside them:

```ts
/** ACCOUNT_BALANCES_POLL_SECS. Measured 2026-08-09: p50 gap 5.79h, max 6.01h
 *  across 19 passes -- the configured value, confirmed. */
account_balances: 21_600,
/** HOTKEY_ALPHA_POLL_SECS. Measured 2026-08-09: 13 passes, max gap 23.64h … */
hotkey_alpha: 86_400,
```

`src/lane-alarm.ts` references `producer-cadence.ts` only in a passing comment — it imports nothing from it. The table-freshness watchdog already derives its bounds this way (`missedTicksMs(producer, n)`), so the pattern is established; lane-alarm is the one place that still guesses.

**Suggested change:** in `laneAlarmPlan`, prefer `PRODUCER_CADENCE_SECS[LANE_PRODUCER[lane]]` where a producer is declared, and fall back to the observed mean gap only for lanes that have none. That makes the declared cadence authoritative and leaves uncalibrated lanes behaving exactly as they do today.

Worth adding a test that a lane with a declared producer does **not** alarm inside that producer's interval — the assertion that would have caught this.

## Why it surfaced now

#10674 gave each lane its own `$exception` fingerprint. Before that, all four shared `watchdog:lane-alarm:Error` and the storm guard dropped three of every four alarms in a tick — so this pattern was invisible: you would have seen one lane, once, and read it as a blip. The first tick after that fix showed four.

### LINKED ISSUE #10744: fix(neon): cap what the write buffer may hold -- DO storage rejects the bulk upserts
## Problem

The write buffer was disabled in #10734 because Durable Object storage rejected the bulk writes
outright:

```
Internal error in Durable Object storage caused object to be reset;
reference = 986ji19jem6epm87bigf7u4t; prune did not land; tally withheld
    — nominator-positions, 2026-08-11 10:23:40Z
```

That is the **platform refusing the shape**, not a transient. No retry helps, and the two fixes that
preceded it (#10722 bounded `list()` + missing re-arm, #10729 enqueue retry) correctly addressed
different failures but could not address this one.

**The cause is the design's own.** A neuron or account-position upsert is thousands of rows and
megabytes of JSON. Chunking that into 96 KiB values and committing them in a single `put()` is more
than DO storage wants to hold at once.

## Fix: cap what may be buffered

Anything over `MAX_BUFFERED_PAYLOAD_BYTES` (64 KiB) goes **direct** to Neon and never touches DO
storage.

The number matters less than the split it creates:

- **The writes that pin the compute are small.** `tao_usd_index` every 60s, the ledger lanes every few
  seconds — all well under the cap, all still batched. That is the entire purpose of the buffer.
- **The megabyte upserts are infrequent.** `neurons` every 15 minutes. Sending those direct costs one
  connection per quarter hour and removes the whole class of failure.

**Per statement, not per lane.** One lane issues both shapes — the same `neurons` lane writes a small
watermark row and a thousands-row upsert — so a per-lane flag would either lose the batching on the
small writes or keep the failure on the big ones.

**Measured in bytes**, not code units: an SS58 address is ASCII but a subnet name need not be, and a
cap counted in characters would let exactly the payloads this exists to stop straight through.

## Acceptance criteria

- [ ] A small statement is still buffered.
- [ ] An oversized statement never reaches DO storage.
- [ ] The cap is byte-measured.
- [ ] With no direct runner available, an oversized statement still buffers rather than being dropped.
- [ ] After re-enabling: consecutive flushes with a backlog, and no DO storage errors in PostHog.

### LINKED ISSUE #10755: incident(neon): the buffer's enqueue was O(n) -- a full storage scan per write, 2.4s CPU, object reset
## Root cause, from Cloudflare's own logs

The write buffer was disabled after three apparently-separate failures. The Worker logs show they were
all one bug:

```
POST /enqueue    outcome: exception
cpuTimeMs: 2424    wallTimeMs: 2699
entrypoint: NeonWriteBufferHub    executionModel: durableObject
→ Internal error in Durable Object storage caused object to be reset
```

**2.4 seconds of CPU to append one row.**

`enqueueStatement` opened with `countPending(storage)`, and `countPending` does a full
`storage.list()` over every chunk in the buffer, then sorts and groups the keys — **on every single
enqueue**.

## Why it was a spiral, not a slow path

1. Each enqueue scans the whole backlog → CPU grows with the backlog
2. The platform resets the object mid-request
3. The reset happens **before `setAlarm`**, so nothing arms the flush
4. Nothing drains → the backlog grows → the next scan is worse

That single O(n) call produced every symptom chased separately:

| Symptom | Chased as | Actually |
|---|---|---|
| `Internal error in DO storage` | #10744 — payload too large | CPU burn from the scan |
| Enqueues stopped landing | #10729 — deploy resets | Object reset by the scan |
| Alarm never fired | #10722 — missing re-arm | Reset happened before `setAlarm` |

Each of those fixes was correct for the failure it targeted. None could fix this, because none touched
the scan.

## Fix

Maintain `pending` as a counter in storage. The enqueue becomes O(1): one `get`, one `put`, no `list`.

The counter is **reconciled, not just decremented** — a clean drain writes the true value rather than
subtracting, because a counter that can only be adjusted eventually lies.

## Acceptance criteria

- [ ] An enqueue never calls `storage.list()` — asserted directly.
- [ ] The ceiling still refuses, now from the counter.
- [ ] A clean drain reconciles the counter to zero even after deliberate drift.
- [ ] After re-enabling: `cpuTimeMs` per enqueue in single-digit ms, and consecutive flushes.

### LINKED ISSUE #10760: graphql: the 42 differences between the generated schema and the published one that are not nullability

### LINKED ISSUE #10763: fix(neon): an orphaned alarm blocks the buffer forever -- '=== null' is not the right guard
## Problem

With #10755 fixed — the O(n) enqueue that burned 2.4s of CPU and got the object reset — the buffer
went live at 12:36:59 and **still did not drain**. Thirteen minutes later:

```
enqueues              succeeding (tao_usd buffered from 12:36:47)
buffered-lane errors  0
DO storage errors     0
flushes               0
```

Everything healthy, nothing draining. That isolates the last fault cleanly.

## Cause

`enqueueStatement` armed only when `getAlarm() === null`. An alarm whose time has **passed without the
handler running** is orphaned — and #10755's CPU spiral reset the object mid-flight often enough to
leave one behind. The guard read that orphan as "an alarm is pending, nothing to do".

Nothing ever fired it. Nothing ever replaced it. The buffer had no path to a first drain even once
every other fault was gone.

## Fix

Arm when the alarm is `null` **or** its time has passed:

```ts
const existingAlarm = await storage.getAlarm();
if (existingAlarm === null || existingAlarm <= now) { … }
```

Comparing against `now` rather than re-arming unconditionally keeps what the original guard was
protecting: a live alarm is left alone, so a steady stream of enqueues cannot push the flush further
away with every write.

## And: an empty drain now records a verdict

Returning silently on `groups.length === 0` is **why three incidents were hard to tell apart** — the
absence of a `neon:buffer-flush` verdict meant both "the alarm never fired" and "it fired and found
nothing". One cheap row per wake makes that answerable in a single query instead of by inference.

## Acceptance criteria

- [ ] A past alarm is replaced.
- [ ] A live alarm is left alone.
- [ ] An alarm exactly at `now` counts as stale.
- [ ] An empty drain records a verdict.
- [ ] After deploy: consecutive flushes, and buffered-lane data landing.

### LINKED ISSUE #10769: graphql: the 300 nullability tightenings need evidence per field, not one aggregate

### LINKED ISSUE #10772: graphql: eight Query fields generate with NO arguments, and the gate that should have caught it skipped them
The generated schema now reproduces the published one field for field (#10760) with per-field evidence for every nullability tightening (#10769). So I swapped `src/graphql.ts`'s `buildSchema(SDL)` for the generated schema and ran the resolver suite.

**75 of 1101 tests failed. Only 5 were nullability.** The other 70 were `400` — the generated schema *rejects queries the tests send*, because eight Query fields come out with **no arguments at all**:

```
subnets              pub: 34 arguments    gen: (none)
subnet               pub: netuid: Int!    gen: (none)     <-- required arg, gone
providers            pub: 8 arguments     gen: (none)
provider             pub: id: String!     gen: (none)
opportunity_boards   pub: limit: Int      gen: (none)
health               pub: (none)          gen: (none)
saved_query          pub: id: String!     gen: (none)
extrinsic            pub: ref: String!    gen: (none)
block_chain_events   pub: block_number    gen: (none)
```

## The cause: one field meaning two things

`QueryBinding.route` documented `route: null` as "the field reads a published artifact directly rather than mirroring an `/api/v1` route **with a component**". Six bindings carried it. **Every one of the six has a route, with parameters, whose response names a component:**

| field | claimed | actual route | params | route's component | type declared over |
|---|---|---|---|---|---|
| `subnets` | null | `/api/v1/subnets` | 30 | `SubnetsArtifact` | `SubnetsArtifact` ✅ |
| `subnet` | null | `/api/v1/subnets/{netuid}` | 1 | `SubnetDetailArtifact` | `SubnetIndexEntry` |
| `providers` | null | `/api/v1/providers` | 9 | `ProvidersArtifact` | `ProvidersArtifact` ✅ |
| `provider` | null | `/api/v1/providers/{slug}` | 1 | `ProviderArtifact` | `Provider` |
| `health` | null | `/api/v1/health` | 6 | `HealthSummaryArtifact` | `HealthSummaryArtifact` ✅ |
| `opportunity_boards` | null | `/api/v1/registry/leaderboards` | 2 | `RegistryLeaderboardsArtifact` | `OpportunityBoards` |

`route` names the **request**; the component question is about the **response**. Conflating them meant `deriveQueryArguments` had nothing to read — and `buildSchema(SDL)` has no resolver hook that could add an argument back, so whatever the generator omits is a capability a GraphQL caller simply cannot reach.

Three more (`saved_query`, `extrinsic`, `block_chain_events`) name routes the OpenAPI document does not contain: `/api/v1/queries/{id}` (no `queries` path at all), `/api/v1/extrinsics/{ref}` (the spec has `{hash}`), `/api/v1/blocks/{block_number}/chain-events` (the spec has `{ref}`). Those last two are deliberate GraphQL argument names, not typos — the descriptions say so — but a route the spec lacks derives nothing either way.

## Every gate that could have caught this skipped it

`validate:graphql-query-arguments` printed **"189 field(s) reproduce their route's parameters exactly, 9 skipped"** and passed. The skip condition is `!binding.route || !openapi.paths?.[binding.route]` — the same shape as the null-returning walk #10718 removed, and the same failure: **a gate that skips is a gate that passes.**

## What splitting the two facts exposes

Repointing the six at their real routes (kept on `claude/graphql-sdl-cutover`, with a `reshapes` reason on the three whose response genuinely differs) takes the gate from *9 skipped, passing* to **3 skipped, 39 named divergences**:

- **12 × `Int` vs `Float`** (`subnets.min_/max_*`) — the same class #10760 resolved for response fields, now on the request side.
- **2 × `cursor: String` vs `Int`** (`subnets`, `providers`) — a real, documented divergence: GraphQL keeps an opaque string keyset where REST takes an integer offset. The binding description already says so.
- **1 rename**: `provider(id:)` vs the route's `slug`.
- **5 × `not_*` negation filters** on `subnets` that no route parameter derives.
- **2 renames**: `subnets.max_readiness`/`min_readiness` vs the route's `max_integration_readiness`/`min_integration_readiness`.
- **10 route capabilities GraphQL does not publish** — `health`'s six filters (netuid/status/sort/order/limit/cursor), `subnets.netuids`, `subnets.q`, `opportunity_boards.board`, `providers.fields`. A GraphQL caller cannot filter the health rollup at all today.

Each needs a decision, not a declaration: which side is right, and where GraphQL is missing a route capability, whether to publish it. Declaring all 39 to make the gate green would be the exemption list that hides what it names.

## Done when

- `route` means the request and only the request; the response-side fact is its own declaration
- `validate:graphql-query-arguments` skips **0** fields
- the generated Query root reproduces the published one argument for argument
- `src/graphql.ts` builds from the generated artifact and `src/graphql-sdl.ts` is deleted

The type side is finished — `report:graphql-schema-diff` is at 0 other differences and 0 unreached, and the printed-schema diff beyond arguments is descriptions, where the generated text is *newer* than the SDL's (the SDL's `AccountEntry.total_stake_tao` prose predates the spot-price correction the Zod already documents).

Closes out #10214.

### LINKED ISSUE #10773: fix(identity): the subnet-identity writer came back and the readers did not — all three surfaces still serve the frozen 2026-08-02 export
`subnet_identity_history` got its writer back on 2026-08-11 (#10740 / #10762 and metagraphed-infra#444). **Nothing repointed the read.** All three surfaces still serve the frozen lakehouse export.

Measured at 13:15Z, minutes after the producer's first successful pass:

```
lane subnet-identity      verdict=ok   "129 scanned, 124 written, 5 error(s)"
lane neon:subnet-identity verdict=ok   "248 row(s) in 2 statement(s)"

GET /api/v1/chain/identity-history?limit=5   →  newest change: 2026-07-31T11:00:20.942Z
```

248 rows written, and the newest identity change the API will admit to is eleven days old. Re-read with two different cache keys and `Cache-Control: no-cache` — it is not an edge cache.

## Why

`handleChainIdentityHistory` passes `null` as the tier result and the composer's only remaining leg is the export:

```ts
// NO TIER READ (#10190). METAGRAPH_SUBNET_IDENTITY_SOURCE reads "retired" in every
// deployed config and is absent from DATA_API_FORWARD_FLAGS, so this resolved to
// null on every request.
const data = await answerChainIdentityHistory(env, null, { limit });
```

That comment was correct when it was written. `syncSubnetIdentityToPostgres` was named as the table's sole writer in four places and never existed, so the tier read genuinely was unreachable and #10190 removed it. The premise expired this morning and the code did not.

**Five call sites pass `null`** — REST ×2 (`workers/request-handlers/entities.ts`), MCP ×2 (`src/mcp-server.ts`), GraphQL ×2 (`src/graphql.ts`) — so the gap is every surface, not one.

The lakehouse copy is one of the 22 **frozen** tables: a one-time 2026-08-02 seed that does not advance. Reading it is not "slightly behind"; it is permanently 2026-07-31 no matter what the chain does.

## What this needs

A live leg in `src/identity-history-answer.ts` — the composer, not the surfaces. That module exists precisely because #9153 added the lakehouse leg to one surface and left MCP and GraphQL reporting `entry_count: 0`; a live leg added at one call site would recreate exactly that.

Cascade order: caller-supplied tier → **live** → frozen → schema-stable empty.

Two properties the fix has to hold:

1. **A live MISS falls through, it does not answer.** The table fills from 2026-08-11 forward and the export holds everything before it, so a netuid whose identity last changed in July has zero live rows and a real frozen timeline. An empty live read treated as authoritative would publish "this subnet has never changed its identity" for most of the network.
2. **The live query uses the cold tier's total order** (`block_number DESC, netuid ASC, id DESC`). The two legs answer the same question and a caller cannot tell which one did; disagreeing orders would surface as rows shuffling whenever the live store went cold.

`subnet_identity_history` is already in `NEON_SOLE_STORE_TABLES` in all three configs, so this is a read-store declaration plus the cascade — not the six-place new-table change.

## Also seen, not covered here

The producer's own pass reports `5 error(s)` out of 129 subnets. Worth its own look; it does not block the repoint.

### LINKED ISSUE #10781: schemas: three operation registries for one concept, and they already disagree
Three registries describe one concept, and they already disagree.

| registry | owns | shape |
|---|---|---|
| `API_ROUTES` (`src/contracts.ts`) | REST paths + artifact_path | path, artifact_path |
| `QUERY_BINDINGS` (`schemas-src/graphql/published-names.ts`) | GraphQL fields | field, route, returns, description |
| the MCP tool registry (`schemas-src/mcp-tools/**`) | tool names + inputSchema | name, inputSchema, handler |

## Why this blocks everything downstream

#10772 is the proof they drift. Six `QUERY_BINDINGS` entries carried `route: null` meaning "mirrors no route with a component" — and every one had a route in `API_ROUTES`, with parameters, whose response named a component. Three more named paths the OpenAPI document does not contain (`/api/v1/queries/{id}` has never existed; `/api/v1/extrinsics/{ref}` is published as `{hash}`). Nothing caught it because the gate skipped exactly the fields whose route it could not resolve.

Any shared inbound codec or validator built on three registries has to reconcile them at request time, which is the drift this work exists to delete.

## What to do

One `OPERATIONS` registry keyed by an operation id every surface resolves to, carrying: the canonical input schema, the response component, and per-surface exposure (REST path / GraphQL field / MCP tool name — any of them absent where that surface does not expose it).

The overlap is heavy but not total, and the registry must model that rather than assume it:
- GraphQL-only: `opportunity_boards`, `chainEvents`, `saved_query` (no REST route at all)
- REST-only: `format=csv` variants
- MCP-only tools

## What to avoid

- **Do not force a 1:1 mapping.** A surface not exposing an operation is normal; make it representable, not an exception.
- **Do not let one field mean two things.** `route: null` meant "no route" AND "response does not name the component" simultaneously, and both facts rotted behind the ambiguity. Separate facts get separate fields.
- **Do not build it as a lookup that can decline silently.** A registry miss must be a typed error, not `null` — that is how `dataComponent` returned null for two different reasons (#10718).

## Done when

One registry; `API_ROUTES`, `QUERY_BINDINGS` and the tool registry derive from it or are gone; and a gate asserts every surface's exposure resolves to an operation that exists.

Part of #10780.

---

## Completion requirements (all must hold)

- [ ] One `OPERATIONS` registry keyed by an operation id every surface resolves to, carrying the canonical input schema, the response component, and per-surface exposure
- [ ] `API_ROUTES`, `QUERY_BINDINGS` and the MCP tool registry **derive from it or are deleted** — not "kept in sync with"
- [ ] A surface not exposing an operation is **representable**, not an exception (GraphQL-only `opportunity_boards`/`chainEvents`/`saved_query`; REST-only `format=csv`; MCP-only tools)
- [ ] **No field means two things.** `route: null` meant "no route" AND "response does not name the component"; separate facts get separate fields
- [ ] A registry miss is a **typed error**, never `null` — the `dataComponent` failure (#10718) must not reappear
- [ ] A gate asserts every surface exposure resolves to an operation that exists, and fails on an orphan in either direction
- [ ] `npm run validate` exit 0, `tsc` clean, full suite green, `codecov/patch` ≥ 99%

JSONbored:
---

<details><summary>ui: 1.0.0</summary>

### ⚠ BREAKING CHANGES

* **api:** one name for the 50th percentile across the contract ([#10989](https://github.com/JSONbored/metagraphed/issues/10989))

### Features

* **economics:** cost to participate — what a subnet declares, and what entry actually costs ([#11041](https://github.com/JSONbored/metagraphed/issues/11041)) ([1ab4bb4](https://github.com/JSONbored/metagraphed/commit/1ab4bb412727cf3bac0bd94292d9f6df4d7b210f))
* **economics:** miner fairness — do a subnet's registered miners actually earn ([#10997](https://github.com/JSONbored/metagraphed/issues/10997)) ([97261c1](https://github.com/JSONbored/metagraphed/commit/97261c13a2d79887961466d78eff5c8cfd6c5597)), closes [#10931](https://github.com/JSONbored/metagraphed/issues/10931)
* **economics:** split subnet emission by recipient class, and give the rao helpers one implementation ([#10947](https://github.com/JSONbored/metagraphed/issues/10947)) ([edf8c26](https://github.com/JSONbored/metagraphed/commit/edf8c268488805d4bbde040014a3679f81fa5968))
* **registry:** treasury readings — what a subnet's own source declares ([#11025](https://github.com/JSONbored/metagraphed/issues/11025)) ([7cd0ab7](https://github.com/JSONbored/metagraphed/commit/7cd0ab76902a7598590f51a342da89d42205d7f4)), closes [#10933](https://github.com/JSONbored/metagraphed/issues/10933)

### Bug Fixes

* **ci:** make the posthog CLI's CDN download unable to fail an install ([#10922](https://github.com/JSONbored/metagraphed/issues/10922)) ([bb88884](https://github.com/JSONbored/metagraphed/commit/bb88884af48559ea84ae2bff1bae9d692a9cd70d))
* **e2e:** re-record the fixtures the p50 rename left behind ([#11035](https://github.com/JSONbored/metagraphed/issues/11035)) ([0f3d824](https://github.com/JSONbored/metagraphed/commit/0f3d8244d2c8a7ecf9ca1aba23e41ce840c4ef95)), closes [#11034](https://github.com/JSONbored/metagraphed/issues/11034)
* **e2e:** the overflow sweep stops reading live production ([#10968](https://github.com/JSONbored/metagraphed/issues/10968)) ([f639887](https://github.com/JSONbored/metagraphed/commit/f639887161d9a2eeee5ad02c67ec1f9231e719d6))
* **seo:** withhold the three unbounded per-entity spaces from crawlers, on both hosts ([#11005](https://github.com/JSONbored/metagraphed/issues/11005)) ([8787848](https://github.com/JSONbored/metagraphed/commit/8787848f90d3c8506ee6f5473f5b4eb9e87a8e0a)), closes [#11002](https://github.com/JSONbored/metagraphed/issues/11002)
* **ui:** /revenue was in the sitemap and nowhere else ([#10977](https://github.com/JSONbored/metagraphed/issues/10977)) ([f9ecad2](https://github.com/JSONbored/metagraphed/commit/f9ecad2fdbab9ae7896e0e6e4e49baa1a7f3af59)), closes [#10927](https://github.com/JSONbored/metagraphed/issues/10927)
* **ui:** a rate-limited read is throttled, not broken ([#11012](https://github.com/JSONbored/metagraphed/issues/11012)) ([ffa5f9a](https://github.com/JSONbored/metagraphed/commit/ffa5f9a7629469f68911cb168ae8828eea261fbe)), closes [#11000](https://github.com/JSONbored/metagraphed/issues/11000)
* **ui:** one chain-stream socket per tab, with a grow-only topic union ([#10961](https://github.com/JSONbored/metagraphed/issues/10961)) ([2e2b512](https://github.com/JSONbored/metagraphed/commit/2e2b512af7f46a37ab734f50124446eedce3f3c9)), refs [#10606](https://github.com/JSONbored/metagraphed/issues/10606)
* **ui:** one chain-stream socket per topic set, not one per call site ([#10952](https://github.com/JSONbored/metagraphed/issues/10952)) ([4ab50f4](https://github.com/JSONbored/metagraphed/commit/4ab50f4adc3d90557e327788b8aa9a15d9e0b7d7)), refs [#10606](https://github.com/JSONbored/metagraphed/issues/10606)
* **ui:** stop telling users their data comes from a database we deleted ([#10934](https://github.com/JSONbored/metagraphed/issues/10934)) ([d079c84](https://github.com/JSONbored/metagraphed/commit/d079c84e02fd7edc3679c874833185194b74b03b)), closes [#10911](https://github.com/JSONbored/metagraphed/issues/10911)

### Documentation

* **method:** publish the attribution rules, and make the schema enforce them ([#10962](https://github.com/JSONbored/metagraphed/issues/10962)) ([be364d3](https://github.com/JSONbored/metagraphed/commit/be364d38aae0e7b4c962a23a38071ceb7b214c6c))

### Refactors

* **api:** one name for the 50th percentile across the contract ([#10989](https://github.com/JSONbored/metagraphed/issues/10989)) ([8d4271a](https://github.com/JSONbored/metagraphed/commit/8d4271ab1887c2b0f0641c6c04f5093721a00a13))
</details>

<details><summary>platform: 2.0.0</summary>

### ⚠ BREAKING CHANGES

* **api:** one name for the 50th percentile across the contract ([#10989](https://github.com/JSONbored/metagraphed/issues/10989))

### Features

* **api:** the entity routes get the tripwire, warn-first, exactly as [#10897](https://github.com/JSONbored/metagraphed/issues/10897) taught ([#11044](https://github.com/JSONbored/metagraphed/issues/11044)) ([4ffb9ad](https://github.com/JSONbored/metagraphed/commit/4ffb9ade2f2958a7c32a767a0ec5ae533b8b519c)), closes [#10984](https://github.com/JSONbored/metagraphed/issues/10984)
* **api:** the revenue window is a caller's choice, resolved in one place ([#10974](https://github.com/JSONbored/metagraphed/issues/10974)) ([446d567](https://github.com/JSONbored/metagraphed/commit/446d567a065c71cae5b961c32fa9fbd05d5f98eb)), closes [#10925](https://github.com/JSONbored/metagraphed/issues/10925)
* **api:** the two composite subnet routes take a sections= lever ([#10959](https://github.com/JSONbored/metagraphed/issues/10959)) ([41e8ceb](https://github.com/JSONbored/metagraphed/commit/41e8ceb31b2915dcdff211df838c307b3b27c059)), closes [#10600](https://github.com/JSONbored/metagraphed/issues/10600) [#10958](https://github.com/JSONbored/metagraphed/issues/10958)
* **economics:** cost to participate — what a subnet declares, and what entry actually costs ([#11041](https://github.com/JSONbored/metagraphed/issues/11041)) ([1ab4bb4](https://github.com/JSONbored/metagraphed/commit/1ab4bb412727cf3bac0bd94292d9f6df4d7b210f))
* **economics:** miner fairness — do a subnet's registered miners actually earn ([#10997](https://github.com/JSONbored/metagraphed/issues/10997)) ([97261c1](https://github.com/JSONbored/metagraphed/commit/97261c13a2d79887961466d78eff5c8cfd6c5597)), closes [#10931](https://github.com/JSONbored/metagraphed/issues/10931)
* **economics:** split subnet emission by recipient class, and give the rao helpers one implementation ([#10947](https://github.com/JSONbored/metagraphed/issues/10947)) ([edf8c26](https://github.com/JSONbored/metagraphed/commit/edf8c268488805d4bbde040014a3679f81fa5968))
* **graphql:** join the nullability evidence to the tightenings it licenses ([#10770](https://github.com/JSONbored/metagraphed/issues/10770)) ([d4f21ed](https://github.com/JSONbored/metagraphed/commit/d4f21ed2618619064706ae009c41e6fa535ec1d6)), closes [#10769](https://github.com/JSONbored/metagraphed/issues/10769)
* **graphql:** reach the four unreached types, and let each fact be fixed where it lives ([#10761](https://github.com/JSONbored/metagraphed/issues/10761)) ([ce285b9](https://github.com/JSONbored/metagraphed/commit/ce285b9e4a759f7cc2d0cd73b321e31e130aad58)), closes [#10760](https://github.com/JSONbored/metagraphed/issues/10760)
* **graphql:** serve the schema the Zod builds, and delete the hand-written SDL ([#10863](https://github.com/JSONbored/metagraphed/issues/10863)) ([07e484d](https://github.com/JSONbored/metagraphed/commit/07e484dbf76d769ad500f9cf181a0e5e759c9812))
* **graphql:** the argument surface, generated -- and the runtime change it exposes ([#10824](https://github.com/JSONbored/metagraphed/issues/10824)) ([5d2a655](https://github.com/JSONbored/metagraphed/commit/5d2a65555c80a01dab0bd197da03b2b975d917ee)), closes [#10772](https://github.com/JSONbored/metagraphed/issues/10772)
* **history:** serve the days below Neon's floor from the lakehouse ([#10799](https://github.com/JSONbored/metagraphed/issues/10799)) ([9948910](https://github.com/JSONbored/metagraphed/commit/9948910517539bd83cdd154eea66aa5050c159e8))
* **mcp:** let the four paged tools pass the page their routes publish ([#10779](https://github.com/JSONbored/metagraphed/issues/10779)) ([50994c1](https://github.com/JSONbored/metagraphed/commit/50994c140b9d69d56e298eed7754711d31fe3c77))
* **mcp:** parse every tool result against the schema that tool publishes ([#10860](https://github.com/JSONbored/metagraphed/issues/10860)) ([4240b36](https://github.com/JSONbored/metagraphed/commit/4240b3672f68965342b2ea34164fcafce25a270e)), closes [#10789](https://github.com/JSONbored/metagraphed/issues/10789)
* **mcp:** the last 21 published parameters a tool could not pass ([#10803](https://github.com/JSONbored/metagraphed/issues/10803)) ([afa608c](https://github.com/JSONbored/metagraphed/commit/afa608cdcf6037f27e3a0645c7b3cd55217eb422)), closes [#10793](https://github.com/JSONbored/metagraphed/issues/10793)
* **neon:** give the subnet-ownership lane the two tables it writes to ([#10821](https://github.com/JSONbored/metagraphed/issues/10821)) ([73c87f1](https://github.com/JSONbored/metagraphed/commit/73c87f17be5a2d764f56a1bbed4abb2ec582ebd4))
* **neon:** store the shares the chain gives us, and normalise in SQL ([#10837](https://github.com/JSONbored/metagraphed/issues/10837)) ([781aa53](https://github.com/JSONbored/metagraphed/commit/781aa53e9bfb240ca6d95275e67a9b2c235c9019))
* **queues:** finish the migration — origin checks and revenue probes ([#10715](https://github.com/JSONbored/metagraphed/issues/10715)) ([#10727](https://github.com/JSONbored/metagraphed/issues/10727)) ([71e6d02](https://github.com/JSONbored/metagraphed/commit/71e6d02f93fdb0a96babdea7880404e7bf4a36c9))
* **registry:** treasury readings — what a subnet's own source declares ([#11025](https://github.com/JSONbored/metagraphed/issues/11025)) ([7cd0ab7](https://github.com/JSONbored/metagraphed/commit/7cd0ab76902a7598590f51a342da89d42205d7f4)), closes [#10933](https://github.com/JSONbored/metagraphed/issues/10933)
* **revenue:** the owner-cut disposition reads the flow streams it names ([#11004](https://github.com/JSONbored/metagraphed/issues/11004)) ([0c6a4eb](https://github.com/JSONbored/metagraphed/commit/0c6a4eb0133434030756034024e5e2fbbfe1b75a)), closes [#10930](https://github.com/JSONbored/metagraphed/issues/10930)
* **schemas:** a published schema may not accept what it does not describe ([#10853](https://github.com/JSONbored/metagraphed/issues/10853)) ([0e3bbd7](https://github.com/JSONbored/metagraphed/commit/0e3bbd7553fea1a7cdc0e442c8333cdae21077dc))
* **schemas:** a spread must be typed or declared -- the nullability walk's blind spot becomes a rule ([#10875](https://github.com/JSONbored/metagraphed/issues/10875)) ([93bd8ec](https://github.com/JSONbored/metagraphed/commit/93bd8ec91a558a51a33306b298e2cf8713b67e6a))
* **schemas:** census the producer-side fallbacks the contract does not state ([#10877](https://github.com/JSONbored/metagraphed/issues/10877)) ([18fe5e7](https://github.com/JSONbored/metagraphed/commit/18fe5e7fdafd632602f5a845ac09c84a2d1afacc))
* **schemas:** one canonical input schema, three codecs, one table ([#10844](https://github.com/JSONbored/metagraphed/issues/10844)) ([c4363e8](https://github.com/JSONbored/metagraphed/commit/c4363e8fca50bf3976682531dcdc37a504984d61)), closes [#10787](https://github.com/JSONbored/metagraphed/issues/10787)
* **schemas:** one operation registry, and the surface tables become projections of it ([#10828](https://github.com/JSONbored/metagraphed/issues/10828)) ([1d73a17](https://github.com/JSONbored/metagraphed/commit/1d73a17d661148f3d8dcb8414701475b6a127183)), closes [#10781](https://github.com/JSONbored/metagraphed/issues/10781)
* **schemas:** the compiler as the nullability oracle, and the seventeen it found ([#10841](https://github.com/JSONbored/metagraphed/issues/10841)) ([1dd6bba](https://github.com/JSONbored/metagraphed/commit/1dd6bbac63331679dc296b3aca1a5cc96bcacc61))
* **sync:** give nominator_positions a prune domain, so self-stake can share it ([#10848](https://github.com/JSONbored/metagraphed/issues/10848)) ([88d9dff](https://github.com/JSONbored/metagraphed/commit/88d9dff6bd692e5a356795eaef530c85f481fa56))
* **sync:** route subnet-ownership through Hyperdrive, and stop the identity history rewriting its own timestamps ([#10839](https://github.com/JSONbored/metagraphed/issues/10839)) ([40a8e62](https://github.com/JSONbored/metagraphed/commit/40a8e62f1690763f254ed4a2d25ee2562647a5ec))
* **types:** the census of hand-written duplicates, and the rule that keeps it at zero ([#10829](https://github.com/JSONbored/metagraphed/issues/10829)) ([e6510ae](https://github.com/JSONbored/metagraphed/commit/e6510ae902ace072adc19c2c72de2c31ae9a4f92)), closes [#10784](https://github.com/JSONbored/metagraphed/issues/10784)

### Bug Fixes

* **alarm:** a dead-letter lane names what was lost, not a cadence it does not have ([#10980](https://github.com/JSONbored/metagraphed/issues/10980)) ([a1d5e5f](https://github.com/JSONbored/metagraphed/commit/a1d5e5fab463b6892ef9adbd7afb26da8de21ced))
* **alarm:** a lane with no declared producer gets no floor, and alarms early ([#10991](https://github.com/JSONbored/metagraphed/issues/10991)) ([cbf151e](https://github.com/JSONbored/metagraphed/commit/cbf151e1d118ffc2d91ffe43de8075668ab5a947))
* **api:** a failed cache write is a miss, not an incident ([#11038](https://github.com/JSONbored/metagraphed/issues/11038)) ([4e58da0](https://github.com/JSONbored/metagraphed/commit/4e58da0f675645d11f9757efd188931b02a20983)), closes [#11022](https://github.com/JSONbored/metagraphed/issues/11022)
* **api:** a projected response is shorter on purpose, so absence is not drift ([#10978](https://github.com/JSONbored/metagraphed/issues/10978)) ([92ce2e2](https://github.com/JSONbored/metagraphed/commit/92ce2e20c3027459c983671df723a773ef9ce9be))
* **api:** an unresolved extrinsic is not settled, and was advertising itself as storable ([#11016](https://github.com/JSONbored/metagraphed/issues/11016)) ([138e4ca](https://github.com/JSONbored/metagraphed/commit/138e4ca4be5683ae7379a21444b4b94a1b4e8b2f))
* **api:** declare the per-row USD twins the economics overlay stamps ([#10935](https://github.com/JSONbored/metagraphed/issues/10935)) ([05d2312](https://github.com/JSONbored/metagraphed/commit/05d23124a2ac3458e55d77a557f09d8c25a7bb67))
* **api:** declare the serve-time stamps the .strict() flip refused, and make the refusal alarm ([#10898](https://github.com/JSONbored/metagraphed/issues/10898)) ([815a2de](https://github.com/JSONbored/metagraphed/commit/815a2de04414ec5177f73f38d9311b600c20bbd1))
* **api:** the account family had no rate limit at all ([#11026](https://github.com/JSONbored/metagraphed/issues/11026)) ([2b92816](https://github.com/JSONbored/metagraphed/commit/2b92816fcb76cbd2a6f36febc4ada0cc5a6587fe)), closes [#11017](https://github.com/JSONbored/metagraphed/issues/11017)
* **api:** the tripwire could not see 109 routes, and what it could not see had drifted ([#10985](https://github.com/JSONbored/metagraphed/issues/10985)) ([a3f5954](https://github.com/JSONbored/metagraphed/commit/a3f5954e87d633bfdd7874518964a46343b2a1de)), closes [#10965](https://github.com/JSONbored/metagraphed/issues/10965)
* **api:** the tripwires validate what is sent, not what the handler built ([#10976](https://github.com/JSONbored/metagraphed/issues/10976)) ([35d9c77](https://github.com/JSONbored/metagraphed/commit/35d9c7709242a00c9b46176356ef82da97ac9b52))
* **chain-firehose:** name which cap refused, without telling the caller ([#10945](https://github.com/JSONbored/metagraphed/issues/10945)) ([6f52367](https://github.com/JSONbored/metagraphed/commit/6f5236782d9a6b2c7a220979c437e5a6a754b7ad)), refs [#10606](https://github.com/JSONbored/metagraphed/issues/10606)
* **ci:** make the posthog CLI's CDN download unable to fail an install ([#10922](https://github.com/JSONbored/metagraphed/issues/10922)) ([bb88884](https://github.com/JSONbored/metagraphed/commit/bb88884af48559ea84ae2bff1bae9d692a9cd70d))
* **crons:** delete three data-api crons whose handlers were deleted with D1 ([#10823](https://github.com/JSONbored/metagraphed/issues/10823)) ([384a4d9](https://github.com/JSONbored/metagraphed/commit/384a4d9ed4f6eb065c24145768dcd60ba50a0add))
* **crons:** make the api Worker's */15 branch explicit, so an unknown cron stops running a producer ([#10827](https://github.com/JSONbored/metagraphed/issues/10827)) ([bd7d795](https://github.com/JSONbored/metagraphed/commit/bd7d795a4b7e811d9ea0091492dede247a5b98b3))
* **e2e:** re-record the fixtures the p50 rename left behind ([#11035](https://github.com/JSONbored/metagraphed/issues/11035)) ([0f3d824](https://github.com/JSONbored/metagraphed/commit/0f3d8244d2c8a7ecf9ca1aba23e41ce840c4ef95)), closes [#11034](https://github.com/JSONbored/metagraphed/issues/11034)
* **e2e:** the overflow sweep stops reading live production ([#10968](https://github.com/JSONbored/metagraphed/issues/10968)) ([f639887](https://github.com/JSONbored/metagraphed/commit/f639887161d9a2eeee5ad02c67ec1f9231e719d6))
* **emission-gate:** pin a sample to one finalized block, and rotate whole samples ([#10742](https://github.com/JSONbored/metagraphed/issues/10742)) ([#10743](https://github.com/JSONbored/metagraphed/issues/10743)) ([9b728bc](https://github.com/JSONbored/metagraphed/commit/9b728bcbc23979fedaf3291b842fb0c0e309f6ff))
* **firehose:** a deploy should not 500 a subscriber, and a crawler should not hold a stream ([#11036](https://github.com/JSONbored/metagraphed/issues/11036)) ([762ecc5](https://github.com/JSONbored/metagraphed/commit/762ecc514c85c810ce574b5aba2520b5589a4143))
* **freshness:** confirm a redirected stale before alarming, and read the stamp gap in the direction that is actually invariant ([#10918](https://github.com/JSONbored/metagraphed/issues/10918)) ([e1bd435](https://github.com/JSONbored/metagraphed/commit/e1bd435ab9cdb92a6d4f7b788ce63c1e9f2a1590))
* **graphql:** say that postgresTierRequest takes a query string, not a row ([#10792](https://github.com/JSONbored/metagraphed/issues/10792)) ([1e82ffb](https://github.com/JSONbored/metagraphed/commit/1e82ffbc947e4b420c3af7e2529686427d81de53)), closes [#10782](https://github.com/JSONbored/metagraphed/issues/10782)
* **graphql:** sync subnet_endpoints' SDL description to the registry ([#10847](https://github.com/JSONbored/metagraphed/issues/10847)) ([c8ea118](https://github.com/JSONbored/metagraphed/commit/c8ea11829f9b012e46e20f526caa34d2c868c1b2))
* **icon-proxy:** record negatives server-side, not in a Cache-Control header ([#11031](https://github.com/JSONbored/metagraphed/issues/11031)) ([de04c99](https://github.com/JSONbored/metagraphed/commit/de04c99aae949271813c7616127033bfcaf5e877))
* **identity:** build the subnet-identity sync route, the writer that never existed ([#10740](https://github.com/JSONbored/metagraphed/issues/10740)) ([2a943f5](https://github.com/JSONbored/metagraphed/commit/2a943f5b55e854d69b3ec50b34e96efe40b6a862))
* **identity:** give subnet_identity_history a Neon table again ([#10728](https://github.com/JSONbored/metagraphed/issues/10728)) ([1d8cdf4](https://github.com/JSONbored/metagraphed/commit/1d8cdf4a4d9b6a7d65ac7f4758ddf159dc28869c))
* **identity:** give the change diff a real baseline, and stop the flag comment lying about it ([#10915](https://github.com/JSONbored/metagraphed/issues/10915)) ([ae37d98](https://github.com/JSONbored/metagraphed/commit/ae37d987795c5a47156f8105946ccd081589234d))
* **identity:** point the subnet-identity read tier at Neon ([#10771](https://github.com/JSONbored/metagraphed/issues/10771)) ([c58a94d](https://github.com/JSONbored/metagraphed/commit/c58a94dc6ed620801878600f0afa2bce008a595e))
* **identity:** repoint the identity timelines at the live store — the writer came back and the readers did not ([#10774](https://github.com/JSONbored/metagraphed/issues/10774)) ([3101049](https://github.com/JSONbored/metagraphed/commit/31010499c0bee2023e7df888ed797171c14c1fd4)), closes [#10773](https://github.com/JSONbored/metagraphed/issues/10773)
* **identity:** take captured_at on the wire, as every sibling sync route does ([#10762](https://github.com/JSONbored/metagraphed/issues/10762)) ([0ac1c3e](https://github.com/JSONbored/metagraphed/commit/0ac1c3ea1036a35a77019bc6bc0ae7e98b99c29d))
* **lakehouse:** the catalog is validated at runtime, not cast to ([#11008](https://github.com/JSONbored/metagraphed/issues/11008)) ([1d42f67](https://github.com/JSONbored/metagraphed/commit/1d42f67c2a8b284718b05e799e6d50c087354581))
* **mcp:** complete the PostHog $mcp_* wire contract and label every caller ([#10885](https://github.com/JSONbored/metagraphed/issues/10885)) ([63921ce](https://github.com/JSONbored/metagraphed/commit/63921ce27121ad01ae54bdd70d1f003bbfce75d8))
* **mcp:** derive get_subnet_health's output from the route artifact, and let the drift alarm carry its diagnosis ([#10914](https://github.com/JSONbored/metagraphed/issues/10914)) ([81ffb3f](https://github.com/JSONbored/metagraphed/commit/81ffb3f7a991f30f9f3f296fe6216fe5e648bb24))
* **mcp:** get_subnet_health serves the route's own row schema, not a copy ([#10905](https://github.com/JSONbored/metagraphed/issues/10905)) ([0c5c941](https://github.com/JSONbored/metagraphed/commit/0c5c941aff67c2285931ba7e6151a5451f4fb2f4))
* **mcp:** give the shared loaders the data REST gives them, and make omission a compile error ([#10957](https://github.com/JSONbored/metagraphed/issues/10957)) ([f353416](https://github.com/JSONbored/metagraphed/commit/f35341602c9b1d07cacc811bd9f33f1f0e789481)), closes [#10926](https://github.com/JSONbored/metagraphed/issues/10926)
* **mcp:** pass the live lanes to /freshness, so two healthy ones stop reading missing ([#10822](https://github.com/JSONbored/metagraphed/issues/10822)) ([b3026f9](https://github.com/JSONbored/metagraphed/commit/b3026f9174b5eeb0b612ff359ba53cdd527a316a)), closes [#10816](https://github.com/JSONbored/metagraphed/issues/10816)
* **mcp:** restore the drift alarm's diagnosis, and pin it so it cannot vanish again ([#10979](https://github.com/JSONbored/metagraphed/issues/10979)) ([99da63f](https://github.com/JSONbored/metagraphed/commit/99da63f6cdba83c0ed2a76fc77b22cbdb7fd7aa2))
* **monitor:** a tripwire refusal is drift wearing a 500, not an outage ([#10988](https://github.com/JSONbored/metagraphed/issues/10988)) ([0ae7518](https://github.com/JSONbored/metagraphed/commit/0ae7518e8fda5a1301f058cb1a6217fa260ea0fb))
* **neon:** cap what the write buffer may hold, and send the rest direct ([#10745](https://github.com/JSONbored/metagraphed/issues/10745)) ([bc8b70e](https://github.com/JSONbored/metagraphed/commit/bc8b70e7d15d4a1d37754183a70f83984969534c)), closes [#10744](https://github.com/JSONbored/metagraphed/issues/10744)
* **neon:** disable the write buffer -- a third failure, and the alarm is the cause ([#10750](https://github.com/JSONbored/metagraphed/issues/10750)) ([49c2f13](https://github.com/JSONbored/metagraphed/commit/49c2f133ef9c0c3755c04ca32e97b7f32398bd5a))
* **neon:** file a flush's lane verdict under the key the write path uses ([#10852](https://github.com/JSONbored/metagraphed/issues/10852)) ([00a0503](https://github.com/JSONbored/metagraphed/commit/00a0503297d8c36d859ee1e283b4170eeebd5ec0)), closes [#10851](https://github.com/JSONbored/metagraphed/issues/10851)
* **neon:** let a buffered lane's -pass/-prune verdict be cleared ([#10833](https://github.com/JSONbored/metagraphed/issues/10833)) ([11312b7](https://github.com/JSONbored/metagraphed/commit/11312b7827c3a3070870d510ca1abb2c95d70214)), closes [#10830](https://github.com/JSONbored/metagraphed/issues/10830)
* **neon:** make the buffer's enqueue O(1) -- the scan was resetting the object ([#10756](https://github.com/JSONbored/metagraphed/issues/10756)) ([e46a722](https://github.com/JSONbored/metagraphed/commit/e46a722debe81b6850c7f284045b82ffea785a5e)), closes [#10755](https://github.com/JSONbored/metagraphed/issues/10755)
* **neon:** replace an orphaned alarm instead of respecting it ([#10764](https://github.com/JSONbored/metagraphed/issues/10764)) ([014054a](https://github.com/JSONbored/metagraphed/commit/014054ab62f8d416f87c82089642c6b78bad1ac3)), closes [#10763](https://github.com/JSONbored/metagraphed/issues/10763)
* **neon:** retry an enqueue when a deploy resets the Durable Object ([#10730](https://github.com/JSONbored/metagraphed/issues/10730)) ([635652c](https://github.com/JSONbored/metagraphed/commit/635652c42a0be6491509cf8089fd839b9dc618e1))
* **neon:** stop the flush's reconcile from scanning the backlog ([#10776](https://github.com/JSONbored/metagraphed/issues/10776)) ([8dd047a](https://github.com/JSONbored/metagraphed/commit/8dd047a7d9a9d7b7ddcd92fde6ff15251c70b163))
* **neon:** turn the write buffer off -- it is failing inside DO storage ([#10734](https://github.com/JSONbored/metagraphed/issues/10734)) ([b696e6d](https://github.com/JSONbored/metagraphed/commit/b696e6d0794f8bf57dfc6eeb56f90f53eae22b30))
* **observability:** file the buffered snapshot's sub-lane verdicts at enqueue ([#10890](https://github.com/JSONbored/metagraphed/issues/10890)) ([13c60f2](https://github.com/JSONbored/metagraphed/commit/13c60f2b34f36a05c659633fc513425bb0250aaa))
* **observability:** record the prune/tally sub-lane verdicts at enqueue, and retire the bare raw-capture-state key ([#10908](https://github.com/JSONbored/metagraphed/issues/10908)) ([7d4a039](https://github.com/JSONbored/metagraphed/commit/7d4a039554f997aead1b5e778f9529b1aacd1cce))
* **observability:** retire two more [#10851](https://github.com/JSONbored/metagraphed/issues/10851) fossils, and declare self-stake's cadence instead of inferring it ([#10936](https://github.com/JSONbored/metagraphed/issues/10936)) ([a2c6e5c](https://github.com/JSONbored/metagraphed/commit/a2c6e5cd7003bab55495f1b6c2daefe5a5e82429))
* **observability:** the four live fault classes the PostHog sweep found ([#10825](https://github.com/JSONbored/metagraphed/issues/10825)) ([039a65c](https://github.com/JSONbored/metagraphed/commit/039a65c1e3f490379d6c25323943888b3d9d5700)), closes [#10813](https://github.com/JSONbored/metagraphed/issues/10813)
* **observability:** the stake-quote drift refusal files the fault too ([#10902](https://github.com/JSONbored/metagraphed/issues/10902)) ([9f41aef](https://github.com/JSONbored/metagraphed/commit/9f41aef19bc441ad65f49f25a96605546d52f6e7)), closes [#10901](https://github.com/JSONbored/metagraphed/issues/10901)
* **queues:** a lane that can say it ran, why it retried, and what it lost ([#10778](https://github.com/JSONbored/metagraphed/issues/10778)) ([7158691](https://github.com/JSONbored/metagraphed/commit/715869179858dc002af70a6688c59a7508206fbf))
* **r2-sql:** break on a repeated upstream 5xx, so one outage costs one exception ([#10741](https://github.com/JSONbored/metagraphed/issues/10741)) ([#10747](https://github.com/JSONbored/metagraphed/issues/10747)) ([0207a58](https://github.com/JSONbored/metagraphed/commit/0207a58c88afb278a45d6dee9b208bbc4f5662b7))
* **registry:** the link-rot lane required an uptime config to check a link ([#11032](https://github.com/JSONbored/metagraphed/issues/11032)) ([6ea79a0](https://github.com/JSONbored/metagraphed/commit/6ea79a037a6130e91906caf89e349c9092be28ed)), closes [#11007](https://github.com/JSONbored/metagraphed/issues/11007)
* **revenue:** ack a probe failure a retry cannot change ([#10857](https://github.com/JSONbored/metagraphed/issues/10857)) ([5c90c8e](https://github.com/JSONbored/metagraphed/commit/5c90c8ea5defe01fc5225b01ca0eb047df5444ec)), closes [#10856](https://github.com/JSONbored/metagraphed/issues/10856)
* **revenue:** eligible means extractable, not merely annotated ([#10785](https://github.com/JSONbored/metagraphed/issues/10785)) ([b468647](https://github.com/JSONbored/metagraphed/commit/b468647ab94f9b7339f5ba3ad0fdb735f9be8d69)), closes [#10783](https://github.com/JSONbored/metagraphed/issues/10783)
* **revenue:** let a declaration say WHERE its rows live ([#10887](https://github.com/JSONbored/metagraphed/issues/10887)) ([7e0c7e6](https://github.com/JSONbored/metagraphed/commit/7e0c7e614e8a1b60fede84200964858eac548a2e))
* **revenue:** price the emission through the live TAO/USD index, and decline instead of publishing a zero rate ([#10920](https://github.com/JSONbored/metagraphed/issues/10920)) ([abceb86](https://github.com/JSONbored/metagraphed/commit/abceb86b38d82b214fe57f8bd6935ab77366e081))
* **schema:** four hand-rolled copies collapsed onto the vocabulary they restate ([#10986](https://github.com/JSONbored/metagraphed/issues/10986)) ([2f09910](https://github.com/JSONbored/metagraphed/commit/2f099109d9be241967d135b5f57a0ecbf1aedea2))
* **schema:** the concentration scorecard is one vocabulary, not two ([#10982](https://github.com/JSONbored/metagraphed/issues/10982)) ([379d3af](https://github.com/JSONbored/metagraphed/commit/379d3af1afcb5fe36dac2c0cf5f7bb485475ef3f))
* **seo:** withhold the three unbounded per-entity spaces from crawlers, on both hosts ([#11005](https://github.com/JSONbored/metagraphed/issues/11005)) ([8787848](https://github.com/JSONbored/metagraphed/commit/8787848f90d3c8506ee6f5473f5b4eb9e87a8e0a)), closes [#11002](https://github.com/JSONbored/metagraphed/issues/11002)
* **telemetry:** assert the capture invariants on every recorder, and fix the three it caught ([#10949](https://github.com/JSONbored/metagraphed/issues/10949)) ([2d56aeb](https://github.com/JSONbored/metagraphed/commit/2d56aeba617bf7865b0800f4291b884dae5a999c)), refs [#10606](https://github.com/JSONbored/metagraphed/issues/10606)
* **telemetry:** count a calling Worker as a caller, not as the shared fallback ([#10953](https://github.com/JSONbored/metagraphed/issues/10953)) ([eecf47f](https://github.com/JSONbored/metagraphed/commit/eecf47ffbc88900332ca431d9cfc8e506834c3f7))
* **telemetry:** count REST callers, and stop one unflagged event minting a person per session ([#10942](https://github.com/JSONbored/metagraphed/issues/10942)) ([4c77918](https://github.com/JSONbored/metagraphed/commit/4c7791839ea60c41335d8f4c4db37af0c9c0125b)), refs [#10606](https://github.com/JSONbored/metagraphed/issues/10606)
* **telemetry:** count the caller, not the connection -- and close the last two profile leaks ([#10944](https://github.com/JSONbored/metagraphed/issues/10944)) ([fddac34](https://github.com/JSONbored/metagraphed/commit/fddac34f27f99f62b62fda52fe565825e622926b)), refs [#10606](https://github.com/JSONbored/metagraphed/issues/10606)
* **tiers:** METAGRAPH_SUBNET_IDENTITY_SOURCE is unread -- make the tree say only that ([#10899](https://github.com/JSONbored/metagraphed/issues/10899)) ([ad0c284](https://github.com/JSONbored/metagraphed/commit/ad0c2848cf56f9529e64f292a1365b08af7db239)), closes [#10893](https://github.com/JSONbored/metagraphed/issues/10893)
* **types:** generate wss-lb's Env types, and make the roster check itself ([#10862](https://github.com/JSONbored/metagraphed/issues/10862)) ([817950e](https://github.com/JSONbored/metagraphed/commit/817950e21884e40f9adf64f4c5c86b1e08262ec5)), closes [#10861](https://github.com/JSONbored/metagraphed/issues/10861)
* **ui:** /revenue was in the sitemap and nowhere else ([#10977](https://github.com/JSONbored/metagraphed/issues/10977)) ([f9ecad2](https://github.com/JSONbored/metagraphed/commit/f9ecad2fdbab9ae7896e0e6e4e49baa1a7f3af59)), closes [#10927](https://github.com/JSONbored/metagraphed/issues/10927)
* **ui:** a rate-limited read is throttled, not broken ([#11012](https://github.com/JSONbored/metagraphed/issues/11012)) ([ffa5f9a](https://github.com/JSONbored/metagraphed/commit/ffa5f9a7629469f68911cb168ae8828eea261fbe)), closes [#11000](https://github.com/JSONbored/metagraphed/issues/11000)
* **ui:** one chain-stream socket per tab, with a grow-only topic union ([#10961](https://github.com/JSONbored/metagraphed/issues/10961)) ([2e2b512](https://github.com/JSONbored/metagraphed/commit/2e2b512af7f46a37ab734f50124446eedce3f3c9)), refs [#10606](https://github.com/JSONbored/metagraphed/issues/10606)
* **ui:** one chain-stream socket per topic set, not one per call site ([#10952](https://github.com/JSONbored/metagraphed/issues/10952)) ([4ab50f4](https://github.com/JSONbored/metagraphed/commit/4ab50f4adc3d90557e327788b8aa9a15d9e0b7d7)), refs [#10606](https://github.com/JSONbored/metagraphed/issues/10606)
* **ui:** stop telling users their data comes from a database we deleted ([#10934](https://github.com/JSONbored/metagraphed/issues/10934)) ([d079c84](https://github.com/JSONbored/metagraphed/commit/d079c84e02fd7edc3679c874833185194b74b03b)), closes [#10911](https://github.com/JSONbored/metagraphed/issues/10911)
* **watchdog:** judge lane silence by the max gap and the declared cadence ([#10726](https://github.com/JSONbored/metagraphed/issues/10726)) ([7326b60](https://github.com/JSONbored/metagraphed/commit/7326b60373192a77531d31143195c25a52f310cb)), closes [#10723](https://github.com/JSONbored/metagraphed/issues/10723)
* **watchdogs:** the SafeMode monitor was blind for a week, and three alarms that could not name their own cause ([#10767](https://github.com/JSONbored/metagraphed/issues/10767)) ([b304402](https://github.com/JSONbored/metagraphed/commit/b304402b588499e077c5bc3d57cdffa0f42547a4))
* **watchdog:** take the freshness census from Neon's schema, not the frozen D1 fixture ([#10826](https://github.com/JSONbored/metagraphed/issues/10826)) ([cc3e02c](https://github.com/JSONbored/metagraphed/commit/cc3e02cddd27c814ba4c971deb7ca8977352e32d)), closes [#10817](https://github.com/JSONbored/metagraphed/issues/10817)

### Performance

* **api:** chain-events TTL follows the tier and the scope, not a constant ([#11014](https://github.com/JSONbored/metagraphed/issues/11014)) ([171c37a](https://github.com/JSONbored/metagraphed/commit/171c37a3c15bdf0460ffc3341bfdf3080423dbaa)), closes [#11003](https://github.com/JSONbored/metagraphed/issues/11003)
* **api:** serve block and extrinsic detail from the edge cache ([#11010](https://github.com/JSONbored/metagraphed/issues/11010)) ([7294e60](https://github.com/JSONbored/metagraphed/commit/7294e604c5f358490436f6b02f3e55e3c0889dd0)), closes [#11001](https://github.com/JSONbored/metagraphed/issues/11001)
* **api:** serve health from the snapshot-stamped overlay cache ([#11033](https://github.com/JSONbored/metagraphed/issues/11033)) ([cfd7341](https://github.com/JSONbored/metagraphed/commit/cfd7341d00b88e20f0b68b2e6547a41d76e46a1c)), closes [#11023](https://github.com/JSONbored/metagraphed/issues/11023)
* **api:** the two block sub-resources [#11010](https://github.com/JSONbored/metagraphed/issues/11010) left behind ([#11028](https://github.com/JSONbored/metagraphed/issues/11028)) ([65eb3a8](https://github.com/JSONbored/metagraphed/commit/65eb3a8d1227384f440e2a92f97db7e9282b6b9f)), closes [#11018](https://github.com/JSONbored/metagraphed/issues/11018)
* **neon:** re-enable the write buffer now that oversized statements bypass it ([#10746](https://github.com/JSONbored/metagraphed/issues/10746)) ([f92ac63](https://github.com/JSONbored/metagraphed/commit/f92ac6363a44f1618fb0d0415d082a3851a6d8a8))
* **neon:** turn the write buffer on, with the O(n) enqueue fixed ([#10758](https://github.com/JSONbored/metagraphed/issues/10758)) ([0439f6a](https://github.com/JSONbored/metagraphed/commit/0439f6a9d80e6401c97fa3dc76827f572263a270))
* **startup:** cut the static edge that re-pinned both mega-modules, and gate the budget ([#10903](https://github.com/JSONbored/metagraphed/issues/10903)) ([6240045](https://github.com/JSONbored/metagraphed/commit/624004500a29a2ef8d6a6f9e1f36467ec8202ff3))

### Documentation

* **contract:** required describes the unprojected response, and the contract now says so ([#11009](https://github.com/JSONbored/metagraphed/issues/11009)) ([7fc1398](https://github.com/JSONbored/metagraphed/commit/7fc1398a2c87446868df7454d7b9f80a4eaa4bbd)), closes [#10960](https://github.com/JSONbored/metagraphed/issues/10960)
* **gates:** record the .refine() measurement -- the movable cross-argument set is empty ([#10883](https://github.com/JSONbored/metagraphed/issues/10883)) ([f7d9291](https://github.com/JSONbored/metagraphed/commit/f7d9291964b931f1c7bf455a515b75ed51bd6a50))
* **method:** publish the attribution rules, and make the schema enforce them ([#10962](https://github.com/JSONbored/metagraphed/issues/10962)) ([be364d3](https://github.com/JSONbored/metagraphed/commit/be364d38aae0e7b4c962a23a38071ceb7b214c6c))
* **neon:** the dual-write era's comments catch up with its deletion ([#10912](https://github.com/JSONbored/metagraphed/issues/10912)) ([53d7d10](https://github.com/JSONbored/metagraphed/commit/53d7d1002a470ccf28d1778ca5ec7b1f27ee4286))
* **schemas:** record the fallback census triage -- the classes decided, not re-decidable ([#10880](https://github.com/JSONbored/metagraphed/issues/10880)) ([44aaf79](https://github.com/JSONbored/metagraphed/commit/44aaf7937d749ccd89316e54c5fb8541356a8ef9))
* **schemas:** record why the codec table is data, not .transform() closures ([#10879](https://github.com/JSONbored/metagraphed/issues/10879)) ([c7cebd3](https://github.com/JSONbored/metagraphed/commit/c7cebd3a97d819fe261e35dda8a1e5fc25a47d48))

### Refactors

* **api:** one name for the 50th percentile across the contract ([#10989](https://github.com/JSONbored/metagraphed/issues/10989)) ([8d4271a](https://github.com/JSONbored/metagraphed/commit/8d4271ab1887c2b0f0641c6c04f5093721a00a13))
</details>

---


JSONbored:
🤖 Created releases:

- [ui-v1.0.0](https://github.com/JSONbored/metagraphed/releases/tag/ui-v1.0.0)
- [platform-v2.0.0](https://github.com/JSONbored/metagraphed/releases/tag/platform-v2.0.0)

:sunflower:

PR #11154, opened 2026-08-14
fix(api): count the decline the labeller cannot see

### LINKED ISSUE #11153: fix(api): ten paginated routes serve a declined page as end-of-feed
## What happens

A cold-tier read declines a too-deep offset and the route answers `200` with a schema-stable empty — `count: 0`, `next_cursor: null` — which is **byte-identical to end-of-feed**, and edge-cacheable. A client paginating the feed stops there and silently drops the tail.

Found by driving every registered route at a depth past `OFFSET_EMULATION_CAP`:

```
/api/v1/validators/{hotkey}/nominators
/api/v1/subnets/{netuid}/hyperparameters/history
/api/v1/subnets/{netuid}/events
/api/v1/subnets/{netuid}/identity-history
/api/v1/accounts/{ss58}/events
/api/v1/accounts/{ss58}/extrinsics
/api/v1/accounts/{ss58}/transfers
/api/v1/accounts/{ss58}/identity-history
/api/v1/blocks/{ref}/extrinsics
/api/v1/blocks/{ref}/events
```

Ten routes, and exactly ten `if (offset > OFFSET_EMULATION_CAP) return null;` sites across the cold-tier readers — a 1:1 mapping.

## Why the existing guard is blind to it

`handleRequest` labels a degraded answer by snapshotting failure generations around the dispatch, deliberately from outside so the label is "unforgettable rather than 40 more places to remember". A query that *fails* moves `currentR2SqlFailureGeneration()` and gets labelled automatically.

The offset cap is checked **before any SQL is built**. No request is sent, nothing fails, no generation moves — so the response ships unlabelled. `tests/degraded-answer-is-labelled.test.ts`'s whole-route-table sweep keys off those same generations and is therefore structurally incapable of seeing this class.

#11149 fixed the extrinsics feed by wrapping that one route's fallback in `unmeasured`. Correct locally, wrong in shape: there are **92** `?? build…([])` operands in the handlers, and per-route wrapping is the thing this labelling architecture exists to avoid.

## Fix

Count the decline at the check, not at the call sites. `offsetBeyondEmulationCap(offset)` replaces the bare comparison at all ten reader sites and records it; `degradedSnapshot` carries it as a fourth counter and `degradedSince` folds it into **unmeasured** — not transient, since the same offset declines identically for the whole TTL, so the answer stays cacheable and merely stops claiming to be measured.

A reader added later reports its decline without its author knowing the labelling exists.

## Guard

A second sweep using **depth** as the discriminator: drive every registered route twice, and if it queries the lakehouse at a shallow offset but goes silent at a deep one, it declined without querying — and that answer must be labelled. No route list to maintain; a route that never pages the lakehouse is skipped by its own behaviour.

JSONbored:
## The defect, found by sweep rather than by guess

Ten paginated routes answered a **declined** page as a bare, edge-cacheable `200` whose body is byte-identical to end-of-feed (`count: 0`, `next_cursor: null`). A client paginating the feed stops there and silently drops the tail.

```
/api/v1/validators/{hotkey}/nominators      /api/v1/accounts/{ss58}/events
/api/v1/subnets/{netuid}/hyperparameters/history   /api/v1/accounts/{ss58}/extrinsics
/api/v1/subnets/{netuid}/events             /api/v1/accounts/{ss58}/transfers
/api/v1/subnets/{netuid}/identity-history   /api/v1/accounts/{ss58}/identity-history
/api/v1/blocks/{ref}/extrinsics             /api/v1/blocks/{ref}/events
```

Ten routes; exactly ten `if (offset > OFFSET_EMULATION_CAP) return null;` sites in the cold-tier readers. A 1:1 mapping.

## Why the existing guard could not see it

`handleRequest` labels degradation by snapshotting failure generations around the dispatch — deliberately from outside, so the label is "unforgettable rather than 40 more places to remember". A query that *fails* moves `currentR2SqlFailureGeneration()` and is labelled automatically.

The offset cap is checked **before any SQL is built**: nothing is sent, nothing fails, no generation moves. The existing whole-route-table sweep keys off those same generations, so it is structurally incapable of catching this class.

## Why not extend #11149's approach

#11149 wrapped the extrinsics feed's fallback in `unmeasured`. Right locally, wrong in shape — there are **92** `?? build…([])` operands in the handlers, and per-route wrapping is exactly what this architecture exists to avoid.

So this counts the decline **at the check**. `offsetBeyondEmulationCap(offset)` replaces the bare comparison at all ten reader sites and records it; `degradedSnapshot` carries a fourth counter and `degradedSince` folds it into **unmeasured** — not transient, because the same offset declines identically for the whole TTL, so the answer stays cacheable and merely stops claiming to be measured. A reader added later reports its decline without its author knowing the labelling exists.

#11149's wraps stay: they cover pre-query declines this counter cannot see — the `call_hash` arm that skips the tier outright, and a filter `feedPredicates` refuses to express.

## The durable part

A second sweep using **depth** as the discriminator: drive every registered route twice, and if it queries at a shallow offset but goes silent at a deep one, it declined without querying and must be labelled. No route list to maintain — a route that never pages the lakehouse is skipped by its own behaviour, so a route added tomorrow is covered without editing the file.

## Verification

- The new sweep **fails without the wiring**, naming all ten routes; passes with it.
- Direct unit test on the primitive: a servable depth must not report a decline (or the marker means nothing), and each decline counts.
- **Patch coverage by diff intersection: 0 uncovered changed lines** across all 8 source files.
- 388 tests pass across the affected surface (cold tier, blocks, extrinsics, events, edge cache, analytics handlers); `tsc`, prettier, eslint clean; unreferenced-exports holds at 731; `validate-r2-sql-scan-bounds` green.

Closes #11153

### DIFF
--- a/src/account-feeds-cold-tier.ts
+++ b/src/account-feeds-cold-tier.ts
@@ -101,7 +101,7 @@ import { storeAll } from "./analytics-live.ts";
 import { decodeCursor, encodeCursor } from "./cursor.ts";
 import { r2SqlQuery, safeBlockNumber, safeSs58Literal } from "./r2-sql.ts";
 import type { R2SqlReader } from "./r2-sql.ts";
-import { OFFSET_EMULATION_CAP } from "./r2-sql-blocks.ts";
+import { offsetBeyondEmulationCap } from "./r2-sql-blocks.ts";
 import { ACCOUNT_EVENTS_COLUMNS } from "../generated/lakehouse/types.ts";
 import type { AccountEventsRow } from "../generated/lakehouse/types.ts";
 import { readStore } from "./read-store.ts";
@@ -178,7 +178,7 @@ export async function loadAccountTransfersColdTier(
   if (limit === null || offset === null || limit <= 0) return null;
   // R2 SQL has no OFFSET; past this depth the over-fetch stops being a
   // reasonable trade and declining beats serving a page that is quietly wrong.
-  if (offset > OFFSET_EMULATION_CAP) return null;
+  if (offsetBeyondEmulationCap(offset)) return null;
 
   // An unusable address is a decline, not an unfiltered scan of every account.
   const addr = safeSs58Literal(ss58);
@@ -700,7 +700,7 @@ export async function loadValidatorNominatorsColdTier(
   const limit = safeBlockNumber(query.limit);
   const offset = safeBlockNumber(query.offset ?? 0);
   if (limit === null || offset === null || limit <= 0) return null;
-  if (offset > OFFSET_EMULATION_CAP) return null;
+  if (offsetBeyondEmulationCap(offset)) return null;
 
   const sort = query.sort ?? DEFAULT_NOMINATOR_SORT;
   // A sort this tier cannot express would otherwise silently serve the default
--- a/src/account-identity-cold-tier.ts
+++ b/src/account-identity-cold-tier.ts
@@ -13,7 +13,7 @@ import { buildAccountIdentity, IDENTITY_FIELDS } from "./account-identity.ts";
 import { buildAccountIdentityHistory } from "./account-identity-history.ts";
 import { decodeCursor, encodeCursor } from "./cursor.ts";
 import { r2SqlQuery, safeBlockNumber, safeSs58Literal } from "./r2-sql.ts";
-import { OFFSET_EMULATION_CAP } from "./r2-sql-blocks.ts";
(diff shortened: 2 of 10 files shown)


JSONbored:
Superseded: main moved (#11141 touched the same cold-tier readers), so this was rebuilt on current main rather than rebased-and-force-pushed under an auto-merging gate. Reopening as a fresh PR with the conflict resolved and everything re-verified.

Issue #11182, opened 2026-08-14
epic(verification): every population expectation must be measured against the source, not against our own sink — four alarms in one night proved it


JSONbored:
Four lane alarms fired for hours on **correct data** on 2026-08-13/14. All four were reader bugs. Every producer was healthy. This epic is the standard that makes that class unrepresentable, and it is a prerequisite for the Pipelines adoption in #10850 — because a streaming ingest with the same verification habits reproduces the same failures with better branding.

## What actually happened

| lane | first diagnosis | actual cause | fix |
|---|---|---|---|
| `validator-nominator-counts` | truncated pass | expectation read off a table that never prunes | #11166 |
| `hotkey-alpha` | truncated pass | `covered` scoped to one pass, `referenced` to all history | #11170 |
| `nominator-positions` | producer dying mid-upload | **two producers share the table**; rule blind to `source` | #11180 |
| `self-stake` | producer dead | healthy; alerts predated the pass completing | — |

Ground truth, counted by walking the chain directly:

```
twox128("SubtensorModule") ++ twox128("Alpha"), state_getKeysPaged to exhaustion
-> 120,253 entries, 121 pages, final page 253 (short page = clean end)
-> 21,547 distinct hotkeys | 24,605 coldkeys | 17,292 (hotkey,netuid) pairs
-> 108,671 rows with netuid != 0, of which ZERO carry shares == 0
```

Our tables said 112,250 hotkeys and 35,073 pairs. **The gap between those two columns is the entire story.**

### S1 — an expectation is measured against the SOURCE
A population constant describes the chain, never our own table. `validator_nominator_counts`, `nominator_positions` and `hotkey_alpha` all accumulate; reading `COUNT(*)` or `COUNT(DISTINCT …)` off them reproduces a number that can only grow while the network shrinks. Record the measurement method in the constant's comment so the next re-anchor is reproducible rather than re-derived.

### S2 — a constant that can rot must say how to re-measure it
Every one of these was correct when written. A stale constant does not fail loudly, it fails *plausibly*. Each must carry the command that regenerates it.

### S3 — a rule must know its table's topology
Three properties changed the correct query and none were declared anywhere: does the table **prune**; does it have **more than one producer** (`nominator_positions.source`); is the unit **rows or keys**. A rule written without knowing these is guessing.

### S4 — both ends of a comparison span the same window
[...]

PR #11188, opened 2026-08-14
feat(watchdog): declare each lane table's topology, so a rule stops guessing

### LINKED ISSUE #11183: V1: declare each lane table's topology beside its schema, so a rule stops guessing
Sub-issue of #11182 (S3).

## The problem

Three properties decide what a correct coverage query looks like, and **none of them are declared anywhere**. Every watchdog re-derives them by reading the writer, and three got it wrong:

| property | why it changes the query | what it cost |
|---|---|---|
| **does it prune?** | an accumulating table's `COUNT(*)` is history, not population | #11166 — expectation read off 112,250 accumulated rows against 21,547 live hotkeys |
| **how many producers?** | `MAX(captured_at)` is whichever ran last | #11180 — a healthy `self-stake` run made the alpha scan look truncated |
| **rows or keys?** | coldkeys hold 1–N positions; row coverage can be high while most accounts are missed | already reasoned per-lane, nowhere stated |

## What to build

A declaration next to `db/schema.sql` — the same source that already drives the generated DB types — naming for each lane table:

- `prunes: true | false` (and if per-key, the key)
- `producers: [...]` and the column that discriminates them, if more than one
- `coverage_unit: rows | <key column>`
- `population_source`: where the expectation is measured (chain storage map, another table, a constant)

Then have the watchdogs **read it** rather than restate it. `nominator-positions` currently imports `POSITION_SOURCE_ALPHA` from the writer, which is the right instinct — this generalises that so the next rule inherits it instead of rediscovering it.

## Standards this must satisfy

- **Generated, not handrolled.** The declaration belongs with the schema and should produce types and a Zod schema the rules consume as fields — a rule that retypes a column name goes blind when the column moves (#11182 S6).
- **No second copy of a vocabulary.** Producer/source values come from the writer's constants (S5).
- **Non-vacuous.** A declaration nothing reads is documentation; the acceptance test is that removing a `source` discriminator from the declaration breaks the `nominator-positions` rule.

## Acceptance

- [ ] every lane table with a staleness watchdog carries a topology declaration
- [ ] each watchdog derives its scoping from it rather than from an inline literal
- [ ] a test proves a wrong declaration changes the rule's verdict (so it is load-bearing, not decorative)

JSONbored:
Closes #11183 (sub-issue of #11182).

## What was missing

Three properties decide the shape of a correct coverage query, and **none was declared anywhere**. Every watchdog re-derived them by reading its writer, and three of four got it wrong:

| property | what it cost |
|---|---|
| **does it prune?** | `validator_nominator_counts` never deletes — expectation read off 112,245 accumulated rows against 21,547 hotkeys on chain (#11166) |
| **how many producers?** | `nominator_positions` has two — a **healthy** `self-stake` run made the full alpha scan look truncated (#11180) |
| **rows or keys?** | coldkeys hold 1–N positions, so row coverage can look high having missed most accounts |

## Reuses the vocabularies that already exist

This is the part I got wrong first and corrected: `lanes` is **`ProducerLane`** from `producer-cadence.ts`, which already pairs each lane with its cadence. `fullScanValue` is the writer's own **`POSITION_SOURCE_ALPHA`**.

Those are two namings of one producer — the lane is `validator_nominators`, the value it *stamps* is `alpha` — so the declaration records the pairing rather than collapsing them into a third name. My first draft used `POSITION_SOURCE_ALPHA` as the producer for tables that have **no `source` column**, which asserted a relationship that doesn't exist just because the strings matched.

## No Zod, deliberately

Matching `producer-cadence.ts`, the sibling that owns the other half of this metadata and carries none. `schemas-src/` is for **published contract components** (`register()`ed in `openapi-registry.ts`); a schema here would `safeParse` a compile-time constant against itself and prove nothing the type doesn't.

Typing `lanes` as `ProducerLane` is strictly **stronger** than the Zod enum it replaced: a bad lane name is now a compile error rather than a runtime one.

## The check with teeth

`validate:lane-topology` compares every column the declaration names against `generated/db/schema.json` — the snapshot introspected from live Neon. Schema-vs-schema, the house pattern. A column renamed upstream fails CI instead of quietly making a rule count the wrong thing.

Registered in `package.json` and `validate.yml` beside the other drift gates, so it actually runs.

Verified it fails on both a non-existent column and a dropped discriminator:

```
nominator_positions.coverageUnit names "ghost_column", which is not a column of nominator_positions
nominator_positions: declares 2 producer lanes but no column to tell them apart
```

## Load-bearing, not decorative

The nominator-positions rule takes its scoping from the declaration via `requireFullScanValue`, which **throws** rather than returning null — binding null into `WHERE source = ?` matches no rows, so the rule would report zero coverage forever, indistinguishable from the truncated pass this began with.

`chain_detail_blocks` is deliberately **absent**: its watchdog is a ticker against an advancing frontier, and `chain_detail` isn't a `ProducerLane`. Forcing a lane name in to satisfy the shape would invent the vocabulary this module exists to avoid inventing.

120 tests pass across the lane suites; `tsc`, prettier, eslint clean; unreferenced-exports holds at 731; **patch coverage by diff intersection: 0 uncovered changed lines**.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -443,6 +443,7 @@ jobs:
           source .github/scripts/run-step.sh
           run_step npm run validate:contract-drift
           run_step npm run validate:db-types-drift
+          run_step npm run validate:lane-topology
           run_step npm run validate:lakehouse-types-drift
           run_step npm run validate:lakehouse-readers
           run_step npm run validate:store-type-parity
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
     "validate:types": "node scripts/validate-types.ts",
     "validate:contract-drift": "node scripts/validate-contract-drift.ts",
     "validate:db-types-drift": "node scripts/validate-db-types-drift.ts",
+    "validate:lane-topology": "node scripts/validate-lane-topology.ts",
     "validate:lakehouse-types-drift": "node scripts/validate-lakehouse-types-drift.ts",
     "validate:operational-surface-parity": "node scripts/validate-operational-surface-parity.ts",
     "validate:schema-enums": "node scripts/validate-schema-enums.ts",
--- a/scripts/validate-lane-topology.ts
+++ b/scripts/validate-lane-topology.ts
@@ -1,0 +1,124 @@
+// The lane topology declaration must describe tables that actually exist
+// (#11183).
+//
+// `src/lane-table-topology.ts` states things Postgres cannot know -- that a
+// producer prunes only its own rows, that a partial pass is partial in
+// coldkeys. Those are declarations and stay declarations. What this gates is
+// the half that IS checkable: every column the declaration names must exist in
+// `generated/db/schema.json`, the snapshot introspected from live Neon.
+//
+// WHY THAT MATTERS MORE THAN IT SOUNDS. The declaration is consumed by coverage
+// rules to scope their queries. A column renamed upstream would not fail those
+// rules loudly -- `WHERE source = $1` against a dropped column errors, but a
+// coverageUnit pointing at a column that moved would quietly count the wrong
+// thing, which is precisely the class of failure #11166/#11170/#11180 were.
+//
+// The consistency rules below are the other half: they refuse declarations that
+// are internally impossible, so a table cannot claim two producers without
+// saying how to tell them apart.
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import path from "node:path";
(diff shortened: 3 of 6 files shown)


JSONbored:
Superseded: the pipeline↔validate.yml parity gate (#1767) caught that `validate:lane-topology` was registered in `validate.yml` and `package.json` but not in `scripts/pipeline.ts`, so it ran in CI and not in the local `npm run check`. Fixed in both step lists and reopened rather than pushed under an auto-merging gate.

Issue #11146, opened 2026-08-14
epic(parity): "every subnet's API, exactly like using it directly" is not what we serve — three mechanisms measured, and what each one costs


JSONbored:
Raised from a real operational report on SN105 (Beam), then measured across the whole fleet on 2026-08-14. The SN105 findings are not SN105 findings.

## Claim under test

The MCP is meant to be Bittensor in a box: an agent reaches every supported subnet's API through us with parity of shape, coverage and security. Three separate mechanisms break that, and each fails silently.

## 1. The catalogue has no method dimension at all

`schemas/subnet-manifest.schema.json` `$defs.surface` declares `id, name, kind, url, provider, auth_required, authority, public_safe, auth, notes, probe, quality_signals, rate_limit, rate_limit_notes, revenue, review, schema_status, schema_url, source_urls, verification` — and `additionalProperties: false`. **There is no `method`.**

So a registered surface is structurally a GET-able URL. `POST /orchestrators/register` and `GET /orchestrators/register` cannot be told apart, and registering the former today would publish a URL the prober would call with GET (405) and an agent would read as "this route exists, fetch it".

Measured over 51 subnets whose registered OpenAPI spec is reachable and parseable:

| | count |
|---|---|
| paths declared across the fleet | 2,607 |
| paths registered as surfaces | 1,392 (53%) |
| **non-GET operations declared** | **1,448** |
| non-GET operations representable in the registry today | **0** |

Worst offenders (paths registered / declared): sn-81 0/206, sn-75 0/206, sn-12 2/214, sn-3 3/207, sn-22 0/35, sn-60 0/34, sn-52 0/20, sn-118 0/13, sn-64 55/168, sn-105 29/56.

## 2. Execution gates on a captured schema that has no scheduled producer

`call_subnet_surface` Phase 2 can already call a non-GET route — but only when that exact path+method is declared in the surface's **captured** schema. So the capture IS the parity mechanism.

`sync-subnets.yml`, which runs `pipeline:refresh` (and with it `schemas:snapshot`), is **manual-only by design** (ADR 0006 retired its schedule). Its last three runs failed (2026-07-15); the last success before that was 2026-06-14.

Every captured schema in production is **11 days old** (median 11, max 11 — the lane last produced on ~2026-08-03).

## 3. `drift_status` compares our copy to our copy

[...]

Issue #8967, opened 2026-08-02
auth(mcp): decide the MCP access model — the OAuth provider, Unkey tiers, and in-args credentials all exist, none wired to /mcp


JSONbored:
Spun out of epic #8959's "related, deliberately not a sub-issue" list — this is the access-model decision the observability epic explicitly did not own. Filed separately because it is product scope, but the audit that surfaced it is the same one.

## Problem

`/mcp` is fully public. `handleMcpRequest` enforces a rate limit (`enforceMcpRateLimit`, `MCP_RATE_LIMITER` at 100/60s) and nothing else — no `Authorization` check, no bearer verification, no `WWW-Authenticate` challenge, no `/.well-known/oauth-protected-resource`. Every one of the 167 tools is callable by anyone on the internet, and 30 days of `$mcp_initialize` data confirms that is exactly who calls it: the top clients are registry crawlers and probes (`glimind-probe` 1,912, `mcpregistry` 661, `glama` 188, `smithery-probe` 168), with real agent traffic (`claude-code` 615, `Anthropic/ClaudeAI` 190) mixed in indistinguishably.

Meanwhile the repo already contains, built and deployed but never connected to MCP:

1. **A complete OAuth 2.1 authorization server** — `@cloudflare/workers-oauth-provider` ^0.8.2, `src/github-oauth.ts`, routes `/authorize` + `/oauth/callback/github` (`workers/api.ts:1646-1650`), `OAUTH_KV` binding. It was built *for* MCP clients; `src/mcp-server.ts` never references it.
2. **A self-serve API-key system with tiers** — wallet-signature login mints `mg_...` keys via Unkey (`workers/data-api.ts` wallet-auth section, `src/unkey-client.ts`), with per-account tiers (`rpc_accounts.tier`, default `free`) and an internal tier-promotion route. Today those keys gate REST surfaces only.
3. **Per-tool credential passing** — `auth_required` subnet surfaces take a caller-supplied `credential` *inside tool arguments* (`src/mcp-server.ts:11161, 11240-11363`), which is the pattern OAuth-aware MCP clients are moving away from.

So three auth systems exist in various states of wiredness, and the MCP surface — the one with big plans — uses none of them. The risk is not abstract: every capability added to the server ships instantly to anonymous internet traffic at 100 req/min per caller, the annotation fix (#8964) only *labels* dangerous tools, and there is no way to give a partner, a paying tier, or an internal agent more capability than a crawler.

## What this issue is

[...]

PR #11218, opened 2026-08-14
feat(mcp): the paid-depth seam and its structured refusal

### LINKED ISSUE #11179: design(monetization): paid MCP access — one server, tiered depth, and where x402 fits
Pinning the design discussed 2026-08-14, so it exists as a decision record before anyone builds. Also absorbs the four Commerce rows of the agent-readiness scan (ACP / AP2 / MPP / UCP / x402), which only make sense once this exists.

## The shape

**One MCP server, not two.** A second server forks the catalogue, docs, analytics, and the contract-drift discipline the zero-drift epic just tightened. The plumbing already exists on one server: `/mcp` accepts `Authorization: Bearer mg_...` today, responses already expose `x-ratelimit-tier` and `x-api-key-block-reason`, and the server card + auth.md already document key-based auth ("authenticating buys throughput, not reach").

**Tier gates depth, freshness and rate — never tool visibility.** This session measured twice what absence does: unregistered mutations read as "no such route", empty schema documents read as "no such API". A paid tool invisible to free users is a tool nobody discovers or upgrades for. Every tool stays listed for everyone; a free call that crosses paid depth gets a structured `payment_required` error naming the tier and the upgrade path — agents relay that to their humans, which makes the MCP itself the sales channel.

**Free keeps the public good** (registry, discovery, health, basic subnet data) — it is the adoption funnel and ADR 0027's reasoning stands. **Paid takes the economic signals** now being built: miner-earnings percentiles, USD emission splits, screening depth, long history windows, plus rate and credential-store perks.

## Where x402 fits

The scan's Commerce rows describe machine payments. x402 (HTTP 402 + per-call payment) is the one that matches an MCP with agent callers: an agent with a wallet pays per call instead of holding a subscription — no signup funnel, which is exactly the friction agents cannot navigate. The `payment_required` error above should be DESIGNED so its shape can later carry an x402 challenge without a second error vocabulary. Do not build the payment leg first; build the tier seam so the challenge has somewhere to live.

## Done means (in order, each shippable)

1. A tier resolver at the MCP dispatch seam (key → tier, anonymous → free), with the tier on `$mcp_tool_call` analytics so pricing is designed against real usage.
2. The structured `payment_required` refusal, published in the error vocabulary like every other code, with tests proving a free call over the line gets it and a paid call does not.
3. A first paid depth boundary on ONE surface (candidate: history window length on the economics/history tools), chosen from analytics, not intuition.
4. x402/commerce evaluated only after 1–3 exist and the scan's Commerce rows re-read against a real tier seam.

JSONbored:
Closes #11179.

**Tier gates depth, never visibility.** Every tool stays listed and callable at every tier; what a key buys is how far back a call may read. A paid tool hidden from free callers is a tool nobody discovers or upgrades for — the same absence-reads-as-nonexistence failure measured twice already (unregistered mutations in #11146, empty captured schemas in #11148).

## What shipped

**Phase 1 was already in place** and is verified, not rebuilt: `applyTieredRateLimit` resolves the tier at the MCP dispatch seam, and `authTier` has ridden every `$mcp_tool_call` since #8967 — which is exactly why phase 3 could be decided from data.

**Phase 2 — the structured refusal.** `payment_required` rides the existing tool-error path with a `payment` block: `tier`, `required_tier`, `boundary`, `limit`, `requested`, `upgrade_url`. The message carries the upgrade URL too, so a client that renders only prose still shows its human something actionable. The block is shaped so an x402 challenge later joins **it** rather than introducing a second error vocabulary. Published where agents read: the tool description and `public/skills/bittensor/SKILL.md`.

**Phase 3 — the first boundary, chosen from usage.** Over the 30 days to 2026-08-14, the economics family carried the **most distinct callers of any tool** (`get_subnet_economics`: 52 callers, above `get_subnet_health`'s 47 on a third the volume) — economic history is the demand centre the issue predicted. Effectively **all** measured traffic is anonymous, so the free depth stays a real product rather than a teaser: `7d`/`30d`/`90d` are open to every caller, and only `1y` and `all` — the two whole-rollup scans — require a paid key.

**Phase 4 (x402)** is explicitly the evaluate-after step in the issue; the seam it needed now exists, and the refusal shape is the place a challenge will live.

## Two traps closed by tests

- **An unrecognised tier clears nothing.** `"enterprise"`, `"__proto__"`, `""` all rank below free — otherwise a typo in an account record becomes free access to every paid depth.
- **An unbounded window crosses every finite ceiling.** `all` parses to `days: null`; treating null as "no request" would have made the *deepest* read the one thing that passed free.

## Validation

- `typecheck`, `lint`, `format:check`, `validate:mcp` (240 tools), `validate:unreferenced-exports` (731, at the ceiling — not raised), `validate:module-state-resets`
- Full suite: **918 files / 20,932 tests passed**
- Diff-intersection coverage: **0 uncovered lines and 0 uncovered branch arms** across `src/mcp-tier-gate.ts` and the changed `src/mcp-server.ts` lines
- The boundary is proven end-to-end through a real `tools/call`: anonymous `1y` → `payment_required` with the block; the same anonymous caller at `30d` → never gated; the gated tool still appears in `tools/list`

### DIFF
--- a/public/.well-known/agent-skills/index.json
+++ b/public/.well-known/agent-skills/index.json
@@ -3,7 +3,7 @@
   "skills": [
     {
       "description": "Use when a developer asks what a Bittensor subnet does, whether it's up right now, how to call/integrate its API, or what mining/validating one costs and earns — e.g. \"which subnet does image generation\", \"is subnet 7 healthy\", \"call the Beam API for me\", \"does mining subnet 3 need a GPU\", \"what does the median miner on subnet 13 make\". Can EXECUTE subnet API calls, not just describe them. Backed by metagraphed (api.metagraph.sh), the live operational + integration registry for ~129 subnets.",
-      "digest": "sha256:78c9793623bcba9c35286490b65b144c8b4f9be58cb9c44a6f27f93d143acd86",
+      "digest": "sha256:80c8a2ae9fe5138253dea4a5f4462d513f7f7242062c128796e1e4687d6cdc6c",
       "name": "bittensor",
       "type": "skill-md",
       "url": "https://api.metagraph.sh/skills/bittensor/SKILL.md"
--- a/public/skills/bittensor/SKILL.md
+++ b/public/skills/bittensor/SKILL.md
@@ -107,6 +107,12 @@ Cursor / other clients: add an MCP server with url
   in any loop.
 - **Missing something?** Call `get_more_tools` with your goal in `context` —
   it records the capability gap so it can actually get built.
+- **`payment_required` names its own fix.** Every tool is callable at every
+  tier; what a key buys is depth. A call past a paid boundary (today: history
+  windows longer than 90 days on `get_economics_trends`) answers
+  `payment_required` with a `payment` block carrying `required_tier`,
+  `boundary` and `upgrade_url`. Retry inside the free depth, or relay the
+  upgrade path — the refusal is actionable, not a dead end.
 
 ## Develop before mainnet (local → testnet → mainnet)
 
--- a/src/mcp-server.ts
+++ b/src/mcp-server.ts
@@ -1228,6 +1228,10 @@ import {
   parseEconomicsTrendsWindow,
 } from "./economics-trends.ts";
 import {
+  FREE_HISTORY_WINDOW_DAYS,
+  requireTierForDepth,
+} from "./mcp-tier-gate.ts";
+import {
   DEREGISTRATION_UNAVAILABLE_CODE,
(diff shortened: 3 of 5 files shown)


JSONbored:
Closing on red `codecov/patch` rather than pushing over it. One partial branch: `ctx?.authTier ?? "anonymous"` invented a fallback this file explicitly rejects two thousand lines down (`applyTieredRateLimit` always sets a tier), so the arm was unreachable by construction. Removed rather than ignored; reopening with 0 uncovered lines and 0 partial branches.

PR #11220, opened 2026-08-14
feat(registry): declare the capture cadence, the drift basis, and the parity ledger

### LINKED ISSUE #11146: epic(parity): "every subnet's API, exactly like using it directly" is not what we serve — three mechanisms measured, and what each one costs

JSONbored:
Closes #11146.

The epic measured three mechanisms that each silently broke "every subnet's API, exactly like using it directly". Phase 3 (the surface `method` dimension) landed in #11210 and the first declared mutation in #11212; this completes the rest in one change.

## 1. Freshness is measurable

The capture lane's cadence is declared once (`SCHEMA_CAPTURE_CADENCE_HOURS = 48`) and published on the schema index as `capture_cadence_hours`. The lane stays manual by ADR 0006 — declaring the cadence does not run it, it makes its **silence measurable**, which is the difference between an unpublished gap and a published one.

**No age is baked into the artifact, deliberately.** The index is served for hours after it is built, so a build-stamped age is wrong on arrival; it is also not reproducible across two builds; and because `buildTimestamp()` returns the 1970 epoch placeholder in every local build, an earlier revision of this branch published a **twelve-day-old capture as fresh** — caught in a built artifact, and the reason the design changed. Every entry already carries `observed_at`, so the subtraction belongs to the reader, whose clock is the only one that says *now*.

## 2. Drift says what it measures

Captured entries carry `drift_basis: "previous-capture"`. `drift_status` compares our snapshot to **our previous snapshot**, never upstream — which is why 23 of 24 measurably-drifted subnets reported `unchanged`. The payload now says so rather than implying the opposite.

## 3. The parity ledger

Each gaps row gains `schema_parity`: captured paths and declared mutations against registered route surfaces and registered mutations, `flagged` when the catalogue registers fewer routes than the subnet declares.

Derived from schema **index** entries only — captured documents exist solely in credentialed builds, so a rollup reading them would not be reproducible — with the capture lane now stamping `non_get_operation_count` so both sides of the mutation ledger are measured. Counts absent on pre-stamp entries report **null, never zero**.

**First measurement:** 53 subnets measured, **27 flagged** — sn-3 registers 8 routes against 208 declared paths; sn-105 registers 29 (including the `POST` surface from #11212, which the ledger counts as its 1 registered mutation) against 53.

## Validation

- `typecheck`, `lint`, `format:check`
- `validate` (129 subnets / 3,330 surfaces), `validate:api` (215 routes), `validate:openapi`, `validate:contract-drift`, `validate:published-names`, `validate:unreferenced-exports` (**731 — at the ceiling, not raised**; the earlier revision's new type exports were removed rather than ratcheted), `validate:schema-enums`, `validate:schema-vocabularies`, `validate:mcp` (240 tools), `validate:schemas`, `validate:no-hand-written-mjs`
- Full suite green, including `artifacts-build-determinism` and the schema-index-preservation suite — both re-run against a **pristine committed baseline**, which is how the build-clock defect above was found
- Two consecutive builds from a pristine baseline: byte-identical

### DIFF
--- a/generated/graphql/schema.ts
+++ b/generated/graphql/schema.ts
@@ -4440,6 +4440,35 @@ type GapsArtifactGaps {
   gaps: Gaps!
   gap_severity: String
   gap_priority: Int
+  schema_parity: GapsArtifactGapsSchemaParity
+}
+
+type GapsArtifactGapsSchemaParity {
+  """Captured machine-readable specs backing this measurement."""
+  captured_schema_count: Int!
+
+  """
+  Paths the subnet's captured spec(s) declare, summed across captured specs.
+  """
+  captured_path_count: Int!
+
+  """
+  Declared POST/PUT/PATCH/DELETE operations. NULL when the captured entries predate the capture-time stamp -- unmeasured, not zero.
+  """
+  declared_non_get_count: Int
+
+  """
+  Registered concrete route surfaces (subnet-api/sse/data-artifact; the openapi spec surface itself is not a route).
+  """
+  registered_route_surface_count: Int!
+
+  """Registered surfaces declaring a non-GET method (#11146 phase 3)."""
+  registered_non_get_count: Int!
+
+  """
+  True when the subnet declares more paths than the catalogue registers as routes -- a caller reading only the registry cannot tell which routes are missing. Judge its currency from the entry's \`observed_at\` against the schema index's \`capture_cadence_hours\`.
+  """
+  flagged: Boolean!
 }
 
 type EvidenceList {
--- a/generated/graphql/types.ts
+++ b/generated/graphql/types.ts
@@ -2745,7 +2745,24 @@ export type GapsArtifactGaps = {
   gaps: Gaps;
   name: Scalars['String']['output'];
   netuid: Scalars['Int']['output'];
+  schema_parity?: Maybe<GapsArtifactGapsSchemaParity>;
   slug: Scalars['String']['output'];
+};
+
+export type GapsArtifactGapsSchemaParity = {
+  __typename?: 'GapsArtifactGapsSchemaParity';
+  /** Paths the subnet's captured spec(s) declare, summed across captured specs. */
+  captured_path_count: Scalars['Int']['output'];
+  /** Captured machine-readable specs backing this measurement. */
+  captured_schema_count: Scalars['Int']['output'];
+  /** Declared POST/PUT/PATCH/DELETE operations. NULL when the captured entries predate the capture-time stamp -- unmeasured, not zero. */
+  declared_non_get_count?: Maybe<Scalars['Int']['output']>;
+  /** True when the subnet declares more paths than the catalogue registers as routes -- a caller reading only the registry cannot tell which routes are missing. Judge its currency from the entry's `observed_at` against the schema index's `capture_cadence_hours`. */
+  flagged: Scalars['Boolean']['output'];
+  /** Registered surfaces declaring a non-GET method (#11146 phase 3). */
+  registered_non_get_count: Scalars['Int']['output'];
+  /** Registered concrete route surfaces (subnet-api/sse/data-artifact; the openapi spec surface itself is not a route). */
+  registered_route_surface_count: Scalars['Int']['output'];
 };
 
 /** Registry-wide interface gap report page. Mirrors GET /api/v1/gaps (and MCP list_gaps). */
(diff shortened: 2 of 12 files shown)


JSONbored:
Closing on red CI rather than pushing over it. The schema index turned out to be un-annotatable by a PR: its committed copy is a deploy-owned capture cache, and `artifacts-build-schema-index` asserts a rebuild reproduces it exactly, so any added field breaks the gate that protects 58 captured schemas.

Reworked so that artifact is untouched — the cadence moves to the `schema_parity` block (a normal built artifact, proven deterministic), and the drift-basis honesty becomes the published `drift_status` description at the point of consumption. Reopening green.

PR #11260, opened 2026-08-15
fix(watchdog): the alarm closed two issues for lanes that are 76 hours dead

JSONbored:
`loadLatestLaneHealth` resolved a tie by row order. On the first day the lane alarm could actually write, that cost two **false recoveries** — the worst thing an alarm can report, because it is indistinguishable from the outage being over.

## The tie is real, not theoretical

`lane_health` has no key. `recordLaneVerdict` INSERTs, and its prune deletes `checked_at < ?` — **strictly** less — so two rows written at the same stamp both survive and both match `MAX(checked_at)`. The reader then did `out[lane] = row` unconditionally, so "the newest verdict" was whichever row the driver happened to yield last. Nothing orders that.

Several lanes are stamped with the **producer's** pass time rather than the check time, so the stamp *freezes* while the producer is down and every later write lands on top of it. Two writers then collide on one millisecond:

| writer | verdict | detail |
| --- | --- | --- |
| sync flush | `ok` | `127 statement(s) flushed` |
| staleness watchdog | `unknown` | `no verdict for 4597m` |

Same lane. Same millisecond.

## What it cost, measured

**2026-08-15 03:58Z.** The alarm read the `ok` side of that tie and closed both issues it had opened thirty minutes earlier:

```
#11252 validator-nominator-counts — closed citing ok at 2026-08-11T23:27:52.032Z
#11253 nominator-positions        — closed citing ok at 2026-08-11T23:35:17.052Z
```

`GET /api/v1/self-health` served **`unknown` for both lanes at those exact milliseconds** — sampled 8 times, stable. Both lanes had been silent for 76 hours and still are. The detail even recomputes each tick (`4597m` → `4606m`) while `checked_at` never moves, which is the visible signature of rows piling up at one frozen stamp.

## The rule

**Newer wins; only a tie is broken, and it breaks to the worst verdict.** That is what `migrations/neon/0006` already says about the column — collapsing `unknown` into `ok` *"would report an unmeasured lane as a healthy one"* — applied to the read.

**Both halves matter.** Ordering by severity *alone* is a different bug in the opposite direction: an older finding would outrank a genuine recovery. The first draft of this fix did exactly that, and the regression test caught it — which is why that test is in the diff and why it says so. A real recovery is not a tie at all: the producer ran, so its stamp advances and its row is the unique max.

`VERDICT_SEVERITY` is `Record<LaneVerdict, number>` rather than a lookup with a default, so a verdict added to `LANE_VERDICTS` stops this compiling until somebody decides where it sits. A default would silently rank a new verdict as the safest thing in the table — the direction that loses findings.

## What this does not fix

The underlying stamping (`checked_at` = the producer's pass time, not the check time) and the resulting unbounded row growth at a frozen stamp are both still there. This makes the *read* safe against them, which is the part that was actively closing real alarms. Changing the stamp semantics touches every consumer of `age_ms` and is not something to do in the same change.

## Verification

- 167 tests across the six lane/self-health suites, all passing.
- **Patch coverage 100%** — 7/7 statements, 9/9 branches on the changed lines, by diff intersection with `coverage-final.json`.
- Both tie orderings asserted explicitly (`ok,unknown` and `unknown,ok`), so the test cannot pass by accident of row order — which is the exact failure being fixed.
- `typecheck`, `lint`, `format:check`, `validate:contract-drift`, `validate:schemas`, `validate:unreferenced-exports`, `validate:boundary-casts` green.

### DIFF
--- a/src/lane-health.ts
+++ b/src/lane-health.ts
@@ -226,19 +226,80 @@ export async function loadLatestLaneHealth(
       // predicate shared by every caller, and so a row that outlives its
       // deletion is still visible to anyone querying the table directly.
       if (isRetiredLane(lane)) continue;
-      out[lane] = {
+      const record: LaneHealthRecord = {
         lane,
         verdict: normalizeVerdict(row.verdict),
         age_ms: toIntOrNull(row.age_ms),
         detail: row.detail == null ? null : String(row.detail),
         checked_at: toIntOrNull(row.checked_at) ?? 0,
       };
+      // TIES ARE REAL, AND THIS USED TO RESOLVE THEM BY ROW ORDER.
+      //
+      // The MAX above is not unique. `lane_health` has no key, recordLaneVerdict
+      // INSERTs, and its prune deletes `checked_at < ?` -- strictly less -- so
+      // two rows written at the SAME stamp both survive and both match. This
+      // loop then assigned unconditionally, so "the newest verdict" was
+      // whichever row the driver happened to yield last, which nothing orders.
+      //
+      // It is not hypothetical. Several lanes are stamped with the PRODUCER's
+      // pass time rather than the check time, so the stamp freezes while the
+      // producer is down and every later write lands on top of it -- a sync
+      // flush writing `ok` ("127 statement(s) flushed") beside a staleness
+      // watchdog writing `unknown` ("no verdict for 4597m"), same lane, same
+      // millisecond.
+      //
+      // Measured 2026-08-15 03:58Z, on the first day the lane alarm could
+      // write: it read the `ok` side of that tie for `nominator-positions` and
+      // `validator-nominator-counts` and CLOSED both issues as recovered, while
+      // /api/v1/self-health served `unknown` for the same lanes at the same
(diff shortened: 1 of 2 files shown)


JSONbored:
**The full cycle just completed in production, which pins the cost.**

```
03:28:47Z  opened  #11252 validator-nominator-counts
03:28:48Z  opened  #11253 nominator-positions
03:58:49Z  CLOSED  both — tie resolved to the `ok` row, "2 recovered"
04:28:46Z  opened  #11261 validator-nominator-counts   ← same lanes
04:28:46Z  opened  #11262 nominator-positions          ← same fault
```

Neither lane changed state at any point — both have been silent since 2026-08-11 and still are. So the tie costs a **flap**, not just a wrong reading: the alarm tells a maintainer a dead lane recovered, then re-files the same finding under a new number an hour later. Two issue numbers burned per cycle, and the recovery comment on #11252/#11253 is a false statement left in the record.

The `0 alarming, 2 recovered` verdict at 03:58 and `2 alarming, 0 recovered` at 04:28 are the same two lanes, from the same unchanged rows.

PR #11323, opened 2026-08-15
fix(ci): prose is not a consumer, in a doc or in a comment

### LINKED ISSUE #10517: ci: the UI route-coverage gate proves a route is MENTIONED, not rendered — and it is a hard rule now
## `validate:ui-route-coverage` proves a route is MENTIONED, not that it is rendered

Now that #10300 drove the ceiling to `MAX_UNRENDERED_ROUTES = 0`, this check is a hard rule on every PR: a new route with no `apps/ui` reference fails CI. That makes the gap between what it measures and what it claims worth stating.

It concatenates every `.ts/.tsx/.mdx/.json` under `apps/ui/src` and `apps/ui/content` and regex-tests each route path against the blob. So **any occurrence of the string counts**. Measured against the real exported `unrenderedRoutes`, not by reading the code:

```
COUNTS AS RENDERED  a real fetch
COUNTS AS RENDERED  a bare code comment          // TODO: someday render /api/v1/chain/burn
COUNTS AS RENDERED  an ApiSourceFooter path      paths={["/api/v1/chain/burn"]}
COUNTS AS RENDERED  prose in an .mdx doc
```

The `ApiSourceFooter` case is the one that will actually happen. That component takes a `paths` array of plain strings, adding a route to it is the natural thing to do when you touch a page, and it satisfies the gate while rendering nothing. The gate would go green on a PR that added a route, listed it in a footer, and shipped no consumer.

## A second, softer blind spot: conditional mounts

A reference inside a component that is mounted behind a collapsed disclosure also counts. I hit this in #10300 itself — `CrowdloansPanel` first landed inside `{showAnalytics && …}` on the explorer page, and `showAnalytics` defaults to `false`. The sweep reported 0 unrendered while the surface was effectively unseen. Caught by audit, not by the gate (fixed in `ace7a167d`).

Static analysis can't reasonably decide "is this mounted on a path a user reaches", so this half is probably not fixable — but it is worth writing down that a green check does not mean a visible panel.

## Why this matters more at 0 than it did at 25

At a ceiling of 25 the check was a **ratchet**: it stopped the backlog growing, and the exact strength of each reference barely mattered. At 0 it is being read as a **guarantee** — "every published route is rendered" — and it does not support that reading. A gate whose claim is stronger than its evidence is the shape this repo keeps finding elsewhere (#10222's lane health, the degraded-header sweep); worth not building another one.

## Options, roughly in order of cost

1. **Exclude the known string-only sites from the source blob.** `ApiSourceFooter`'s `paths` arrays are the big one. Cheap, removes the case most likely to occur in practice, and does not pretend to solve the rest.
2. **Require the reference to appear in a query/fetch context** — e.g. only count matches inside `apiFetch(...)`/`queryOptions` call sites, or only in `lib/metagraphed/queries.ts`. Stronger, and arguably the right shape: a route is rendered because something *fetches* it. Would need every existing reference re-checked against the narrower rule first.
3. **Leave it, and fix the wording.** Rename to `validate:ui-route-references` and change the success line from "are rendered" to "are referenced by apps/ui", so nobody reads a guarantee the check cannot make.

(3) is worth doing regardless of whether (1) or (2) lands — the current message is the part that overstates.

Found while auditing my own work on #10300 rather than from a failure; nothing is currently gamed, and all 205 routes really are rendered today.

JSONbored:
Finishes #10517's narrowing. The check matched route paths against a
concatenation of every file under `apps/ui`, so any occurrence counted --
and two kinds of occurrence were not references at all.

`.mdx` went in the change that rendered
`/api/v1/accounts/{ss58}/identity-history`; that route was green at a ceiling of
zero on a row in `docs/accounts.mdx` alone.

COMMENTS GO HERE, and I found them the way the issue predicted somebody would:
by writing one. The component added for that route explains itself by naming the
route, and that sentence alone held the gate up when the query beneath it was
mutated away. `stripCodeComments` removes `//` and `/* */` before matching.

STRING-AWARE ACROSS ALL THREE QUOTE FORMS, which is the whole difficulty and the
one way this change could be worse than no change. This codebase composes paths
as `` `/api/v1/accounts/${ss58}/events` ``, so a stripper that does not know
backticks deletes the references the check exists to find -- silently, with the
count climbing and every entry looking like a real gap. Removing backtick
handling fails a named test.

The narrowing surfaced exactly one route, `/api/v1/networks`, and the right
answer was not to render it. It is a capability matrix -- which chains are
addressable and which route families each serves -- and the thing that needs it
is a client deciding what to call. Rendering it in the UI to satisfy a checker
would be building a surface for the checker, so it joins the feeds under
`EXTERNAL_CONSUMER_ROUTES`: named per route, one justification each, and
explicitly not "routes we could not be bothered to render" -- that is what the
ceiling is for.

WHAT REMAINS UNSEEN is whether a reference is MOUNTED. An unmounted component
whose query names the route still passes, and static analysis cannot reasonably
decide "is this on a path a user reaches" -- #10517 says so and I agree. The
messages say "appear nowhere in apps/ui" rather than "rendered", so the claim
and the evidence now match.

Closes #10517

### DIFF
--- a/scripts/validate-ui-route-coverage.ts
+++ b/scripts/validate-ui-route-coverage.ts
@@ -75,6 +75,27 @@ const FEED_PREFIX = "/api/v1/feeds/";
 /** Network-prefixed testnet variants mirror their mainnet twin's rendering. */
 const NETWORK_PREFIX = "/api/v1/{network}/";
 
+/**
+ * Routes whose consumer is an API CALLER rather than our own UI.
+ *
+ * The same exemption the feeds already have, named per route because this one
+ * cannot be recognised by prefix. `/api/v1/networks` is a capability matrix --
+ * which chains are addressable and which route families each one actually
+ * serves -- and the thing that needs it is a client deciding what to call, not
+ * a person reading a page. Rendering it in the UI to satisfy a checker would be
+ * building a surface for the checker.
+ *
+ * EXPLICIT AND JUSTIFIED, one line each, because this is the mechanism most
+ * likely to rot into "the list of routes we could not be bothered to render".
+ * A route belongs here when its consumer is external BY DESIGN -- not when it
+ * is merely unrendered today, which is what the ceiling is for.
+ */
+const EXTERNAL_CONSUMER_ROUTES: readonly string[] = [
+  // Answers "what can this deployment serve", for a caller choosing a network
+  // prefix. Its rendered form is the docs page that documents it.
+  "/api/v1/networks",
+];
+
 const escapeLiteral = (segment: string): string =>
   segment.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
 
@@ -99,7 +120,9 @@ export function publishedRoutes(openapi: {
     .map(([route]) => route)
     .filter(
       (route) =>
-        !route.startsWith(NETWORK_PREFIX) && !route.startsWith(FEED_PREFIX),
+        !route.startsWith(NETWORK_PREFIX) &&
+        !route.startsWith(FEED_PREFIX) &&
+        !EXTERNAL_CONSUMER_ROUTES.includes(route),
     )
     .sort();
 }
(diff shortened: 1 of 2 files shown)


JSONbored:
Heads up — #11322 landed the `.mdx` half of this an hour ago, so this branch will conflict. It is based on `15f1d271a`, which is one commit before it.

What is already on main:

- `.mdx` dropped from the evidence blob, and `ApiSourceFooter paths={[…]}` arrays stripped (measured: the footer strip removes **0** routes today, the `.mdx` drop removes **1**).
- `PROSE_ONLY_ROUTES` — a named register holding exactly `/api/v1/accounts/{ss58}/identity-history`, printed on every run rather than folded into a total.

**The part that will bite on rebase:** the register has a staleness check. If your branch renders `/api/v1/accounts/{ss58}/identity-history`, the gate fails with

```
PROSE_ONLY_ROUTES lists 1 route(s) that apps/ui now references. Delete them from the list…
```

That is intended — the list is supposed to shrink when the gap closes, so it cannot rot into the allowlist `MAX_UNREFERENCED_ROUTES` refuses to be. Render the route **and** delete the entry in the same change and it goes green.

Your comment-stripping is the piece main does not have, and it is worth landing on its own — I did not touch comments. Worth re-measuring after the rebase though: with `.mdx` already excluded, the number of routes held up by a comment alone may be different from what you measured against the older blob.

JSONbored:
Superseded by the branch rebuilt on current main — this one conflicts after #11311 and the PROSE_ONLY_ROUTES work landed.

Issue #10960, opened 2026-08-13
contract: a projected response omits properties its component marks required (fields today, sections after #10600)


JSONbored:
## The measurement

Live against production, 2026-08-13:

```
GET /api/v1/subnets/1/metagraph?fields=uid   ->  200, 256 rows
row keys: ["uid"]
```

`SubnetMetagraphArtifact.neurons.items` publishes:

```json
"required": ["uid", "hotkey", "coldkey", "active", "validator_permit"]
```

So a documented, supported request returns rows missing **four properties the component
marks required**. A client generated from `openapi.json` and validating responses rejects a
200 the server considers correct.

## Why it matters more now

This is not new — `fields` has shipped on 5 routes for a long time — but #10600 adds
`sections=` on the two composite subnet routes, which projects at the DOCUMENT level. There
`SubnetDetailArtifact` marks `subnet`, `surfaces`, `gaps` and `candidate_surfaces` required,
so `?sections=economics` returns a document missing four required properties by the same
mechanism.

#10600 deliberately followed the established `fields` behaviour rather than inventing a
different answer for one parameter. That decision is only defensible if the underlying
inconsistency gets settled once, for both.

## The options, none of which is free

1. **Loosen `required` to the envelope** on every projectable component. Honest about what
   the surface serves; costs every non-projecting caller the guarantee that `subnet` is
   present.
2. **Publish a projected variant** (`SubnetDetailArtifactProjected`, `NeuronProjected`) and
   reference it from the projecting operation. Keeps the strict contract for the default
   response; doubles the component count and needs the operation to declare which variant a
   given request yields — which OpenAPI cannot express as a function of a query value.
3. **Accept and document it**, stating in each projectable component's description that
   `required` describes the unprojected response. Cheapest; leaves generated validators
   wrong.

The MCP surface already answers this one way (#9884: a row schema is `.partial()` iff its
tool advertises `fields`), which is option 1 applied at the row level. REST does not follow
it, and the divergence is itself worth resolving — the two surfaces are generated from the
same schemas.

## Not urgent, but not invisible either

Nothing is broken for a caller who does not validate. The failure mode is a generated client
that does, and the response-validation tripwire
(`METAGRAPH_VALIDATE_RESPONSES`) would report drift on these routes if it were turned on for
a projecting request.

JSONbored:
Correction to this issue's last paragraph, which I wrote and which is wrong.

[...]

Issue #9720, opened 2026-08-07
get_tao_usd and get_emission_pipeline have no way to ask for less


JSONbored:
Both tools return their full body or nothing. Measured on production today:

| Tool | Bytes | Why |
|---|---|---|
| `get_tao_usd` | **143,126** | 1,428 `points` in the series; every summary a caller usually wants (`latest`, `change_usd`, `change_pct`, `point_count`, `priced_point_count`, `oldest_observed_at`) is already a top-level scalar beside them |
| `get_emission_pipeline` | **56,280** | 129 subnets × 16 fields, and `?netuid` is the only filter — it narrows to ONE subnet or leaves all 129 |

(Both were larger when first reported — 297,674 B and 116,802 B — so the numbers move with the window and the day. The shape is the problem, not the reading.)

An agent that wants "what is TAO worth right now" pays 143 KB (~36K tokens) for one number it can already see at the top of the body. An agent comparing emission gating across a handful of subnets has no way to say which subnets or which fields.

`/api/v1/economics` already has exactly the pattern both are missing — `fields`, `limit`, `cursor`, `sort`, `order` — so this is applying an established shape, not inventing one.

## Proposal

**`/api/v1/network/tao-usd`** — an `include_points` toggle. The summary is a complete answer on its own; the series is the optional part.

**`/api/v1/chain/emission-pipeline`** — `sort` / `order` / `limit` / `fields` over the `subnets` array, matching `/api/v1/economics`. Sorting by `final_share` is the one a caller most wants and currently cannot have: #9707 established that `final_share` (TAO actually received), not `emission_share` (the stage-1 price share), is the number that answers "where is the emission".

## Defaults

REST keeps serving what it serves today — a browser can stream 143 KB and a context window cannot, so **the MCP tool carries the narrowing default and the REST route does not**. That is the same asymmetry #9701 established for `list_candidates`, for the same reason.

Issue #11330, opened 2026-08-15
build: apps/ui typecheck is red on main — four call sites pass a `subnets` arg no query and no route accepts


JSONbored:
`npx tsc --noEmit -p apps/ui/tsconfig.json` fails on `main` with four instances
of the same error, and the change behind it does not do what it intends.

```
apps/ui/src/components/metagraphed/home-watched-module.tsx(286,82)
apps/ui/src/routes/-validators-index-page.tsx(96,79)
apps/ui/src/routes/-validators-index-page.tsx(131,73)
apps/ui/src/routes/-validators-index-page.tsx(219,81)

error TS2353: Object literal may only specify known properties, and 'subnets'
does not exist in type '{ sort?: GlobalValidatorSort; limit?: number }'.
```

The call sites read:

```ts
validatorsQuery({ sort: "total_stake", limit: ALL_VALIDATORS_LIMIT, subnets: false })
```

and the query is:

```ts
export const validatorsQuery = ({
  sort = "subnet_count",
  limit = 20,
}: { sort?: GlobalValidatorSort; limit?: number } = {}) => …
  apiFetch(…, { params: { sort, limit }, signal })
```

`subnets` is accepted by neither the signature nor the `params` it forwards.

## It would not work if it were threaded through

The route rejects the parameter outright:

```
GET /api/v1/validators?limit=5&subnets=false   400   194 bytes
GET /api/v1/validators?limit=5                 200   10,949 bytes
```

So the intent — "stop dehydrating rows the hubs never render" — is currently a
no-op in both directions: the argument never leaves the client, and the server
would refuse it if it did. The hub still ships the full payload.

## Where it came from

`-validators-index-page.tsx` was last touched by #11321 and #11327, so this
belongs to the SEO hub work rather than to anything I am on. Flagging rather
than fixing: the right shape depends on whether `/api/v1/validators` should grow
a projection parameter or whether the hub should trim client-side, and that is a
contract decision for whoever owns those PRs.

## Worth noting about how it got in

CI is green on those PRs. `apps/ui` typechecking needs the workspace packages
built first, so a bare `tsc` in that directory reports a wall of unrelated
`QUERY_PARAMETER_ENUMS` errors and this one is easy to lose in it — which is
probably also why it is easy to miss locally. Whatever fixes this is worth
pairing with a check that actually runs the built-workspace typecheck, or the
next one lands the same way.

JSONbored:
[...]

Issue #11315, opened 2026-08-15
seo(w3): /validators ships 1.29 MB of HTML for 1,650 words


JSONbored:
Part of #11313 (W3). One PR.

## The defect

Measured against production 2026-08-15:

| page | HTML | of which inline `<script>` | rendered words |
|---|---|---|---|
| `/validators` | **1,293 KB** | **988 KB** | 1,650 |
| `/subnets` | **511 KB** | 333 KB | 1,296 |
| `/apis` | 688 KB | — | 890 |
| `/subnets/{n}` | ~346–472 KB | 19 KB | ~471 |
| `/` | 98 KB | 31 KB | 377 |

`/validators` ships **1.29 MB of HTML to render 1,650 words**, and three quarters of it is dehydrated loader JSON restating data the page already rendered into markup. `/subnets` is the same shape.

These are the two hubs W1 is about to point every category query at. Mobile-first indexing and Core Web Vitals both take this directly, and on a 4G phone a 1.3 MB document is seconds of LCP before a single row paints.

## Deliverable

Cut the SSR payload on `/validators`, `/subnets` and `/apis`.

## Requirements

- Stop dehydrating loader data the page has already rendered into markup. The duplication is the bug, not the row count.
- **Budget: < 150 KB HTML for any hub**, enforced in CI next to the existing bundle-budget check.
- The internal anchors from #11231 (129 subnets) and #11232 (138 providers) **must survive** — verified by the existing `crawlable-subnet-index` spec, which is already in the `interaction` project.
- Measure before and after in node, not a shell pipeline: `grep` here is a ugrep wrapper that skips long-line SSR HTML as binary and silently reports zero matches.

## Don't

- Fix this by paginating or virtualising away the internal links. That reintroduces #11231 — 99 of 129 subnet pages having no inbound link anywhere on the site — and is the single most expensive regression available here.
- Move rendering to the client. A crawler must still see the rows in the SSR HTML.
- Trade the fix for a slower TTFB: this is a payload problem, not a compute one.

## Acceptance

- [ ] `/validators`, `/subnets`, `/apis` each under 150 KB of HTML
- [ ] CI budget check covering every hub, derived from the hub list
- [ ] `crawlable-subnet-index.spec.ts` still green — all 129 subnet and 138 provider anchors present
- [ ] Before/after payload table in the PR body, measured against production and the built Worker

JSONbored:
[...]

PR #11331, opened 2026-08-15
feat(ui): render the identity timeline, and stop counting comments as references

### LINKED ISSUE #11079: audit: tao-usd fires DRIFTED on include_points=false — the seam misses the #9720 projection lever
The #11046 warn log's only fingerprint is now dispositioned, deterministically: `[METAGRAPH_AUDIT_RESPONSES] tao-usd DRIFTED (warn)` fires on **every** `GET /api/v1/network/tao-usd?include_points=false` — the published #9720 "send me less" toggle (and the MCP `get_tao_usd` tool's default). The tripwire reports `data.points: expected array, received undefined`, which is exactly the omission the caller asked for.

`auditResponse` (workers/api.ts) derives its `projected` signal from the request URL — but only from `fields`/`sections`. `include_points` is the same class of lever: a caller-requested omission the schema's `required` (which describes the unprojected response, per #10960/#11009) deliberately does not model. The landed doctrine (#10978) is to pass the projection signal so only genuine ABSENCE is forgiven — never to loosen the schema.

**Fix:** treat `include_points=false` as `projected` at the audit seam. `parseBooleanParam` is strict (`"true"`/`"false"` only; anything else 400s before the seam sees a 200), so string equality is exact. Pin both directions in tests/audit-responses.test.ts: absence forgiven under the param, still DRIFTED without it.

Not the #11046 flip — this clears its criterion-2 blocker. The flip still waits on its quiet-day window.

JSONbored:
The warn log's only fingerprint (tao-usd, 3x in the first post-deploy hour)
reproduces deterministically: every GET /api/v1/network/tao-usd
?include_points=false -- the #9720 'send me less' toggle, and the MCP
get_tao_usd tool's default -- was fingerprinted for omitting data.points,
the exact omission the caller asked for.

auditResponse derives its projected signal from the request URL, but only
read fields/sections. include_points is the same class of lever, and the
schema's required deliberately describes the unprojected response (#10960),
so the seam now carries the third signal. parseBooleanParam is strict --
anything but the literal 'false' defaults to true or 400s before the seam
sees a 200 -- so string equality is exactly the handler's own condition.
Both directions pinned: the omission is forgiven under the toggle, and the
same absence without it is still drift.

Closes #11079

### DIFF
--- a/apps/ui/src/lib/metagraphed/queries.ts
+++ b/apps/ui/src/lib/metagraphed/queries.ts
@@ -208,6 +208,8 @@ import type {
   SubnetIdleStake,
   ChainIdleStake,
   ChainIdleStakeSubnet,
+  AccountIdentityHistory,
+  AccountIdentityHistoryEntry,
   SubnetIdentityHistory,
   SubnetWeightSetter,
   SubnetWeightSetters,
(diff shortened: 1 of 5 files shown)


JSONbored:
Superseded by a branch built additively on main. This one carried whole-file restores that silently reverted 58 lines of #11315/#11326 — see #11330 and #11345.

PR #11340, opened 2026-08-15
fix(ui): remove an argument no query and no route accepts, which is failing CI on main

### LINKED ISSUE #11079: audit: tao-usd fires DRIFTED on include_points=false — the seam misses the #9720 projection lever

JSONbored:
The warn log's only fingerprint (tao-usd, 3x in the first post-deploy hour)
reproduces deterministically: every GET /api/v1/network/tao-usd
?include_points=false -- the #9720 'send me less' toggle, and the MCP
get_tao_usd tool's default -- was fingerprinted for omitting data.points,
the exact omission the caller asked for.

auditResponse derives its projected signal from the request URL, but only
read fields/sections. include_points is the same class of lever, and the
schema's required deliberately describes the unprojected response (#10960),
so the seam now carries the third signal. parseBooleanParam is strict --
anything but the literal 'false' defaults to true or 400s before the seam
sees a 200 -- so string equality is exactly the handler's own condition.
Both directions pinned: the omission is forgiven under the toggle, and the
same absence without it is still drift.

Closes #11079

### DIFF
--- a/apps/ui/src/components/metagraphed/home-watched-module.tsx
+++ b/apps/ui/src/components/metagraphed/home-watched-module.tsx
@@ -283,8 +283,8 @@ function WatchedValidators({ hotkeys }: { hotkeys: string[] }) {
   // Same sort + limit as the /validators index, so this shares that query's
   // cache key instead of firing a second 2000-row fetch.
   const validators =
-    useQuery(validatorsQuery({ sort: "total_stake", limit: ALL_VALIDATORS_LIMIT, subnets: false }))
-      .data?.data?.validators ?? [];
+    useQuery(validatorsQuery({ sort: "total_stake", limit: ALL_VALIDATORS_LIMIT })).data?.data
+      ?.validators ?? [];
   const watched = new Set(hotkeys);
 
   const { flashed, flash } = useRowFlash();
--- a/apps/ui/src/routes/-validators-index-page.tsx
+++ b/apps/ui/src/routes/-validators-index-page.tsx
@@ -93,8 +93,7 @@ export function ValidatorsPage() {
         context="validators"
         fallback={<TableSkeleton rows={10} columns={8} />}
         retryQueryKeys={[
-          validatorsQuery({ sort: "total_stake", limit: ALL_VALIDATORS_LIMIT, subnets: false })
-            .queryKey,
+          validatorsQuery({ sort: "total_stake", limit: ALL_VALIDATORS_LIMIT }).queryKey,
         ]}
       >
         <ValidatorsDirectory density={density} onDensityChange={onDensityChange} />
@@ -128,7 +127,7 @@ function ValidatorsDirectory({
   // key never changes with UI state, so sorting/searching re-uses the cached
   // full set instead of refetching.
   const res = useSuspenseQuery(
-    validatorsQuery({ sort: "total_stake", limit: ALL_VALIDATORS_LIMIT, subnets: false }),
+    validatorsQuery({ sort: "total_stake", limit: ALL_VALIDATORS_LIMIT }),
   ).data;
   const all = res.data.validators;
   const generatedAt = res.meta?.generated_at ?? null;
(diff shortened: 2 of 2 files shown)


JSONbored:
## Hold — the premise does not reproduce, and this reverts a saving that is live in production

I ran the exact command from #11330 against a tree containing #11327 (`58fe34e76`), and again after confirming `validatorsQuery` is byte-identical to `origin/main`'s:

```
$ npx tsc --noEmit -p apps/ui/tsconfig.json ; echo $?
0
```

`main` declares the parameter (`apps/ui/src/lib/metagraphed/queries.ts`):

```ts
export const validatorsQuery = ({
  sort = "subnet_count",
  limit = 20,
  subnets = true,
}: { sort?: GlobalValidatorSort; limit?: number; subnets?: boolean } = {}) =>
```

The signature quoted in #11330 is the pre-#11327 one. `3c8e14f5d` (the newest commit on main) does not touch this file or any of the four call sites.

## `subnets` was never meant to reach the server

Its absence from `params` is the mechanism, not the bug:

```ts
params: { sort, limit },      // deliberately not `subnets`
...
data: subnets
  ? data
  : { ...data, validators: data.validators.map((v) => ({ ...v, subnets: [] })) },
```

The fetch happens during SSR, so it never crosses the user's wire. The cost #11315 was about is the **dehydrated** rows the server inlines into the HTML, and that is what the flag strips. #11327's body states this directly — the API has no field projection, `?fields=` answers 400 — which is why the trim is client-side, and why `subnets` is in the query key so two shapes cannot share a cache entry.

## Production, fetched just now

`https://metagraph.sh/validators`:

| | |
|---|---|
| `stake_alpha` occurrences | **0** |
| `emission_alpha` occurrences | **0** |
| `total_stake_tao` occurrences | **1,027** |
| document | **1,241,424 bytes** |

Both fields that exist only inside `subnets[]` are gone, the row fields the table renders are still there at 1,027 rows, and the document is ~85 KB below the 1,328,128 bytes #11327 measured beforehand — the ~87 KB that PR predicted.

**Removing the argument sets `subnets` back to its `true` default and re-inlines all of it**, on `/validators` and on the homepage's watched module, which shares the cache key.

## What I could be wrong about

If CI genuinely failed, the useful artifact is the **commit sha it ran at** — #11327 lands the signature and the call sites in one commit, so a tree with one and not the other is not a state `main` was ever in, but a mid-rebase or a stale `node_modules`/tsbuildinfo could produce it locally. `packages/ui-kit` also has to be built before `apps/ui` typechecks at all, and a missing build fails with a different error (TS7016) that is easy to read past.

If you have that sha and it reproduces, I will fix it properly — but the fix would be to restore whatever the tree was missing, not to drop the argument.

JSONbored:
Merged before this landed — filed as #11343 with the after-measurement.

Production `/validators` went from **1,241,424 B** to **1,328,004 B** (+86,580), back to the pre-#11327 size within 124 bytes, with `stake_tao` and `validator_trust` — fields that exist only inside `GlobalValidatorSubnet` — at 1,159 occurrences each. Three fetches with distinct cache-busters agreeing.

**Correction to my comment above:** I cited `stake_alpha`/`emission_alpha` at 0 occurrences as evidence the trim was working. Those names are from #11327's PR body and appear under neither setting — the fields are `stake_tao`/`emission_tao`. That check proved nothing either way. The byte delta is the evidence and it holds.

Issue #10285, opened 2026-08-09
feat(api): subnet deregistration ranking — which subnet is next to be pruned


JSONbored:
Found while researching #10259. taostats models subnet churn predictively and we have no equivalent surface.

## The chain mechanic

From taostats' own docs: **"when a new subnet is registered, the subnet with the lowest EMA price is deregistered."** So every registration evicts someone, and which subnet is nearest the threshold is knowable in advance.

They expose exactly that:

- `/subnet-deregistration-ranking` — *"pruning rank defines the order at which subnets will be replaced by new registrations"*
- `/subnet-deregistration-ranking-historical` — *"daily ranking of lowest performing subnets"*

## What we have and do not have

We now record deregistration **after the fact** (`subnet_lifecycle`, #10262/#10263) — when a subnet left, and whether the event predates capture. What we cannot answer is the question a subnet owner actually asks: **"how close am I to being pruned?"**

The inputs are already in Neon: `subnet_snapshots` carries EMA/moving price per subnet, and the emission data is there. The ranking is an ordering over data we hold, not a new capture lane — this is a serving gap, not a data gap.

## Shape

A ranked list of live subnets ordered by pruning proximity, plus a daily historical series so a subnet can see its own trajectory toward or away from the threshold. The historical half is the more valuable one: a single day's rank is noise, a trend is a warning.

Worth confirming the exact quantity the runtime compares (EMA price vs moving price vs emission share) against the pallet before publishing an ordering — publishing a rank that disagrees with the chain's own pruning order would be worse than publishing none.

Related: #10259 (why deregistered subnets are retained rather than pruned), #10263 (the lifecycle surface).

JSONbored:
## The pallet rule, read from source — and a correction: this IS a data gap

The issue asks to "confirm the exact quantity the runtime compares before publishing an ordering." Done. The rule is `Subtensor::get_network_to_prune()` in `pallets/subtensor/src/coinbase/root.rs`, and it needs **four** inputs, not one.

### The rule

Iterate `NetworksAdded`, and for each netuid:

1. skip if not added, or if it is **root**;
2. skip if `current_block < NetworkRegisteredAt[netuid] + NetworkImmunityPeriod` — **subnets have an immunity period**;
3. take `price = get_moving_alpha_price(netuid)`;
4. keep the lowest price; **on a tie, the EARLIER `NetworkRegisteredAt` wins**.

[...]

PR #11341, opened 2026-08-15
feat(api): the deregistration-rank series, replayed from stored inputs (#10296)

### LINKED ISSUE #10296: feat(api): daily deregistration-rank history — a single day's rank is noise, a trend is a warning
## The half of #10285 that needs a lane

#10285 asked for two things: the current deregistration ranking, and a **daily historical series** so a subnet can see its trajectory toward or away from the threshold. The current ranking ships in the PR closing #10285; this is the other half, split out because it needs storage the first half does not.

The issue's own argument for why the history is the more valuable side still stands:

> a single day's rank is noise, a trend is a warning

A subnet owner reading `rank: 94` learns almost nothing. Reading "94, was 71 a month ago, and the gap to the bar has halved" learns exactly what they need.

## Why it could not ride along

The current ranking is a pure projection over the pinned economics blob — no table, no lane, computed per request. A series is not: it needs a row per subnet per day, which means a table, a writer, and a cron.

It also **accrues only forward**. Unlike the hyperparameters backfill (#5597), the inputs cannot be reconstructed cheaply from history: reproducing a past day's ranking means reading `SubnetMovingPrice`, `NetworkRegisteredAt` and `SubnetMechanism` for every netuid *at that day's block*, one archive sweep per day. Possible, but it is a backfill project rather than a lane.

## Shape

- a daily row per subnet: `rank`, `comparison_price`, `immune`, `blocks_until_prunable`, and the block it was computed at;
- `ranked_count` beside it, since a rank of 94 means different things in a field of 100 and a field of 128;
- served as `/api/v1/subnets/{netuid}/deregistration-ranking/history` (one subnet's trajectory) and/or a network-wide daily snapshot.

Worth deciding whether this belongs as columns on the existing `subnet_snapshots` daily rollup — which already persists per-subnet pipeline state daily and already carries `pipeline_block` — rather than as a new table. That rollup already has the moving price; it would need the two new reads landed by #10285's PR plus the immunity period.

**Note the repeats-previous-observation trap** that `/subnets/{netuid}/emission-pipeline/history` documents: the daily writer carries the last capture forward when a fresh one has not landed, so two consecutive points can be the same observation. A rank series inherits that exactly, and must publish `distinct_observations` beside `point_count` for the same reason — otherwise a rank that was simply not re-measured reads as a rank that held steady.

Follow-up to #10285.

JSONbored:
## The write half has been running since 2026-08-10 and nothing read it

#10285 asked for two things and shipped one: the pallet's pruning order **as of one block**. This is the other half, on #10285's own argument:

> a single day's rank is noise, a trend is a warning

`rank: 94` tells a subnet owner almost nothing. "94, was 71 a month ago" tells them what to do.

`subnet_deregistration_daily` has six clean days in it — 129 netuids each, one pinned block per day, 127 priced — and `grep subnet_deregistration_daily src/ workers/` returned one hit: the table-name constant.

## The rank is not stored, and this is why that matters

The lane persists the four **measured** inputs plus the block they were pinned at, never the derived rank, so a later correction to the pallet rule reaches the whole series instead of leaving a record of the old rule's answers. This route is the other half of that design: **it replays the ranking on read.**

Through `rankDeregistration`, not `projectDeregistrationRanking`. The latter takes an economics blob and renames `moving_price_pinned` → `moving_price` on the way in, while the stored column is spelled `moving_price` — and a near-miss there produces a plausible ranking that is subtly **not the pallet's**. `rankDeregistration`'s input is a 1:1 name match with the stored columns, so there is no shape to get wrong.

**Rank is relative, so a one-subnet series reads every subnet's rows.** It does not exist in one netuid's row and cannot be read off one. Each day is loaded whole, ranked, then narrowed — ~129 rows a day, ~23,000 at the widest window, on the migration's `(snapshot_date, netuid)` index.

## Verified against production before the module existed

Two independent replays that agreed exactly: the pallet rule expressed in SQL over the whole table, and the builder run over the full 129-row day.

```
2026-08-15   112 ranked   16 immune   next_to_deregister 36
             sn36 rank 1   sn74 rank 38   sn70/sn86/sn16 immune, rank null
```

**And that replay found what the series is for.** Netuid 70:

| day | registered_at_block | moving_price | standing |
|---|---|---|---|
| 08-10 | 7,787,562 | 0.001424 | **rank 1** |
| 08-11 | 7,787,562 | 0.001357 | **rank 1** |
| 08-12 | **8,825,571** | **4.0e-8** | immune, 863,870 blocks |
| 08-13 | 8,825,571 | 0.000439 | immune |
| 08-14 | 8,825,571 | 0.001523 | immune |
| 08-15 | 8,825,571 | 0.002866 | immune |

A registration block **130 blocks before that day's pin**. It was deregistered and re-registered, exactly as the ranking had it two days earlier, and its price has climbed back every day since. Netuid 36 inherited rank 1 and has fallen on all six days: 0.0018149 → 0.0014414.

None of that is visible in a single day's answer. Those rows are the test fixtures.

## Three things the surface refuses to smooth

**`rank` is NULL while immune** — an immune subnet holds no position in the prunable order, and reporting one invents a standing the chain never granted. `immune` rides beside it so a caller can tell that from an unreadable rank. `validate-api` asserts the invariant against the live response: immune ⟹ no rank, prunable ⟹ `1 ≤ rank ≤ ranked_count`.

**A rank without its field size is not a fact.** 94 of 100 and 94 of 128 are different standings, so `ranked_count` never travels alone.

**A rank that was not re-measured looks exactly like a rank that held steady**, and only one of those is good news. `pinned_block` rides on every point, each declares `repeats_previous_observation`, and `distinct_observations` counts distinct pinned blocks — never ranks compared for equality.

`comparison_price` (what the pallet compares — a flat 1.0 for a Stable subnet) sits beside the raw `moving_price`, so the substitution is visible rather than inferred. Netuid 86 is the case that makes it matter: no price at all, which the pallet's `ValueQuery` compares as 0 — **rank 1 on price alone** — and immune, so it is not in the order at all. That is the mis-ordering #10285 was filed about, pinned as a test.

## The UI

The panel sits directly under the current standing it gives direction to, in the existing "Lifecycle & deregistration standing" section. It says **"closer to the bar by 31"** rather than an arrow or a signed number, because rank 1 is *next to be pruned* — so a rank that fell is a subnet in more danger, and every visual shorthand for that reads backwards to half the people who see it. An immune day reads as `immune`, not as `—`.

## Registration

| | |
|---|---|
| contracts | `artifact()` + `route()` + `MAINNET_ONLY_ROUTE_PATHS` |
| router | path pattern, dispatch branch, `isMainnetOnlyApiPath` |
| storage | `R2_ONLY_PATTERNS` — a baked copy would freeze one version of the pallet rule, which is the thing the lane's design exists to prevent |
| query | `schemas-src/route-queries.ts`, so an undeclared `?window=` 400s |
| MCP | `get_deregistration_ranking_history` + route-map + `COLLECTIONS_UNEXERCISED_REASONS` (the harness binds no Postgres) |
| docs | `backend-artifact-contracts.md` + regenerated `apps/ui` reference page |
| UI | types, query, panel, mounted, `validate:ui-route-coverage` back at 0 |

The `unreferenced-exports` ratchet caught four new exports; rather than raising it, `DeregistrationHistoryArtifact`/`Point` now type the builder against the published contract, `GetDeregistrationHistoryInput` types the MCP handler, and the unused output alias is gone. **Still 731, at the ceiling.**

## Validation

```
tsc --noEmit  (root + apps/ui)          clean
eslint + prettier  (root + apps/ui)     clean
npm run build                           openapi + types + client regenerated
26 validators incl. api, contract-drift,
  ui-route-coverage, mcp, mcp-route-map,
  docs, unreferenced-exports            all pass
vitest  9 adjacent suites                 154 passed (30 new)
vitest  apps/ui                         1,407 passed (11 new)
patch coverage (git diff -U0 ∩ lcov)    72/72 lines, 69/69 branches = 100%
```

Closes #10296


JSONbored:
`validate` caught a real duplicate on the first run, and it was right: the artifact envelope here is the SAME ten-key shape `/subnets/{netuid}/emission-pipeline/history` declares — `point_count`, `distinct_observations`, `oldest_day`/`newest_day`, `first_captured_day`, `degraded`.

That is not a coincidence of field names, it is one honesty contract for a one-subnet daily series, so it is now declared once in `schemas-src/routes/daily-subnet-series.ts` and both routes build their envelope from it around their own point schema. The per-route DESCRIPTIONS stay per-route — the shape is shared, but what `point_count` counts on a given surface is that surface's own sentence.

`npm run build` produces **no change to `public/`** beyond the new route, so `PipelineHistoryArtifact` is byte-identical to what it published before the refactor. The duplicate count fell 5 → 4.

Issue #10818, opened 2026-08-11
data: three tables are written on a schedule and read by nothing


JSONbored:
Part of #10806.

Three tables are written on a schedule and read by nothing in `src/` or `workers/`:

| table | writer | reader | in `NEON_SOLE_STORE_TABLES`? | in `TABLE_FRESHNESS`? |
|---|---|---|---|---|
| `subnet_deregistration_daily` | cron `17 5 * * *` → `src/subnet-deregistration-daily.ts:176` | **none** | **no** | **no** |
| `subnet_identity` | `src/hyperparams-identity-neon-write.ts:148` | **none** | yes | **no** |
| `attribution_candidates` | `src/attribution-sweep.ts:238` | **none** | yes | **no** |

`subnet_deregistration_daily` is invisible on **three** axes at once — no reader, outside the
ownership flag, outside the freshness sweep — so a lane that silently stops writing it would be
noticed by nothing. `/api/v1/chain/deregistration-ranking` is computed live from the economics blob,
not from this table.

Storing inputs rather than derived ranks is a deliberate, well-argued choice
(`src/subnet-deregistration-daily.ts:1-31`) and this issue does **not** propose reversing it. The gap
is that an accumulate-for-later table with no consumer and no watchdog is indistinguishable from a
dead lane.

Note `attribution_candidates` is currently **empty** (0 rows) despite `attribution_sweeps` having 128
rows written as recently as 16:27 — worth confirming that is expected before deciding.

## Options

1. Give each a reader (the honest one for `subnet_deregistration_daily` — it exists to answer a
   question nobody is asking it yet)
2. Keep them accumulating, but add them to `TABLE_FRESHNESS` (#10817) so silence is visible
3. Retire

## Acceptance

- [ ] Each of the three has a reader, or a declared "accumulating, watched" status
- [ ] `attribution_candidates` being empty is explained or fixed

PR #11350, opened 2026-08-15
feat(review): the attribution sweep's queue, in front of the human it is for (#11227)

### LINKED ISSUE #11227: attribution: candidates will now accumulate with no reader — the lane's whole point is putting them in front of a human
#10818 fixed the sweep so it actually fetches sources. It will now start writing rows to `attribution_candidates`, and **nothing reads that table**.

```
grep -rn attribution_candidates src workers scripts
  src/attribution-sweep.ts:232    INSERT INTO attribution_candidates   <- the writer
  src/read-store-tables.ts:105    (declared readable)
  src/table-freshness-watchdog.ts (watched, maxAgeMs null, coveredBy attribution-sweep)
  scripts/validate-archive-policy.ts:134  "serving" tier
```

No route, no handler, no artifact. The module's own header is explicit about why that matters:

> An ss58 string appearing on a team's page does not make the address theirs [...] Clearing `docs/nametag-evidence-bar.md` is a human judgement, and **this lane exists to put candidates in front of one**, not to skip it.

A candidate nobody can see is the same as no candidate, so until there is a reader the fix in #10818 converts one silent failure (finding nothing) into another (finding things nobody looks at).

## What it needs

A review surface, not a public one. These are unverified strings scraped off third-party pages — the evidence bar is explicitly a human judgement, and publishing them as though they were attributions is the exact error the lane refuses to make. The existing `review/*` artifacts are the natural home; `list_review_gaps` and `list_enrichment_evidence` are the shape.

Should carry, per candidate: the ss58, the `source_url` it was found on, the netuid, and `first_seen`/`last_seen` — so a reviewer can go and look at the page rather than trust the row.

## Worth checking when it lands

Whether the sweep produces a usable signal-to-noise ratio at all. The known false positive is a validator hotkey inside an API response — somebody else's key, published by them — and `subnet-api` is 1,948 of the 2,900 sweepable surfaces. If most candidates are that, the fan-out is worth narrowing before a human is asked to read them.

Follows #10818. Related: #11226.

JSONbored:
## The lane's whole point was putting candidates in front of a human, and nothing could see them

#10818 fixed the sweep so it actually fetches sources. It started writing to `attribution_candidates`, and `grep` found exactly one reader: the table-name constant. No route, no handler, no artifact.

`src/attribution-sweep.ts`'s own header on why that is not a small gap:

> Clearing `docs/nametag-evidence-bar.md` is a human judgement, and **this lane exists to put candidates in front of one**, not to skip it.

So the fix converted one silent failure — finding nothing — into another: finding things nobody looks at.

## Every row is a lead, and the surface is built so it cannot read as one

An ss58 appearing in the text of a page a subnet published does not make the address theirs. The common false positive is a **validator hotkey inside an API response** — somebody else's key, published by them — and nothing in the table distinguishes it from a treasury.

- lives under `/review/`, beside the boards whose job is handing a person something to check
- `source_url` is **required** on every candidate, because the review *is* opening it
- the UI renders the address **whole**, not as a truncated badge: a reviewer compares it against the page
- the section header says *lead*, and the empty state says what an empty queue means

## The listing rule is re-derived at read time, from the table

The sweep already drops a source yielding more than `LISTING_ADDRESS_CAP` distinct addresses — "this page is a listing" is a fact about the page. That cap is enforced on **write**. It is derived again here, over the data, for three reasons:

**Rows outlive rules.** Measured against production 2026-08-15:

```
4,913 rows   87 sources   3,946 addresses   55 subnets
  25 pre-cap sources account for 4,751 of them
     https://77.creativebuilds.io/allHolders      1,230 addresses
     https://provider.engy.ai/api/subnet/v1/miners  388
     https://mantis123.com/dashboard/api/snap/miners 386
```

A reader that trusts the writer serves a queue that is 97% strangers' keys forever, and no backfill makes that untrue of the *next* rule change.

**The cap can move** — its own comment calls it "a judgement, not a law", calibrated on an empty band in one day's distribution. A read-time derivation moves the whole history with it.

**It costs nothing.** A `GROUP BY` over a table already being read.

Applying it leaves **162 rows across 49 subnets from 62 sources** — which the module's own comment independently predicted at "61 URLs and 161 candidates across 48 subnets", a day apart. That is the difference between a queue and a landfill.

## The suppression is published, never silent

`suppressed_count`, `suppressed_source_count` and `listing_address_cap` ride on every response. A filter a caller cannot see is one they cannot check — and the suppressed share is the number to watch: if it stops falling, the sweep's fan-out needs narrowing before a human is asked to read more. The issue asks for exactly that check, and this is what makes it measurable.

## Counts are unbounded, the list is bounded

`reviewable_count` is measured over the whole table beside a `?limit=`-trimmed `candidates` array. The older registry list routes publish no totals at all, so counting a limited fetch silently reports the limit as the population; `validate-api` now asserts `candidate_count <= reviewable_count` against the live response, and the UI reads the total off the payload rather than from `rows.length`.

A failed COUNT degrades the counts to `null` — never to the page length, and never to `0`, which would assert the sweep has found nothing. That negative is the lane's most important result and must not be manufactured by a broken read.

## Two things the gates caught, both worth having

**`ExternalLink`, not a bare anchor.** `no-restricted-syntax` refused the raw `<a href>`. It is right, and it matters more here than usual: these URLs are **untrusted third-party strings scraped off pages the registry does not control**, and that component is what runs `safeExternalUrl` over them.

**The tool's bounds are the route's.** `mcp-input-parity` caught a hand-written `offset` that was looser than the route's published `maximum`. The input schema now reads `ROUTE_QUERY_SCHEMAS["/api/v1/review/attribution-candidates"].shape.*` — a looser bound on one surface is simply the way around the other's.

## Registration

contracts (artifact + route + `MAINNET_ONLY_ROUTE_PATHS`), router branch + `REGISTRY_ONLY_API_PATHS`, `R2_ONLY_PATTERNS` (a baked copy would freeze the queue **and** bake the rule's output rather than the rule), `route-queries.ts`, `validate-api`, MCP tool + route-map + `COLLECTIONS_UNEXERCISED_REASONS`, `backend-artifact-contracts.md`, regenerated reference page, and a section on `/gaps` beside the other review boards.

`unreferenced-exports` is **731, at the ceiling** — the new types went to work typing the builder against the published contract rather than raising it.

## Validation

```
tsc --noEmit  (root + apps/ui)          clean
eslint + prettier  (root + apps/ui)     clean
npm run build                           openapi + types + client regenerated
26 validators                           all pass
vitest  10 adjacent suites                538 passed (28 new)
vitest  apps/ui                         2,308 passed
patch coverage (git diff -U0 ∩ lcov)    55/55 lines, 79/79 branches = 100%
```

## Not in scope

The 4,751 pre-cap rows still sit in the table. The read-time rule makes them harmless to the queue, so this is storage rather than correctness — but pruning them belongs to the lane, not the reader, and a `DELETE` against production is a maintainer's call rather than a PR's.

Closes #11227


JSONbored:
`validate` caught the published prose, and it was right.

The route description said *"the common false positive is a validator hotkey inside an API response"*. `scan:public-safety` flags space-joined `validator hotkey` in prose deliberately — the snake_case `validator_hotkey` field name is exempt as benign Bittensor API naming, but the space-joined phrase is exactly the shape it exists to catch, and `public/llms-full.txt` publishes the route description verbatim.

Reworded to say the same thing without the flagged construction, on every surface that carries it — contracts, the response schema, the MCP tool description, and `backend-artifact-contracts.md`:

> the common false positive is a hotkey belonging to a validator, appearing inside an API response that validator publishes — somebody else's key, on their own page

That is arguably clearer about *why* it is a false positive, which is the point of the sentence. Rebuilt, rescanned (`Public-safety scan passed`), and regenerated the reference page.

Issue #10709, opened 2026-08-11
arch(cron): dispatch on declared cadence, not on a literal cron string — one free minute remains


JSONbored:
`dispatchScheduled` in `workers/api.ts` keys on the **literal cron string** — 75 `if (cron === X_CRON)` branches against 39 declared cron entries. Two lanes sharing an expression means the first branch wins and the second **silently never runs**.

## The immediate constraint

Only **two minutes were free out of 60**, and both are now spoken for:

```
free for a new every-hour cron: [17, 26]
```

17 went to the origin-reachability lane (#10548). **One slot remains.** The lane after that has nowhere to go, and the failure mode when someone reuses a minute is silence, not an error.

The uniqueness test caught exactly this during #10548 — a new lane collided with `EMISSION_DRIFT_CHECK_CRON` (`9,39 * * * *`). It catches exact duplicates after the fact; it cannot create a slot.

## The shape of the fix

One dispatcher cron plus a lane registry — `{ name, everyMinutes, run }` — with the dispatcher running whichever lanes are due. That decouples "how often should this run" from "which minute is free" and removes the collision class entirely.

`lane_health` already records `(lane, verdict, checked_at)` per run, so **last-run time is already available** — `MAX(checked_at) WHERE lane = ?`. The cadence data needed exists.

## The tradeoff that needs deciding before this is built

The dispatcher's **tick frequency bounds every lane's cadence**, and running several lanes in one invocation shares one Worker execution budget.

- A single hourly tick (minute 26) means no lane can run more than hourly, and three lanes sharing one invocation. The attribution sweep alone does up to 64 outbound fetches, the origin lane up to 36 — plausibly over budget together.
- Running **one lane per tick** keeps each invocation small but divides the tick rate among the lanes: 4 ticks/hour across 3 lanes is ~1.3 runs/hour each.
- Finer granularity needs a more frequent tick, and every frequent expression (`*/5`, `*/15`) is already taken — so the dispatcher would have to piggyback inside an existing branch, coupling its reliability to that lane's.

Migrating the three newest lanes frees minutes 17, 24 and 56, which gives the dispatcher up to 4 ticks/hour to work with. That is enough to start, and it is the natural first migration: they are the three I know best and the only ones with no legacy behaviour to preserve.

## Why this is not a drive-by change

[...]

Issue #10329, opened 2026-08-09
health: account_identity's freshness bound is HALF its producer's cadence, so it alarms for 12 hours of every 24


JSONbored:
Found while auditing `lane_health` after #10232 landed. `table-freshness` is reporting **`stale`** right now:

```
table-freshness   stale   account_identity 21.6h > 12h
```

It is not stale. The producer runs on a **24-hour** cadence.

## The numbers

`metagraphed-infra`, `roles/indexer-rust/tasks/main.yml:129`:

```yaml
ACCOUNT_IDENTITY_POLL_SECS: "{{ account_identity_poll_secs | default('86400') }}"
```

No group_vars override exists, so the default is what runs: **86,400s = 24h**. `src/table-freshness-watchdog.ts:169` declares:

```ts
account_identity: { column: "captured_at", kind: "ms", maxAgeMs: 12 * HOUR, reason: "identity sync" },
```

**0.5 ticks.** The table is *guaranteed* to breach the bound for the second half of every cycle, forever. Confirmed against Neon: `account_identity` holds 502 rows at `captured_at = 2026-08-08T23:09:12Z`, and the `account-identity` lane's own verdict for that pass is `ok | 129 scanned, 456 written, 0 error(s)`. The producer is healthy; the bound is wrong.

This is #9301's shape — an alarm that is always on is an alarm nobody reads — and it is sitting next to real signals in the same watchdog's output.

## The rest of the table is fine, which is worth saying

Every poller-backed bound in `TABLE_FRESHNESS`, measured against the `*_POLL_SECS` the container actually reads:

| table | bound | cadence | ticks |
|---|---|---|---|
| **account_identity** | **12h** | **24h** | **0.50** ⟵ |
| neurons / neuron_daily / account_position_daily | 2h | 0.25h | 8.00 |
| subnet_hyperparams | 12h | 1h | 12.00 |
| account_balances (+passes) | 24h | 6h | 4.00 |
| hotkey_alpha (+passes) | 60h | 24h | 2.50 |
| nominator_positions / validator_nominator_counts (+passes) | 36h | 24h | 1.50 |

One entry below one tick; nothing else under 1.5. So this is a single wrong number, not a systemic mis-sizing.

## But two `reason` strings are wrong, and that is how the number got wrong

```ts
account_balances: { maxAgeMs: 24 * HOUR, reason: "12h producer; matches ACCOUNT_BALANCES_STALENESS_THRESHOLD_MS x2" },
[...]

Issue #9632, opened 2026-08-06
top-holders: holdings columns compose live but republish once a day, and the comment justifying that cadence is now false


JSONbored:
## Measured

Served right now (`GET /api/v1/accounts/top-holders?sort=total_tao`):

```
captured_at  2026-08-06T01:34:44.014Z
5EYCAe5jLQhn6ofDSvqF6iY53erXNkwhyE1aCEgvi1NNs91F  free_tao 5455237.07854851
```

Live in D1 at the same moment (`account_balances`, `captured_at` 1785991125923 = 04:38:45Z, a **complete** pass of 364,568/364,568):

```
5EYCAe5jLQhn6ofDSvqF6iY53erXNkwhyE1aCEgvi1NNs91F  free_tao 5455012.470851503
```

A **~225 TAO** difference on the top holder, and the served figure is three hours behind data that is already complete and sitting in the table. By the end of a day it is up to 24 hours behind.

## Why

`TOP_HOLDERS_FLOW_CRON = "34 1 * * *"` — once daily. That cadence was chosen deliberately and the reasoning is recorded in `workers/config.ts`:

> the windows are 7/30/90 days, and the sibling holdings columns it sits beside are a fixed 2026-08-02 snapshot

**That second clause is no longer true.** #9502 and its follow-ups retired the frozen artifact for `free_tao`, `delegated_tao` and `total_tao`; they are composed live from `account_balances` (refreshes every ~1–4 h) and `hotkey_alpha` (~20 min). The premise that made daily obviously sufficient was removed, and the cadence was never revisited.

This is a stale comment doing real damage: it is the rationale the next person reads before deciding whether the cron is right.

## The shape of the fix

One cron currently publishes two things with very different costs and very different natural cadences:

| columns | cost | how fast the source moves |
|---|---|---|
| `net_flow_7d/30d/90d` | **1.65 GB** scan — `GROUP BY coldkey` over the 90-day `account_events` window (measured 2026-08-05) | 7/30/90-day windows — daily is genuinely right |
| `free_tao`, `delegated_tao`, `total_tao` | D1 reads, cheap | every ~1–4 hours |

So the fix is **not** to run the whole lane more often — that multiplies 1.65 GB/day by the new frequency for columns whose windows do not move. It is to let the holdings columns republish on their own cadence without re-running the flow scan.

[...]

PR #11353, opened 2026-08-15
test(cron): a lane's declared cadence must be the cron it runs on

JSONbored:
## A lane declares its interval twice, and nothing compared them

A lane whose producer is a cron in this repo states how often it runs in two places:

```ts
// workers/config.ts
export const TOP_HOLDERS_HOLDINGS_REFRESH_CRON = "49 */3 * * *";

// src/producer-cadence.ts
top_holders_holdings: 10_800,
```

I added both, in the same change (#9632), this afternoon. **Nothing would have failed if one had said three hours and the other six.**

## The consequence is a wrong alarm, not a late lane

`PRODUCER_CADENCE_SECS` is what every staleness watchdog sizes its bound from — `missedTicksMs(lane, n)`. A cadence that disagrees with the cron does not make the producer slow. It makes the **alarm** wrong, in whichever direction the error points:

| drift | result | precedent |
|---|---|---|
| declared **too long** | the bound never fires; a dead lane is silent | #10566 |
| declared **too short** | a healthy lane reads `stale` for most of every cycle | #10329, #9301 |

Both are invisible from either file, because each one is internally correct. #10329 is exactly this: `account_identity` bounded at twelve hours against a twenty-four-hour poller — half a tick, so the lane read `stale` for the back half of every cycle. That cost a real alarm and the trust in it.

## `cronIntervalSecs` returns null rather than a guess

The parser handles the shapes this repo's lane crons actually use, and refuses everything else:

```
"34 1 * * *"                → 86400     "1,16,31,47 * * * *"  → null  (uneven)
"49 */3 * * *"              → 10800     "0,1,2 * * * *"       → null  (uneven across the wrap)
"26 * * * *"                →  3600     "0 */5 * * *"         → null  (5 ∤ 24)
"11,41 * * * *"             →  1800     "*/10 */2 * * *"      → null  (two cadences)
"3,13,23,33,43,53 * * * *"  →   600     "0 0 * * 1"           → null  (narrows the day)
```

The wrap case is the one worth naming: `0,1,2 * * * *` is evenly spaced *within* the hour and leaves a 57-minute gap once an hour. A parser that only checked consecutive gaps would call that a one-minute cadence and put a one-minute bound under a watchdog — the failure this exists to prevent, arriving through the tool meant to prevent it. So an unmeasurable shape is reported as unmeasurable and the caller decides.

## Deliberately partial, and it says so

`WORKER_CRON_LANES` names its two lanes explicitly rather than deriving the set. Most of `PRODUCER_CADENCE_SECS` is transcribed from metagraphed-infra's ansible defaults — the module header already says this repo cannot check those — and inventing a local expression for them would be a second copy of the thing this stops. A test pins the map's contents, so dropping a lane from it removes the check **loudly** rather than quietly.

## Proved it can fail before landing it

A gate that has only ever passed is not evidence. Setting the cadence to `21_600` against the 3-hourly cron:

```
AssertionError: top_holders_holdings declares 21600s but
TOP_HOLDERS_HOLDINGS_REFRESH_CRON = "49 */3 * * *" fires every 10800s
```

Restored, 22 tests pass.

## Relationship to #10709

That issue proposes replacing 75 `if (cron === X)` branches with a lane registry keyed on declared cadence. This is not that, and does not pre-empt it — it closes one live drift the current design cannot see, and it makes the registry's job smaller by establishing that a lane's cadence and its schedule must agree. Commented there separately with a correction to that issue's free-minute count: the scarce unit is the literal **string**, not the minute, so the space is larger than "one slot".

Not building the dispatcher in this session, on that issue's own advice — getting it wrong stops 39 producers with no error anywhere.

## Validation

```
tsc --noEmit           clean
eslint + prettier      clean
validate:boundary-casts, unreferenced-exports/-modules,
  lane-topology, contract-drift                       pass
vitest  producer-cadence, lane-silence-cadence,
  top-holders-staleness-watchdog, api-crons-have-handlers   84 passed (15 new)
gate proven to fail on a real drift, then restored
```

Refs #10709

### DIFF
--- a/src/producer-cadence.ts
+++ b/src/producer-cadence.ts
@@ -127,6 +127,126 @@ export const PRODUCER_CADENCE_SECS = {
 export type ProducerLane = keyof typeof PRODUCER_CADENCE_SECS;
 
 /**
+ * Which of those cadences belong to a WORKER CRON in this repo, and to which
+ * `workers/config.ts` constant (#10709).
+ *
+ * ## The drift this closes
+ *
+ * Most entries above are transcribed from metagraphed-infra's ansible defaults
+ * — this repo cannot check those, and the header says so. But two of them are
+ * OURS: the lane runs on a cron declared a few files away, so its interval is
+ * stated TWICE, in two languages, with nothing comparing them.
+ *
+ * That is not hypothetical. #9632 added `top_holders_holdings: 10_800` beside
+ * `TOP_HOLDERS_HOLDINGS_REFRESH_CRON = "49 *\/3 * * *"` in the same change, and
+ * nothing would have failed if one had said three hours and the other six.
+ *
+ * ## Why the consequence is an alarm rather than a stale lane
+ *
+ * The number here is what every staleness watchdog sizes its bound from —
+ * `missedTicksMs(lane, n)`. A cadence that disagrees with the cron does not
+ * make the producer late; it makes the ALARM wrong, in whichever direction the
+ * error points:
+ *
+ *   - declared too LONG  → the bound never fires, and a dead lane is silent.
+ *     That is #10566's shape.
+ *   - declared too SHORT → a healthy lane reads `stale` for most of every
+ *     cycle. That is #10329 (`account_identity` bounded at half a tick) and
+ *     #9301, both of which cost a real alarm and the trust in it.
+ *
+ * Both are invisible from either side, because each file is internally correct.
+ *
+ * ## Deliberately partial
+ *
+ * Only the lanes whose producer is a cron in THIS repo can be checked, so this
+ * map names them explicitly rather than deriving the set. An infra-poller
+ * cadence has no local expression to compare against, and inventing one would
+ * be a second copy of the thing this exists to stop.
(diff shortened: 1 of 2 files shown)


JSONbored:
`codecov/patch` was right — two branches of `cronIntervalSecs` had no test, and both are real cron shapes rather than defensive dead ends:

**A stepped minute inside a fixed hour.** `"*/10 1 * * *"` fires six times inside one hour and then not for twenty-three. Reading the minute step as the cadence would bound a daily lane at ten minutes — the #10329 direction, and the worse of the two, because it alarms on a lane that is working.

**An hour range or list.** `"0 1-5 * * *"` and `"0 1,13 * * *"` are ordinary cron and have no single interval; they must fall through rather than be measured by whichever branch is nearest.

Both now assert `null`, with the single-fixed-minute case (`"34 1 * * *"` → 86400) beside them so the assertion is not passing on a function that returns null for everything.

Patch coverage: **31/31 lines, 44/44 branches = 100%**, measured by intersecting `git diff -U0 origin/main...HEAD` with `coverage/lcov.info`.

Issue #9475, opened 2026-08-05
top-holders watchdog suppresses its alert on the one condition that is actually true


JSONbored:
Follow-up to #9464, correcting a judgement call in that PR.

## What #9464 shipped

`src/top-holders-staleness-watchdog.ts` classified the leaderboard artifact and, for one verdict, deliberately declined to notify:

```ts
if (generatedAt === TOP_HOLDERS_FROZEN_GENERATED_AT) {
  return { stale: true, alert: false, reason: "frozen", ... };
}
```

`TOP_HOLDERS_FROZEN_GENERATED_AT` was a committed copy of the artifact's own `generated_at` (`2026-08-02T22:38:17.501738+00:00`). When it matched, the tick wrote a `lane_health` row and sent nothing to PostHog. The stated reasons were exception cost and alarm fatigue.

## Why that was wrong

**The cost argument does not survive arithmetic.** Twice hourly is 48 events/day against a project running ~1M/day — 0.005%. The production storm guard is `POSTHOG_EXCEPTION_STORM_WINDOW_MS: "300000"` (5 minutes), so at 30-minute spacing it never engages either way. There was no cost being traded for anything.

**It made the watchdog silent on the exact defect it was built for.** #9464 exists because a frozen leaderboard went three days unnoticed. Answering that with a row in a table nobody queries lets the next three days pass identically. `lane_health` is the *record* — #9330/#9340 were explicit that it is not a substitute for the notification — and #9464 used it as one.

**Suppressing on a hardcoded timestamp is a known-bad-state allowlist.** It disables the alarm for precisely the condition that is true right now, indefinitely, and a second frozen materialization would have to be hand-added to keep it quiet. Nothing else in this repo suppresses an alarm that way. The two sibling lanes this watchdog is modelled on — #9273 (nominator-positions) and #9301 (validator-nominator-counts) — both alerted on every tick while their lanes were dead, and both lanes got fixed.

The "alarm fatigue" citation was also misapplied. #9273's header warns against a threshold *mis-sized against a healthy producer*, so that a working lane alerts three quarters of the day. That is a false positive. A leaderboard nothing refreshes is a true positive, and the way to silence a true positive is to fix it.

## What to do

Delete the special case and the constant. A stale leaderboard is `stale`, it reports on every tick, and it keeps reporting until the lane gets a producer or the route is withdrawn. `alert` collapses back into `stale`, which also removes a field that only existed to carry the exemption.

[...]

Issue #9301, opened 2026-08-03
nominator_count is frozen at 2026-08-02: the counts lane has a producer but no Cloudflare-native sink


JSONbored:
`nominator_count` on `/api/v1/validators` and `/api/v1/validators/{hotkey}` is served from the lakehouse mirror of `chain.validator_nominator_counts` (#9276). That mirror is frozen and can only decay.

Measured live 2026-08-03 against the lakehouse via R2 SQL:

- 112,550 rows across 112,550 distinct hotkeys, 18 distinct `captured_at` values, newest `2026-08-02T01:38Z` — the retired box's last scan.
- Of the 1,031 validators on `/api/v1/validators?limit=2000`, only **564** get a count; **467** have no row at all.
- `SELECT count(*) ... WHERE nominator_count = 0` returns 0 — the scan never recorded a zero, so an absent row conflates "no nominators" with "not scanned".

Coverage can only fall from here as new validators register.

## Why this is a sink problem, not a producer problem

The producer was already ported off the box. `metagraphed-infra`'s poller Container carries `src/bin/poller/jobs/validator_nominators.rs`, which performs the same full `SubtensorModule::Alpha` scan on a 24h tick. It is one of the five lanes `Dockerfile.poller` leaves disabled, verbatim:

> Postgres-free job set only … The five Postgres-backed lanes stay disabled until they have a **Cloudflare-native sink**.

So the scan needs no rehosting — only its write target does. This is the same shape as #9273/#9292, which gave `nominator_positions` (the *other* output of that same scan) a D1 sink.

## Scope

- A D1 table + writer for `validator_nominator_counts`, reusing `src/neurons-d1-write.ts`'s bound-parameter budget.
- Un-retire `handleValidatorNominatorCountsSync` (503 since #9193) to write there.
- Read it in data-api's `buildGlobalValidators`/`buildValidatorDetail`, which are currently handed a hardcoded empty map / null count. That makes #9276's serving-Worker lakehouse overlay a no-op rather than a conflict — it only ever fills a count the payload left null.

Part of #9146

PR #11356, opened 2026-08-15
fix(alarm): nominator-positions is bounded under its producer's real ceiling

JSONbored:
## The constant claims something production disproves

> 30h is one missed pass plus slack for the scan itself (~4 minutes) and cron jitter. **It fires only once a pass has genuinely been skipped.**

It does not. Measured over 29 consecutive pass gaps of `nominator_positions.captured_at`, today:

```
min 0.66h    p90 24.41h    MAX 34.57h
4 gaps exceed the 24h cadence;  2 exceed the 30h bound

worst: 08-08 23:11 (34.6h),  08-05 06:35 (31.4h),
       08-11 00:47 (25.6h),  08-15 11:47 (24.1h)
```

**A 34.6h gap is not a skipped pass** — a skipped pass on a daily poller is ~48h. It is the pass running **late**, which is what a container running its lanes sequentially does when something ahead of this one runs long.

So the premise the constant rests on — *"a healthy lane presents an age anywhere in [0h, 24h+scan]"* — is measurably false. The observed range is **[0.66h, 34.57h]**.

## What it cost

`lane_health`, 535 ticks:

| verdict | ticks | worst age |
|---|---:|---|
| `ok` | 519 | 29.95h — *just* under the bound |
| `stale` (detail `stale`) | **9** | 30.02h → 33.02h |
| `stale` (detail `partial`) | 7 | 0.45h → 3.43h |

All nine age-based alerts trace to the two late passes above. **Not one of them is a producer that stopped.**

An alarm firing on a working lane 3% of the time is exactly the #9301 failure this constant has already been corrected for once — corrected then from 6h, and still left under the real ceiling. The `ok` worst case of 29.95h against a 30h bound shows how little margin was left.

## 36h, and the cost stated rather than hidden

1.5 ticks. Clears the measured 34.57h maximum by ~1.4h.

A writer that stops entirely is now reported **six hours later** — a day and a half rather than a day and a quarter. Against a daily producer that is still the same day, and it buys an alarm that means something when it fires.

**This is not suppressing a known-bad state**, which #9475 rightly refuses. The lane is healthy in all nine cases; the bound was describing a cadence the producer does not have.

## Pinned into a corridor, not at a number

Two assertions, so it can move neither way without a reason:

```
threshold 108000000ms is at or under the observed 124452000ms worst case
  -- a healthy lane will read stale                        # at 1.25 ticks

a bound at or above two full ticks cannot distinguish a late pass
  from a skipped one                                       # at 2 ticks
```

Both proven to fail before landing, then restored. That is stronger than pinning `36 * HOUR`: it says *why* 36 and what would make it wrong.

## Fixtures derived, not literal

The watchdog's own tests hard-coded `31 * HOUR` and `34 * HOUR` as "stalled", plus `/34\.0 h old/` and `/threshold 30 h/` in a message assertion. Every one of those had already been rewritten once for the 6h → 30h move, and they fail for the *wrong reason* each time the number under test changes. They now derive from `NOMINATOR_POSITIONS_STALENESS_THRESHOLD_MS`, so the next re-size touches the constant and its rationale only.

The `thresholds are unchanged by the refactor` pin in `producer-cadence.test.ts` did its job — it failed the moment the bound moved, which is what a pin against live alarm numbers is for. Widened deliberately, with the measurement written into it.

## Not in scope

The other 7 stale ticks carry `detail: "partial"` at ages 0.45–3.43h — a **coverage** fault on 2026-08-14, not an age one. Different rule, different evidence, left alone.

## Validation

```
tsc --noEmit                                        clean
eslint + prettier                                   clean
validate:unreferenced-exports, lane-topology,
  contract-drift                                    pass
vitest nominator-positions-staleness-watchdog,
  producer-cadence, lane-silence-cadence,
  lane-health                                       95 passed (2 new)
both corridor assertions proven to fail, restored
```

### DIFF
--- a/src/nominator-positions-staleness-watchdog.ts
+++ b/src/nominator-positions-staleness-watchdog.ts
@@ -68,7 +68,7 @@ import { requireFullScanValue } from "./lane-table-topology.ts";
 /**
  * How old the ledger may get before this is a stall.
  *
- * THIRTY HOURS, not the neurons lane's 45 minutes, and the difference is the
+ * THIRTY-SIX HOURS, not the neurons lane's 45 minutes, and the difference is the
  * work: this lane is a full SubtensorModule::Alpha scan (~153k rows across
  * every coldkey on the network), which is why it never shared the 15-minute
  * metagraph cadence even when it ran.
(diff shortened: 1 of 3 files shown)


JSONbored:
## Swept every staleness bound against its producer's real behaviour. Only this one is mis-sized.

Two threshold defects in one afternoon (this, and the birth alarm in #11355) looked like a class, so I checked the whole set rather than assuming. It is **not** a class — which is worth recording, because the obvious next move would have been to loosen several bounds that are doing their job.

Observed pass gaps against each declared bound, plus the age-based `stale` verdicts each actually produced:

| lane | bound | p90 gap | max gap | age-based `stale` ticks | verdict |
|---|---|---|---|---|---|
| `nominator-positions` | 30h | **24.41h** | **34.57h** | **9** | **mis-sized** |
| `account-balances` | 12h | 6.14h | 23.50h | 25 | **correct** |
| `neurons` | 45 min | — | — | 169 | **correct** |
| `hotkey-alpha` | 48h | 23.99h | 33.16h | 0 | correct |
| `validator-nominator-counts` | — | 28.97h | 31.35h | 0 | correct |

## Why `account-balances` is not the same defect, despite 25 alerts

Its 12h bound looks breached by a 23.50h maximum. It is not routinely:

```
55 gaps    median 5.05h    p90 6.14h
exactly ONE gap exceeds 12h:  23.5h on 08-12 18:57
next worst: 12.0h, then 7.1h, 6.5h, 6.2h
```

The producer sits on its 6h cadence and had **one** stall. All 25 stale ticks come from that single incident, which is an alarm working — the bound is nowhere near the routine ceiling. Loosening it would have suppressed a real 23.5h outage.

`neurons` is the same story at larger scale: 169 stale ticks with a worst age of **28.42h** against a 45-minute bound is a multi-day metagraph outage being reported correctly, not a bound that misreads a healthy lane.

## What makes `nominator-positions` different

Its bound sits **inside the routine distribution**, not above it:

```
p90 24.41h   max 34.57h   bound 30h
4 of 29 gaps exceed the 24h cadence; 2 exceed the bound
ok verdicts reach 29.95h -- 3 minutes of margin
```

The tail crosses the bound as ordinary behaviour, so it fires on a working producer repeatedly rather than on one incident. That is the distinction the corridor assertions in this PR encode: a bound must clear the observed ceiling, and stop short of two full ticks where "late" and "skipped" become indistinguishable.

`hotkey-alpha` shows what a correctly-sized bound looks like on the same shape — a 33.16h maximum under a 48h bound, and zero age-based alerts across 230 ticks.

## Method

Gaps are between consecutive DISTINCT `captured_at` values (for the two lanes with a pass ledger, between COMPLETE passes), against `lane_health` grouped by `verdict` and `detail` so age-based stales are separated from `partial` and `no_rows`, which are different rules with different evidence.

JSONbored:
One correction to my own sweep comment, because it affects how the numbers should be read.

A "pass" for this lane is a **4-hour window** (`NOMINATOR_POSITIONS_PASS_WINDOW_MS`), not a single `captured_at` — the coverage query is `COUNT(DISTINCT coldkey) WHERE captured_at >= MAX(captured_at) - window`. So the producer writes a full scan (~21,236 coldkeys, matching `NOMINATOR_POSITIONS_EXPECTED_COLDKEYS = 21,263`) plus small incremental stamps between scans, and the p90 in my table mixes intra-window stamps with between-scan gaps.

**The claim this PR rests on is unaffected**, because the watchdog measures the age of `MAX(captured_at)` — the wall-clock silence — not a pass boundary. The 34.57h maximum is a stretch with no write of any kind, and `lane_health` corroborates it independently without going through `captured_at` at all:

```
ok     worst age 29.95h   <- 3 minutes under the 30h bound
stale  worst age 33.02h   <- sampled every 30 min, so the true peak is 33.0-33.5h
```

Two independent measurements agreeing on ~33-34.6h is the evidence; either alone would be a hypothesis. The corridor assertion uses 34.57h, which is the conservative (larger) of the two.

The `account-balances` / `neurons` comparisons in that table are unaffected — both are one-`captured_at`-per-pass ledgers.

Issue #11328, opened 2026-08-15
SN101 lost 42% of its announced axons on 2026-08-11 and has not recovered — nothing alarmed


JSONbored:
Found while pricing #10805's derivation against `neuron_daily`. Not an axon-removals-family question — this is an operational fact about a live subnet that no surface reported.

## What happened

`neuron_daily`, netuid 101:

| snapshot_date | with_axon | distinct_hotkeys | emission_tao | permits |
| --- | --- | --- | --- | --- |
| 2026-08-09 | 225 | 256 | 295.1673 | 9 |
| 2026-08-10 | 223 | 256 | 295.0668 | 8 |
| **2026-08-11** | **129** | 256 | 295.1792 | 8 |
| 2026-08-12 | 128 | 256 | 295.2016 | 8 |
| 2026-08-15 | 129 | 256 | 295.2016 | 8 |

**94 neurons stopped announcing an axon on one day and have not resumed in the four days since.**

## It is not a deregistration, and not an economic event

- **256 distinct hotkeys throughout.** The miner set did not change — these are the same UIDs held by the same hotkeys before and after. The day-over-day diff confirms it: the 75 transitions that survive the UID-reuse filter all have `hotkey = prev_hotkey`.
- **Emission is flat** at ~295.17 TAO/day across the boundary, and `validator_permit` holds at 8. The subnet keeps paying, and the miners keep earning, without announcing.

## It is not our capture breaking

This is the part that took checking, because "we stopped reading it" looks identical to "they stopped publishing it":

| snapshot_date | axons network-wide | **excluding netuid 101** | subnets captured |
| --- | --- | --- | --- |
| 2026-08-10 | 6,758 | 6,535 | 129 |
| **2026-08-11** | **6,657** | **6,528** | 129 |
| 2026-08-12 | 6,611 | 6,483 | 129 |

Excluding SN101 the series is flat across the boundary — the whole network-wide dip is SN101. All 129 subnets were captured every day, and SN101's other 21 columns (hotkey, coldkey, stake, emission, rank, permits) are present and consistent on 08-11. A poller failure does not surgically drop one field for one subnet while reading the rest of that subnet correctly.

A capture gap would also have recovered. This has held for four days.

## Why it is worth an issue

An axon is how a miner is reachable. 42% of a subnet's miners becoming unreachable-by-announcement, while continuing to earn, is either:

[...]

PR #11383, opened 2026-08-16
fix(watchdog): an unread mechanism is not a withdrawal — say unknown, not why

### LINKED ISSUE #11381: The axon alarm names a mechanism it failed to measure
`loadAxonLossMechanisms` returns `{}` on any failure — a deliberate safe direction for the *read*. But the finding's default `kind` was `announcements-withdrawn`, so an unreadable mechanism did not degrade to "unknown". It degraded to a confident, specific claim:

> announced axons collapsed: SN25 14/80 axons (18%) -- miners that are still registered and still earning stopped publishing an axon, so this is a reachability change nothing else in the fleet reports

On the two subnets currently flagged that sentence is **inverted**. Measured over the watchdog's own 8-day window, per UID:

| subnet | via_reuse | same_hotkey |
| --- | --- | --- |
| 25 | 67 | **0** |
| 102 | 43 | **0** |

Zero miners went dark on either. Every loss was a deregistration whose replacement never served.

The same fallback also suppressed the detail suffix (`lossesSameHotkey` stays `null`, so `axonDetail` prints the bare ratio), which is what production published at `2026-08-15T19:36:02Z` — leaving no way to tell a failed read from a measured withdrawal.

An alarm may escalate on uncertainty. It may not invent the finding it is uncertain about.

## Fix

Add a fourth kind, `mechanism-unknown`, as the starting value; only a successful per-UID read moves it. A read that succeeds and explains nothing (`reuse + same === 0`) takes the same exit — different fact, same claim. The exception message gates the `#11328` withdrawal wording on `some(kind === "announcements-withdrawn")` rather than on being the last branch left, and turnover stops borrowing it.

The enrichment query itself is healthy in production, so this is latent robustness, not a live outage.

JSONbored:
Closes #11381.

`loadAxonLossMechanisms` returns `{}` on any failure — right for the read. But the finding's starting `kind` was `announcements-withdrawn`, so an unreadable mechanism degraded not to "we could not tell" but to a confident, specific claim:

> announced axons collapsed: SN25 14/80 axons (18%) — miners that are still registered and still earning stopped publishing an axon…

On the two subnets flagged at `2026-08-15T19:36:02Z` that sentence was **inverted**:

| subnet | via_reuse | same_hotkey |
| --- | ---: | ---: |
| 25 | 67 | **0** |
| 102 | 43 | **0** |

Zero miners went dark. Every loss was a deregistration whose replacement never served. The same fallback suppressed the detail suffix too, so a failed read and a measured withdrawal published the identical string.

**An alarm may escalate on uncertainty. It may not invent the finding it is uncertain about.**

## Changes

- `mechanism-unknown` is a fourth kind and the **starting value**; only a successful per-UID read moves it. A read that succeeds and explains nothing (`reuse + same === 0`) takes the same exit.
- The detail names the gap instead of printing a bare ratio.
- The #11328 withdrawal wording is gated on `findings.some(kind === "announcements-withdrawn")` rather than on being the last branch left, so neither an unread mechanism nor a turned-over subnet borrows it. What's left gets its own message: how many went away, deliberately not why.
- A measured **tie** still resolves to withdrawal — that's a measured tie, not an absent measurement.

## Pinned

Two tests assert the message both ways, including `assert.doesNotMatch(/stopped publishing an axon/)` on the unknown path. Five existing assertions encoded the old default and were updated rather than worked around — the one named *"an unmeasured mechanism leaves the default and the plain detail"* was the bug written down as a test.

Mutation-verified: restoring the old starting value fails 7 tests; restoring the ungated message fails 1.

All 55 CI validators pass. Latent robustness — the enrichment query is healthy in production.

### DIFF
--- a/src/axon-announcement-watchdog.ts
+++ b/src/axon-announcement-watchdog.ts
@@ -139,11 +139,26 @@ export interface AxonFinding {
    *   the opposite of what happened.
    * - `subnet-turned-over`: the metagraph itself emptied, so the missing axons
    *   are missing NEURONS and membership is the story.
+   * - `mechanism-unknown`: the per-UID read did not answer. THE STARTING VALUE
+   *   (#11381).
    *
-   * Defaults to `announcements-withdrawn` until the mechanism read runs, which
-   * is the conservative direction: it is the reading that asks for a human.
+   * It used to start at `announcements-withdrawn`, on the reasoning that the
+   * alarming reading is the conservative one. That confuses escalating with
+   * asserting. `loadAxonLossMechanisms` returns `{}` on any failure, so an
+   * unreadable mechanism did not become "we could not tell" -- it became the
+   * sentence "miners that are still registered and still earning stopped
+   * publishing an axon", published as fact. On the two subnets flagged
+   * 2026-08-15 that sentence was INVERTED: 67 and 43 losses via UID reuse, and
+   * zero same-hotkey stops on either.
+   *
+   * An alarm may escalate on uncertainty. It may not invent the finding it is
+   * uncertain about.
    */
-  kind: "announcements-withdrawn" | "churn-replaced" | "subnet-turned-over";
+  kind:
+    | "announcements-withdrawn"
+    | "churn-replaced"
+    | "subnet-turned-over"
+    | "mechanism-unknown";
   /** Axon losses in the window whose UID changed hands. Null until measured. */
   lossesViaReuse: number | null;
   /** Axon losses where the same hotkey stopped announcing. Null until measured. */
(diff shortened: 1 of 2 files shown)


JSONbored:
Superseded by #11382, which landed the same fix while this was in flight — and landed all three halves of it: `mechanism-unknown` as the starting value, the `findings.some(kind === "announcements-withdrawn")` gate on the #11328 wording, and a detail string that names the unread case ("mechanism UNREAD — the drop is real, the cause is not established") instead of printing a bare ratio.

I checked for a delta before closing rather than assuming: there is none worth carrying. My version differed only in wording.

My own miss: I read the issue and started implementing without first checking whether a PR was already open against it. Closing rather than forcing a redundant change through.

Issue #11392, opened 2026-08-16
The axon alarm calls a move to an unroutable address 'stopped publishing an axon'


JSONbored:
`classifyAxonMechanism` splits axon losses two ways: the UID changed hands (`churn-replaced`) or it did not (`announcements-withdrawn`). The second is two mechanisms wearing one name, and the wrong one is the majority.

Measured over the full 38 days `neuron_daily` retains, network-wide:

| same-hotkey loss of a routable axon | count |
| --- | ---: |
| **moved to an unroutable address, still announcing** | **166** |
| truly stopped announcing (`axon IS NULL`) | 105 |
| total | 271 |

The moves **outnumber** the stops. For all 166 the alarm says *"miners that are still registered and still earning stopped publishing an axon"* — and they never stopped. They publish an address in RFC 5737 documentation space or RFC 1918 private space that nothing can reach.

## SN126, which nothing has looked at

Per-subnet, same-hotkey losses of routability over 38 days:

| subnet | withdrawals | days | window |
| --- | ---: | ---: | --- |
| **126** | **160** | 14 | 07-13 → 08-11 |
| 101 | 76 | 2 | 07-19 → 08-11 |
| nine others | 35 total | | |

SN126 is the **largest same-hotkey source on the network** — bigger than the SN101 event #11328 was filed for — and it is almost entirely moves. Run the same query with `axon IS NOT NULL` instead of the routable test and SN126 has **one** loss, not 160. Its miners never went dark.

The watchdog never saw it: the event ended 08-11 and the lane shipped 08-15, and a 7-day trailing baseline only catches things while they are happening.

## Fix

A fourth kind, `moved-unroutable`, taken when a measured majority of same-hotkey losses still announce. The finding carries `lossesMovedUnroutable` as a subset of `lossesSameHotkey`, so `sameHotkey - moved` is the count that genuinely went dark.

It **still pages** — reachability really did change and nothing else in the fleet reports it — but with wording that does not claim anyone stopped.

An unmeasured `movedUnroutable` reads as 0, which leaves today's answer rather than inventing a move; and the count is clamped into the total it partitions, so an unreadable driver value cannot render a negative "stopped".

## Bearing on #10805

This is the third mechanism a derived axon-removals family would have to separate. A naive `prev_has AND NOT has_axon` derivation reports **2,186** removals over 38 days; **1,915 are deregistrations** and **166 more are moves**. The honest count of miners that actually stopped announcing is **105**, across 10 subnets.

PR #11395, opened 2026-08-16
feat(axon): derive the axon-removals family from state, split three ways

### LINKED ISSUE #10805: decide the axon-removals family: derive it from axon state, or retire it
The three axon-removal routes — `/api/v1/chain/axon-removals`,
`/api/v1/subnets/{netuid}/axon-removals`, `/api/v1/accounts/{ss58}/axon-removals`,
plus their MCP tools and GraphQL fields — answer a permanent zero on every
scope, for every subject, in every window. They are the only route family in
the repo with no reader on any surface.

They are already honest about it: every one carries
`degraded: { reason: "axon_removals_never_emitted" }` (`src/uncurated-event-streams.ts`),
so a consumer can tell the zero from a measurement. What is still open is
whether the family should be **derived** or **retired**, and that is a contract
decision rather than a code one.

## The event does not exist; the state transition does

Measured against the lakehouse on 2026-08-03 and recorded in
`src/uncurated-event-streams.ts` — `chain.chain_events` is the COMPLETE
pallet-level stream, 898M rows, genesis to head:

| event kind | chain_events | account_events |
| --- | --- | --- |
| `AxonServed` | 2,792,038 | 2,792,038 |
| `NeuronRegistered` | 1,072,413 | 1,072,413 |
| `PrometheusServed` | 18,041 | 0 |
| **`AxonInfoRemoved`** | **0** | **0** |

Zero occurrences, ever. Unlike `PrometheusServed` — a curation gap the indexer
can close — the routes were modelled on an event the Subtensor runtime does not
emit, so no indexer work populates them. That distinction is why the two carry
different degraded reasons.

But an axon going away is a real on-chain fact; it is just a **state
transition** (`AxonInfo` cleared, or its `block` going stale) rather than an
event. And we already capture that state daily:

- `neurons.axon` and `neuron_daily.axon` are real columns (`db/schema.sql`), and
  `axon` is in `NEURON_HISTORY_FIELDS` (`src/neuron-daily-cold-tier.ts`), so
  every neuron's axon is snapshotted per day.
- It is populated and served today. Live, 2026-08-11, via
  `/api/v1/subnets/{netuid}/metagraph`:

  | netuid | neurons | non-null `axon` | example |
  | --- | --- | --- | --- |
  | 64 | 256 | 6 | `43.155.26.33:10010` |
  | 1 | 256 | 4 | `172.67.154.247:443` |
  | 8 | 256 | 31 | `34.65.245.134:8091` |

So the data the family wants exists — as state, in a table we already write
daily and already read through a cold tier.

## Option A — derive it, the way deregistration already is

`src/deregistration-derivation.ts` derives evictions from UID reuse because the
chain does not emit those either; the same shape applies here. A day-over-day
diff of `neuron_daily.axon` per (netuid, uid) gives a removal when a non-null
axon becomes null, and the existing builders (`buildChainAxonRemovals`,
`buildSubnetAxonRemovals`, `buildAccountAxonRemovals`) already take exactly the
per-netuid rollup rows such a lane would emit — they are tested against those
rows today, so the derivation is the only missing piece.

**The design risk is what makes this non-trivial, and it must be settled before
any lane ships.** With 4–31 of 256 neurons carrying a non-null axon, "non-null
became null" has at least three causes:

1. the neuron cleared its axon on chain — a real removal;
2. the poller failed to read it that day — a capture gap;
3. the UID was deregistered and reused — already a *deregistration*, and
   counting it again here would double-report one event under two names.

A naive diff reports all three as teardowns, which manufactures activity on a
card whose whole purpose is to report it. So the derivation needs to establish
that `axon = null` means "no axon announced on chain" rather than "not
captured" — i.e. confirm what the poller writes when the SDK returns the
never-served `0.0.0.0:0` sentinel — and to subtract UID-reuse transitions the
deregistration lane already owns. Neither is hard; both are the work.

Worth pricing first: with ~129 subnets × 256 UIDs × the retained day count, the
diff is small, but the answer it produces is only as good as the capture
completeness above, so measure how often a non-null axon goes null and comes
back before building anything on it.

## Option B — retire the family

Publishing a permanently-degraded route has a cost too: it appears in the
OpenAPI contract, the MCP tool list, the GraphQL schema and the coverage
counts, and every one of those says the surface exists. If the transition is
not worth deriving, saying so in the contract is more honest than a card that
is correct and useless.

Retiring published routes is a contract decision, so this issue does not
presume it — it records that "publish a confident 0 forever" was never one of
the options, and that the state to derive from is there if we want it.

## Not blocking anything

The routes are marked degraded and their tests pin the absence rather than
asserting a ranking no source can produce, so nothing here is urgent or broken.
This exists so the choice is made deliberately rather than by default.

Surfaced while sweeping the dead tier reads (#10190): axon-removals was the one
family whose deletions left no rung anywhere, on any surface.

### LINKED ISSUE #11394: The axon transition definition lives in the alarm, where the API cannot share it
Groundwork for #10805, which stays open for the route wiring.

The rule for "a UID stopped having a reachable axon" exists only inside `loadAxonLossMechanisms` in the axon-announcement watchdog. #10805 derives the same transitions for three public routes, and a second copy is how the alarm and the API come to disagree about what a withdrawal is.

That disagreement is not hypothetical — it is the same defect twice in one day:

- **#11370** — the alarm reported deregistrations as withdrawal. SN25 67/0, SN102 43/0.
- **#11392** — the alarm reported a move to an unroutable address as "stopped publishing an axon". 166 of 271 network-wide.

Putting a third copy behind a public contract, where a consumer cannot see the mistake, is how it becomes permanent.

## What this adds

`src/axon-reachability-changes.ts` — one definition of the transition, and the primitives the routes will shape it with:

- `AXON_TRANSITION_SEQ_SQL` / `AXON_TRANSITION_WINDOW_SQL` — the per-UID day-over-day sequence, built on `ROUTABLE_AXON_SQL` from #11376 so routability has one owner.
- `AXON_CHANGE_PREDICATE_SQL` — a loss of **reachability**, not of the axon.
- `AXON_CHANGE_KIND_SQL` — the three-way split as a CASE, so aggregation and row listing agree by construction. Deregistration is decided **first**: on a reused UID the newcomer's axon says nothing about the miner that left.
- `toAxonReachabilityChange`, `tallyAxonChanges`, `buildAxonChangesChain`, `buildAxonChangesScoped`, `axonChangesWindow`, `axonChangesDerivation`.

The watchdog is repointed at the shared fragments, so there is exactly one definition today rather than one plus a promise.

## Two properties worth naming

**The chain leaderboard ranks by `stopped_announcing`, not by total.** Ranking by total puts SN126's 160 moves above every genuine withdrawal on the network — the precise misreading this family exists to prevent.

**The newest `snapshot_date` is still being written.** Measured 2026-08-16: completed days carry exactly one `captured_at` (~23:46 UTC); the current day carried two, fifteen minutes apart, and the same aggregate returned 1,915 then 1,916 deregistrations forty minutes apart. A UID can read `stopped-announcing` on one request and `moved-unroutable` on the next. So every scope reports `start_date`/`end_date` and an `end_date_settled` flag rather than presenting the trailing day as final — silently dropping it would give a caller asking for 7d six days and no way to tell.

## Not in scope

The three route handlers, their schemas, the MCP and GraphQL mirrors, and the openapi regen. #10805 carries those and the decision recorded on it.

JSONbored:
Closes #11394. Closes #10805.

The three axon-removal routes answered a **permanent zero** on every scope, in every window — they were modelled on `AxonInfoRemoved`, an event with zero occurrences in the complete pallet-level stream, genesis to head. They now answer from `neuron_daily`.

## The obvious derivation is wrong 95% of the time

Over the 38 days the table retains, network-wide, of 2,186 transitions where a UID stopped having a routable axon:

| mechanism | count | |
| --- | ---: | --- |
| deregistration — the UID changed hands | **1,915** | the newcomer never served |
| moved to an unroutable address | **166** | still announcing |
| **truly stopped announcing** | **105** | the only real removal |

So `removals` counts stopped-announcing and **only** that. The other two kinds are carried in `changes`, stated rather than summed in. That distinction is the error #11370 and #11392 each fixed inside the alarm — putting it on a public contract, where consumers cannot see it, would have made it permanent.

**One definition.** The transition lives in `src/axon-reachability-changes.ts`, and the axon-announcement watchdog reads the same fragments, so the alarm and the API cannot drift about what a withdrawal is.

## Ranking follows from the same point

The chain leaderboard orders by **removals**, then total, then netuid. Ordering by total puts SN126's 160 moves above every genuine withdrawal on the network.

The account scope applies the same rule to `total_removals`, `concentration` and `dominant_netuid`. The busiest account in the window had **6 transitions across 5 subnets, every one a deregistration** — the old contract reported that as "6 removals" for an account that tore down nothing. Those subnets still appear, with `removals: 0` and their `changes` stated, so churn is visible rather than hidden.

## Two contract properties, now stated

1. **Daily resolution, 38-day retention bound**, in `derivation`. A longer window is out of range, not an empty result.
2. **The newest `snapshot_date` is still being written.** Completed days carry exactly one `captured_at` (~23:46 UTC); the current day carried **two**, fifteen minutes apart, and the same aggregate returned 1,915 then 1,916 forty minutes later. Each scope reports the days it read and flags the trailing one via `end_date_settled`.

## The degraded marker is gone, and what replaced it is stronger

An unread window has a **NULL `start_date`**; a measured zero does not. A new field can forget to set a marker — it cannot forge dates it never read. The marker's own comment set out the only two honest options, *"derive it from axon-info state… or RETIRE the route family"*. This takes the first.

## Wiring

`METAGRAPH_AXON_CHANGES_SOURCE` forwards to DATA_API, which grows three handlers. The account-event handler **factory gains the tier rung it never had** — which is why every account feed fell to an empty card.

The account filter is applied **outside** the sequence: filtering the CTE removes the row that replaced the hotkey, which would turn a deregistration into a withdrawal.

## Verified against production at every step

- The SQL fragments reproduce **1,916 / 166 / 105**.
- The refactored watchdog query returns **SN25 67/0/0** and **SN126 3/36/36** — all 36 of SN126's recent same-hotkey losses are moves.
- The account query returns the 5-subnet all-deregistration footprint the tests now pin.

**21,588 tests pass across 939 files. Every CI validator passes** — including `openapi-examples` (217/217 worked examples schema-valid), `nullable-overpromises`, `published-names`, `contract-drift` and `graphql-types-drift`, with `openapi.json`, the GraphQL SDL and the generated client all regenerated and committed.

Two gates caught real work rather than being worked around: `graphql-tier-cascade` flagged two exemptions that were no longer true once the handlers reached a tier, and `unreferenced-exports` found the account_events constants the derivation left dead.


JSONbored:
Superseded by #11391, which landed the #10805 derivation while this was in flight — and did it with a debounce this branch lacked (a one-reading absence is a missed poll, not a removal).

Closing rather than merging: re-litigating a merged contract to swap in a parallel implementation would be churn, and #11391's is sound for what it measures.

One measured gap remains, which I'll take as a focused addition on top of #11391 rather than a replacement. Its rule is `axon` non-null → null, so a miner that moves to an address nothing can reach is not a removal — it is not anything:

| | count |
| --- | ---: |
| merged rule, same hotkey | 106 |
| reachable → unreachable, same hotkey | 271 |
| **invisible to the merged rule** | **166** |
| **SN126 under the merged rule** | **1** |
| SN126, what actually happened | **160** |

SN126 is the largest same-hotkey source on the network over 38 days and the feed shows one event. Tracked in the follow-up issue.

Issue #11404, opened 2026-08-16
alarm(lane): lakehouse-seam


JSONbored:
`lakehouse-seam` has reported **stale** on every tick for **1.2h** (2 consecutive verdicts).

| | |
| --- | --- |
| fault | `stale` |
| since | 2026-08-16T09:48:35.043Z |
| observed cadence | 1.0h |
| watchdog said | `the raw capture trails the chain head by 8543 block(s) (threshold 900, ~28.5h of chain): blocks 8848170..8856712 have not been captured as raw bytes at all, so no decode can recover them. Past ~24h the span between the lakehouse's ceiling and Neon's rolling retention floor is held by no store and must be re-read from the chain` |

```sql
SELECT datetime(checked_at/1000,'unixepoch') at, verdict, detail
FROM lane_health WHERE lane = 'lakehouse-seam'
ORDER BY checked_at DESC LIMIT 40;
```

Closed automatically on the first `ok` verdict.
Opened by the lane-health reader (`src/lane-alarm.ts`); the record it reads
is also served at `GET /api/v1/self-health`.

JSONbored:
## Cause found, and it is not the decode container

The raw capture is losing to the chain head because its watermark write is buffered while its read is not.

`raw_capture_state.last_contiguous_block` is where each tick resumes. The write goes through the ten-minute Neon write-behind buffer; `neonWatermarkRead` SELECTs it straight from Postgres. So every tick inside a flush window resumes at the same height and re-captures the same blocks. The R2 key is derived from the block range, so those re-writes are byte-identical — nothing throws, `last_error` stays NULL, and every tick records `ok, 10 captured`.

Measured on production 2026-08-16, mainnet:

| at | block | gain |
| --- | ---: | ---: |
| 11:06:39Z | 8848184 | |
| 11:17:03Z | 8848194 | +10 over 10.4 min |
| 11:27:03Z | 8848204 | +10 over 10.0 min |

Three consecutive windows at exactly `maxBlocks` — 0.96 blocks/min against a chain producing 5.0. `neon:raw-capture-state` recorded "41 statement(s) flushed" in one window: forty-one watermark writes collapsing into one effective value.

The effective rate is `maxBlocks per FLUSH INTERVAL`, not per tick, which is why #11402 made it worse rather than better:

| cron | maxBlocks | ticks per 10 min | blocks per 10 min |
| --- | ---: | ---: | ---: |
| `*/5` | 53 | 2 | 53 |
[...]

PR #11405, opened 2026-08-16
fix(capture): the lane resumed from a watermark the buffer had not flushed yet, and re-read the same blocks

JSONbored:
The capture lane's watermark is the row it reads to decide where the next tick
resumes, and its write was going through the ten-minute write-behind buffer
while `neonWatermarkRead` SELECTed it straight from Postgres. So for a whole
flush interval every tick read back a watermark its own earlier ticks had
already moved past, resumed at the same height, and re-captured the same
blocks. The R2 key is derived from the block range, so the re-writes were
byte-identical: nothing threw, `last_error` stayed NULL, and every tick
recorded `ok, 10 captured`.

The effective rate is `maxBlocks per FLUSH INTERVAL`, not per tick — every tick
in a window writes the same `watermark + maxBlocks` and the last one wins.

## Measured on production, 2026-08-16

`raw_capture_state.last_contiguous_block`, mainnet:

| at | block | gain |
| --- | ---: | ---: |
| 10:56:03Z | 8848179 | |
| 11:06:39Z | 8848184 | +5 over 10.6 min |
| 11:17:03Z | 8848194 | +10 over 10.4 min |

**0.96 blocks/min against a chain producing 5.0.** The second window is exactly
`maxBlocks` — ten blocks per ten-minute flush, which is the arithmetic above
rather than a coincidence.

Meanwhile `lane_health` for `raw-capture:mainnet` reported `ok, 10 captured` on
all eleven of those ticks, and `behind` climbed by exactly +5 a minute — the
chain's own rate, which is what "the watermark did not move at all" looks like
from the outside.

## This is why #11402 made it worse

The two faults compose, and the product is what matters:

| cron | maxBlocks | ticks per 10 min | blocks per 10 min | vs chain 5.0/min |
| --- | ---: | ---: | ---: | ---: |
| `*/5` | 53 | 2 | 53 | 5.3/min |
| `*/1` | 10 | 10 | 10 | 1.0/min |

#11402 shortened the cron so ticks would finish instead of being killed
partway, which is right and is confirmed by the data: ticks now bank all 10
blocks. But `pacedLaneBudget` sizes the budget from the cron interval while the
flush interval stayed at ten minutes, so a fivefold-shorter tick cut throughput
fivefold. The lane went from just above break-even to losing four blocks a
minute, and #11404 fired.

The two fixes are complementary, not competing: with the buffer out of the path
each tick reads a fresh watermark, so `*/1 × 10` is 10 blocks/min against 5 —
about twice the chain, where `*/5` was 1.06x.

## The fix

`raw-capture-state` joins `NEVER_BUFFER_LANES` and comes out of
`NEON_WRITE_BUFFER_LANES` in all three configs, matching how `blocks-head` and
`chain-detail` are already handled. It is the cheapest lane to exempt: two rows
a minute, against `blocks-head` at one every twelve seconds.

`src/raw-capture-sync.ts` already carried this exact warning one layer up —
*"an earlier version nearly shipped writing Neon while still reading a
store-backed store ... it would have re-captured the same blocks every tick,
forever, while both stores looked healthy"*. Buffering re-opened that gap
without touching either function, because it moved the write and left the read.

The flag's own comment argued the lane was safe because
`table-freshness-watchdog` holds it to two hours and a ten-minute flush fits
inside that. That reasoning is now stated as the trap it was: a freshness bound
answers "is this row too old for a READER", and says nothing about whether the
PRODUCER reads it back to decide its next write.

Also corrects `table-freshness-watchdog`'s `raw_capture_state` reason, which
still said "every 5 min" after #11402 moved the cron to every minute.

## Not closing #11404

The lane is ~8,600 blocks behind. At 10/min against 5/min it closes that in
roughly 29 hours, so the seam alarm should stay open until it actually
converges rather than be closed by the commit that makes convergence possible.

## The gate, so the next one is caught rather than remembered

`tests/buffered-lane-read-your-writes.test.ts` sweeps rather than lists. It
pairs every `<X>_NEON_LANE` with the table that lane writes, then asks which
modules IMPORT that lane's writer and also `SELECT ... FROM` that table. A
producer that reads its own lane's table must be in `NEVER_BUFFER_LANES`.

Three properties make it worth having rather than decorative:

- **The pairing is checked, not assumed.** Every lane's table must exist in
  `db/schema.sql`, so a lane whose derived name stopped matching fails loudly
  instead of quietly matching no reader and passing.
- **The two overrides are guarded.** `chain-detail` (four tables) and
  `self-stake` (shares `nominator_positions`) are declared, and each is
  required to have NO like-named table — so an override cannot be used to point
  a lane away from a table it does write.
- **It fails on the real tree.** Removing `raw-capture-state` from
  `NEVER_BUFFER_LANES` makes it report
  `src/raw-capture-sync.ts produces lane "raw-capture-state" and SELECTs
  raw_capture_state`, which is how the sweep was proved to be more than an
  emptiness assertion over a broken matcher.

Swept across `src/` today, `raw-capture-sync.ts` is the only producer that
reads its own lane's table — so this is the whole class, not one instance of it.


JSONbored:
Superseded by #11406, which reaches the same diagnosis and goes further: batched RPC reads are what actually closes the ~8,600-block gap, where this PR only stops it growing.

Not re-litigating a parallel implementation. The two pieces here that #11406 does not carry — the derived `buffered-lane-read-your-writes` sweep, and the stale `RAW_CAPTURE_CRON every 5 min` prose in `table-freshness-watchdog` — will land as a focused follow-up once #11406 merges, so the fix that matters is not held up behind them.

Branch kept for that follow-up; measurements and the failing-gate proof are in the PR body above and quoted on #11406.

Issue #9378, opened 2026-08-04
fix(chain): the testnet capture lane makes 3x the calls the endpoint accepts


JSONbored:
#9371 shipped the testnet raw-capture lane with `maxPerTick: 100`. Measured in production, that is
three times what the public endpoint will serve.

## Measured

Replaying `captureTick`'s exact call pattern (3 sequential RPC calls per block, no delay) against
`test.finney.opentensor.ai`:

```
FAILED after 33 blocks (100 calls) in 23.0s: HTTP 429
```

**Exactly 100 calls.** And the budget is per-client, not per-host — probing
`test.chain.opentensor.ai` immediately afterwards stopped after 4 blocks, because the first probe
had already spent it. So there is no sibling endpoint to spread across.

## Consequences

1. **Two-thirds of every tick is rejected.** The lane asks for 100 blocks (300 calls), gets ~33, and
   throws 200 requests at a public endpoint we do not own purely to be refused. That is the part
   worth fixing regardless of throughput.
2. **The documented drain time is wrong.** #9371 says the backlog "drains in a few hours". Real
   numbers: 33 captured/tick against ~25 produced/tick (12s blocks, 5-min cron) is a net drain of 8
   blocks/tick, so the 8,438-block backlog takes **~88 hours**, not a few.

Correctness is unaffected — the no-gap guarantee holds, because `captureTick` stops at the first
failure and keeps the prefix, so the next tick resumes at the same height. This is a throughput and
good-citizenship defect, not a data one.

## Requirements

1. The per-tick budget must sit inside the measured limit, with headroom — a tick that ends by being
   throttled is one that wasted the calls it made.
2. The floor's documentation must state the real drain time. A comment claiming "a few hours" is
   worse than none: it is the number someone will plan against.
3. The budget must be asserted against the endpoint limit by a test, the same way the combined
   subrequest ceiling already is — otherwise the next person to raise it re-creates this.

## Do NOT

Do not raise the testnet floor to shrink the backlog. The watermark is already past 7,700,000, so
moving the floor up now leaves blocks 7,700,264..newFloor permanently uncaptured — a real gap, which
is the one thing this lane exists to prevent.

PR #11406, opened 2026-08-16
fix(capture): the watermark write was buffered while its read was not, so every tick resumed from a stale value

JSONbored:
The raw-capture lane had fallen **8,621 blocks (~29h)** behind, which stalled the lakehouse and left the site's 24h on-chain volume showing an em-dash. Every signal said ok: ticks ran on schedule, each reported `10 captured`, no exception, `last_error` NULL, `stopped_at` NULL.

## Root cause

`raw-capture-state` was the first lane named in `NEON_WRITE_BUFFER_LANES`, so the watermark's **write** went through the write-behind buffer (`FLUSH_INTERVAL_MS` = 10 minutes) while `neonWatermarkRead` read it **straight from Neon**.

That watermark is not consumed by some other reader — it is where the lane itself resumes one minute later. So for up to a full flush interval the producer read back a value that did not include what it had just captured, re-captured the same range, and enqueued the same write again.

Measured in production 2026-08-16, mainnet:

| | |
|---|---|
| tick cadence | every 60s, each reporting `10 captured` |
| watermark advance | ~5 blocks per **~10 minutes** — one flush interval exactly |
| effective throughput | **~1 block/min** against a chain producing 5 |

The failure is invisible from either side alone: the buffer flushed successfully, the capture captured successfully, and only the *pairing* was wrong.

`raw-capture-state` joins `NEVER_BUFFER_LANES` (checked before the flag) and is removed from the flag in all three worker configs, so the config no longer advertises buffering that cannot happen. A test pins that no shipped `wrangler*.jsonc` names a lane the code refuses — this arrived as a reasonable-looking config entry, and nothing at the write site says its reader is one tick away.

## The read cost, separately

The lane spent three sequential HTTP requests per block, and the endpoint meters **requests**, not JSON-RPC calls. Re-measured against `archive.chain.opentensor.ai`:

| | requests | calls | outcome |
|---|---:|---:|---|
| one call per request | 103 | 103 | **429** |
| fifty calls per request | 140 | **1,400** | no limit, 103.7s |

So a chunk now costs **two requests whatever its size**: one `chain_getBlockHash` over a *list* of heights, then one JSON-RPC batch of bodies and events. The node states its own ceiling (`"Exceeded max limit of 50"`), which at two calls per block makes **25 blocks** the largest chunk it will answer — measured at 0.81 MB and ~1.7s.

Verified against the live archive: **25 blocks in 2 requests**, with `parent_hash` chaining 24/24.

Batch replies are correlated **by id, never by position**. JSON-RPC permits a server to reorder them, and reading positionally would assemble every block in a chunk from another block's bytes — wrong data rather than missing data, which no gap check would catch.

The budget is re-derived in requests rather than calls, and the pacing gap is now a **cycle time** (the read's own duration comes off it), so a tick lands on its request budget instead of on `budget + latency`.

## Also corrected

#9378 concluded the rate limit was per **client**; it is per backend **node**. Exhausting `archive.chain.opentensor.ai` and immediately draining `lite.chain.opentensor.ai` got a full fresh 100. What made the original measurement look client-wide is that `entrypoint-finney.opentensor.ai` is a CNAME to `lite.chain.opentensor.ai` — one machine under two names (`65.109.251.221`), while archive answers from `65.109.254.0`.

The endpoint list stays **failover-only** regardless: one node's allowance already funds hundreds of blocks/minute against a chain producing five. `pacedLaneBudget`'s docstring had been arguing both sides of this since #11401 and now states what was measured.

## Cost note

Un-buffering the watermark adds one small upsert per chunk. While backfilling that is ~16/min per lane; once caught up a tick is one chunk, so ~1/min. `blocks-head` already writes every ~12s and is never-buffered, so the Neon compute could not suspend either way — this buys back correctness at no change in suspension behaviour.

## Verification

- full suite: **947 files, 21,725 passed, 11 skipped, 0 failures**
- `validate:` schemas, lane-topology, startup-budget, contract-drift, workflows, no-passthrough, boundary-casts, type-duplicates, schema-shape-duplicates, double-assertions, migrations, private-boundary — all pass
- patch coverage: no uncovered statements or branches in the changed lines
- live: chunked read verified end-to-end against the production archive endpoint


JSONbored:
Same diagnosis reached independently in #11405 — closing that one in favour of this, since the batched reads here are the part that actually closes the ~8,600-block gap rather than just stopping it growing.

**Your red `validate` is `validate:worker-types-parity`, and it is a one-command fix.** I hit exactly the same thing ten minutes ago. `workers/*.worker-configuration.d.ts` pins each var as a string *literal*, not `string`:

```ts
NEON_WRITE_BUFFER_LANES: "raw-capture-state,neurons,…";
```

so dropping `raw-capture-state` from the flag makes the committed types stale, and the header's config hash means it is caught even where no literal a caller reads has changed.

```bash
npm run types:workers
```

That regenerates all four workers at once and produces a 6-line diff across `workers/worker-configuration.d.ts`, `workers/data-api.worker-configuration.d.ts` and `workers/registry-sync-api.worker-configuration.d.ts`. (Not `npx wrangler types --env-interface Env` — that writes a stray file at the repo root.) It is not in the local `validate:*` subset most of us run by hand, which is why we both pushed red.

**Two things from #11405 that are not in this PR**, which I will land as a focused follow-up once this merges rather than complicate the diff here:

1. `tests/buffered-lane-read-your-writes.test.ts` — a derived sweep rather than a list. It pairs every `<X>_NEON_LANE` with the table that lane writes, then flags any module that imports that lane's writer and also `SELECT`s that table. Each lane's table must exist in `db/schema.sql` so a pairing that stops matching fails loudly instead of matching no reader and passing, and both overrides (`chain-detail`, four tables; `self-stake`, sharing `nominator_positions`) are required to have no like-named table so an override cannot point a lane away from one it does write. Verified to fail on the real tree: removing the exemption makes it report ``src/raw-capture-sync.ts produces lane "raw-capture-state" and SELECTs raw_capture_state``.
2. `src/table-freshness-watchdog.ts:184` still reads `reason: "RAW_CAPTURE_CRON every 5 min"` after #11402 moved the cron to `*/1`.

Measurements from #11405 that corroborate this PR's premise, in case they are useful in the body — `raw_capture_state`, mainnet, three consecutive windows at exactly `maxBlocks`:

| at | block | gain |
| --- | ---: | ---: |
| 11:06:39Z | 8848184 | |
| 11:17:03Z | 8848194 | +10 over 10.4 min |
| 11:27:03Z | 8848204 | +10 over 10.0 min |

0.96 blocks/min against a chain producing 5.0, with `neon:raw-capture-state` recording "41 statement(s) flushed" in one window and `raw-capture:mainnet` reporting `ok, 10 captured` on every tick.

Issue #11550, opened 2026-08-21
feat(api): publish a bounded subnet price-share composition timeline


JSONbored:
## Problem

The homepage currently has only a point-in-time economics ranking. Rendering an OpenCode-style historical composition chart from that snapshot would be misleading, while requesting history per series in the browser would add avoidable latency and unstable client-side aggregation.

The original proposed cross-subnet `total_stake_alpha` stack is invalid: alpha is denominated independently by subnet and must never be summed across netuids.

The historical `emission_share` column is a legacy economics-artifact normalization: alpha price divided by the sum of reported alpha prices. It includes Root when Root reports a price and does not retain the historical runtime eligibility inputs required to reconstruct v440 Stage 1. It must never be described as runtime Stage 1, stake, market cap, or final TAO emission.

## Scope

- Add a typed, mainnet-only `GET /api/v1/chain/subnet-price-share-composition` endpoint with up to 56 closed UTC daily points.
- Read existing `subnet_snapshots` history in one bounded indexed query. Select a deterministic fixed cohort of up to six **netuids** from the newest eligible date, with `netuid` tie-breaking; never reshuffle colors or series identity by day.
- Use only existing `emission_share` as an explicitly **artifact-normalized moving-price share**. Preserve Root in its stored denominator; do not post-hoc filter or re-normalize historical rows where runtime eligibility is unknown.
- Mark the result `estimated_observed_price_set`. Legacy daily snapshots have no completed-pass manifest, so a returned day must require one persisted writer timestamp for numeric shares, unique netuids, valid six-decimal share units, and a total within the six-decimal source rounding envelope. This detects certain mixed writes; it is not a source artifact ID or proof of complete chain coverage.
- Derive `Other artifact-normalized price share` as the exact six-decimal residual of that normalized unit, never as a persisted bucket, unselected-row sum, or claimed network total.
- Do not join identity history: a stable series is a netuid, not a claim that a reused netuid is one project throughout the timeline.
- Omit mixed, malformed, off-grid, partial-normalization, uncaptured, and cap-truncated raw dates rather than inventing zeroes or a complete-looking bar.
- Use existing worker edge-cache, route-contract, OpenAPI, docs, GraphQL, MCP, and mainnet-addressing patterns. Keep the payload compact for an interactive composition timeline.

## Acceptance criteria

- The response contains a stable ordered netuid series list plus a bounded chronological sequence of real daily points.
[...]

Issue #11547, opened 2026-08-21
feat(ui): enforce data-document page archetypes and analytical interaction grammar


JSONbored:
## Problem

Shared tokens alone cannot keep route code from rebuilding dense dashboard walls. The current UI has competing page grammars, and its analytical interaction model is too generic to express a truthful temporal composition chart without bespoke route CSS.

## Scope

- Introduce opinionated shared page archetypes for directory, dossier, stream, record, data-report, operations, tool-workspace, and editorial routes.
- Constrain each header/archetype to a truthful information hierarchy instead of allowing every hero slot on every route.
- Add a first-class composition-timeline primitive to the shared design system: narrow dotted daily lanes, sharp stacked prism segments, compact anchored inspection, keyboard/touch selection, reduced motion, and a nonvisual summary.
- Define the timeline’s specific cross-filter contract: at rest all real series remain vividly legible; inspecting a day turns other days graphite; inspecting one segment keeps that named series vivid across days and turns unrelated segments graphite. This applies to temporal composition only, not to every ranking chart.
- Keep product mint for calls to action, focus, and live status—not as the composition-chart hover effect.
- Provide a shared desktop-first spacing/gutter contract and specimen coverage in the design-primitives route.

## Acceptance

- New core-route work adopts an explicit archetype rather than a local header/panel composition.
- The shared composition timeline supports real series identity, an `Other` residual, stable palette assignment, compact metric-only tooltip content, and the documented column-versus-series inspection states.
- The primitive uses its own semantic chart tokens, so no route needs one-off geometry, hover, or palette CSS to achieve the data-document grammar.
- Controls live adjacent to the question they change, and raw details remain progressively disclosed.
- The primitives pass keyboard, touch, reduced-motion, contrast, and responsive tests.

## Boundaries

This creates reusable visual and interaction contracts only. The composition data endpoint is #11550; homepage integration remains #11544; other route-family migrations remain separate issues and PRs.

JSONbored:
[...]

Issue #11544, opened 2026-08-21
feat(ui): rebuild the homepage as a visual Bittensor data landing page


JSONbored:
Parent: #11517

## Problem

The current homepage reads as a cleaned-up explorer: it stacks copy, controls, and small panels before it gives a visitor one compelling visual understanding of Bittensor. It does not yet deliver the calm, visual-first OpenCode-style composition requested for Metagraphed, and its spacing collapses at responsive boundaries.

## Scope

- Rebuild the homepage through the shared design system as the visual reference implementation for the explorer redesign.
- Create one generous, mint-lit signal-field hero inspired by Render Workflows' spatial drama, with Metagraphed-owned composition and no copied assets.
- Apply the transferable OpenCode data-page grammar: centered editorial measure, concise title/one-line read, a primary live visual, restrained framed analytical modules, and progressive disclosure for detail.
- Replace redundant copy and KPI/card walls with a clear live Bittensor signal visual plus concise ranked/route-entry modules.
- Keep the existing Metagraphed mint as the primary action/focus color; use the categorical chart palette only to distinguish real analytical series.
- Define responsive behavior intentionally at desktop, tablet, and phone: generous safe gutters, recomposed sections rather than squeezed desktop grids, no horizontal clipping, and accessible reduced-motion/focus behavior.

## Acceptance criteria

- At 1280x800, the above-fold experience has one visual thesis, visible live data, and at most one primary plus one secondary action.
- At 768x1024 and 375x812, content has comfortable gutters and no text, controls, or charts are cut off or merely compressed.
- Below-the-fold content is organized as a small visual data narrative; raw detail is intentionally disclosed rather than front-loaded.
- The homepage mounts `/api/v1/chain/subnet-price-share-composition` as the real source for the composition chart and removes that route's temporary `PROSE_ONLY_ROUTES` exception in the same PR.
- All visual changes are made through shared tokens/primitives, covered by focused tests, responsive overflow checks, and the required before/after screenshot matrix.

## Out of scope

- Copying OpenCode or Render brand assets, copy, logos, or proprietary implementation.
- Migrating the other major route families in this PR; they follow once this homepage reference implementation is accepted.

JSONbored:
Superseded by #11618 under epic #11604 (Design System v2), which replaces the Metagraph Terminal plan with the measured contract from the 2026-08-22 redesign research. The requirements that still apply were carried over into the replacement issue(s); this issue is closed so the tracker has one plan.

PR #11556, opened 2026-08-21
feat(ui): establish responsive data document grammar

JSONbored:
Establishes the shared temporal-composition primitive the data-document grammar
needs, and rebuilds the landing page's one live visual on top of it.

**Status: open for preview and review, not for merge.** The homepage is a
reference implementation in progress, not a finished surface.

## What this adds

`CompositionTimeline` in `packages/ui-kit` — the primitive #11547 asks for, which
the generic `InteractiveDataField` could not express. That component stacks
segments but has no concept of a *series*, so the cross-filter contract could
only have been faked with route-local geometry and hover CSS.

The contract, verified in a real browser rather than asserted:

| state | vivid segments | graphite |
| --- | ---: | ---: |
| at rest | 392 | 0 |
| inspecting one day | 7 | 385 |
| inspecting one series | 56 | 336 |
| after `Escape` | 392 | 0 |

- narrow dotted daily lanes over a rail; sharp stacked categorical segments
- **vivid at rest** — legible before it is touched, not dull until hovered
- inspecting a day greys the *other* days and keeps that day's full palette;
  inspecting a series keeps it vivid across every day without singling out a day
- column-as-button: `role`, `aria-pressed`, one roving tab stop across 56 lanes,
  arrow/Home/End/Escape, pointer, touch, `prefers-reduced-motion`
- a real `<table>` carrying the same numbers, so the marks are never the only
  way to read the values
- product mint is spent only on the focus ring, never to express a datum

## Two palette rules that came out of looking at it

Walking the 11-tone ramp in order gave the six-subnet cohort **four adjacent
violets** that could not be told apart — consecutive ramp entries are
near-neighbours in hue and blend exactly where stacked segments touch.
`compositionToneAt` strides the ramp instead.

The derived residual was taking the ramp's last entry — a **red**, which reads as
an alert state on a segment meaning only "everything else". A residual now
carries no categorical tone at all, and the type enforces it rather than a
comment asking nicely.

## Hero

The rejected composition ran a wide title facing a boxed column that held both
the supporting sentence and the primary action. That made the sentence read as a
caption for the backdrop and buried the action in the quietest corner of the
page. Copy now descends in one column on a shared left edge, with the lattice as
spatial counterweight rather than a facing panel. At 1280×800 the first live data
module enters the viewport.

## Defects found by rendering, not by reading

- The visually-hidden table **escaped the viewport** below 1280px. `sr-only` hides
  by setting `width: 1px`, which a table ignores without `table-layout: fixed` —
  and even then the `<caption>`, held on one line by `white-space: nowrap`,
  dragged the box out to 1,133px.
- The field's surfaces were keyed on a `.dark` class **the landing shell never
  sets**, so a graphite page drew a near-white inspector. Tokens now derive from
  the active scope and are correct in the light app, the dark app, and the
  landing.
- `isolation: isolate` on every lane trapped the readout in a local stacking
  context, letting later lanes paint straight through its text and slice the
  percentages in half.

## Fixtures

`GET /api/v1/chain/subnet-price-share-composition` is implemented on this stack
but **is not deployed** — `api.metagraph.sh` returns 404 for it (verified against
a 200 control on `/api/v1/chain/activity`). The homepage's only live visual reads
it, so the specs get an **explicitly synthetic fixture built from the committed
contract**, in its own human-readable file, announced on stdout whenever it is
served, and loaded last so any real recording wins. It is deliberately *not* in
`ssr-supplement.json`: that file is recorded production bytes, and `record()`
refuses to persist even a genuine non-2xx on the grounds that a recorded error is
worse than a gap.

Two genuine gaps are **recorded from production**, not invented:

- `/api/v1/registry/leaderboards` — absent, so `/leaderboards` rendered an error
  state at every viewport.
- `/api/v1/subnets?limit=500` — the raised fetch ceiling needs it, and its absence
  was silently falling back to a smaller recording. That left **128 of 129 subnet
  pages with no inbound link** in the server-rendered index and
  `/subnets/category/privacy` with zero subnets. Pre-existing on this branch.

## Validation

- `npm run test:e2e --workspace=apps/ui` — **250 passed, 0 failed**
- all nine `#11315` payload ratchets pass, **thresholds untouched** (`git diff` on
  `indexable-routes.spec.ts` is empty)
- unit: 151 (ui-kit) + 2,341 (apps/ui)
- lint 0 errors both workspaces, prettier clean, typecheck clean
- `validate:ui-route-coverage`, `validate:contract-drift` pass

## Live preview

**https://e9634e07-metagraphed-ui.zeronode.workers.dev/** — version `e9634e07`,
built by Workers Builds from `c29d2bc5b`. A 0%-traffic version; `metagraph.sh`
routing is untouched. Each build publishes a new version, so the URL changes per
push.

One module does not light up there: `/api/v1/chain/subnet-price-share-composition`
is 404 in production (the endpoint is implemented lower in this stack but not
deployed), so the composition chart shows its honest unavailable state against
the live API. Everything else runs on real data. The chart itself is exercised
end-to-end in the e2e suite against the labelled synthetic fixture, and the
`#11547` interaction contract is asserted there.

## Not done yet

- **No before/after screenshot matrix.** Deferred until the surface settles.
  Worth knowing when it lands: light and dark are *identical* on this page —
  `mg-shell--landing` scopes graphite tokens in both themes by existing design —
  so that matrix will legitimately carry duplicate rows.
- The hero keeps one primary action; the reference has zero, with its CTA in the
  header strip.

Refs #11547, #11544, #11550 — none of their acceptance criteria are complete, so
this deliberately closes nothing.


JSONbored:
Closing as part of superseding the Metagraph Terminal stack with epic #11604 (Design System v2). The ten `codex/*` PRs form one linear chain that cannot merge independently, three have red CI, one is conflicting, and at the stack tip the profile routes still carry the legacy grammar (41 StatTiles / 33 SectionAnchors / 10 tables on the account profile). The replacement work lands as one PR per sub-issue of #11604, starting from `main`. The `CompositionTimeline` primitive from this branch is the named cherry-pick source for #11608 (StackedColumns); the branch is kept for that. Branch left in place; not deleted.

PR #11575, opened 2026-08-22
feat(ui): give subnet discovery three explicit task modes

### LINKED ISSUE #11520: feat(ui): make subnet discovery focused and progressive
Parent: #11517

## User problem

The Subnets page combines discovery, rankings, domains, churn, taxonomy, exports, comparison, saved views, and advanced table preferences in one default experience. A user cannot tell whether this is for browsing projects, researching economics, or finding an integration.

## Scope

- Make the default subnet view a focused discovery experience.
- Surface name, netuid, plain-language purpose, health/readiness, verified public interface availability, and a small economics summary before advanced metrics.
- Establish intentional Browse, Research, and Compare/customize modes while keeping shareable URLs.
- Reposition category, API, domain, ranking, lifecycle, and revenue lenses as clear secondary paths.

## Acceptance

- A non-expert can find a buildable subnet using search plus a small set of understandable filters.
- The default view does not expose every column, export, density, or compare control.
- Market/economic fields are labeled in plain language with contextual help.
- Card/list behavior is first-class on mobile; wide-table behavior is deliberate on desktop.

## Constraints

Keep canonical route behavior and existing deep links intact. Do not duplicate the SEO content work in #11320.

JSONbored:
Rebuilds `/subnets` around the three jobs people actually arrive with, instead
of doing all of them at once.

**Open for preview and review, not for merge.**

## The problem, concretely

The default view carried eight measurement columns, a column customizer, a
density switch, a CSV export, a sparkline-window control and a compare
control — and nowhere did it say what any subnet *does*. A reader who had
never seen the registry could not tell whether the page was for browsing
projects, researching economics, or finding something to integrate with.

## Three task modes, in the design system

`DirectoryMode` — **Browse / Research / Compare** — lands in `packages/ui-kit`
as an underlined task strip rather than one more utility pill, because
choosing the job is the page's primary decision, not a display preference. It
lives in the shared kit so the validator, account and chain directories adopt
the same grammar instead of each inventing one.

It is a different axis from `ViewMode` (table/grid/matrix): a mode decides
which questions the page answers, a view decides how the rows are drawn.

## Browse is genuinely reduced

| | Browse | Research |
| --- | --- | --- |
| columns | UID, Name, **Interfaces**, Alpha price, Health | UID, Name, Emission, Alpha price, 7d %, Total stake, Health, Age (+ customizer) |
| purpose line | ✅ every row | — |
| column customizer | — | ✅ |
| density switch | — | ✅ |
| CSV export | — | ✅ |
| sparkline window | — | ✅ |
| saved views | — | ✅ |
| share / copy link | ✅ | ✅ |

Every row now leads with a plain-language answer to "what is this?", from the
subnet's own `description` and falling back to the registry's
`derived_description` — the one thing the table never showed and the first
question a non-expert has. In practice that reads as *"Verified inference for
frontier open models"*, *"Making every camera intelligent"*, *"Reason
Mining"*.

The header says **Interfaces**, not the registry's own word *Surfaces*.

The standalone 24h/7d column is gone: the price cell already printed the same
delta, so every row carried it twice.

**Share survives in all three modes.** Hiding it was the first thing I tried
and it was wrong — the issue asks for shareable URLs, and a reader who has
just narrowed a Browse list is exactly the person who wants to send it. Only
the CSV export is a Research affordance.

The purpose line is on the phone's cards too. A card is the phone's primary
row, not a terse version of the page.

## Deep links

`mode` defaults to `browse` and is stripped from the URL at that value, so
every existing `/subnets` link keeps working and lands on the focused view.
It stays on this route rather than the shared `tableSearchSchema`, which four
routes use — giving three of them a mode they do not implement would put a
dead control in their URLs. `?view=`, `?q=`, `?density=`, `?section=` are all
unchanged and covered by a test.

## Also

`description` / `derived_description` are now declared on the `Subnet`
interface. The wire has always carried them (`schemas-src/routes/subnets.ts`);
the hand-written interface had not, so they could only be read back through
its index signature as `unknown`.

`subnets-directory-controls.spec.ts` asserted the old contract — that the
filter sheet always carries Display and Export — so it now asserts the
mode-aware one, plus a new case proving Research really does restore what
Browse withholds.

The new spec is registered in `playwright.config.ts`'s `interaction`
`testMatch`. That file's own comment warns an unmatched spec silently never
runs; mine did, until I added it.

## Validation

- `npm run test:e2e --workspace=apps/ui` — **260 passed, 0 failed**, including
  8 new `#11520` cases
- unit: 151 (ui-kit) + 2,347 (apps/ui)
- root + both workspace linters clean, prettier clean, typecheck clean
- `validate:` schema-vocabularies, unreferenced-exports/modules, contract-drift,
  ui-route-coverage, private-boundary, boundary-casts, artifact-budgets,
  workflows, public-safety scan — all pass

## Not done yet

No before/after screenshot matrix — this surface is still moving, and the
stack it sits on is unmerged. Flagging rather than quietly skipping it.

Closes #11520


JSONbored:
Closing as part of superseding the Metagraph Terminal stack with epic #11604 (Design System v2). The ten `codex/*` PRs form one linear chain that cannot merge independently, three have red CI, one is conflicting, and at the stack tip the profile routes still carry the legacy grammar (41 StatTiles / 33 SectionAnchors / 10 tables on the account profile). The replacement work lands as one PR per sub-issue of #11604, starting from `main`. Branch left in place; not deleted.

Issue #11522, opened 2026-08-20
feat(ui): make validator discovery and profiles decision-ready


JSONbored:
Parent: #11517

## User problem

Validator pages present staking-facing metrics, operator grouping, watch/compare controls, and a mobile glossary without a clear decision flow. On phone, the guide appears before the directory.

## Scope

- Replace the phone-first long glossary with a concise evaluation primer and an optional metric glossary.
- Clarify group-by-operator behavior and the meaning/limits of APY, take, stake concentration, and nominators.
- Make compare a deliberate user mode rather than an always-present row-level burden.
- Reorganize validator profile content into summary, subnet performance, rewards/history, and advanced/alert sections.

## Acceptance

- At 375px, search and initial validator results appear before a long metric reference.
- Decision-relevant fields carry freshness and methodology context.
- Grouping and comparison are understandable without prior product knowledge.
- The UI avoids recommendation language and makes risk/limitations visible.

## Constraints

Do not construct transactions or imply financial advice. Preserve current data and shareable directory filters.

JSONbored:
Superseded by #11616 and #11617 under epic #11604 (Design System v2), which replaces the Metagraph Terminal plan with the measured contract from the 2026-08-22 redesign research. The requirements that still apply were carried over into the replacement issue(s); this issue is closed so the tracker has one plan.

PR #11580, opened 2026-08-22
feat(ui): rank validator operators, not their 1,022 keys

JSONbored:
Ranks the **148 named operators** instead of the 1,022 hotkeys behind them.

**Open for preview and review, not for merge.** Partial against #11522 — this
is the directory half; the profile reorganisation and the metric-glossary work
are still open, so it deliberately closes nothing.

## The problem

One operator was scattered across dozens of rows — Rizzo across 105, Yuma 86,
Kraken 81 — so the question the page exists to answer got *harder* the more
keys a good operator ran. Named operators hold **85.7%** of observed stake, so
the operator is the useful unit and the keys are its detail.

Expanding a row shows the keys it runs, each still linking to its own page.

## Geometry

`RankedRailList` — rank, value, name, then a bar proportional to the leader.
Value first, no column headers, no cards; the rail and the row rhythm do the
work. It lives in ui-kit so `/subnets` and the other directories can use the
same comparison.

## What may be summed, and what may not

This was the actual design problem, not the layout.

| fact | treatment | why |
| --- | --- | --- |
| stake | **summed** | additive, so the operator total is a real total |
| take | **range** | Yuma runs both 9% and 18%; a blended 13.5% is a rate nobody is charged |
| nominators | **absent** | one delegator can back several of an operator's keys, so summing would claim more distinct people than exist |
| subnet reach | **(key, subnet) positions** | never a union over `subnets[]` — see below |

The list response **caps `subnets[]` at 10 entries**, and 12 of 1,022 rows are
truncated: tao.bot reports `subnet_count: 116` with 10 array entries. A union
over those arrays would have silently printed 10. Positions come from each
key's own `subnet_count`, which is exact whatever the array contains.

Expanding an operator makes the take range self-justifying: you can see the one
Yuma key at 9% and the rest at 18%.

## Crawlability, kept

Collapsing 1,022 rows to 149 removes every inbound validator link — the exact
regression #11231 was filed for, and #11315's guard caught it on the first run.

The disclosure is a native `<details>`, so all **590** keys stay in the
server-rendered DOM whether or not anyone opens them, and browsers skip layout
for closed details. Paying for that properly took three passes:

| | avg key link | page |
| --- | ---: | ---: |
| first attempt | 312 B | 1,269 KiB (over ceiling) |
| precomputed strings | 280 B | 1,251 KiB (over) |
| positional spans + CSS focus ring | **177 B** | **1,192 KiB** |

React separates adjacent JSX interpolations with a `<!-- -->` marker; at 590
links those markers alone were ~30 KiB. The rest came from dropping three
per-row class attributes for positional selectors and moving the focus ring
into CSS.

Net: the page carries 590 crawlable links for **2 KiB more** than it previously
spent carrying none, and stays under the ceiling this stack already lowered.

## Deep links

`mode` defaults to `browse` and is stripped from the URL at that value, so
existing `/validators` links keep working. `sticky-table-header` now points at
`?mode=research`, where the per-key table lives.

## Validation

- `npm run test:e2e --workspace=apps/ui` — **261 passed, 0 failed**
- unit: 151 (ui-kit) + 2,357 (apps/ui), including 10 new cases pinning what may
  and may not be aggregated
- root + both workspace linters clean, prettier clean, typecheck clean
- `validate:` unreferenced-exports/modules, schema-vocabularies,
  ui-route-coverage, contract-drift, boundary-casts — all pass

## Not done yet

- No before/after screenshot matrix; the surface is still moving.
- The validator **profile** page (`/validators/$hotkey`) is untouched — #11522
  also asks for summary / subnet performance / rewards / advanced sections
  there, and that is the next slice.
- APY is modelled as a range but not yet surfaced; take was the one that
  mattered first.

Refs #11522


JSONbored:
Closing as part of superseding the Metagraph Terminal stack with epic #11604 (Design System v2). The ten `codex/*` PRs form one linear chain that cannot merge independently, three have red CI, one is conflicting, and at the stack tip the profile routes still carry the legacy grammar (41 StatTiles / 33 SectionAnchors / 10 tables on the account profile). The replacement work lands as one PR per sub-issue of #11604, starting from `main`. `RankedRailList`, `CompositionBreakdown`, `MeasureBand` and the operator-identity reducer from this branch are the named cherry-pick sources for #11609, #11607 and #11616; the branch is kept for that. Branch left in place; not deleted.

Issue #11521, opened 2026-08-20
feat(ui): reorganize subnet profiles around trust, intent, and freshness


JSONbored:
Parent: #11517

## User problem

A subnet profile currently places developer integration, economics, validation, ownership, health, alerts, evidence, and raw metagraph data in one broad default experience. Important values can also have materially different freshness without a single explanation.

## Scope

- Redesign the entity-profile summary around what the subnet does, whether it is trustworthy/usable, and the visitor's next task.
- Establish clear pathways for Use its API, Research economics, and Participate/validate.
- Promote a succinct freshness/source summary and attach freshness to decision-relevant metrics.
- Move raw metagraph, ownership, alert setup, audit history, and deep economics into intentional advanced sections.

## Acceptance

- The first viewport explains the subnet, present health/readiness, and the most relevant next action.
- API/endpoints are easy to reach for buildable subnets.
- Mixed-freshness metrics are not visually presented as one simultaneous snapshot.
- Tabs/sections remain navigable on 375px and 768px without clipped labels.

## Constraints

Keep evidence, provenance, and advanced tabs available; this is a hierarchy redesign, not a loss of explorer depth.

JSONbored:
Superseded by #11612 under epic #11604 (Design System v2), which replaces the Metagraph Terminal plan with the measured contract from the 2026-08-22 redesign research. The requirements that still apply were carried over into the replacement issue(s); this issue is closed so the tracker has one plan.

PR #11582, opened 2026-08-22
feat(ui): give the subnet profile one honest composition and a level hero

JSONbored:
Three defects on `/subnets/<netuid>`, all found by looking at the rendered
page, plus the one composition it can honestly draw.

**Open for preview and review, not for merge.** Partial against #11521 —
`Refs`, not `Closes`.

## What I got wrong first

I assumed this page was another card wall. It isn't — the stacked work already
reorganised it around Build / Research / Participate with per-measure
freshness, so most of #11521's acceptance was already met. Worth saying
plainly, because it changed what this PR is.

## The hero was bottom-aligned against a taller rail

Measured: the copy column is 280px, the signal rail 421px. Bottom-aligning
them left the rail's first label floating **142px above the title** —
technically aligned, visibly wonky. Both columns now share a top edge, so the
eyebrow chip sits level with "Availability".

Only the subnet masthead passes `aside`, so this is scoped in practice despite
living in the shared hero.

## The trust rails meant three different things

| rail | old colour rule |
| --- | --- |
| Availability | reliability threshold → positive/warning/negative |
| Build readiness | always brand (mint) |
| Source coverage | always neutral (grey) |

So colour carried three meanings at once, and a **perfect 4/4 source coverage
rendered grey beside a 96/100 in mint** — reading as the weakest of the three.

They now encode magnitude and nothing else. Availability's verdict is already
carried by the health chip beside the title, which is where a semantic colour
belongs.

## The link row was a paragraph of navigation

Six labelled links plus two actions, under a hero that already carries three
primary actions. Icons now, with the name in `aria-label`/`title` — nothing
lost to a screen reader or a hover, only the visual noise.

## One honest composition

**What this subnet exposes**, by surface kind — SN1 is 32 subnet APIs of 47
surfaces. Counts of distinct surfaces are additive and share a unit, so unlike
this subnet's economics they genuinely divide a whole. It answers the question
a builder arrives with, which was previously a bare number in a paths row.

Registry vocabulary is translated on the way out: nobody should need to know
that `subnet-api` is an enum member to read a chart.

`/api/v1/subnets/{netuid}/surfaces` had no recorded fixture. Recorded one.

## Validation

- `npm run test:e2e --workspace=apps/ui` — **270 passed, 0 failed**, including
  5 new `#11521` cases
- one of those reads the drawn slice widths back out of the DOM and checks them
  against the printed percentages — the shares are derived from the same values
  that size the bar, and the test proves that rather than assuming it
- unit 2,357; root + both workspace linters clean; prettier, typecheck clean
- `validate:` unreferenced-exports/modules, ui-route-coverage, contract-drift,
  boundary-casts, artifact-budgets — pass

## Not done yet

- No before/after screenshot matrix.
- Left column still carries a tall gap under its actions, because the rail is
  the taller of the two columns. Tightening the rail is a follow-up.
- `#11521` also asks for raw metagraph / ownership / alert setup / audit
  history to move into intentional advanced sections; the tabs already do much
  of that, and auditing the rest is the remaining slice.

Refs #11521


JSONbored:
Closing as part of superseding the Metagraph Terminal stack with epic #11604 (Design System v2). The ten `codex/*` PRs form one linear chain that cannot merge independently, three have red CI, one is conflicting, and at the stack tip the profile routes still carry the legacy grammar (41 StatTiles / 33 SectionAnchors / 10 tables on the account profile). The replacement work lands as one PR per sub-issue of #11604, starting from `main`. Branch left in place; not deleted.

PR #11587, opened 2026-08-22
feat(ui): make the type system a system; one composition and one task strip

### LINKED ISSUE #11520: feat(ui): make subnet discovery focused and progressive

### LINKED ISSUE #11522: feat(ui): make validator discovery and profiles decision-ready

JSONbored:
Closes #11522
Closes #11520

Measured against opencode.ai/data rather than argued about. The gap was never colour — it was that there was **no system**, and fifteen local decisions read as fifteen local decisions.

## What the reference actually does

| | opencode.ai/data | metagraphed (before) |
|---|---|---|
| Font families | 1 (all 970 nodes) | 1 ✓ |
| Distinct font sizes | 7 — three carry 93% | 8 |
| Base/workhorse size | **11px** | **9.5px** — fractional |
| `letter-spacing` values | **1** (`normal`, all 970) | **15 bespoke** |
| Border radii | **1** (4px, 37 uses) | 5 |
| Ranked bar fill | plain white on `rgb(48,48,48)` | graphite ✓ |
| Where the 11-colour ramp lives | **the stacked composition only** | nowhere |

Their ranked bars have the same long-tail stubs ours do. The ramp is reserved for one chart, where every colour is a real category.

## 1 · Type is now a system

`letter-spacing` was set in 22 places with 15 bespoke values — ten negative, `-0.02em` through `-0.07em`, all doing the one job of tightening a heading, each picked by eye at a different moment. No two headings on the site were tracked alike. Now one `--mg-tracking-display`, one `--mg-tracking-label`, `normal` everywhere else. **Zero bespoke values remain.**

`--mg-type-micro` was **9.5px** and was the workhorse — 2,114 of 2,588 text nodes on /validators. A sub-10px fractional size doing the job of body copy is most of why the pages read as cramped. 10px now; eight distinct sizes drop to seven.

**A correction to my own diagnosis:** I assumed /subnets' 145 full-radius elements were pill badges. They aren't — 129 are 8×8 status dots, which *should* be circles. Measuring first prevented a pointless edit across 182 call sites. Only `.mg-chip` was a real pill.

## 2 · /validators gets the composition it lacked

Nothing on the page was entitled to categorical colour, so it had none. **Concentration.** is a real part-to-whole on the shared `CompositionBreakdown`: stake is TAO-denominated for every key, so unlike a subnet's alpha these values genuinely sum. It says what a ranked rail cannot — the top ten hold **63.3%** — once the rail's tail is 140 rows of stubs.

The residual defaulted to flat `#9a9a9a`, **brighter than most of the ramp on a dark ground**, so the one slice carrying no meaning was the loudest block on the bar and, at 36.7%, the largest. It derives from the ink now.

**The page was already contradicting itself.** There was another concentration bar lower down ranked per **key** while the directory above it ranks **operators** — the same page said 63.3% and 59.1%, and an operator running two keys in the top ten could appear twice in its own chart. Removed, with its private ten-colour array.

### The table

`30d Δ` was clipped off the right edge at 1280 — **not** from column widths. Those are weights `TableColGroup` turns into percentages, so editing them re-proportions rather than resizes. The cause was `min-w-[1100px]`, sized for the nine-column set and never revisited.

```
before:  table 1100  container 1086  overflow 14
after:   table 1086  container 1086  overflow  0
```

Dominance → opt-in (total stake restated as a share of the same total). Total stake moves beside the name it ranks instead of four columns from the ordering the table was already in.

## 3 · /subnets: one task strip, not two nested ones

Four stacked control layers before a single subnet: a `Registry | Rankings` tablist, a module heading and caption, a `Browse | Research | Compare` tablist nested inside it, then search + five chips + a Filters drawer. **Two of the four were tablists answering the same question at two levels.**

The code already knew — the comment above the outer strip read *"Rankings is a sibling task, not a separate destination."* It was drawn as one anyway.

One strip now: **Browse · Research · Compare · Rankings.** `DirectoryModeTabs` is generic over its value type rather than widening the shared `DirectoryMode` union, so /subnets adds a fourth task without /validators inheriting a mode it cannot serve.

First data row: **y=841 → y=760**, one whole tab row gone.

## Tests

Rankings had **no e2e coverage at all** — the full suite passed unchanged after the restructure, which is how I know. Two new guards now pin the single strip and the Rankings round-trip.

Two test corrections worth naming:
- my rankings guard first asserted `section === "registry"` after switching back, and failed: the router strips the param at its default, so it was testing default-encoding, not the tab.
- `leaderboards-csv-export-menu.test.ts` indexed the first `{search.section === "rankings" ? (` and sliced forward; when the ActionBar moved into the module's `actions` prop that anchor matched a different conditional and the slice came back empty. It now finds the ActionBar from the menu it asserts about.

## Verification

- 277/277 e2e passed
- typecheck, root + both workspace lints, prettier, 2,357 unit tests clean
- `unreferenced-exports` (731, at ceiling), `ui-route-coverage`, `contract-drift`, `boundary-casts`, `schema-vocabularies` pass

## Known, not done

`?section=rankings` still opens onto a **card wall** — ten bordered boards, plus its own kicker/h2/paragraph heading stack and pill-shaped window toggles. That is the next thing.

**Do not merge** — visual review pending.


JSONbored:
Closing as part of superseding the Metagraph Terminal stack with epic #11604 (Design System v2). The ten `codex/*` PRs form one linear chain that cannot merge independently, three have red CI, one is conflicting, and at the stack tip the profile routes still carry the legacy grammar (41 StatTiles / 33 SectionAnchors / 10 tables on the account profile). The replacement work lands as one PR per sub-issue of #11604, starting from `main`. `AnalyticsSection` from this branch is the named cherry-pick source for #11607; the branch is kept for that. Branch left in place; not deleted.

